PR #23, opened 2018-07-25
docs: add how to install on Linux systems

dongsupark:
**Description**

On Linux systems, it's not possible to make use of the hyperkit driver of minikube.

It's also needed to run `run.sh --skip-minikube-start` to avoid hyperkit errors.

---

---

### DIFF
--- a/docs/kyma/docs/031-gs-local-installation.md
+++ b/docs/kyma/docs/031-gs-local-installation.md
@@ -16,7 +16,7 @@ To run Kyma locally, download these tools:
 - [Minikube](https://github.com/kubernetes/minikube) 0.28.0
 - [kubectl](https://kubernetes.io/docs/tasks/tools/install-kubectl/) 1.9.0
 - [Helm](https://github.com/kubernetes/helm) 2.8.2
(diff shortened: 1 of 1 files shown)


pkosiec:
Resolves #27

docs/kyma/docs/031-gs-local-installation.md
```diff
@@ -46,11 +47,18 @@ Follow these steps to "always trust" the Kyma certificate on macOS:
...
   ```
   cmd/run.sh
   ```
 
+ (For Linux users) First start a minikube cluster, and run the `run.sh` script, because the hyperkit driver is not available on Linux:
+
+  ```
+  minikube start
+  cmd/run.sh --skip-minikube-start
+  ```
```
pwittchen: How about improving `cmd/run.sh` script? 

I mean, we can do something like that:

```bash
if [ `uname` = "Linux" ]; then
  minikube start
  SKIP_MINIKUBE_START=true
fi
```

in the beginning of the script.
After that it should work seamlessly both on macOS and Linux out of the box.

dongsupark:
Ok, https://github.com/kyma-project/kyma/pull/41 looks reasonable.

PR #38, opened 2018-07-26
allowing to install kyma with minikube version newer than 0.28.0

pwittchen:

- [x] I have tested my changes. (I've tested it with minikube 0.28.2 on macOS)
- [x] I have updated the relevant documentation. (no docs was required)
---

**Description**

- Before this change, I wasn't able to install kyma with minikube version 0.28.2
- In this PR, I'm allowing to install kyma with minikube version newer than 0.28.0 (previously it was possible to install it with 0.28.0 version **only**)

### DIFF
--- a/installation/scripts/minikube.sh
+++ b/installation/scripts/minikube.sh
@@ -120,7 +120,7 @@ function checkIfMinikubeIsInitialized() {
 function checkMinikubeVersion() {
     local version=$(minikube version | awk '{print  $3}')
 
-    if [[ "${version}" != *"${MINIKUBE_VERSION}"* ]]; then
+    if [[ "${version}" < *"${MINIKUBE_VERSION}"* ]]; then
         echo "Your minikube is in v${version}. v${MINIKUBE_VERSION} is supported version of minikube. Install supported version!"
         exit -1
     fi


mszostok:
As you can read:
> Minikube is still under active development, and features may change at any time. 

source: https://github.com/kubernetes/minikube/releases/tag/v0.28.2

Additionally, the sem versioning says:
>Major version zero (0.y.z) is for initial development. Anything may change at any time. The public API should not be considered stable.

source: https://semver.org/#spec-item-4

so this check is by design. For example, currently, we are using the deprecated bootstrapper from minikube ([localkube](https://github.com/kyma-project/kyma/blob/master/installation/scripts/minikube.sh#L188)). Probably in next releases, guys from minikube will remove it. When you will change this check, then we will allow the user to install kyma on the different version and it can fail but the user will not see any warnings from our site. With this check, we protect the user from making this mistake.

If it's really important for you then we can start a discussion and maybe add a new flag which will allow you to install kyma even if versions are different (do it at your own risk) or maybe add check which will validate only the major and minor numbers and allows you to increment the patch version because I'm assuming that incrementing patch number should not break the flow but it's still a little bit risky and need to be discussed.

pwittchen:
It's not important for me. I can update this script locally and install app with the newest minikube anyway. If there are relevant issues with this script update, you can close PR without merging.

PR #117, opened 2018-08-03
Cache vendor in CI

clebs:

---

**Description**

- Add caching of vendor folders when building golang components

### DIFF
--- a/components/api-controller/Jenkinsfile
+++ b/components/api-controller/Jenkinsfile
@@ -41,10 +41,14 @@ podTemplate(label: label) {
                             }
                         }
 
-                        stage("install dependencies $application") {
-                            execute("make resolve")
+                        cache(maxCacheSize: 256, caches: [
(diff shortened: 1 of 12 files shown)


Abd4llA:
@clebs When will this cache gets invalidated ? When the step produces different files in hash?
how does it determine that vendor comes from hash but e.g. bin/<new bin> doesn't?

clebs:
@Abd4llA this is a PoC for now, so I do not have full results yet.
Nevertheless, this cache is not invalidated, it gets updated on every build when `dep ensure` updates the contents of the `vendor` folder.

At the start of  the dependencies stage it is loaded, gets updated and is stored again at the end of the stage.

>how does it determine that vendor comes from hash but e.g. bin/ doesn't?
I am afraid I did not understand the question, could you rephrase it?

clebs:
Closing PR, the cache plugin is not viable for a multiple parallel job scenario.
Parallel jobs write to the same cache causing race conditions, I/O errors and overwrites. No way found to have dedicated cache for each job.

PR #133, opened 2018-08-07
Do not allow nested hostnames in Api

sjanota:

---

**Description**

- api-controller:
  * [x] CRD hostname validation is changed: hostname can be complete address (with domain) or simply first part of it. For example CRD should accept both 'my-service' and 'my-service.kyma.local'
  * [x] Controller reads DOMAIN_NAME env variable and adds it to hostname if it's provided in simple form
  * [x] CRD validates hostname against DOMAIN_NAME
  * [x] DOMAIN_NAME is set in deployment template
  * [x] [cosmetic] Remove getting Api object from kubernetes twice in onCreate 

**Issue link**

### DIFF
--- a/components/api-controller/pkg/controller/v1alpha2/crd.go
+++ b/components/api-controller/pkg/controller/v1alpha2/crd.go
@@ -100,7 +100,7 @@ func Crd() *k8sApiExtensions.CustomResourceDefinition {
 }
 
 const (
-	hostnamePattern = `^(([a-zA-Z0-9]|[a-zA-Z0-9][a-zA-Z0-9\-]*[a-zA-Z0-9])\.)*([A-Za-z0-9]|[A-Za-z0-9][A-Za-z0-9\-]*[A-Za-z0-9])$`
+	hostnamePattern = `^([A-Za-z0-9]|[A-Za-z0-9][A-Za-z0-9\-]*[A-Za-z0-9])$`
 	urlPattern      = `^(?:https?:\/\/)?(?:[^@\/\n]+@)?(?:www\.)?([^:\/\n]+)`
 )
 
--- a/components/api-controller/pkg/controller/v1alpha2/crd.go
+++ b/components/api-controller/pkg/controller/v1alpha2/crd.go
@@ -100,7 +100,7 @@ func Crd(domainName string) *k8sApiExtensions.CustomResourceDefinition {
 }
 
 const (
-	hostnamePatternFormat = `^([A-Za-z0-9]|[A-Za-z0-9][A-Za-z0-9\-]*[A-Za-z0-9])(\.%s)?$`
+	hostnamePatternFormat = `^([A-Za-z0-9]|[A-Za-z0-9][A-Za-z0-9\-]{0,62}[A-Za-z0-9])(\.%s)?$`
 	urlPattern      = `^(?:https?:\/\/)?(?:[^@\/\n]+@)?(?:www\.)?([^:\/\n]+)`
 )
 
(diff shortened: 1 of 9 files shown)


kubadz:
There are two problems - one with constantly updating all resources every n seconds (visible in logs after changes introduced in commit 1590cc4de38fa89fc5358c220de7b0a409fff31d). Probably not a bug, but we should investigate it. If not a bug we shouldn't log these events.

The second problem appears after updating the hostname of created Api. Status of resources is somehow reset after performing update and "successfully" saving the status. That can be seen in api-controller logs after performing next update.

components/api-controller/pkg/controller/v1alpha2/crd.go
```diff
@@ -100,10 +100,14 @@ func Crd() *k8sApiExtensions.CustomResourceDefinition {
 }
 
 const (
-	hostnamePattern = `^(([a-zA-Z0-9]|[a-zA-Z0-9][a-zA-Z0-9\-]*[a-zA-Z0-9])\.)*([A-Za-z0-9]|[A-Za-z0-9][A-Za-z0-9\-]*[A-Za-z0-9])$`
+	hostnamePatternFormat = `^([A-Za-z0-9]|[A-Za-z0-9][A-Za-z0-9\-]*[A-Za-z0-9])(.%s)*$`
```
pbochynski: I'm not a regexp master, but I think there is a lot of invalid hostnames that will pass the validation. Try the following:
- `host/kyma.local`
- `host.kyma*local`
- `host.kyma.local.kyma.local.kyma.local`
All of them will be valid hostnames for domain kyma.local.
To avoid it please:
- replace `*` with `?` for second group
- replace `.` with `\.` in second group
- escape `.` in domain name

sjanota: You're absolutely right.

sjanota:
@kubadz 
According to [sample-controller](https://github.com/kubernetes/sample-controller/blob/master/controller.go#L135) first problem is the expected behaviour. I removed my change from the PR so periodical updates are now filtered out again.

sjanota:
@kubadz 
Regarding the second problem: [here](https://github.com/kyma-project/kyma/blob/master/components/api-controller/pkg/controller/v1alpha2/controller.go#L306) we are modifying status field and setting it to the same value that is seen later when the Api is broken. In [sample-controller](https://github.com/kubernetes/sample-controller/blob/master/controller.go#L321) it is also mentioned that implementation should never ever modify value received from cache. I added deep copying of objects when they are put in the queue and it solved the problem (at least I'm not able to reproduce it anymore).

components/api-controller/pkg/controller/v1alpha2/crd.go
```diff
@@ -100,10 +100,15 @@ func Crd() *k8sApiExtensions.CustomResourceDefinition {
 }
 
 const (
-	hostnamePattern = `^(([a-zA-Z0-9]|[a-zA-Z0-9][a-zA-Z0-9\-]*[a-zA-Z0-9])\.)*([A-Za-z0-9]|[A-Za-z0-9][A-Za-z0-9\-]*[A-Za-z0-9])$`
+	hostnamePatternFormat = `^([A-Za-z0-9]|[A-Za-z0-9][A-Za-z0-9\-]*[A-Za-z0-9])(\.%s)?$`
 	urlPattern      = `^(?:https?:\/\/)?(?:[^@\/\n]+@)?(?:www\.)?([^:\/\n]+)`
```
piotrmsc: According to RFC, each label of the hostname must be up to 63 characters

PR #214, opened 2018-08-17
Document the micro services programming model

montaro:


montaro:
@abbi-gaurav I'm working now on referring for the guide in the review. But I'd like to address your review comments:
- How can I access the headers?
This is verbosely elaborated in the `JS` example.
- What does the headers mean? 
They are all documented by category right after the example
- How can I access the payload?
This is also documented and explained in the `JS` code example
- Where can I access the schema of the payload?
I'm not sure what do you mean exactly because it's basically a `JSON` object (event payload) so it shows how to access the payload/where is it in the request, elaborated with a foobar example

abbi-gaurav:
@montaro 
* It is confusing when we have `kyma-*` headers and there is no description there and a link to some object that seems to be saying something else.

* For the schema, you need to provide information about how a developer writing a microservice write logic that is aligned with the payload structure.

PR #224, opened 2018-08-20
Add helm chart for PodPreset admission controller installation

mszostok:

---

**Description**

- Provide a helm chart which will install the admission controller for PodPresets.
This helm chart installs the controllers extracted from the main k8s repository. The source code you can find [here](https://github.com/jpeeler/podpreset-crd)

We need it because on the managed cluster we cannot enable the core k8s admission controllers for PodPreset because they are in the alpha state.

Additionally is worth to mention that Jeff Peeler implemented a new component called controller-manager which is responsible for restarting the Deployments when PodPreset was changed. This functionality is new and was not present in k8s implementation, so for now I disabled it.

**How to test**

Just install the Kyma locally or use the cluster (I've prepared the gopher cluster, you can download the kubeconfig from the **33** on-demand/gopher Jenkins Job)

_Steps_
1. Create a pod preset from API Group: **settings.svcat.k8s.io/v1alpha1**
```bash
kubectl create -f https://gist.github.com/mszostok/f56be101e7459c64ccb88c23cd23e777
```

2. Create a deployment.
```bash
kubectl create -f https://gist.github.com/mszostok/e5a68e6a9e9d743eed680a00d4612701/raw
```

3. Check that pod has annotation about the PodPreset action and has injected env and mounted volume.
```bash
kubectl describe po website
```

**NOTE:** _Still we need to clarify where to store the Docker images for the controller_

**Differences between this helm chart and source [repo](https://github.com/jpeeler/podpreset-crd)**
- create proper ClusterRoles, and introduce separate account for each deployment
- autogeneration of proper CA and TLS keys

### DIFF
--- a/resources/cluster-essentials/charts/pod-preset/values.yaml
+++ b/resources/cluster-essentials/charts/pod-preset/values.yaml
@@ -1,0 +1,19 @@
+webhook:
+  replicaCount: 1
+  image:
+    repository: mszostok/service-catalog-admission-webhook
+    tag: 0.0.1
+    pullPolicy: IfNotPresent
+  verbosity: 6
+  service:
+    type: NodePort
+    port: 443
+
+controller:
+  enabled: false
+  image:
+    repository: mszostok/service-catalog-podpreset-controller
+    tag: 0.0.1
+    pullPolicy: IfNotPresent
+
+
--- a/resources/cluster-essentials/charts/pod-preset/values.yaml
+++ b/resources/cluster-essentials/charts/pod-preset/values.yaml
@@ -2,7 +2,7 @@ webhook:
   replicaCount: 1
   image:
     repository: mszostok/service-catalog-admission-webhook
-    tag: 0.0.1
+    tag: 0.0.2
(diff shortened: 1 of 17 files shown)


resources/cluster-essentials/charts/pod-preset/values.yaml
```diff
@@ -0,0 +1,19 @@
+webhook:
+  replicaCount: 1
+  image:
+    repository: mszostok/service-catalog-admission-webhook
```
aszecowka: need to be clarified where this image should be placed

aszecowka: additionally, it would be great to have README file. In such README I would like to know how to build new docker image.
Version 0.0.1 is build from which revision?

mszostok: done

mszostok: as I wrote in the description, this is a temporary solution until we will be able to download the official ones

PR #233, opened 2018-08-22
Move kyma components list from CM to installation CR

crabtree:

---

**Description**

- Move kyma components list from ConfigMap to installation CR
- Extend components list from simple list of strings to list of structs to handle more parameters describing component


components/installer/pkg/config/installation-data.go
```diff
@@ -95,12 +93,11 @@ func (installationData *InstallationData) ShouldInstallComponent(componentName s
...
-	split := strings.Split(strings.Replace(cList, " ", "", -1), ",")
-	output := make(map[string]struct{}, len(split))
-	for _, c := range split {
-		if c != "" {
-			output[c] = struct{}{}
+func convertToMap(cList []v1alpha1.KymaComponent) map[string]v1alpha1.KymaComponent {
+	output := make(map[string]v1alpha1.KymaComponent, len(cList))
+	for _, c := range cList {
+		if c.Name != "" {
+			output[c.Name] = c
```
aszecowka: is it possible that this list contains duplicates? What in such case?

components/installer/pkg/config/installation-data_test.go
```diff
@@ -3,26 +3,27 @@ package config
...
+			fakeComponentList := []v1alpha1.KymaComponent{}
 			testMap := convertToMap(fakeComponentList)
 
 			So(len(testMap), ShouldEqual, 0)
 		})
 
 		Convey("Should the string contain just one element, the map must be of length 1 as well", func() {
 
-			fakeComponentList = "core"
+			fakeComponentList := []v1alpha1.KymaComponent{
```
aszecowka: you need to update test name, because now it is not a string. Please check other cases.

components/installer/pkg/config/installation-data.go
crabtree: For now it is not a problem, because installation process is still hardcoded so every component will be installed only once. In further changes this needs to be addressed.

components/installer/pkg/config/installation-data_test.go
crabtree: True, fixed

PR #303, opened 2018-08-28
Remote Environment Broker: Middleware to propagate from http.Request to osb.Context brokerNamespace

aszecowka:

---

**Description**
Remote Environment Broker can be registered as a Cluster-scoped broker or Namespaced-Broker. To distinguish this 2 situations we need to extract broker namespace from http.Request host.


components/remote-environment-broker/internal/config/config.go
```diff
@@ -28,10 +28,10 @@ type Config struct {
 	Logger                     logger.Config
 	Port                       int              `default:"8080"`
 	Storage                    []storage.Config `valid:"required"`
-	BrokerName                 string           `valid:"required"`
 	BrokerRelistDurationWindow time.Duration    `valid:"required"`
+	ClusterScopedBrokerName    string           `valid:"required"`
 	// Before adjusting ui-api-layer and tests, this flag by default should be set to true.
-	ClusterScopedBroker bool `default:"true"`
+	ClusterScopedBroker bool ` valid:"required"`
```
aszecowka: our Configuration does not support properly defaults for boolean type. go-defaults will set ClusterScopedBroker to true if this value were equal zero value (it means false). Pointers are also not supported.

components/remote-environment-broker/internal/config/flavor_test.go
```diff
@@ -0,0 +1,25 @@
...
+
+import (
+	"testing"
+
+	"github.com/kyma-project/kyma/components/remote-environment-broker/internal/config"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestBrokerFlavorGetNsFromBrokerURL(t *testing.T) {
```
mszostok: IMO it will be more clear if you will have just two functions `TestBrokerFlavorGetNsFromBrokerURLSuccess` and `TestBrokerFlavorGetNsFromBrokerURLFailure`
for me `t.Run` is nice for table test and if you, for example, want to group more failure/success test case in one method.
e.g.
```go

func TestBrokerFlavorFailure(t *testing.T) {
	t.Run("first failure case", func(t *testing.T) {
		//  some test
	})

	t.Run("second failure case", func(t *testing.T) {
		//  some test
	})
}
```

but grouping it in this way
```go
func TestBrokerFlavor(t *testing.T) {
	t.Run("success case", func(t *testing.T) {
		//  some test
	})

	t.Run("failure case", func(t *testing.T) {
		//  some test
	})

	t.Run("success case", func(t *testing.T) {
		//  some test
	})

	t.Run("failure case", func(t *testing.T) {
		//  some test
	})
}
```
is less readable, but it is **only my opinion** which I respect in my tests

components/remote-environment-broker/internal/broker/middleware.go
```diff
@@ -4,16 +4,48 @@ import (
...
 
 	osb "github.com/pmorie/go-open-service-broker-client/v2"
+	"github.com/sirupsen/logrus"
 )
 
 // OSBContextMiddleware implements Handler interface
-type OSBContextMiddleware struct{}
+type OSBContextMiddleware struct {
+	brokerFlavor brokerFlavorProvider
+	log          *logrus.Entry
```
mszostok: please use interface `logrus.FieldLogger` instead of pointer to struct, same for the arg in `NewOsbContextMiddleware` constructor

components/remote-environment-broker/internal/broker/middleware.go
```diff
@@ -4,16 +4,48 @@ import (
...
 
 // ServeHTTP adds content of Open Service Broker Api headers to the requests
-func (OSBContextMiddleware) ServeHTTP(rw http.ResponseWriter, r *http.Request, next http.HandlerFunc) {
+func (m *OSBContextMiddleware) ServeHTTP(rw http.ResponseWriter, r *http.Request, next http.HandlerFunc) {
+	brokerNamespace := ""
+	if m.brokerFlavor.IsClusterScoped() == false {
+		var err error
+		brokerNamespace, err = m.brokerFlavor.GetNsFromBrokerURL(r.Host)
+		if err != nil {
+			errMsg := "misconfiguration, broker is running as a namespace-scoped, but cannot extract namespace from request host"
```
mszostok: "(..) from request Host header"

components/remote-environment-broker/internal/broker/middleware.go
```diff
@@ -4,16 +4,48 @@ import (
...
+			m.log.Error(errMsg, err)
+			writeErrorResponse(rw, http.StatusInternalServerError, errMsg, "")
```
mszostok: the errMsg is copied to the `Error` field in struct. The description of this field is *Error is a machine readable info on an error.* 
But there is also another field called `Desc`, with such description: *A meaningful error message explaining why the request failed.*
so what about that?
```go
writeErrorResponse(rw, http.StatusInternalServerError, "OSBContextMiddlewareError", errMsg)
```

components/remote-environment-broker/internal/broker/middleware_test.go
```diff
@@ -0,0 +1,88 @@
...
+	req := httptest.NewRequest(http.MethodGet, "https://core-reb.kyma-system.svc.cluster.local/v2/catalog", nil)
+	rw := httptest.NewRecorder()
+	nextCalled := false
+	// WHEN
+	sut.ServeHTTP(rw, req, func(nextRw http.ResponseWriter, nextReq *http.Request) {
+		nextCalled = true
+		osbCtx, ex := osbContextFromContext(nextReq.Context())
+		// THEN
+		assert.True(t, ex)
+		assert.True(t, osbCtx.ClusterScopedBroker)
```
mszostok: should we also add:
```go
assert.Empty(t, osbCtx.BrokerNamespace)
```
is there a contract which says if ClusterScopedBroker is set to `true` then BrokerNamespace should not be populated? IMO should

components/remote-environment-broker/internal/broker/middleware_test.go
```diff
@@ -0,0 +1,88 @@
...
+	defer mockBrokerFlavor.AssertExpectations(t)
+
+	mockBrokerFlavor.On("IsClusterScoped").Return(false)
+	mockBrokerFlavor.On("GetNsFromBrokerURL", mock.Anything).Return("", errors.New("some error"))
+	logSink := spy.NewLogSink()
+	sut := NewOsbContextMiddleware(mockBrokerFlavor, logSink.Logger)
+	// WHEN
+	req := httptest.NewRequest(http.MethodGet, "https://core-reb.kyma-system.svc.cluster.local/v2/catalog", nil)
+	rw := httptest.NewRecorder()
+	// THEN
```
mszostok: IMO `// THEN` should be moved to line 85

components/remote-environment-broker/internal/config/flavor.go
```diff
@@ -0,0 +1,37 @@
+package config
+
+import (
+	"fmt"
+	"regexp"
+)
+
+// BrokerFlavor aggregates information about mode REB is running: if it is cluster scoped or namespaced scoped
+type BrokerFlavor struct {
```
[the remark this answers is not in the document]
aszecowka: currently it is used by middleware but in the future new methods will be added, for example for generation of service url for ns-broker - to have only in one place regexp. So I will create package internal/mode.

components/remote-environment-broker/internal/config/flavor.go
```diff
@@ -0,0 +1,37 @@
...
+	clusterBrokerName  string
+	nsBrokerURLPattern *regexp.Regexp
+}
+
+// NewBrokerFlavorFromConfig creates BrokerFlavor from configuration
+func NewBrokerFlavorFromConfig(cfg *Config) *BrokerFlavor {
+	return &BrokerFlavor{
+		clusterBrokerName:  cfg.ClusterScopedBrokerName,
+		clusterScoped:      cfg.ClusterScopedBroker,
+		nsBrokerURLPattern: regexp.MustCompile("reb-ns-for-([a-z][a-z0-9-]*)\\."),
```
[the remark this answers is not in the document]
aszecowka: I don't like this name, so I switched to Compile and returning error.

components/remote-environment-broker/internal/config/flavor.go
```diff
@@ -0,0 +1,37 @@
...
+}
+
+// GetNsFromBrokerURL extracts namespace from broker URL
+func (bf *BrokerFlavor) GetNsFromBrokerURL(url string) (string, error) {
+	out := bf.nsBrokerURLPattern.FindStringSubmatch(url)
+	if len(out) == 0 {
+		return "", fmt.Errorf("url:%s does not match pattern %s", url, bf.nsBrokerURLPattern.String())
+	}
+	// out[0] = matched regexp, out[1] = matched group in bracket
+	return out[1], nil
```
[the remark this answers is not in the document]
aszecowka: It will never happen that we match whole expression and don't match subexpression, because regexp is defined in that way that group match is required [a-z].
But to not confuse others I change if

components/remote-environment-broker/internal/broker/middleware.go
aszecowka: from request.go:
```
// For server requests Host specifies the host on which the
	// URL is sought. Per RFC 2616, this is either the value of
	// the "Host" header or the host name given in the URL itself.
	// It may be of the form "host:port". For international domain
	// names, Host may be in Punycode or Unicode form. Use
	// golang.org/x/net/idna to convert it to either format if
	// needed.
	//
	// For client requests Host optionally overrides the Host
	// header to send. If empty, the Request.Write method uses
	// the value of URL.Host. Host may contain an international
	// domain name.
	Host string
```
so it will be safer to write "from request"

components/remote-environment-broker/internal/broker/dto.go
```diff
@@ -5,7 +5,7 @@ import (
 	"github.com/pkg/errors"
 )
 
-// DTOs for Open Service Broker v2.12 API
+// DTOs for Open Service BrokerService v2.12 API
```
mszostok: some mistake during auto-refactor, check also other places like middleware.go

PR #328, opened 2018-08-29
add manage-component.sh script

jakkab:

---

**Description**

- add manage-component.sh script, that creates the `components.yaml` file

### DIFF
--- a/installation/scripts/manage-component.sh
+++ b/installation/scripts/manage-component.sh
@@ -1,0 +1,40 @@
+#!/bin/bash
+
+set -o errexit
+
+ROOT_PATH=$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )
+
+COMPONENT_NAME="$1"
+ENABLED="$2"
+STATUS="disabled"
+FILE_NAME="components.yaml"
+FILE_PATH=$ROOT_PATH/../${FILE_NAME}
+
+# Check if the provided value is a valid boolean
+if [[ ! ("${ENABLED}" == "true" || "${ENABLED}" == "false") ]]; then
+    echo "\"${ENABLED}\" is not a boolean. Please provide a boolean value!"
+    exit 1
+fi
+
+# Set status
+if [ "${ENABLED}" == "true" ]; then
+    STATUS="enabled"
+fi
+
+# Create the components.yaml file if it does not exist
+if [ ! -f "${FILE_PATH}" ]; then
+    echo "Generating components.yaml file"
+    touch "${FILE_PATH}"
+fi
+
+# Remove previous entry in case the provided key exists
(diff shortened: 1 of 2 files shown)


jakkab:
PowerShell counterpart of this script will be implemented in #330

installation/scripts/manage-component.sh
```diff
@@ -0,0 +1,40 @@
...
+set -o errexit
+
+ROOT_PATH=$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )
+
+COMPONENT_NAME="$1"
+ENABLED="$2"
+STATUS="disabled"
+FILE_NAME="components.yaml"
+FILE_PATH=$ROOT_PATH/../${FILE_NAME}
+
```
crabtree: Please add curly brackets around `$ROOT_PATH`

installation/scripts/manage-component.sh
```diff
@@ -0,0 +1,40 @@
...
+# Create the components.yaml file if it does not exist
+if [ ! -f "${FILE_PATH}" ]; then
+    echo "Generating components.yaml file"
+    touch "${FILE_PATH}"
+fi
+
+# Remove previous entry in case the provided key exists
+if grep -Fq ${COMPONENT_NAME} ${FILE_PATH}; then
+    sed -i '' '/'"${COMPONENT_NAME}"'/d' ${FILE_PATH}
```
crabtree: Have you tested this sed expression on linux? Will it still be valid?

installation/scripts/manage-component.sh
jakkab: done!

installation/scripts/manage-component.sh
jakkab: done!

installation/scripts/manage-component.sh
```diff
@@ -0,0 +1,47 @@
...
+# Set status
+if [ "${ENABLED}" == "true" ]; then
+    STATUS="enabled"
+fi
+
+# Create the components.yaml file if it does not exist
+if [ ! -f "${FILE_PATH}" ]; then
+    echo "Generating components.yaml file"
+    touch "${FILE_PATH}"
+fi
```
piotrmsc: Looking at this script I assume right now it's designed to be called for each component. In such case checking 'if file exist' for each component that it was invoked with is suboptimal.
Perhaps you can move this part to the place where this script will be invoked or change this script to accept list of component names and iterate over it so this check will be performed only once.

strekm: this script is designed to be fired manually as well as by CI. We do not want to introduce complicated logic checking if the file exists. For example we want to enable to components extra on CI. So we will introduce to separate steps to do that instead of implementing complicated condition to evaluate if the file exists first. @jakkab please check if file creation is necessary (on mac os there's no need to create it)

PR #349, opened 2018-08-30
Adjust CR documentation to the new guidelines

klaudiagrz:

### DIFF
--- a/docs/api-gateway/docs/011-api-custom-resource.md
+++ b/docs/api-gateway/docs/011-api-custom-resource.md
@@ -35,8 +35,6 @@ This table analyses the elements of the sample CR and the information it contain
 
 | Field   |      Mandatory?      |  Description |
 |:----------:|:-------------:|:------|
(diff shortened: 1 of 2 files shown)


docs/api-gateway/docs/011-api-custom-resource.md
```diff
@@ -35,8 +35,6 @@ This table analyses the elements of the sample CR and the information it contain
 
 | Field   |      Mandatory?      |  Description |
 |:----------:|:-------------:|:------|
-| **apiVersion** |  **YES** | Defined basing on the **group** and **version** fields of the CRD **spec** section. |
```
tomekpapiernik: Why delete these fields? They are a part of every CR.

klaudiagrz: Yup, that's why @derberg decided not to include them in the table as they are obvious.

derberg: exactly, these are basic fields that are part of every resource, not only CRDs

tomekpapiernik: okaaaay

PR #392, opened 2018-08-31
disable-failing-ui-tests

y-kkamil:
this disables randomly failing ui tests


resources/core/.helmignore
```diff
@@ -18,4 +18,6 @@
 # Various IDEs
 .project
 .idea/
-*.tmproj
\ No newline at end of file
+*.tmproj
+#
```
derberg: empty comment

PR #394, opened 2018-08-31
Update Jaeger version to 1.6.0

montaro:


resources/core/charts/jaeger/templates/deployment.yaml
```diff
@@ -43,5 +43,8 @@ spec:
               path: "/"
               port: {{ .Values.jaeger.uiPort }}
             initialDelaySeconds: 5
-          resources:
+        {{- if .Values.jaeger.memory }}
```
marcobebway: Why not check for `Values.jaeger.memory.maxTraces`?

montaro: Because I meant to check mainly for the deployment type, if it is in memory, then all these list of options, which is basically only one item for now.

Issue #351, opened 2018-08-30
Fix missing nginx trace span


abbi-gaurav:
In kyma, trace spans are created for traffic coming from nginx to downstream services. But for some reason nginx trace spans are not created anymore, and we cannot see them on jaeger UI.

Follow the next steps to reproduce:
 * success scenario:
   * checkout an old version of kyma (e.g. on *August 10th, 2018*)
   * run kyma
   * issue a publish request through nginx
   * verify that you can see nginx trace spans on jaeger UI.
 * failure scenario:* checkout the latest version of kyma
   * run kyma
   * issue a publish request through nginx
   * verify that you cannot see nginx trace spans on jaeger UI.

Hint:
 * The opentracing configuration is defined by configmaps.
 * Check if the correct configmap is getting applied.

Acceptance criteria:
* The issue is fixed and nginx trace spans are visible on jaeger ui
* Change default service name from nginx to one that is better aligned with Kyma architecture. Refer [nginx ingress tracing](https://kubernetes.github.io/ingress-nginx/user-guide/third-party-addons/opentracing/).
  * Check with framefrogs what should be the name.

abbi-gaurav:
* Created PR to fix the missing nginx issue.

* Changing the name is still pending and needs to be discussed with framefrogs. Since I will be going on vacation and most likely won't have time to follow it up, so moving it back to **To DO** for the name alignment and change.

PR #430, opened 2018-09-03
Fix missing nginx tracing span

abbi-gaurav:
The nginx tracing span was missing since the controller-configmap.yaml has an if condition in the helm chart and the flags to enable the tracing were inside the if block. Moving them out of the if block as enabling tracing does not depend upon the controller.headers flag

---

**Description**

-
-
-

**Issue link**

See also #351


abbi-gaurav:
screen shot 2018-09-03 at 16 45 02

Trace after the change.

PR #436, opened 2018-09-04
Add common error handling to UI API

### LINKED ISSUE #271: Error handling in ui-api-layer
Confirm these statements before you submit the issue:

- [x] I have searched open and closed issues for duplicates.
- [x] I have read the contributing guidelines.
---

**Description**
Incorrect error handling in ui-api-layer component.

##### Losing the error context
When we are creating/deleting objects in services then we are inlining the k8s command and returning the same error, see: https://github.com/kyma-project/kyma/blob/master/components/ui-api-layer/internal/domain/servicecatalog/instance_service.go#L186

This is not a good idea, because we are losing the context of given error, because of that, we should check if the error occurred and if yes, then wrap it with `errors.Wrap()`
_Example_
```go
func (svc *instanceService) Create(params instanceCreateParameters) (*v1beta1.ServiceInstance, error) {
   // function body...
  err := svc.client.ServicecatalogV1beta1().ServiceInstances(params.Namespace).Create(&instance)
  if err != nil {
    return errors.Wrapf(err, "while creating service instance %q", instance.Name)
  }
  return nil
}
```

##### Relying directly on k8s errors
When we will introduce changes mention above then we need to also handle somehow the cases where we already rely on errors returned directly from k8s, e.g. see:
https://github.com/kyma-project/kyma/blob/master/components/ui-api-layer/internal/domain/servicecatalog/binding_resolver.go#L40

this problem can be solved by returning our own *not found/already exists* error and provide a function for detecting them: *IsNotFound/IsAlreadyExists*

##### Lack of correct recognition of errors
Not all mutations have implemented a proper error handling. We should always return a different error in case of *already exists* and *not found* err to help the end user to investigate the problem with the mutation failed. Now if any error occur then we are always returnig general error 
```
"Cannot create instance `%s` in environment `%s
```
see:
https://github.com/kyma-project/kyma/blob/master/components/ui-api-layer/internal/domain/servicecatalog/instance_resolver.go#L38-L42

michal-hudy:

---

**Description**

- Add common error handling to UI API
-
-

**Issue link**

Fixes #271

### DIFF
--- a/components/ui-api-layer/internal/domain/apicontroller/api_resolver.go
+++ b/components/ui-api-layer/internal/domain/apicontroller/api_resolver.go
@@ -25,7 +25,7 @@ func newApiResolver(lister apiLister) *apiResolver {
 func (ar *apiResolver) APIsQuery(ctx context.Context, environment string, serviceName *string, hostname *string) ([]gqlschema.API, error) {
 	apis, err := ar.apiLister.List(environment, serviceName, hostname)
 	if err != nil {
-		glog.Error(errors.Wrapf(err, "while listing %s for service name %s, hostname %s", pretty.APIs, serviceName, hostname))
+		glog.Error(errors.Wrapf(err, "while listing %s for service name %v, hostname %v", pretty.APIs, serviceName, hostname))
 		return nil, gqlerror.New(err, pretty.APIs, gqlerror.WithEnvironment(environment))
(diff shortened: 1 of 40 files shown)


michal-hudy:
@aszecowka Thanks for noticing this. Fixed

PR #463, opened 2018-09-04
Add hostname validation to api-controller

### LINKED ISSUE #386: Verify uniqueness of hostnames in API resources in api-controller
Confirm these statements before you submit the issue:

- [x] I have searched open and closed issues for duplicates.
- [x] I have read the contributing guidelines.
---

**Description**

Implement validation of hostnames uniqueness in API resources in api-controller and set status invalid in case of duplicated hostname.

**Reasons**

In the API resources you are defining hostnames under which the API will be accessible. Then api-controller creates Virtual Service, which is an Istio resource used, among others, to expose an app on Gateway under some hostname. Currently, Istio delegates to a user the responsibility to ensure that the blocks of routes in two virtual services don't conflict with each other. In our case, it will always be conflicting as we are exposing services on host and match requests on every path by using "/.*" regex. It may lead to service exposure not working while a user doesn't get any feedback.

**Attachments**

kubadz:

---

**Description**

- Implement validation of hostnames uniqueness in API resources in api-controller and set status invalid in case of duplicated hostname.
- Add info about required environment variable "DOMAIN_NAME" to the readme file

**Issue link**

Resolves #386


components/api-controller/README.md
```diff
@@ -37,12 +37,13 @@ $ go run cmd/controller/main.go
...
-| API_CONTROLLER_LOG_LEVEL | No | `info` | Show detailed logs in the application.
-| DEFAULT_ISSUER | Yes | - | Used to set default issuer in NetworkPolicy
-| DEFAULT_JWKS_URI | Yes | - | Used to set default jwksUri in NetworkPolicy
-| GATEWAY_FQDN | Yes | - | Used to set gateway in VirtualServices spec
+| Name | Required | Default | Description | Possible Values |
+|------|----------|---------|-------------|-----------------|
+| API_CONTROLLER_LOG_LEVEL | No | `info` | Show detailed logs in the application. | `info`, `debug`
+| DEFAULT_ISSUER | Yes | - | Used to set default issuer in NetworkPolicy | any string |
+| DEFAULT_JWKS_URI | Yes | - | Used to set default jwksUri in NetworkPolicy | any string |
+| GATEWAY_FQDN | Yes | - | Used to set gateway in VirtualServices spec | any string |
```
klaudiagrz: spec -> `specification` ?
`Used to set Gateway in the VirtualServices specification.`

components/api-controller/README.md
```diff
@@ -37,12 +37,13 @@ $ go run cmd/controller/main.go
...
+| DOMAIN_NAME | Yes | - | Used to set hostname in VirtualServices spec when short version of hostname is provided | any string |
```
klaudiagrz: spec -> specification ?
`Used to set a hostname in the VirtualServices specification if a short version of the hostname is provided.`

components/api-controller/pkg/apis/gateway.kyma.cx/meta/v1/types.go
```diff
@@ -34,6 +36,10 @@ func (s StatusCode) IsError() bool {
 	return s == Error
 }
 
+func (s StatusCode) IsUpdateFailure() bool {
+	return s == Error
```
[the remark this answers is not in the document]
kubadz: You are right with the first one. `IsEmpty` function seems correct.

components/api-controller/pkg/controller/networking/v1/networkingv1.go
```diff
@@ -18,23 +20,41 @@ import (
...
 	return &istioImpl{
 		istioNetworkingInterface: i,
+		kubernetesInterface:      k,
 		istioGateway:             istioGateway,
 	}
 }
 
+type HostnameNotAvailableError struct {
+	errorMsg string
+}
```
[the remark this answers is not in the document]
kubadz: Good idea. Done.

components/api-controller/README.md
```diff
@@ -37,12 +37,13 @@ $ go run cmd/controller/main.go
 
 Use these environment variables to configure the application:
 
-| Name | Required | Default | Description |
-|-----|---------|--------|------------|
-| API_CONTROLLER_LOG_LEVEL | No | `info` | Show detailed logs in the application.
-| DEFAULT_ISSUER | Yes | - | Used to set default issuer in NetworkPolicy
-| DEFAULT_JWKS_URI | Yes | - | Used to set default jwksUri in NetworkPolicy
-| GATEWAY_FQDN | Yes | - | Used to set gateway in VirtualServices spec
+| Name | Required | Default | Description | Possible Values |
```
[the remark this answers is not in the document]
kubadz: Done.

components/api-controller/pkg/apis/gateway.kyma.cx/meta/v1/types.go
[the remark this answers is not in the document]
sjanota: My bad, you're right.

components/api-controller/README.md
kubadz: Done.

components/api-controller/README.md
kubadz: Done.

components/api-controller/pkg/controller/networking/v1/networkingv1.go
```diff
@@ -157,3 +190,35 @@ func gatewayResourceFrom(vscv *istioNetworkingApi.VirtualService) *kymaMeta.Gate
...
+
+		for _, vs := range vsList.Items {
+			for _, occupiedHostname := range vs.Spec.Hosts {
+				if occupiedHostname == hostname {
+					//if hostname is not used by virtualservice actually assigned to CR
+					if vs.UID != assignedVirtualServiceUID {
+						return false, nil
+					}
+
+				}
```
piotrmsc: return here, we don;t need to iterate more

components/api-controller/pkg/controller/networking/v1/networkingv1.go
```diff
@@ -157,3 +190,35 @@ func gatewayResourceFrom(vscv *istioNetworkingApi.VirtualService) *kymaMeta.Gate
...
+		vsList, err := a.istioVirtualServiceInterface(ns.GetName()).List(k8sMeta.ListOptions{})
+
+		if err != nil {
+			return false, err
+		}
+
+		for _, vs := range vsList.Items {
+			for _, occupiedHostname := range vs.Spec.Hosts {
+				if occupiedHostname == hostname {
+					//if hostname is not used by virtualservice actually assigned to CR
```
piotrmsc: `					//if hostname is used by virtualservice actually assigned to CR
					if vs.UID == assignedVirtualServiceUID {
						return true, nil
					}
                                          return false, nil
`

components/api-controller/pkg/controller/networking/v1/networkingv1.go
```diff
@@ -157,3 +190,35 @@ func gatewayResourceFrom(vscv *istioNetworkingApi.VirtualService) *kymaMeta.Gate
 		Version: vscv.ResourceVersion,
 	}
 }
+
+func (a *istioImpl) isHostnameAvailable(metaDto meta.Dto, hostname string, assignedVirtualServiceUID types.UID) (bool, error) {
+
```
piotrmsc: ensure hostname is not empty - ie. case with kubectl edit without validate flag

components/api-controller/pkg/controller/networking/v1/networkingv1.go
kubadz: Done.

components/api-controller/pkg/controller/networking/v1/networkingv1.go
kubadz: Done, as it is the same suggestion as in the previous comment.

PR #548, opened 2018-09-06
Improve ResourceQuotaStatus endpoint

### LINKED ISSUE #461: ResourceQuotaStatus follow up
**Description**
We can improve our API endpoint for checking ResourceQuotaStatus to provide more information to the user.

**Acceptance criteria**
- ResourceQuota name which has been exceeded in the message.
- Return in message the ReplicaSets and StatefulSets names array which are blocked by the ResourceQuota.
- Rename endpoint to resourceQuotasStatus.
- Refactor the `ResourceQuotaStatus` function to its own resolver.
- Documentation about message structure with example with all possible information

polskikiel:
**Description**

- ResourceQuota name which has been exceeded in the returned message.
- Return in message the ReplicaSets and StatefulSets names array which are blocked by the ResourceQuota.
- Rename endpoint to resourceQuotasStatus.
- Refactor the ResourceQuotaStatus function to its own resolver.
- Documentation about message structure with example with all possible information
- Changed implementation of calculating the required space for the next replica to use LimitRanges.


polskikiel:
@PK85 I added the responses examples to the proposal: https://github.com/kyma-project/kyma/pull/548/files#diff-9531813a8e4702f287d475dd05cccf1b

components/ui-api-layer/internal/domain/k8s/resourcequotastatus_service.go
```diff
@@ -106,45 +114,66 @@ func (svc *resourceQuotaStatusService) CheckResourceQuotaStatus(environment stri
...
+	}
+	return limits, nil
+}
+
+func (*resourceQuotaStatusService) checkAvailableResources(limits map[v1.ResourceName]*resource.Quantity, resourcesToCheck []v1.ResourceName, resourceQuotas []*v1.ResourceQuota) []exceededRef {
+	result := make([]exceededRef, 0)
+	for _, rq := range resourceQuotas {
+		for _, name := range resourcesToCheck {
+			hard, hardExists := rq.Spec.Hard[name]
+			used := rq.Status.Used[name]
```
mszostok: Scenario:
1. User creates the ResourceQuota **ABC**
2. In etcd ResourceQuota **ABC** is available and have the `hard` object set under `spec` field
3. Controller for some reason cannot handle created **ABC**  ResourceQuota, so controller will not populate the `Used` section


When sth like that will take place then you will have probably panic in `if` statement

components/ui-api-layer/internal/domain/k8s/resourcequotastatus_resolver.go
```diff
@@ -0,0 +1,39 @@
...
+		statusSvc: resourceQuotaStatusService,
+	}
+}
+
+type resourceQuotaStatusResolver struct {
+	statusSvc *resourceQuotaStatusService
+}
+
+func (r *resourceQuotaStatusResolver) ResourceQuotasStatus(ctx context.Context, environment string) (gqlschema.ResourceQuotasStatus, error) {
+	resourcesToCheck := []v1.ResourceName{
```
mszostok: As discussed. IMO this is misleading because here is a slice where I can supposedly define what kind of resource should be taken into account when we are checking the resource quotas status. But when we will go to the service then we will discover that this is not true because some part of the calculation is hardcoded. My proposition is to remove from here this parametrization and move it directly to service. After that resolver will look like that:
```go
func (r *resourceQuotaStatusResolver) ResourceQuotasStatus(ctx context.Context, environment string) (gqlschema.ResourceQuotasStatus, error) {
	exceeded, err := r.statusSvc.CheckResourceQuotaStatus(environment)
	if err != nil {
		glog.Error(errors.Wrapf(err, "while getting %s [environment: %s]", pretty.ResourceQuotaStatus, environment))
		return gqlschema.ResourceQuotasStatus{}, gqlerror.New(err, pretty.ResourceQuotaStatus, gqlerror.WithEnvironment(environment))
	}

	return exceeded, nil
}
```

mszostok:
My two cents here:
- We need to test scenario when:
  - ResourceQuota exists but there are not LimitRanges in given ns
  - Both ResourceQuota and LimitRanges does not exist in given ns

Address problem when ResourceQuota exists but there are not LimitRanges in given ns. In such case, we will have an error on replicaset but his error will not be shown in UI because it's not a problem with quota exceeded. Probably discuss this with @PK85.

components/ui-api-layer/internal/domain/k8s/resourcequotastatus_service.go
```diff
@@ -153,16 +182,29 @@ func (*resourceQuotaStatusService) containersUsage(containers []v1.Container, re
 		for _, name := range resourcesToCheck {
 			switch name {
 			case v1.ResourceRequestsMemory:
-				limits[name].Add(*container.Resources.Requests.Memory())
+				if container.Resources.Requests.Memory().IsZero() {
+					limits[name].Add(*defaultLimits[name])
```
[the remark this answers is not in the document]
mszostok: As discussed between @piotrmiskiewicz, @polskikiel and me. We should use the struct which will simplify the solution. As a value use the `resources.Quantity`

components/ui-api-layer/internal/domain/k8s/resourcequotastatus_resolver_test.go
```diff
@@ -0,0 +1,83 @@
...
+	"github.com/kyma-project/kyma/components/ui-api-layer/internal/domain/k8s/automock"
+	testingUtils "github.com/kyma-project/kyma/components/ui-api-layer/internal/testing"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"k8s.io/api/core/v1"
+	"k8s.io/client-go/informers"
+	"k8s.io/client-go/kubernetes/fake"
+)
+
+func TestResourceQuotaStatusResolver_ResourceQuotaStatus_HappyPath(t *testing.T) {
```
mszostok: please change this tests. You should only mock the Service nothing more.
You should have two tests:
 - happy path (error = nil)
 - failure (error != nil )

components/ui-api-layer/internal/domain/k8s/resourcequotastatus_resolver_test.go
```diff
@@ -0,0 +1,37 @@
...
+	// GIVEN
+	statusChecker := automock.NewResourceQuotaStatusChecker()
+	statusChecker.On("CheckResourceQuotaStatus", fixNamespaceName()).Return(gqlschema.ResourceQuotasStatus{Exceeded: false}, nil)
+
+	resolver := newResourceQuotaStatusResolver(statusChecker)
+
+	// WHEN
+	_, err := resolver.ResourceQuotasStatus(context.Background(), fixNamespaceName())
+	require.NoError(t, err)
```
mszostok: missing // THEN 
missing assertion for returned status from `ResourceQuotasStatus` method

components/ui-api-layer/internal/domain/k8s/resourcequotastatus_resolver_test.go
```diff
@@ -0,0 +1,37 @@
...
+func TestResourceQuotaStatusResolver_ResourceQuotaStatus_Error(t *testing.T) {
+	// GIVEN
+	statusChecker := automock.NewResourceQuotaStatusChecker()
+	statusChecker.On("CheckResourceQuotaStatus", fixNamespaceName()).Return(gqlschema.ResourceQuotasStatus{}, errors.New("something went wrong"))
+
+	resolver := newResourceQuotaStatusResolver(statusChecker)
+
+	// WHEN
+	status, err := resolver.ResourceQuotasStatus(context.Background(), fixNamespaceName())
+	require.Error(t, err)
```
mszostok: missing // THEN

components/ui-api-layer/internal/domain/k8s/resourcequotastatus_service.go
```diff
@@ -23,81 +21,90 @@ type statefulSetLister interface {
...
-		deployKind = "Deployment"
-	)
+const (
+	rsKind  = "ReplicaSet"
+	stsKind = "StatefulSet"
+)
 
-	// You exceeded the ResourceQuota if at least one of the `.status.used` values equals or is bigger than its equivalent from the `.spec.hard` value.
+func (svc *resourceQuotaStatusService) CheckResourceQuotaStatus(environment string) (gqlschema.ResourceQuotasStatus, error) {
+	resourcesToCheck := []v1.ResourceName{
```
mszostok: I'm still not convinced to it but I can live with it

PR #566, opened 2018-09-07
Switch is local env to generic overrides

strekm:
**Description**

- Switch is local env to generic overrides


components/installer/pkg/overrides/provider.go
```diff
@@ -217,8 +219,18 @@ func NewLegacyProvider(overrideData OverrideData, installationData *config.Insta
 	}
 }
 
-func getStaticFileOverrides(installationData *config.InstallationData, chartDir string) StaticFile {
-	if installationData.IsLocalInstallation {
+func (p legacyProvider) getStaticFileOverrides(overrides Map, chartDir string) StaticFile {
+	isLocalEnv := FindOverrideValue(overrides, "global.isLocalEnv")
+	strIsLocalEnv := fmt.Sprintf("%v", isLocalEnv)
```
Tomasz-Smelcerz-SAP: Why not just use type assertion here?

Issue #453, opened 2018-09-04
Update Service-Catalog to v0.1.31 


aszecowka:
We create some issues for Service-Catalog:
https://github.com/kubernetes-incubator/service-catalog/issues/2306
https://github.com/kubernetes-incubator/service-catalog/issues/2291
https://github.com/kubernetes-incubator/service-catalog/issues/2276

and PRs:
https://github.com/kubernetes-incubator/service-catalog/pull/2333
https://github.com/kubernetes-incubator/service-catalog/pull/2337

Update ServiceCatalog if new version released, additionally:
- check if above issues have been solved and released

mszostok:
#### Status on Sept, 11:
- the newest release: **v0.1.31**
- only the defaultRelistInterval was fixed for ns scoped brokers ([issue]( https://github.com/kubernetes-incubator/service-catalog/issues/2306)) and available in the newest release

PR #623, opened 2018-09-11
Update the ServiceCatalog Helm Chart to the 0.1.31 version

mszostok:
**Description**

- Update the Service Catalog Helm Chart to the newest [v0.1.31 version](https://github.com/kubernetes-incubator/service-catalog/tree/v0.1.31)

**Related issue(s)**
- https://github.com/kyma-project/kyma/issues/453

Tested locally and via CI

### DIFF
--- a/resources/core/charts/service-catalog/charts/catalog/values.yaml
+++ b/resources/core/charts/service-catalog/charts/catalog/values.yaml
@@ -1,6 +1,6 @@
 # Default values for Service Catalog
 # service-catalog image to use
-image: quay.io/kubernetes-service-catalog/service-catalog:v0.1.28
+image: quay.io/kubernetes-service-catalog/service-catalog:v0.1.31
 # imagePullPolicy for the service-catalog; valid values are "IfNotPresent",
 # "Never", and "Always"
 imagePullPolicy: Always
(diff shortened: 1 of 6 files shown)


resources/core/charts/service-catalog/charts/catalog/values.yaml
```diff
@@ -80,6 +80,32 @@ apiserver:
       useEmbedded: false
       # etcd URL(s); override this if NOT using embedded etcd
       servers: https://core-service-catalog-etcd-client.kyma-system.svc.cluster.local:2379
+      # etcd image to use
```
aszecowka: if we are not using embedded etcd, does it make sense to specify those values here? It can be a misleading a little bit. On the other hand, it is easier to update service-catalog chart; but in such case, we should add a short comment that those values refers only to embedded etcd.  If we want to keep this values, we should probably copy etcd-pvc.yaml from the original chart.

mszostok: I agree with you. IMO the more important is to have the easy update of the service-catalog chart because we will do that quite often. So I kept those values and also I added `etcd-pvc.yaml`.  Comment also added.

Issue #616, opened 2018-09-10
Orchestrator master and PR branches refactor


clebs:
The current Kyma orchestrator needs refactoring to accommodate the release process that will come in later.
Specifically the following needs to be refactored:
- Master images will now be pushed to the `develop` folder in the registry
- Master images will now have a short commit hash as version
- PR images will now be pushed to the `pr` folder in the registry
- Version overrides now support registry directories

PR #624, opened 2018-09-11
Refactor pipeline to define versions and docker images in a release compatible way

clebs:
**Description**

- Version overrides now support registry directories
- Master images will now be pushed to the `develop` folder in the registry
- Master images will now have a short commit hash as version
- PR images will now be pushed to the `pr` folder in the registry

**Related issue(s)**
Implements #616

### DIFF
--- a/components/binding-usage-controller/Jenkinsfile
+++ b/components/binding-usage-controller/Jenkinsfile
@@ -5,9 +5,9 @@ def application = "binding-usage-controller"
 
 def isMaster = params.GIT_BRANCH == 'master'
 
-def dockerPushRoot = isMaster 
-    ? "${env.DOCKER_REGISTRY}"
-    : "${env.DOCKER_REGISTRY}snapshot/"
+def dockerPushRoot = isMaster
+    ? "${env.DOCKER_REGISTRY}develop/"
+    : "${env.DOCKER_REGISTRY}pr/"
 
 def dockerImageTag = isMaster
     ? params.APP_VERSION
--- a/components/binding-usage-controller/Jenkinsfile
+++ b/components/binding-usage-controller/Jenkinsfile
@@ -2,26 +2,20 @@
 
 def label = "kyma-${UUID.randomUUID().toString()}"
 def application = "binding-usage-controller"
-
-def isMaster = params.GIT_BRANCH == 'master'
-
-def dockerPushRoot = isMaster
-    ? "${env.DOCKER_REGISTRY}develop/"
-    : "${env.DOCKER_REGISTRY}pr/"
-
-def dockerImageTag = isMaster
-    ? params.APP_VERSION
-    : params.GIT_BRANCH
+def dockerPushRoot = "${env.DOCKER_REGISTRY}${params.PUSH_DIR}"
+def dockerImageTag = params.APP_VERSION
 
 echo """
 ********************************
 Job started with the following parameters:
 DOCKER_REGISTRY=${env.DOCKER_REGISTRY}
+PUSH_DIR=${params.PUSH_DIR}
 DOCKER_CREDENTIALS=${env.DOCKER_CREDENTIALS}
 GIT_REVISION=${params.GIT_REVISION}
 GIT_BRANCH=${params.GIT_BRANCH}
 APP_VERSION=${params.APP_VERSION}
 APP_FOLDER=${env.APP_FOLDER}
+FULL_BUILD=${params.FULL_BUILD}
 ********************************
 """
 
@@ -47,7 +41,7 @@ podTemplate(label: label) {
                             execute "./before-commit.sh ci"
                         }
 
-                        if (isMaster) {
+                        if (params.FULL_BUILD) {
                             stage("IP scan $application (Sourceclear)"){
                                 withCredentials([string(credentialsId: 'SRCCLR_API_TOKEN', variable: 'SRCCLR_API_TOKEN')]) {
                                     execute("make scan","SRCCLR_API_TOKEN=$SRCCLR_API_TOKEN")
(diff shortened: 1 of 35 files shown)


components/binding-usage-controller/Jenkinsfile
```diff
@@ -5,9 +5,9 @@ def application = "binding-usage-controller"
 
 def isMaster = params.GIT_BRANCH == 'master'
 
-def dockerPushRoot = isMaster 
-    ? "${env.DOCKER_REGISTRY}"
-    : "${env.DOCKER_REGISTRY}snapshot/"
+def dockerPushRoot = isMaster
```
aszecowka: if PR images end up in pr directory, why "master" images are not placed in the "master" directory? :)

clebs: Well, it is debatable, but registry folders may not map one to one with branches. It is called develop because it contains development builds. For now only master builds target that dir, but that might change.

I will present this in the SIG and we can discuss it further if needed.

orchestrator.Jenkinsfile
```diff
@@ -278,38 +280,67 @@ String projectVersion(project) {
...
-      gateway-tests: ${versions['tests/gateway-tests']}
-      test-environments: ${versions['tests/test-environments']}
-      kubeless-test-client: ${versions['tests/kubeless-test-client']}
-      api-controller-acceptance-tests: ${versions['tests/api-controller-acceptance-tests']}
-      connector-service-tests: ${versions['tests/connector-service-tests']}
-      metadata-service-tests: ${versions['tests/metadata-service-tests']}
-      event-bus-tests: ${versions['tests/event-bus']}
-
+"""
+global.docs.version: ${versions['docs']}
```
strekm: instead of ": " it should be "=". in generate-*-config.sh we create configmap (to be precise: configmap out of env file) out of it and kubernetes will complain.

clebs: Yes that is a mistake, I still was with the YAML map in mind.
Thanks for spotting that out! Fixed now.

orchestrator.Jenkinsfile
```diff
@@ -278,38 +280,63 @@ String projectVersion(project) {
 @NonCPS
 def versionsYaml(versions) {
```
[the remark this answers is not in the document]
piotrmsc: @clebs  I think we're missing dex as well?

orchestrator.Jenkinsfile
```diff
@@ -280,40 +301,69 @@ String projectVersion(project) {
...
+global.api-controller-acceptance-tests.version=${versions['tests/api-controller-acceptance-tests']}
+global.api-controller-acceptance-tests.dir=${versions['tests/api-controller-acceptance-tests'] == env.BRANCH_NAME ? 'pr' : 'develop'}
+global.connector-service-tests.version=${versions['tests/connector-service-tests']}
+global.connector-service-tests.dir=${versions['tests/connector-service-tests'] == env.BRANCH_NAME ? 'pr' : 'develop'}
+global.metadata-service-tests.version=${versions['tests/metadata-service-tests']}
+global.metadata-service-tests.dir=${versions['tests/metadata-service-tests'] == env.BRANCH_NAME ? 'pr' : 'develop'}
+global.event-bus-tests.version=${versions['tests/event-bus']}
+global.event-bus-tests.dir=${versions['tests/event-bus'] == env.BRANCH_NAME ? 'pr' : 'develop'}
+global.test-logging.version=${versions['tests/logging']}
+global.test-logging.dir=${versions['test-logging'] == env.BRANCH_NAME ? 'pr' : 'develop'}
```
rakesh-garimella: Not sure if `versions['test-logging']` if this is right

orchestrator.Jenkinsfile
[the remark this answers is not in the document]
clebs: The list of components was taken from master, and `dex` is not there. The goal of this PR is to refactor the CI to acommodate the upcoming release process, not adding new components.

If there are other components to be added, please ask the responsible team to do so in a separate PR.

orchestrator.Jenkinsfile
clebs: Yes, that was a mistake...
Thanks for catching that one! Fixed.

resources/core/charts/environments/templates/4-deployment.yaml
```diff
@@ -22,7 +22,11 @@ spec:
       serviceAccountName: {{ template "name" . }}-service-account
       containers:
         - name: {{ .Chart.Name }}
+        {{ if .Values.global.environments.dir }}
```
strekm: can we implement it in cleaner way? To build dir variable in a way that this if else statement won't be needed here: e.g. `dir="develop/"` so in chart we would have something like: `image: {{ .Values.global.containerRegistry.path }}/{{ .Values.global.environments.dir }}environments:{{ .Values.global.environments.version }}` and we would handle `dir=""` in the same way.

components/ui-api-layer/Jenkinsfile
```diff
@@ -49,7 +44,7 @@ podTemplate(label: label) {
                             execute("./before-commit.sh ci")
                         }
 
-                        if (isMaster) {
+                        if (params.GIT_BRANCH == 'master') {
```
michal-hudy: According to other `Jenkinsfiles` it should be `params.FULL_BUILD`.

components/ui-api-layer/Jenkinsfile
```diff
@@ -66,7 +61,7 @@ podTemplate(label: label) {
                         stage("push image $application") {
                             sh "docker tag ${dockerPushRoot}${application}:latest ${dockerPushRoot}${application}:${dockerImageTag}"
                             sh "docker push ${dockerPushRoot}${application}:${dockerImageTag}"
-                            if (isMaster) {
+                            if (params.FULL_BUILD) {
```
michal-hudy: According to other `Jenkinsfiles` it should be `params.GIT_BRANCH == 'master'`.

resources/core/charts/environments/templates/4-deployment.yaml
clebs: Applied the change, please take another look.
Thanks!

components/ui-api-layer/Jenkinsfile
clebs: Yes, that was a mistake, thanks for catching it!
Fixed.

PR #627, opened 2018-09-11
Delete password schema field from redis bundle

polskikiel:
**Description**

- Deleted password schema's field from redis bundle, it will be autogenerated now.

Part of kyma-project/console#86


polskikiel:
Partially resolves kyma-project/console#86

PR #679, opened 2018-09-13
Create configmap from versions.env file

### LINKED ISSUE #650: Create overrides config-map from versions.env file
AC:
- ConfigMap with overrides should be created based on versions.env file

strekm:
Resolves #650

### DIFF
--- a/installation/scripts/generate-local-config.ps1
+++ b/installation/scripts/generate-local-config.ps1
@@ -3,7 +3,7 @@ $CURRENT_DIR = Split-Path $MyInvocation.MyCommand.Path
 $CONFIG_TPL_PATH = "${CURRENT_DIR}\..\resources\installer-config-local.yaml.tpl"
 $CONFIG_OUTPUT_PATH = (New-TemporaryFile).FullName
 
-$VERSIONS_FILE_PATH="${CURRENT_DIR}\..\versions.env"
+$VERSIONS_FILE_PATH = "${CURRENT_DIR}\..\versions.env"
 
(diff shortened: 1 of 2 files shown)


installation/scripts/generate-local-config.ps1
```diff
@@ -3,6 +3,8 @@ $CURRENT_DIR = Split-Path $MyInvocation.MyCommand.Path
 $CONFIG_TPL_PATH = "${CURRENT_DIR}\..\resources\installer-config-local.yaml.tpl"
 $CONFIG_OUTPUT_PATH = (New-TemporaryFile).FullName
 
+$VERSIONS_FILE_PATH="${CURRENT_DIR}\..\versions.env"
```
crabtree: Missing spaces around =

installation/scripts/generate-local-config.ps1
```diff
@@ -24,6 +26,18 @@ Invoke-Expression -Command $cmd
 
 ##########
 
+Write-Output "Configuring versions ..."
+
+if([System.IO.File]::Exists($VERSIONS_FILE_PATH)){
+    $cmd = "kubectl create configmap versions --from-env-file=${VERSIONS_FILE_PATH} -n `"kyma-installer`""
```
crabtree: Please make the output of PS script similar to sh

installation/scripts/generate-local-config.ps1
strekm: done

installation/scripts/generate-local-config.ps1
strekm: in sh shouldn't be that echo instead

Issue #356, opened 2018-08-30
Enable secure access to the Jaeger UI


abbi-gaurav:
Currently, the access to the Jaeger UI is only possible via `kubectl port-forward`

This is not a good developer experience. The access should be secure OOB in a seamless fashion.

 

AC:

* The Jaeger UI is accessed in a secure fashion using pluggable Identity Provider.
* The technical solution should not fail Security Threat Modelling.

abbi-gaurav:
Verified the UI access working securely on local and cluster.

PR #681, opened 2018-09-13
Secure Jaeger UI

radufa:
**Description**

- Add a "keycloack proxy" as a sidecar to Jaeger 
- Expose Jaeger UI only through its "keycloack" sidecar
- Extend "dex" to support a new static client for Jaeger UI integration 

**Related issue(s)**

Implements #356

### DIFF
--- a/resources/core/charts/jaeger/templates/virtualservice.yaml
+++ b/resources/core/charts/jaeger/templates/virtualservice.yaml
@@ -7,7 +7,7 @@ spec:
   hosts:
   - jaeger.{{ .Values.global.domainName }}
   gateways:
-  - core-{{ .Chart.Name }}-gateway
+  - {{ .Values.global.istio.gateway.name }}
   http:
   - match:
     route:
@@ -15,18 +15,3 @@ spec:
         host: core-{{ .Chart.Name }}
         port:
           number: {{ .Values.service.uiPort }}
----
-apiVersion: networking.istio.io/v1alpha3
-kind: Gateway
-metadata:
-  name: core-{{ .Chart.Name }}-gateway
-spec:
-  selector:
-    istio: ingressgateway
-  servers:
-  - port:
-      number: 80
-      name: http
-      protocol: HTTP
-    hosts:
-    - jaeger.{{ .Values.global.domainName }}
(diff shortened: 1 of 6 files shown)


resources/core/charts/jaeger/templates/virtualservice.yaml
```diff
@@ -8,14 +7,26 @@ spec:
   hosts:
   - jaeger.{{ .Values.global.domainName }}
   gateways:
-  - {{ .Values.global.istio.gateway.name }}
+  - core-{{ .Chart.Name }}-gateway
```
pbochynski: Why can't you use gateway defined for kyma? https://github.com/kyma-project/kyma/blob/master/resources/core/charts/gateway/templates/gateway.yaml

resources/core/charts/jaeger/templates/virtualservice.yaml
```diff
@@ -8,14 +7,26 @@ spec:
...
+---
+apiVersion: networking.istio.io/v1alpha3
+kind: Gateway
+metadata:
+  name: core-{{ .Chart.Name }}-gateway
+spec:
+  selector:
+    istio: ingressgateway
+  servers:
+  - port:
```
pbochynski: Why do you need http gateway?

resources/core/charts/jaeger/templates/virtualservice.yaml
radufa: We have defined a new “security context” only for Jaeger UI which is independent of the others already existing. Therefore, we have a new “static client” for dex, see the related PR.

Jaeger UI will be used by developers in addition to admins or support. Compromising it should not imply that the others are compromised too.

resources/core/charts/jaeger/templates/virtualservice.yaml
radufa: To be more restrictive in the reachability from outside. All the other ports exposed by Jaeger are available only from calls coming from services running inside of Kyma. We will define Istio authorization rules for them later (we have an issue for this) .

radufa: Changed

abbi-gaurav:
See also #356

PR #692, opened 2018-09-13
Consistent codes in connector-service (#539)

### LINKED ISSUE #539: Response codes should be consistent in the Connector Service
**Description**

Response codes should be consistent in the Connector Service:

- client-certs endpoint should return 403 on the wrong token
- client-certs endpoint should return 400 on wrong CSR
- messages sent to a user should be more specific and user-friendly. Counterexample: `illegal base64 data at input byte 5`

**Expected result**

- Proper response codes returned
- The acceptance tests updated

Maladie:
**Description**

- Resolves #539

**Related issue(s)**

### DIFF
--- a/components/connector-service/internal/certificates/certificates.go
+++ b/components/connector-service/internal/certificates/certificates.go
@@ -43,7 +43,7 @@ func (cu *certificateUtility) decodeBytesFromBase64(bytes []byte) (decodedData [
 	data := make([]byte, base64.StdEncoding.DecodedLen(len(bytes)))
 	_, err := base64.StdEncoding.Decode(data, bytes)
 	if err != nil {
-		return nil, apperrors.Internal("Error while decoding base64 bytes. Internal error from base64 library: %s", err)
+		return nil, apperrors.Internal("There was an error while parsing the base64 content. An incorrect value was provided.")
 	}
 
 	return data, nil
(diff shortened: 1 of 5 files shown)


components/connector-service/internal/certificates/certificates.go
```diff
@@ -43,7 +43,7 @@ func (cu *certificateUtility) decodeBytesFromBase64(bytes []byte) (decodedData [
 	data := make([]byte, base64.StdEncoding.DecodedLen(len(bytes)))
 	_, err := base64.StdEncoding.Decode(data, bytes)
 	if err != nil {
-		return nil, apperrors.Internal("Error while decoding base64 bytes: %s", err)
+		return nil, apperrors.Internal("Error while decoding base64 bytes. Internal error from base64 library: %s", err)
```
janmedrek: 'Internal error from base64 library' does not sound good, perhaps something like 'Invalid base64 string' would be better?

The same goes for latter occurrence in this file and the one in tests

docs/application-connector/docs/assets/connectorapi.yaml
```diff
@@ -97,6 +97,12 @@ paths:
             application/json:
               schema:
                 $ref: '#/components/schemas/csrResponse'
+        '400':
+          description: 'Invalid CSR.'
```
janmedrek: The CSR may be actually valid, it just does not meet our subject fields requirements, I would suggest doing something like 'Improper subject values in CSR', you may want to double-check it with @lszymik

docs/application-connector/docs/assets/connectorapi.yaml
```diff
@@ -97,6 +97,12 @@ paths:
...
+            content:
```
janmedrek: This indentation is wrong, it should be the on same level as description, you can see the error in Swagger Editor: https://editor.swagger.io/

janmedrek:
Also, please do sign CLA :)

PR #701, opened 2018-09-14
Change Redis bundle micro plan to used embedded storage

mszostok:
**Description**

- Change the Redis bundle micro plan to used embedded storage, that will make it more useful for test-case scenarios


mszostok:
@aszecowka Good point, added :)

PR #755, opened 2018-09-17
Generic overrides

### LINKED ISSUE #300: Documentation about generic overrides
Consider: default overrides, config map entries, secret entries.

Tomasz-Smelcerz-SAP:
**Description**

- Added documentation about providing overrides for Helm Charts to Installer

**Related issue(s)**
Resolves #300

### DIFF
--- a/docs/kyma/docs/037-gs-installation-overrides.md
+++ b/docs/kyma/docs/037-gs-installation-overrides.md
@@ -57,12 +57,13 @@ test:
         memory: "1Gi"
 ```
 
-If you want to override this value, for example to "2Gi", proceed as follows:
-- Create a ConfigMap in `kyma-installer` namespace, labelled with: `installer:overrides` (or reuse an existing one)
+To override this value, for example to "2Gi", proceed as follows:
+- Create a ConfigMap in the `kyma-installer` Namespace, labelled with: `installer:overrides` (or reuse an existing one).
 - Add an entry `test.acceptance.ui.requests.memory: 2Gi` to the map.
 
-Once the installation starts, Installer generates overrides based on the map entries and value of "2Gi" will be used instead of default "1Gi" from the chart `values.yaml` file.
-For values that should be kept in Secrets, just define a Secret object instead of ConfigMap with the same key and value. Don't forget to label the Secret with `installer:overrides` and remember that values in Secrets must be base64-encoded.
+Once the installation starts, the Installer generates overrides based on the map entries. The system uses the value of "2Gi" instead of the default "1Gi" from the chart `values.yaml` file.
+
+For overrides that the system should keep in Secrets, just define a Secret object instead of a ConfigMap with the same key and value. Don't forget to label the Secret with `installer:overrides`. Remember that values in Secrets must be base64-encoded. (TODO: same key and value, but base64...)
 
 
 #### Sub-chart overrides
--- a/docs/kyma/docs/037-gs-installation-overrides.md
+++ b/docs/kyma/docs/037-gs-installation-overrides.md
@@ -95,21 +95,25 @@ spec:
 	  - "--tokenExpirationMinutes={{ .Values.deployment.args.tokenExpirationMinutes }}"
 ```
 
-The following fragment of `values.yaml` file in `connector-service` chart defines default value for `tokenExpirationMinutes`:
+The following fragment of the `values.yaml` file in `connector-service` chart defines the default value for `tokenExpirationMinutes`:
+
 ```
 deployment:
   args:
     tokenExpirationMinutes: 60
 ```
 
-If you want to override this value, for example to "90", do as follows:
-- Create a ConfigMap in kyma-installer namespace, labelled with: `installer:overrides` (or reuse existing one)
+To override this value, such as to change "60 to "90", do the following:
+
+- Create a ConfigMap in the `kyma-installer` Namespace labeled with `installer:overrides` or reuse existing one.
 - Add an entry `application-connector.connector-service.deployment.args.tokenExpirationMinutes: 90` to the map.
 
-Notice that the user-provided override key is now composed from two parts:
-  - Chart "path" inside top-level `core` chart: **application-connector.connector-service**
-  - Original template value reference from the chart (without .Values. prefix): **deployment.args.tokenExpirationMinutes**
-Once the installation starts, Installer generates overrides based on the map entries and value of "90" will be used instead of default "60" from the chart `values.yaml` file.
+Notice that the user-provided override key now contains two parts:
+
+- The chart "path" inside the top-level `core` chart: `application-connector.connector-service`
+- The original template value reference from the chart without the .Values. prefix: `deployment.args.tokenExpirationMinutes`.
+
+Once the installation starts, the Installer generates overrides based on the map entries. The system uses the value of "90" instead of the default value of "60" from the `values.yaml` chart file.
 
 
 #### Global overrides
--- a/docs/kyma/docs/037-gs-installation-overrides.md
+++ b/docs/kyma/docs/037-gs-installation-overrides.md
@@ -10,7 +10,7 @@ This document describes how to configure Installer with override values for Helm
 ## Overview
 
 The Installer is a Kubernetes Operator that uses Helm to install Kyma components.
-Helm provides an *overrides* feature to customize installation of charts (e.g: to configure environment-specific thing).
+Helm provides an *overrides* feature to customize installation of charts (e.g: to configure environment-specific values).
 When using Installer for Kyma installation, users can't interact with Helm directly (installation is not an interactive process).
 To customize Kyma installation, Installer exposes a generic mechanism for configuring Helm overrides called **user-defined** overrides.
 
--- a/docs/kyma/docs/037-gs-installation-overrides.md
+++ b/docs/kyma/docs/037-gs-installation-overrides.md
@@ -39,16 +39,15 @@ Kyma uses component-specific overrides only for the installation of specific com
 
 #### Top-level charts overrides
 
-Overrides for top-level charts are straightforward. Just use template value from the chart (without leading ".Values." prefix) as the entry key in ConfigMap/Secret.
+Overrides for top-level charts are straightforward. Just use the template value from the chart (without leading ".Values." prefix) as the entry key in the ConfigMap or Secret.
 
 Example:
 
-There's a `core` top-level chart, that is installed by the Installer.
-In one of it's templates there's a following line with a value reference:
+The Installer uses a `core` top-level chart that contains a template with the following value reference:
 ```
 memory: {{ .Values.test.acceptance.ui.requests.memory }}
 ```
-In order to resolve this template, there's a default value for "test.acceptance.ui.requests.memory" in the `values.yaml` file of the chart.
+The chart's default value `test.acceptance.ui.requests.memory` in the `values.yaml` file resolves the template.
 The following fragment of `values.yaml` shows this definition:
 ```
 test:
--- a/docs/kyma/docs/037-gs-installation-overrides.md
+++ b/docs/kyma/docs/037-gs-installation-overrides.md
@@ -13,7 +13,7 @@ The Installer is a Kubernetes Operator that uses Helm to install Kyma components
 Helm provides an *overrides* feature to customize the installation of charts, such as to configure environment-specific values.
 When using Installer for Kyma installation, users can't interact with Helm directly. The installation is not an interactive process.
 
-To customize Kyma installation, Installer exposes a generic mechanism for configuring Helm overrides called **user-defined** overrides.
+To customize the Kyma installation, the Installer exposes a generic mechanism to configure Helm overrides called **user-defined** overrides.
 
 
 ## User-defined overrides
--- a/docs/kyma/docs/037-gs-installation-overrides.md
+++ b/docs/kyma/docs/037-gs-installation-overrides.md
@@ -1,0 +1,114 @@
+---
+title: Helm overrides for Kyma installation
+type: Getting Started
+---
+
+Kyma components are packaged as [Helm](https://github.com/helm/helm/tree/master/docs) charts and are installed by the [Installer](../../../components/installer/README.md).
+This document describes how to configure Installer with overrides for Helm [charts](https://github.com/helm/helm/blob/master/docs/charts.md).
+
+
+## Overview
+
+The Installer is a Kubernetes Operator that uses Helm to install Kyma components.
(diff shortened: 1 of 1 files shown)


docs/kyma/docs/037-gs-installation-overrides.md
```diff
@@ -0,0 +1,168 @@
...
+  - Original template value reference from the chart (without .Values. prefix): **deployment.args.tokenExpirationMinutes**
+Once the installation starts, Installer generates overrides based on the map entries and value of "90" will be used instead of default "60" from the chart `values.yaml` file.
+
+
+#### Global overrides
+
+Some important parameters are usually shared across the charts.
+Helm convention to provide these is by using `global` override key.
+
+For example, to define `global.domain` override, just use "global.domin" as the name of the key in ConfigMap/Secret for Installer.
```
strekm: typo: should be "domain" instead of "domin"

docs/kyma/docs/037-gs-installation-overrides.md
```diff
@@ -0,0 +1,168 @@
...
+      requests:
+        memory: "1Gi"
+```
+
+If you want to override this value, for example to "2Gi", proceed as follows:
+- Create a ConfigMap in `kyma-installer` namespace, labelled with: `installer:overrides` (or reuse an existing one)
+- Add an entry `test.acceptance.ui.requests.memory: 2Gi` to the map.
+
+Once the installation starts, Installer generates overrides based on the map entries and value of "2Gi" will be used instead of default "1Gi" from the chart `values.yaml` file.
+For values that should be kept in Secrets, just define a Secret object instead of ConfigMap with the same key and value. Don't forget to label the Secret with `installer:overrides` and remember that values in Secrets must be base64-encoded.
```
dpolitesap: For values that the system should keep in Secrets, just define a Secret object instead of a ConfigMap with the same key and value. Don't forget to label the Secret with `installer:overrides`. Remember that values in Secrets must be base64-encoded.

docs/kyma/docs/037-gs-installation-overrides.md
```diff
@@ -0,0 +1,168 @@
...
+
+
+#### Sub-chart overrides
+
+Overrides for sub charts follow the same convention as top-level charts, but require additional information about sub-chart location.
```
dpolitesap: Overrides for sub-charts follow the same convention as top-level charts. However, overrides require additional information about sub-chart location.

When a sub-chart contains the `values.yaml` file, the information about the chart location is not necessary because the chart and it's `values.yaml` file are on the same level in the directory hierarchy.

The situation is different when the Installer installs a chart with sub-charts.
All template values for a sub-chart must be prefixed with a sub-chart "path" that is relative to the top-level "parent" chart.

This is not an Installer-specific requirement. The same considerations apply when you provide overrides manually using the `helm` command-line tool.

docs/kyma/docs/037-gs-installation-overrides.md
```diff
@@ -0,0 +1,168 @@
...
+When a sub-chart contains `values.yaml` file, the information about chart location is not necessary because chart and it's `values.yaml` file are on the same level in directory hierarchy.
+The situation is different when the Installer installs a chart with sub-charts.
+All template values for a sub-chart must be prefixed with a sub-chart "path" (relative to top-level "parent" chart).
+This in not Installer-specific requirement, the same considerations apply when you provide overrides manually via `helm` command-line tool.
+
+Example:
```
dpolitesap: Here is an example.

docs/kyma/docs/037-gs-installation-overrides.md
```diff
@@ -0,0 +1,168 @@
...
+
+There's a `core` top-level chart, that is installed by the Installer.
```
dpolitesap: There's a `core` top-level chart, that the Installer installs.
There's an `application-connector` sub-chart in `core` with another nested sub-chart: `connector-service`.
In one of its templates there's a following fragment (shortened):

docs/kyma/docs/037-gs-installation-overrides.md
```diff
@@ -0,0 +1,168 @@
...
+  containers:
+  - name: {{ .Chart.Name }}
+	args:
+	  - "/connectorservice"
+	  - '--appName={{ .Chart.Name }}'
+	  - "--domainName={{ .Values.global.domainName }}"
+	  - "--tokenExpirationMinutes={{ .Values.deployment.args.tokenExpirationMinutes }}"
+```
+
+The following fragment of `values.yaml` file in `connector-service` chart defines default value for `tokenExpirationMinutes`:
```
dpolitesap: The following fragment of the `values.yaml` file in `connector-service` chart defines the default value for `tokenExpirationMinutes`:

docs/kyma/docs/037-gs-installation-overrides.md
```diff
@@ -0,0 +1,168 @@
...
+```
+deployment:
+  args:
+    tokenExpirationMinutes: 60
+```
+
+If you want to override this value, for example to "90", do as follows:
```
dpolitesap: To override this value, such as to change "60 to "90", do the following:
- Create a ConfigMap in the `kyma-installer` Namespace labeled with `installer:overrides` or reuse existing one.
- Add an entry `application-connector.connector-service.deployment.args.tokenExpirationMinutes: 90` to the map.

docs/kyma/docs/037-gs-installation-overrides.md
```diff
@@ -0,0 +1,168 @@
...
+- Create a ConfigMap in kyma-installer namespace, labelled with: `installer:overrides` (or reuse existing one)
+- Add an entry `application-connector.connector-service.deployment.args.tokenExpirationMinutes: 90` to the map.
+
+Notice that the user-provided override key is now composed from two parts:
```
dpolitesap: Notice that the user-provided override key now contains two parts:
  - The chart "path" inside the top-level `core` chart `application-connector.connector-service`
  - The original template value reference from the chart without the `.Values.` prefix  `deployment.args.tokenExpirationMinutes`.

Once the installation starts, the Installer generates overrides based on the map entries. The system uses the value of "90" instead of the default value of "60" from the `values.yaml` chart file.

docs/kyma/docs/037-gs-installation-overrides.md
```diff
@@ -0,0 +1,168 @@
...
+  - Chart "path" inside top-level `core` chart: **application-connector.connector-service**
+  - Original template value reference from the chart (without .Values. prefix): **deployment.args.tokenExpirationMinutes**
+Once the installation starts, Installer generates overrides based on the map entries and value of "90" will be used instead of default "60" from the chart `values.yaml` file.
+
+
+#### Global overrides
+
+Some important parameters are usually shared across the charts.
```
dpolitesap: There are several important parameters usually shared across the charts.
Helm convention to provide these requires the use of the `global` override key.
 For example, to define the `global.domain` override, just use `global.domain` as the name of the key in ConfigMap or Secret for the Installer.

docs/kyma/docs/037-gs-installation-overrides.md
dpolitesap: See previous comment above.

docs/kyma/docs/037-gs-installation-overrides.md
```diff
@@ -0,0 +1,168 @@
...
+Helm convention to provide these is by using `global` override key.
+
+For example, to define `global.domain` override, just use "global.domin" as the name of the key in ConfigMap/Secret for Installer.
+Once the installation starts, Installer is merging all map entries and collect all global entries together (under `global` top-level key) for the purpose of installation.
```
dpolitesap: Once the installation starts, the Installer merges all of the map entries and collects all of the global entries under the `global` top-level key to use for installation.

docs/kyma/docs/037-gs-installation-overrides.md
```diff
@@ -0,0 +1,168 @@
...
+
+
+## Merging and conflicting entries
+
+When installer encounters two overrides with the same key prefix, it tries to merge them.
```
dpolitesap: When the Installer encounters two overrides with the same key prefix, it tries to merge them.
If both of them represent a map (they have nested sub-keys), their nested keys are recursively merged.
If at least one of keys points to a final value, the Installer performs the merge in a non-deterministic order, so either one of the overrides is rendered in the final yaml file.

It is important to avoid overrides having the same keys for final values.

docs/kyma/docs/037-gs-installation-overrides.md
```diff
@@ -0,0 +1,168 @@
...
+If both of them represent a map (they have nested sub-keys), their nested keys are recursively merged.
+If at least one of keys points to a final value,  Installer performs the merge in non-deterministic order, so either one or the other override is rendered to the final YAML.
+Considering that, **avoid overrides having the same keys for final values**.
+
+### Example of non-conflicting merge:
+
+Having two overrides with a common *key prefix* ("a.b"):
```
dpolitesap: Two overrides with a common *key prefix* ("a.b"):

docs/kyma/docs/037-gs-installation-overrides.md
```diff
@@ -0,0 +1,168 @@
...
+
+```
+"a.b.c": "first"
+"a.b.d": "second"
+```
+
+Installer yields correct output:
```
dpolitesap: The Installer yields correct output:

docs/kyma/docs/037-gs-installation-overrides.md
```diff
@@ -0,0 +1,168 @@
...
+```
+a:
+  b:
+    c: first
+    d: second
+```
+
+### Example of conflicting merge:
+
+Having two overrides with *the same* key ("a.b"):
```
dpolitesap: Two overrides with *the same* key ("a.b"):

docs/kyma/docs/037-gs-installation-overrides.md
```diff
@@ -0,0 +1,168 @@
...
+
+```
+"a.b": "first"
+"a.b": "second"
+```
+
+Installer yields either:
```
dpolitesap: The Installer yields either:

docs/kyma/docs/037-gs-installation-overrides.md
```diff
@@ -0,0 +1,168 @@
...
+
+```
+a:
+  b: "second"
+```
+
+
+## Values and types
+
+Installer generally recognizes all override values as **strings** and internally render overrides to Helm as YAML stream with only **string** values.
```
dpolitesap: Installer generally recognizes all override values as strings. It internally renders overrides to Helm as a yaml stream with only string values.

There is one exception to this rule with respect to handling booleans:  

The system converts "true" or "false" strings that it encounters to a corresponding boolean value (true/false).

docs/kyma/docs/037-gs-installation-overrides.md
```diff
@@ -0,0 +1,172 @@
...
+
+Installer merges all overrides recursively into a single YAML stream and passes it to Helm during the Kyma installation/upgrade operation.
+
+
+### Common vs component overides
+
+The Installer looks for available overrides each time a component installation or update operation is due.
+Overrides for the component are composed from two sets: **common** overrides and **component-specific** overrides.
+
+Kyma uses common overrides for the installation of all components. Only ConfigMaps or Secrets marked with the installer:overrides Label contain the definition of overrides (TODO: Fix it).
```
Tomasz-Smelcerz-SAP: The intention of this sentence is that common overrides are labelled only with "installer:overrides" label, no other label is required. I think that this sentence does not quite reflect this condition.

docs/kyma/docs/037-gs-installation-overrides.md
```diff
@@ -0,0 +1,172 @@
...
+        memory: "1Gi"
+```
+
+To override this value, for example to "2Gi", proceed as follows:
+- Create a ConfigMap in the `kyma-installer` Namespace, labelled with: `installer:overrides` (or reuse an existing one).
+- Add an entry `test.acceptance.ui.requests.memory: 2Gi` to the map.
+
+Once the installation starts, the Installer generates overrides based on the map entries. The system uses the value of "2Gi" instead of the default "1Gi" from the chart `values.yaml` file.
+
+For overrides that the system should keep in Secrets, just define a Secret object instead of a ConfigMap with the same key and value. Don't forget to label the Secret with `installer:overrides`. Remember that values in Secrets must be base64-encoded. (TODO: same key and value, but base64...)
```
Tomasz-Smelcerz-SAP: There's a problem here. We should define a secret with the same key and value, but the value must be base64 so it's not the same, isn't it? Well, for the one who creates the object (Secret) it's different, because it's encoded base64. But for the _system_, for the software entities that will read this secret, this value _is_ the same, because Kubernetes automatically "unwraps" base64 encoding while reading Secrets.
We have to find a nice way to describe it. Perhaps this would be OK:
... just define a Secret object instead of a ConfigMap with the same key and a base64-encoded value.

docs/kyma/docs/037-gs-installation-overrides.md
```diff
@@ -0,0 +1,184 @@
...
+
+```
+a:
+  b: "second"
+```
+
+
+## Values and types
+
+Installer generally recognizes all override values as strings. It internally renders overrides to Helm as a YAML stream with only string values.
```
Tomasz-Smelcerz-SAP: I've written "YAML" on purpose (instead of "yaml") since this is the name of the data format and that's the convention used in technical documents (JSON is another example).

docs/kyma/docs/037-gs-installation-overrides.md
dpolitesap: Thanks for the clarification. Let's keep what you had before.  Otherwise you can use this:  

Kyma uses common overrides for the installation of all components. Common overrides (ConfigMaps or Secrets) have the label 'installer:overrides' and contain the definition of overrides. They require no additional label.

docs/kyma/docs/037-gs-installation-overrides.md
dpolitesap: I agree.  

For overrides that the system should keep in Secrets, just define a Secret object instead of a ConfigMap with the same key and a base64-encoded value. Be sure to label the Secret with `installer:overrides`.

docs/kyma/docs/037-gs-installation-overrides.md
dpolitesap: ok

pbochynski:
@tomekpapiernik Is there something missing to close this pull request?

PR #768, opened 2018-09-18
POC - Kyma charts bundled in installer

strekm:
**Description**

- POC - Kyma charts bundled in installer

### DIFF
--- a/components/installer/deploy/installer/kyma.Dockerfile
+++ b/components/installer/deploy/installer/kyma.Dockerfile
@@ -1,0 +1,5 @@
+FROM eu.gcr.io/kyma-project/installer:0.3.383
+
+LABEL source="git@github.com:kyma-project/kyma.git"
+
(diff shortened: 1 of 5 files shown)


components/installer/deploy/installer/kyma.Dockerfile
```diff
@@ -0,0 +1,5 @@
+FROM eu.gcr.io/kyma-project/installer:0.3.383
+
+LABEL source="git@github.com:kyma-project/kyma.git"
+
+COPY /kyma /kyma/injected
```
pbochynski: All sources are copied?

strekm: no, in our local scrips only resources and installation folders are copied. (in this POC i used local deployment for installer and kyma/resources and kyma/installation is already copied to installer folder)

Issue #602, opened 2018-09-10
Ui-api-layer tests fail on the cluster after DEX SCI integration


aszecowka:
**Description**
- CI has access to proper credentials that ui-api-layer should use, so just use them :)
- Make ui-api-layer configurable for credentials in use and best use by default those values https://github.com/kyma-project/kyma/blob/master/resources/core/values.yaml#L30
- There are different flows depending if you are getting token for a static user or some specific IDP, so make sure tests support both cases. Depending on a given condition that is already used here https://github.com/kyma-project/kyma/blob/master/tests/acceptance/dex/dex_test.go#L46. But of course do not skip tests on cluster

**Reason**
Ui-api-layer tests fail during stability checking because of wrong credentials

derberg:
After discussing with @PK85 we agreed that for now we can skip this test on stability check until we do not do stability check on static users. This is kind of dependent on #197 

We will come back to the ticket in 4 weeks once other important priorities are fulfilled.

derberg:
/remind me to check this issue in 4 weeks

derberg:
/remind me to check this issue in 4 weeks

derberg:
long term stability testing should not require testing kyma that is already integrated with some identity provider. This will no longer be a valid issue

PR #794, opened 2018-09-19
Skip UI API Acceptance Tests when SCI is enabled

michal-hudy:
**Description**

- Skipped UI API Acceptance Tests when SCI is enabled

**Related issue(s)**
See #602

### DIFF
--- a/tests/ui-api-layer-acceptance-tests/dex/dex.go
+++ b/tests/ui-api-layer-acceptance-tests/dex/dex.go
@@ -1,0 +1,28 @@
+package dex
+
+import (
+	"io/ioutil"
+	"log"
+	"os"
+	"strings"
+
+	"github.com/pkg/errors"
+)
+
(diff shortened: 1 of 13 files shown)


tests/ui-api-layer-acceptance-tests/dex/dex.go
```diff
@@ -0,0 +1,34 @@
...
+	case os.IsNotExist(err):
+		return ""
+	default:
+		exitOnError(err, "while reading DEX configuration file")
+	}
+
+	return string(data)
+}
+
+func exitOnError(err error, context string) {
```
pkosiec: I would like to see a different if statement:

func exitOnError(err error, context string) {
	if err == nil {
		return
	}
        log.Fatal(errors.Wrap(err, context))
}

also, do we need it at all, if we can inline it? 🙂

PR #796, opened 2018-09-19
Fix service catalog acceptance test

polskikiel:
**Description**

- Added retry pooling loop for the broker's service classes.

**Related issue(s)**

### DIFF
--- a/tests/acceptance/servicecatalog/helpers_test.go
+++ b/tests/acceptance/servicecatalog/helpers_test.go
@@ -17,7 +17,7 @@ import (
 	"k8s.io/client-go/rest"
 )
 
-func getServiceBrokerServices(t *testing.T, url string) []osb.Service {
+func getCatalogForBroker(t *testing.T, url string) []osb.Service {
 	config := osb.DefaultClientConfiguration()
 	config.URL = url
 
--- a/tests/acceptance/servicecatalog/servicecatalog_test.go
+++ b/tests/acceptance/servicecatalog/servicecatalog_test.go
@@ -1,20 +1,21 @@
 package servicecatalog_test
 
 import (
+	"fmt"
 	"os"
 	"testing"
 	"time"
 
+	"github.com/kyma-project/kyma/components/remote-environment-broker/pkg/apis/remoteenvironment/v1alpha1"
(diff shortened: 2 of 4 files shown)


tests/acceptance/servicecatalog/helpers_test.go
```diff
@@ -62,6 +62,24 @@ func awaitCatalogContainsServiceClasses(t *testing.T, timeout time.Duration, ser
 	}
 }
 
+// awaitBrokerServiceClass awaits for service broker's not empty catalog response.
```
aszecowka: can you try to use wait.AtMost function to be more consistent? Please remember that AtMost has a small bug, so ask @mszostok to copy fix from him. Additionally, maybe renaming will be required: wait.ForFuncAtMost?

polskikiel: Ok, I will sync with @mszostok and align this solution to use shared waiter.

tests/acceptance/servicecatalog/servicecatalog_test.go
```diff
@@ -56,11 +63,16 @@ func TestServiceCatalogContainsRemoteEnvironmentBrokerServiceClass(t *testing.T)
 
 		deleteClusterServiceClassesForRemoteEnvironment(t, re)
 	}()
-	brokerServices = getServiceBrokerServices(t, os.Getenv(remoteEnvironmentBrokerURLEnvName))
 
 	// then
-	assert.NotEmpty(t, brokerServices)
-	// timeout must be greater than the broker relistDuration
+	wait.ForFuncAtMost(t, func() error {
```
aszecowka: this test is a little bit different than for helm broker, here we create new RE and we should not only check length but if the list contains newly created RE.
Additionally, I would recommend renaming method: getServiceBrokerServices to getCatalogForBroker

tests/acceptance/servicecatalog/wait/wait.go
```diff
@@ -5,7 +5,7 @@ import (
 	"time"
 )
 
-func AtMost(t *testing.T, fn func() error, timeout time.Duration) {
+func ForFuncAtMost(t *testing.T, fn func() error, timeout time.Duration) {
```
aszecowka: this function has a flaw, I asked you to fixed it.

Issue #723, opened 2018-09-17
Refactor remote environment source functionality


PK85:
AC:
- Rename api.groups of RemoteEnvironment, EnvironmentMapping and EventActivation
  - Adjust domain of **remoteenvironments.applicationconnector.kyma-project.io**
  - Adjust domain of **eventactivations.applicationconnector.kyma-project.io**
  - Adjust domain of **environmentmappings.applicationconnector.kyma-project.io**
    - group name is adjusted in the CRD
    - all usage is adjusted
    - documentation and examples are adjusted
- RE changes:
  - Remove source attribute. Use *RemoteEnvironment name* for EventActivation creation as **sourceId** field
  - Add **name** and **description** attribute per service
- add per every service ability to provide labels and put them into *serviceClass metadata* (related issue, see: https://github.com/kyma-project/console/issues/127). They will be displayed via ui-api-layer by convention, just three fields will be displayed.
```
labels:
        //helmbroker case
        showcase: true
        //helmbroker case
        local: true
        //RE case
        connected-app: ec-prod
```
- documentation and examples are adjusted

PK85:
We will do one PR for all components changes, continue work with https://github.com/kyma-project/kyma/pull/715

mszostok:
- [ ] REB client and business logic update @mszostok
- [ ] REB acceptance test and example update @piotrmiskiewicz 
- [ ] UI-API-Layer update
- [ ] REB documentation update

PR #798, opened 2018-09-19
Update RemoteEnvironment and EventActivation queries after breaking changes

mszostok:
**Description**

- Update RemoteEnvironment and EventActivation queries after breaking changes

**Related issue(s)**
 - https://github.com/kyma-project/kyma/issues/723


components/ui-api-layer/internal/domain/remoteenvironment/re_converter.go
```diff
@@ -30,26 +30,21 @@ func (c *remoteEnvironmentConverter) ToGQL(in *v1alpha1.RemoteEnvironment) gqlsc
 	dto := gqlschema.RemoteEnvironment{
 		Name:        in.Name,
 		Description: in.Spec.Description,
-		Source: gqlschema.RemoteEnvironmentSource{
```
aszecowka: was it done on purpose, removing resource filed from graphql (and not including sourceId)

jesusreal: In adittion, shouldn't we also add `Labels JSON`?

components/ui-api-layer/internal/domain/remoteenvironment/re.go
```diff
@@ -47,13 +47,13 @@ func New(restConfig *rest.Config, reCfg Config, informerResyncPeriod time.Durati
 	}
 
 	informerFactory := externalversions.NewSharedInformerFactory(client, informerResyncPeriod)
-	remoteEnvironmentGroup := informerFactory.Remoteenvironment().V1alpha1()
+	remoteEnvironmentGroup := informerFactory.Applicationconnector().V1alpha1()
```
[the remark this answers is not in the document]
mszostok: right, fixed

components/ui-api-layer/internal/domain/remoteenvironment/re_converter.go
mszostok: It's done on purpose. Now only EventActivation has SourceId filed

mszostok:
@jesusreal but the question is who will use that? for now, we are not mapping all fields from RE to gql model. Only those which are used on UI. Labels will be used in case of the ServiceCatalog view, so labels should be available on ServiceClasses and this will be done by Wookiee team.

PR #808, opened 2018-09-19
Metadata service must return 404 when service is not existing

### LINKED ISSUE #785: Update of not existing service in Metadata service is causing the Internal Server Error.
**Description**

Update of not existing service in Metadata service is causing the Internal Server Error.
The `PUT` operation is returning the code `500` as soon as the service cannot be found.

**Expected result**

There should be a response code `404` - `Not Found`.

**Actual result**

The response is:
```
{
    "code": 500,
    "error": "failed to read service, service with ID 9c79154c-7188-487c-ad6a-9844f16ec95c not found"
}
```

**Steps to reproduce**

Execute the `PUT` call with an ID which doesn't exist.

```
curl -X PUT \
  https://gateway.wormhole.cluster.kyma.cx/ec-default/v1/metadata/services/9c79154c-7188-487c-ad6a-9844f16ec95c -d '{.....
```

lszymik:
**Description**

- Metadata service must return 404 when updating not existing service

**Related issue(s)**
Resolves #785


tests/metadata-service-tests/test/apitests/metadata_test.go
```diff
@@ -156,6 +156,42 @@ func TestApiMetadata(t *testing.T) {
...
+					DisplayName: "documentation name",
+					Description: "documentation description",
+					Type:        "documentation type",
+					Tags:        []string{"tag1", "tag2"},
+					Docs:        []testkit.DocsObject{{Title: "docs title", Type: "docs type", Source: "docs source"}},
+				},
+			}
+
+			// when
+			statusCode, _ := metadataServiceClient.UpdateService(t, "12345", updatedServiceDefinition)
```
Szymongib: Please add an error check:
```go
statusCode, err := metadataServiceClient.UpdateService(t, "12345", updatedServiceDefinition)

// then
require.NoError(t, err)
...
```

PR #863, opened 2018-09-21
Block broker deletion if there are mappings left

polskikiel:
**Description**

- ServiceBroker will be deleted only when there are no environment mappings left in the broker's namespace.

**Related issue(s)**


components/remote-environment-broker/internal/mapping/controller.go
```diff
@@ -45,6 +45,11 @@ type nsBrokerFacade interface {
 	Exist(destinationNs string) (bool, error)
 }
 
+//go:generate mockery -name=mappingSvc -output=automock -outpkg=automock -case=underscore
+type mappingSvc interface {
```
aszecowka: maybe mappingLister will be better name?

components/remote-environment-broker/internal/mapping/controller.go
```diff
@@ -189,7 +196,7 @@ func (c *Controller) processItem(key string) error {
 			return err
 		}
 		if c.manageNsBrokers {
-			if err := c.ensureNsBrokerNotRegistered(namespace); err != nil {
+			if err := c.ensureEnvMappings(namespace); err != nil {
```
aszecowka: I don't like this method name. On the basis of its name, what is your assumption about its responsibilities?
I would expect that "ensureEnvMapping" creates EnvironmentMapping.
My proposal: `ensureNsBrokerNotRegisteredIfNoMappings`, maybe you will find a better one

components/remote-environment-broker/internal/mapping/controller_test.go
```diff
@@ -147,24 +147,25 @@ func TestControllerRunFailure(t *testing.T) {
 }
 
 type tcNsBrokersEnabled struct {
-	name        string
-	prepareMock func() *automock.NsBrokerFacade
-	errorMsg    string
+	name           string
+	nsBrokerFacade func() *automock.NsBrokerFacade
```
aszecowka: maybe this methods should be called prepareNsBrokerFacade, prepareMappingSvc

components/remote-environment-broker/internal/mapping/controller_test.go
```diff
@@ -234,39 +235,102 @@ func TestControllerProcessItemOnEMDeletionWhenNsBrokersEnabled(t *testing.T) {
...
+						ObjectMeta: metav1.ObjectMeta{
+							Name: "fix",
+						},
+					},
+				}, nil)
+				return mappingSvc
+			},
+		},
+		{
+			name: "broker will be removed if there is no environment mappings in the namespace",
```
aszecowka: is it the same use case as a happy path?

components/remote-environment-broker/internal/mapping/mapping_service_test.go
```diff
@@ -0,0 +1,65 @@
...
+	svc := newMappingService(emInformer)
+
+	// WHEN
+	result, err := svc.ListEnvironmentMappings("prod")
+
+	// THEN
+	require.NoError(t, err)
+	assert.Equal(t, result[0], em1)
+	assert.Equal(t, result[1], em2)
+	assert.Len(t, result, 2)
```
aszecowka: 1. first, check len of result, then access via indexes. 
2. But it will be better to not make assumptions about the order of returned objects, so use Contains method instead.

components/remote-environment-broker/internal/mapping/mapping_service_test.go
```diff
@@ -0,0 +1,65 @@
+package mapping
+
+import (
+	"testing"
+	"time"
+
+	"github.com/kyma-project/kyma/components/remote-environment-broker/pkg/apis/applicationconnector/v1alpha1"
+	fake2 "github.com/kyma-project/kyma/components/remote-environment-broker/pkg/client/clientset/versioned/fake"
```
aszecowka: fake2?

Issue #707, opened 2018-09-14
Move authorization proxy to kyma components


pbochynski:
**Description**
Apiserver related changes were rejected by code owners of kube rbac proxy project (https://github.com/brancz/kube-rbac-proxy).
Therefore we have to move the code to kyma components.

**TODO**
- ensure the minimal requirements are fulfilled (pipeline, tests, etc)
- update authorization proxy proposal to enable selective authorization (avoid double authorization for calls to apiserver)
- mention in the documentation that the code is based on open source project

piotrmsc:
Status update : Docu PR created, PR with dynamic versioning created

PR #876, opened 2018-09-24
Adjust apiserver-proxy to allow k8s apiserver to be secured

piotrmsc:
**Description**
As with the decision to adjust kube-rbac-proxy with our needs for securing k8s apiserver, this PR serves all required code changes on top of kube-rbac-proxy.

- Code changes in the apiserver-proxy which allows k8s apiserver to be secured

**Related issue(s)**
See also #707 >

### DIFF
--- a/components/apiserver-proxy/cmd/proxy/main.go
+++ b/components/apiserver-proxy/cmd/proxy/main.go
@@ -1,22 +1,7 @@
-/*
-Copyright 2017 Frederic Branczyk All rights reserved.
-
-Licensed under the Apache License, Version 2.0 (the "License");
-you may not use this file except in compliance with the License.
-You may obtain a copy of the License at
-
-    http://www.apache.org/licenses/LICENSE-2.0
-
-Unless required by applicable law or agreed to in writing, software
-distributed under the License is distributed on an "AS IS" BASIS,
-WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
-See the License for the specific language governing permissions and
-limitations under the License.
-*/
-
 package main
 
-import (
+import (	
+	"strings"
 	"crypto/tls"
 	stdflag "flag"
 	"fmt"
@@ -28,8 +13,6 @@ import (
 	"os/signal"
 	"syscall"
 
-	"k8s.io/apiserver/pkg/authentication/authenticator"
-
 	"github.com/golang/glog"
 	"github.com/hkwi/h2c"
 	"github.com/kyma-project/kyma/components/apiserver-proxy/internal/authn"
@@ -37,6 +20,7 @@ import (
 	"github.com/kyma-project/kyma/components/apiserver-proxy/internal/proxy"
 	flag "github.com/spf13/pflag"
 	"golang.org/x/net/http2"
+	"k8s.io/apiserver/pkg/authentication/authenticator"
 	k8sapiflag "k8s.io/apiserver/pkg/util/flag"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/rest"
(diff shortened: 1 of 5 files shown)


components/apiserver-proxy/internal/proxy/proxy.go
```diff
@@ -9,13 +9,18 @@ import (
...
 	yaml "gopkg.in/yaml.v2"
+	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/apiserver/pkg/authentication/authenticator"
 	"k8s.io/apiserver/pkg/authentication/user"
 	"k8s.io/apiserver/pkg/authorization/authorizer"
+	apirequest "k8s.io/apiserver/pkg/endpoints/request"
 	clientset "k8s.io/client-go/kubernetes"
 )
 
+const KUBERNETES_SERVICE = "kubernetes.default"
```
sjanota: This constant seems to be unused. If so, please remove it.

components/apiserver-proxy/internal/proxy/proxy.go
```diff
@@ -201,3 +225,15 @@ func (c *Config) DeepCopy() *Config {
 
 	return res
 }
+
+func newRequestInfoResolver() *apirequest.RequestInfoFactory {
+	apiPrefixes := sets.NewString(strings.Trim("/apis", "/"))
```
sjanota: Why not simply:
```
apiPrefixes := sets.NewString("apis", "api")
legacyAPIPrefixes := sets.NewString("api")
```

components/apiserver-proxy/internal/proxy/proxy.go
piotrmsc: It is used in main.go

components/apiserver-proxy/internal/proxy/proxy.go
piotrmsc: yes, it was simplified

components/apiserver-proxy/cmd/proxy/main.go
```diff
@@ -162,21 +146,32 @@ func main() {
...
 	if err != nil {
 		glog.Fatalf("Failed to create rbac-proxy: %v", err)
 	}
 
-	proxy := httputil.NewSingleHostReverseProxy(upstreamURL)
+	proxyForApiserver := strings.Contains(cfg.upstream, proxy.KUBERNETES_SERVICE)
+
+	rp := httputil.NewSingleHostReverseProxy(upstreamURL)
+	
+	if proxyForApiserver {
```
sjanota: Are we going to use it in ant other way then as proxy for apiserver? If not I'd refrain from using conditions here and just hardcode this one case.

components/apiserver-proxy/cmd/proxy/main.go
```diff
@@ -221,11 +216,11 @@ func main() {
 	}
 
 	if cfg.insecureListenAddress != "" {
-		if cfg.upstreamForceH2C {
+		if cfg.upstreamForceH2C && !proxyForApiserver {
```
sjanota: Same here, if we are going to use it only for apiserver this whole if only clutters

components/apiserver-proxy/cmd/proxy/main.go
```diff
@@ -291,9 +286,10 @@ func main() {
 func initKubeConfig(kcLocation string) *rest.Config {
 
 	if kcLocation != "" {
+		println(kcLocation)
```
sjanota: Forgotten debug statement?

components/apiserver-proxy/cmd/proxy/main.go
piotrmsc: yes, it should be possible to use it for another app as well although here it is designed for apisever :)

components/apiserver-proxy/cmd/proxy/main.go
piotrmsc: like above

Issue #670, opened 2018-09-13
The Connector service should return the signed certificate together with root CA certificate


lszymik:
**Description**

The Connector service is working in that way, that the CSR is being processed and the signed client certificate is returned. However, some of the clients require the certificate chain to verify the certificate.

**Reasons**

The certificate validation should be possible.

**Details**

The Connector service response:
```
{
    "crt":"BASE64_ENCODED_CRT"
}
```

will contain the full chain base64 encoded. 

```
-----BEGIN CERTIFICATE-----
MIIESz...
-----END CERTIFICATE-----
-----BEGIN CERTIFICATE-----
MIIFUD...
-----END CERTIFICATE-----
```

**Acceptance criteria**
- The full certificate is returned
- No breaking changes

**Notes**
- Kindly please bear in mind that the order of the certificates matter. The server-side certificate must be attached after the client certificate. It must be double checked before merging to a production system.

- As soon as the implementation will be ready for a review, we need to deploy changes to a cluster and ask integration team for a verification. The manual verification was already done but without a check of the certificate upload validation.

PR #894, opened 2018-09-24
Connector service returns signed cert together with root CA cert (#670)

Maladie:
**Description**

- Connector service now returns root CA cert followed by signed client cert (#670)

**Related issue(s)**


components/connector-service/internal/certificates/certificates.go
```diff
@@ -177,9 +177,17 @@ func (cu *certificateUtility) SignWithCA(caCrt *x509.Certificate, csr *x509.Cert
...
 	}
 
-	clientCrt := &bytes.Buffer{}
+	serverCrt := pemEncodeCrt(caCrt.Raw)
 
-	pem.Encode(clientCrt, &pem.Block{Type: "CERTIFICATE", Bytes: clientCrtRaw})
+	clientCrt := pemEncodeCrt(clientCrtRaw)
 
-	return cu.encodeStringBase64(clientCrt.Bytes()), nil
+	certChain := append(clientCrt.Bytes(), serverCrt.Bytes()...)
```
janmedrek: The good idea would be to extract encoding/appending to separate function that would accept two certificates, it could return whole chain as string/byte array. I am thinking about something like: 
`createCertChain(caCrt, clientCrt []byte) string`

You could also do the same treatment to the template creation in earlier part of this func:
`prepareCertificateTemplate(caCrt *x509.Certificate, csr *x509.CertificateRequest) x509.Certificate`

After this, the SignWithCA func would look something like this:
```
func (cu *certificateUtility) SignWithCA(caCrt *x509.Certificate, csr *x509.CertificateRequest, key *rsa.PrivateKey) (
	crtBase64 string, appError apperrors.AppError) {

	clientCRTTemplate := prepareCertificateTemplate(caCrt, csr)

	clientCrtRaw, err := x509.CreateCertificate(rand.Reader, &clientCRTTemplate, caCrt, csr.PublicKey, key)
	if err != nil {
		return "", apperrors.Internal("Error while creating certificate: %s", err)
	}

        certChain := createCertChain(caCrt.Raw, clientCrtRaw)
	
	return cu.encodeStringBase64(certChain), nil
}
```

components/connector-service/internal/certificates/certificates_test.go
```diff
@@ -473,22 +519,75 @@ func TestCertificateUtility_CheckCSRValues(t *testing.T) {
...
+		caCrt, csr, key := prepareCrtAndKey(certificateUtility)
+
+		// when
+		crtBase64, err := certificateUtility.CreateCrtChain(caCrt, csr, key)
+
+		// then
+		rawCrt, decodeErr := base64.StdEncoding.DecodeString(crtBase64)
+		decodedCrt, singleCertErr := rawCrtTox509Certificates(rawCrt)
+
+		require.NoError(t, singleCertErr)
```
lszymik: This error checks should be in different order. If there will be `err` not nil, then following command will fail. Please verify it with team.

lszymik: It should be tested like that:
```
when
then

when
then

when
then
```

components/connector-service/internal/certificates/certificates_test.go
```diff
@@ -473,22 +519,75 @@ func TestCertificateUtility_CheckCSRValues(t *testing.T) {
...
 
 		// when
-		crtBase64, err := certificateUtility.SignWithCA(caCrt, csr, key)
+		crtBase64, err := certificateUtility.CreateCrtChain(caCrt, csr, key)
 
 		// then
+		rawCrt, decodeErr := base64.StdEncoding.DecodeString(crtBase64)
+		decodedCrt, singleCertErr := rawCrtTox509Certificates(rawCrt)
+
+		require.NoError(t, singleCertErr)
```
lszymik: The tests should be ordered. See the previous comment.

components/connector-service/internal/certificates/certificates_test.go
lszymik: Pushed.

components/connector-service/internal/certificates/certificates_test.go
lszymik: fixed

Issue #797, opened 2018-09-19
Enable Dynamic Versioning for application-connector


Disper:
Description

Example: https://github.com/kyma-project/kyma/blob/master/resources/core/charts/environments/templates/4-deployment.yaml (line 25)

PR #985, opened 2018-09-27
Enable dynamic versioning

Szymongib:
**Description**

- Enable dynamic versioning if Connector Service, Metadata Service, Remote Env Controller and Gateway

**Related issue(s)**
#797


kubadz:
You should also add your tests to https://github.com/kyma-project/kyma/blob/master/release.Jenkinsfile similarily to how it is done in orchestrator jenkinsfile.

PR #1036, opened 2018-10-01
Make all TLS certs secrets

sjanota:
**Description**

- [x] All TLS certificates in kyma should be installed as secrets
- [x] Prepare procedure for certificate update
- [x] Move procedure to security docs when #1219 is merged
- [x] Check if private keys are required in kyma-system and remove them from secrets if not
- [x] Test RE on cluster
- [x] Check if procedure for RE is accurate. Adapt if not

**Related issue(s)**


resources/core/charts/apiserver-proxy/templates/deployment.yaml
```diff
@@ -42,7 +42,7 @@ spec:
         - "--v=10"
         - "--oidc-issuer=https://dex.{{ .Values.global.domainName }}"
         - "--oidc-clientID=kyma-client"
-        - "--oidc-ca-file=/etc/{{ .Values.configmapName}}/ca.crt"
+        - "--oidc-ca-file=/etc/apiserver-proxy-tls-cert/tls.crt"
         ports:
         - containerPort: {{ .Values.containerPort}}
         volumeMounts:
```
kubadz: You need to use secret here as you did in configurations-generator. You should also delete that configmap from values if you are not using it.

piotrmsc:
Like we have discussed, please create one secret with tls in the kyma-system ns and reference it in the required components deployments.

piotrmsc:
In the remote-environment controller please create one secret with the same as in the kyma-system and reference it in the values.yaml of the controller https://github.com/kyma-project/kyma/blob/master/resources/remote-environments/values.yaml#L33

Please also label components which are using tls certificate, it will help with update procedure.

resources/remote-environments/values.yaml
```diff
@@ -30,7 +30,7 @@ service:
 ingress:
   enabled: true
   authSecret: kyma-integration/nginx-auth-ca
-  tlsSecret: nginx-ingress-cert
+  tlsSecret: ingress-tls-cert
```
Szymongib: Same as above - needs to be changed also in Remote Environment Controller.

docs/security/docs/020-update-tls-certificates.md
```diff
@@ -0,0 +1,86 @@
+---
+title: Update TLS certificate
+type: Details
+---
+
+The TLS certificate is a vital security element. This document describes how to update the TLS certificate in Kyma.
+
+>**NOTE:** This procedure can interrupt the communication between your cluster and the outside world for a limited
```
pbochynski: How can we avoid that? What about rolling update?

sjanota: Rolling update should do the trick, but we have only one instance of ingress gateway by default. We would need to increase size of ingress gateway replica set.  It won't work on minikube as we're using hostPort.

sjanota: Rolling update should do the trick, but we have only one instance of ingress gateway by default. We would need to increase size of ingress gateway replica set.  It won't work on minikube as we're using hostPort.

Also istio will be adding hot-reloading of certificates in Q4, as here: https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!msg/istio-users/dyBMP0OlMBY/70MTj8kzBwAJ.

sjanota: Rolling update should do the trick, but we have only one instance of ingress gateway by default. We would need to increase size of ingress gateway replica set.  It won't work on minikube as we're using hostPort.

Also istio will be adding hot-reloading of certificates in Q4, as pointed here: https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!msg/istio-users/dyBMP0OlMBY/70MTj8kzBwAJ.

sjanota: Rolling update should do the trick, but we have only one instance of ingress gateway by default. We would need to increase size of ingress gateway replica set.  It won't work on minikube as we're using hostPort.

Also istio will be adding hot-reloading of certificates in Q4, as pointed here: https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!msg/istio-users/dyBMP0OlMBY/70MTj8kzBwAJ.

sjanota: Follow up: #1372

sjanota: Follow up: #1372

sjanota: Follow up: #1372

Issue #1053, opened 2018-10-01
core-logging pods have memory limit set to 400Mi and 1Gi in local installation


PK85:
Without load logging consumes 31Mi and 15Mi, see:

```
kubectl top pods -n kyma-system
NAME                                        CPU(cores)   MEMORY(bytes)
...
core-logging-logspout-szkhf                              7m           31Mi
core-logging-oklog-0                                     65m          15Mi
...
```

Do we need to allocate 18% of whole kyma local memory in minkube installation (8GB)?
```kubectl describe nodes
  Namespace                  Name                                                               CPU Requests  CPU Limits  Memory Requests  Memory Limits
  ---------                  ----                                                               ------------  ----------  ---------------  -------------
  kyma-system                core-logging-logspout-szkhf                                        125m (3%)     0 (0%)      96Mi (1%)        400Mi (5%)
  kyma-system                core-logging-oklog-0                                               125m (3%)     0 (0%)      96Mi (1%)        1Gi (13%)
```

if this is production value, please adjust local settings (isLocalEnv), try first with memory request 128Mi and limit 256Mi. If possible please set less memory allocation.

Issue #1052, opened 2018-10-01
prometheus-core-0 pod has memory limit set to 2098Mi in local installation


PK85:
prometheus-core-0 pod allocates locally 2098Mi which is 26% of whole kyma local minkube installation (8GB)?
```kubectl describe nodes
  Namespace                  Name                                                               CPU Requests  CPU Limits  Memory Requests  Memory Limits
  ---------                  ----                                                               ------------  ----------  ---------------  -------------
  kyma-system                prometheus-core-0                                                  10m (0%)      10m (0%)    250Mi (3%)       2098Mi (26%)
```

if this is production value, please adjust local settings (isLocalEnv), try first with memory request 128Mi and limit 256Mi.

Losing data locally is acceptable, please let me know if you see other options.

PR #1067, opened 2018-10-02
Adjust local settings, memory resources for Prometheus and Logging

venturasr:
**Related issue(s)**
https://github.com/kyma-project/kyma/issues/1053
https://github.com/kyma-project/kyma/issues/1052

### DIFF
--- a/resources/core/charts/logging/values.yaml
+++ b/resources/core/charts/logging/values.yaml
@@ -43,6 +43,13 @@ oklog:
     limits:
       memory: 1Gi
 
+isLocalEnv:
+  resources:
+    requests:
+      cpu: 125m
+      memory: 128Mi
+    limits:
+      memory: 256Mi
 
 # Enable to add load balanced service.
 # Example annotation will create AWS internal LoadBalancer.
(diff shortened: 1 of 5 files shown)


resources/core/charts/logging/values.yaml
```diff
@@ -43,6 +43,13 @@ oklog:
     limits:
       memory: 1Gi
 
+isLocalEnv:
```
PK85: reduce also second component that takes 400

resources/core/charts/monitoring/charts/prometheus/values.yaml
```diff
@@ -128,6 +128,13 @@ resources:
   limits:
     memory: 2Gi
 
+isLocalEnv:
+  resources:
+    requests:
+      memory: 256Mi
+    limits:
+      memory: 1Gi
```
PK85: have you checked 512Mi?

resources/core/charts/monitoring/charts/prometheus/values.yaml
```diff
@@ -128,6 +128,13 @@ resources:
   limits:
     memory: 2Gi
 
+isLocalEnv:
```
PK85: for that isLocalEnv local variable, please use installer local and cluster overrides

PR #1072, opened 2018-10-02
Adjust memory configuration for helm test

aszecowka:
Adjust memory configuration for helm test


resources/core/templates/tests/test-ui-api-acceptance.yaml
```diff
@@ -24,6 +24,9 @@ spec:
           value: "{{ .Values.test.auth.password }}"
         - name: DEX_CONFIGURATION_FILE
           value: /etc/dex/cfg/config.yaml
+      resources:
```
aszecowka: looking at grafana I noticed that we are really close to the limit (almost 90Mi when the limit is 96Mi) so I decided to make an exception and it increase here.

PR #1085, opened 2018-10-02
Add istio-kyma-patch back

sjanota:
**Description**

Add istio-kyma-patch again. Fixed this time.

**Related issue(s)**


sjanota:
Cluster tests are green

piotrmsc:
LGTM, tests passed on a cluster for me as well. One minor thing to cleanup is prometheus clusterrole,binding and configmap

components/istio-kyma-patch/patch.sh
```diff
@@ -0,0 +1,98 @@
...
+  local configmap=$(kubectl -n istio-system get configmap istio-sidecar-injector -o jsonpath='{.data.config}')
+  local alreadyEnabled=$(grep "policy: enabled" <<< "$configmap")
+  if [[ -n ${alreadyEnabled} ]]; then
+    # Disable automatic injecting
+    configmap=$(sed 's/policy: enabled/policy: disabled/' <<< "$configmap")
+    configmap=$(sed 's/\[\[ \.ProxyConfig\.ZipkinAddress \]\]/zipkin.kyma-system:9411/g' <<< "$configmap")
+
+    # Set limits for sidecar. Our namespaces have resource quota set thus every container needs to have limits defined.
+    # Add limits to already existing resources sections
+    configmap=$(sed 's|    resources:|    resources:\'$'\n      limits: { memory: 128Mi, cpu: 100m }|' <<< "$configmap")
```
mszostok: why u are not specifying `requests` property?

resources/istio/values.yaml
```diff
@@ -394,11 +379,9 @@ pilot:
   traceSampling: 100.0
   # Resources for a small pilot install
   resources:
-    limits:
-      cpu: 500m
-      memory: 2Gi
     requests:
-      memory: 512Mi
+      cpu: 500m
+      memory: 2048Mi
```
mszostok: where u overriding that?

installation/resources/installer-config-local.yaml.tpl
```diff
@@ -49,14 +49,11 @@ data:
   global.alertTools.credentials.slack.channel: ""
   global.alertTools.credentials.victorOps.routingkey: ""
   global.alertTools.credentials.victorOps.apikey: ""
-  gateways.istio-ingressgateway.service.externalPublicIp: ""
-  gateways.istio-ingressgateway.type: "NodePort"
   nginx-ingress.controller.service.loadBalancerIP: ""
   configurations-generator.kubeConfig.clusterName: "kyma.local"
   cluster-users.users.adminGroup: ""
   pilot.resources.limits.memory: 256Mi
   pilot.resources.requests.memory: 128Mi
```
sjanota: Pilot resources were moved to istio-overrides. Remove them

resources/istio/values.yaml
sjanota: This is because we have included raw istio charts. This value is provided in default charts: https://github.com/istio/istio/blob/1.0.2/install/kubernetes/helm/istio/values.yaml#L383 . If This is overridden using  installer overrides for [minikube](https://github.com/kyma-project/kyma/blob/fde018349b43782c8e1b13e4d8ebc8cacf4baed3/installation/resources/installer-config-local.yaml.tpl#L98) and [cluster](https://github.com/kyma-project/kyma/blob/fde018349b43782c8e1b13e4d8ebc8cacf4baed3/installation/resources/installer-config-cluster.yaml.tpl#L100).

components/istio-kyma-patch/patch.sh
sjanota: We are setting limits here, because our namespaces have resource-quota. In such case every container needs to have limits set. Usually it is done by limit-range, but it does not apply to injected containers. Setting requests is not necessary so we skipped that part and used what they provided.

mszostok: "Setting requests is not necessary so we skipped (..)" but why it's not necessary. IMO you should because it will not be verbose enough and you rely on the internal implementation of k8s. Please set it as it was before.

sjanota: Previously `requests.memory` for istio-proxy was set to same value as `limits.memory` in [istio/values.yaml](https://github.com/kyma-project/kyma/blob/master/resources/istio/values.yaml#L36). According to [k8s docs](https://kubernetes.io/docs/tasks/administer-cluster/manage-resources/memory-default-namespace/#what-if-you-specify-a-container-s-limit-but-not-its-request) it is the same as not specifying requests at all. So, accidentally we have the same behaviour as previously :).

mszostok: so, as I said: "it will not be verbose enough"  and you can see my first comment, "By default, the request will be the same as the limit." but this requires additional knowledge and can be changed in future so you also rely on the k8s implementation.

what's more the request for [CPU](https://github.com/kyma-project/kyma/blob/master/resources/istio/values.yaml#L35) was different, so it's not the same as previously

sjanota: fixed

PR #1101, opened 2018-10-03
Give mixer more memory on cluster

sjanota:
**Description**

- Set mixer (istio-policy and istio-telemetry) memory limit to 1Gi on cluster

**Related issue(s)**


piotrmsc:
Please add also an override in the **local** installation for pilot memory: requests  96Mi and limits 256Mi

PR #1112, opened 2018-10-04
Reduce local installation to the minimum possible memory usage

PK85:
**Description**

- ...
- ...
- ...

**Related issue(s)**

### DIFF
--- a/resources/istio/templates/sidecar-injector-configmap.yaml
+++ b/resources/istio/templates/sidecar-injector-configmap.yaml
@@ -18,9 +18,9 @@ data:
       - name: istio-init
         resources:
           limits:
-            memory: 96Mi
-          requests:
             memory: 32Mi
+          requests:
+            memory: 16Mi
 {{- if contains "/" .Values.global.proxy_init.image }}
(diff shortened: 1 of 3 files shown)


sjanota:
Maybe specific momory values could be extracted to appropriate `values.yaml`s and set via overrides?

resources/istio/templates/sidecar-injector-configmap.yaml
```diff
@@ -18,9 +18,9 @@ data:
       - name: istio-init
         resources:
           limits:
-            memory: 96Mi
-          requests:
             memory: 32Mi
+          requests:
+            memory: 16Mi
```
sjanota: Here you set memory for init-container which I think is not that important. Resources for istio proxy are defined in [`resources/istio/values.yaml`](https://github.com/kyma-project/kyma/blob/master/resources/istio/values.yaml#L39). You can set them via overrides so they are environment specific

PK85: Yes, just marked PR with WIP label.

Issue #542, opened 2018-09-06
Refactor the Application Connector Helm Chart


janmedrek:
**Description**

We need to apply modularization to the Kyma. One of the steps is to externalize the Application Connector Helm Chart which will contain all relevant items.

The proposal of changes can be found here: [Proposal](https://github.com/kyma-project/community/blob/master/sig-and-wg/sig-core/proposals/application-connector-as-addon.md)

This is the first attempt to do it. If there will be no possibility to meet all Acceptance Criteria, please create follow-up issues.

Acceptance:
- Helm chart which is installing all Application Connector elements created
- The chart can be installed optionally on top of the Kyma core
- The installation procedure must be created - A user should be provided with a short and consistence documentation on how to deploy the Application Connector. The same global variables must be used
- The installer must be cleared from the Application Connector stuff

PR #1153, opened 2018-10-05
Prepare Application Connector for Modularisation

janmedrek:
**Description**

- AC components moved out of the core to the new Application Connector chart
- REB moved to the new chart
- NGINX moved to the new chart

**Related issue(s)**
#542


janmedrek:
@venturasr Hey, this parameter needs to be provided in order to install helm charts, whether it is true or not depends solely on type of installation (tests config, some of them are not ran in local environment).
In my case it was on-cluster installation - that's why I've set it to false.

Issue #1048, opened 2018-10-01
Enable publish API to use sourceId value from header if present


abbi-gaurav:
### Description
As discussed with @lszymik we can get away from the events service. This service is only enriching the events metadata with the `sourceId` value. 
However, this can also be done at the proxy level (nginx or istio not revelant.) 

**This is also important from resource efficiency perspective**

For publish API in event-bus, it will be required to handle the scenario where `sourceId` is sent via the headers vs in the HTTP payload.

###Order of Precedence
1.  If `sourceId` is present in the payload is highest
1.  else one in the header is used
1. If none present, an error is thrown. As decided by the team, we will return 4xx in case `sourceId` is missing. This needs to be translated to 5xx at the proxy level.

### AC:
- [x] **There is an agreement with Framefrogs about the error propagation and translation. (Do this as the first step before starting code)**
- [x] Event Bus publish API handles the sourceId header value and uses it in the FQTN.
- [x] The validations done in event service are applied in publish.
- [x] The `/re/v1/events` should remain unchanged.
- [x] The tracing for events works as expected.

lszymik:
Thank you very much. Please keep in mind that backward compatibility must be ensured at least for a transition phase :)

lszymik:
@abbi-gaurav we need to change dependencies here. The issue is not blocked, but it is blocking. The header with `sourceId` must be implemented first; then we will be able to test forwarding headers.

Kindly please treat it with a priority.

abbi-gaurav:
Verified with the new image.

## source-id in body
```bash
 curl -i \
>         -H "Content-Type: application/json" \
>         -X POST http://core-publish.kyma-system:8080/v1/events \
>         -d '{"source-id": "stage.commerce.kyma.local", "event-type": "test-eve
nt-bus", "event-type-version": "v1", "event-time": "2018-11-02T22:08:41+00:00",
"data": {"orderCode": "1", "orderPrice": 123.40}}'
HTTP/1.1 200 OK
Content-Type: application/json
Date: Wed, 10 Oct 2018 11:28:42 GMT
Content-Length: 52

{"event-id":"b4f0db60-b87e-4404-8441-5a2b990aaa7b"}
```

### Source-Id as Header
```bash
/ # curl -i \
>         -H "Content-Type: application/json" \
>         -H "Source-Id: stage.commerce.kyma.local" \
>         -X POST http://core-publish.kyma-system:8080/v1/events \
[...]

PR #1192, opened 2018-10-09
Enable Source-Id as a header field

abbi-gaurav:
**Description**
- Adjust publish API to accept source id field from the header
- Add tests and validations for sourceid from payload and header
- Update Makefile to remove push as it is no longer possible to run
- Move vet and test targets to a common Makefile at the root
- Update Jenkins file accordingly.
- Add stages to run and vet the event-bus source code as first stages
- Ensure source id in the payload is given precedence when it is present in both header and payload
- Did minor clean up to remove some unwanted functions and variables
- Update the OpenAPI definition.
- Remove the source field. This was a miss in the api.yaml

**Related issue(s)**
See also #1048

### DIFF
--- a/components/event-bus/Jenkinsfile
+++ b/components/event-bus/Jenkinsfile
@@ -50,8 +50,12 @@ podTemplate(label: label) {
                             + "./...")
                         }
 
-                        stage("build and test - event-bus-publish") {
-                            execute("/cmd/event-bus-publish", "make clean build")
+                        stage("test - event-bus") {
+                            execute("/", "make test")
+                        }
+
+                        stage("build - event-bus-publish") {
+                            execute("/cmd/event-bus-publish", "make clean compile")
                         }
 
                         stage("build image - event-bus-publish") {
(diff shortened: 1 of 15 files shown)


components/event-bus/cmd/event-bus-publish/handlers/handlers.go
```diff
@@ -90,6 +90,11 @@ func handlePublishRequest(w http.ResponseWriter, r *http.Request, publisher *con
 		return
 	}
 	publishRequest, err := parseRequest(body)
+
+	if len(publishRequest.SourceID) == 0 {
```
marcobebway: At this point `err` may not be nil.

Please move this check after calling `ValidatePublish`.

abbi-gaurav: Thanks for the feedback. As discussed the logic will be added after `err` nil check but before `ValidatePublish` as we need to add the sourceId from the header before validating.

abbi-gaurav:
@marcobebway applied the feedback. Please have a look

components/event-bus/Jenkinsfile
```diff
@@ -50,8 +50,12 @@ podTemplate(label: label) {
                             + "./...")
                         }
 
-                        stage("build and test - event-bus-publish") {
-                            execute("/cmd/event-bus-publish", "make clean build")
+                        stage("test - event-bus") {
```
marcobebway: 👍

PR #1206, opened 2018-10-10
Add more logging when ns deletion fails to work

venturasr:
**Description**

This pull request add some logging to test-kubeless, in order to identify a possible reason for the time out when namespaces are being deleted. 

**Related issue(s)**

### DIFF
--- a/tests/kubeless-test-client/test-kubeless.go
+++ b/tests/kubeless-test-client/test-kubeless.go
@@ -27,31 +27,47 @@ func deployK8s(yamlFile string) {
 }
 
 func deleteK8s(yamlFile string) {
-	cmd := exec.Command("kubectl", "delete", "-f", yamlFile)
+	cmd := exec.Command("kubectl", "delete", "-f", yamlFile, "--grace-period=0", "--force", "--ignore-not-found")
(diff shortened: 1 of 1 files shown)


tests/kubeless-test-client/test-kubeless.go
```diff
@@ -41,11 +41,17 @@ func deleteNamespace(namespace string) {
 	for {
 		select {
 		case <-timeout:
+			cmd := exec.Command("kubectl", "get", "ns", namespace, "-o yaml")
```
aszecowka: It will not solve our problem. Printing namespace details tell us only that namespace has "kubernetes" finalizer - but still, we don't know which object from this namespace cannot be removed. 
Could you analize "namespaced_resources_deleter.go" (https://github.com/kubernetes/kubernetes/blob/7f23a743e8c23ac6489340bbb34fa6f1d392db9d/pkg/controller/namespace/deletion/namespaced_resources_deleter.go), especially deleteAllContent method. They are using the dynamic client to detect what should be deleted. Maybe we can do the same and just print resources which are still not deleted after the timeout (6min)

Issue #741, opened 2018-09-17
 Hard-coded passphrases in Dex


pbochynski:
The Dex config map contains a section of static passphrases. This includes hashes of predefined passphrases for high-privileged default users.

**Risk:**

Users who run a Kyma cluster with these default users, will be vulnerable to unauthorized access by arbitrary attackers. This fully compromises confidentiality, integrity, and availability of the respective Kyma cluster.

This risk persists even if the plaintext passphrases were documented elsewhere.

**Suggested mitigation:**

Don't use hard-coded passphrases!

Instead, generate strong, securely random passphrases during the install process. Print these passphrases (and the corresponding users) to the console. Or emit them to a local file (which must only be readable by the current OS user.)

Alternatively, elicit the passphrases interactively, or make them configurable via env-vars or CLI arguments.

derberg:
Please share how you plan to fix this before you do as this must developer friendly:
- printing on local install to console is not enough as we also might have cluster installation 
- this must be introduced well in documentation because it must be easy to use https://github.com/kyma-project/kyma/pull/198/files
- it must be properly communicated so people from projects like https://github.com/kyma-project/console/tree/master/tests know there is additional work need

kubadz:
@derberg 
Sorry, I didn't see your comment. All the documentation related to creating the static users or logging to the UI was updated. There was also an appropriate information in the slack. For users it will be communicated in the changelog of the new release.

meeque:
Sorry, I know this is already closed. But could you have a look at my comment in PR #1245?

PR #861, opened 2018-09-21
Split istio installation process

sjanota:
**Description**

Changes proposed in this pull request:
- custom changes we have applied to isito charts are extracted
- plain istio 1.0.2 charts are vendored in resources irectory
- kubernetes job performing the patch is added as helm chart. it must be run after istio and beore kyma

**Related issue(s)**

See also #382

PR #1235, opened 2018-10-11
Add tool enabling static users generation for Dex

kubadz:
**Description**

- Add tool enabling static users generation for Dex
- Make dex acceptance test use user credentials from env variables instead of hardcoded values
- Remove default user credentials from ui-api-layer-acceptance-tests

That PR has to be merged to release the static-users-generator tool and images of acceptance tests not depending on hardcoded user credentials. That is necessary to introduce another PR with use of that tool in Dex.

PR #1245 contains all the changes necessary to test the static-users-generator tool and changes in acceptance tests introduced in this PR.

**Related issue(s)**

See also #741


tools/static-users-generator/Dockerfile
```diff
@@ -0,0 +1,13 @@
+FROM alpine
```
sjanota: Specify version of alpine. Should be 3.8.

tools/static-users-generator/scripts/config_replace.sh
```diff
@@ -0,0 +1,78 @@
...
+  fi
+
+  PASSWORD=$(echo -n ${PASSWORD} | base64 -d)
+
+  HASH=$(htpasswd -bnBC 10 "" ${PASSWORD} | tr -d ':\n')
+
+  echo "email: ${EMAIL}"
+  echo "username: ${USERNAME}"
+  echo "password: ${PASSWORD}"
+  echo "password encrypted: ${HASH}"
```
sjanota: Do not log passwords :)

kubadz: You are right. I already fixed that in PR #1245, but of course it should be fixed in this PR.

tests/acceptance/dex/dex_test.go
```diff
@@ -117,7 +127,13 @@ func TestSpec(t *testing.T) {
 
 		tokenParts := strings.Split(idToken, ".")
 
-		tokenPayloadEncoded := tokenParts[1] + "="
+		tokenPayloadEncoded := tokenParts[1]
+
+		if len(tokenPayloadEncoded)%3 == 1 {
```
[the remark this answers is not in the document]
sjanota: For reference, I think following is more correct:
```
missingTokenBytes := int(math.Abs(3 - len(tokenPayloadEncoded)%3))
```
Output is the same though ;) no need to change it.

tools/static-users-generator/scripts/config_replace.sh
```diff
@@ -0,0 +1,74 @@
+#!/bin/bash
```
mszostok: your team really like coding in bash?
I still have the same opinion about that as previously...
https://github.com/kyma-project/kyma/pull/861#discussion_r220844279

tools/static-users-generator/Dockerfile
```diff
@@ -0,0 +1,13 @@
+FROM alpine:3.8
```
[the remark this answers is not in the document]
sjanota: There is no image `alpine:3.8.1` according to [docker hub](https://hub.docker.com/_/alpine/). `docker pull alpine:3.8.1` says the same thing.

tools/static-users-generator/Dockerfile
```diff
@@ -0,0 +1,13 @@
...
+
+ENV KUBE_LATEST_VERSION="v1.11.3"
```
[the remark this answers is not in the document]
kubadz: I totally agree that KUBE_VERSION is the better name for that env.

tools/static-users-generator/scripts/config_replace.sh
kubadz: That is not about what we like. The approach with bash script was faster to implement in case of such a small tool. We will be probably implementing this tool in go later.

tools/static-users-generator/Dockerfile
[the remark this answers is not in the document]
kubadz: sjanota is right. I checked and alpine:3.8 is actually a 3.8.1 (with the issue fixed) version of alpine. You can check it by yourself by executing:
```
docker run -it alpine:3.8
cat /etc/alpine-release
```

PR #1236, opened 2018-10-11
Console enable http with redirect to https

maxmarkus:
Added port 80 to gateway for console with https redirect.


resources/core/charts/gateway/templates/gateway.yaml
```diff
@@ -16,4 +16,13 @@ spec:
...
\ No newline at end of file
+    - "*.{{.Values.global.domainName}}"
+  - port:
+      number: 80
+      name: http
+      protocol: HTTP
+    tls:
+      httpsRedirect: true # automatic 301 redirect from http to https
+    hosts:
+    - "console.{{.Values.global.domainName}}"
```
piotrmsc: Instead of listing hosts here please provide a wildcard like we have above.
If we have console and grafana we could have as well dex, jaeger, ui-api etc. so it's better have it like with https

PR #1245, opened 2018-10-12
Use tool enabling static users generation for Dex

kubadz:
**Description**

- Add admin user credentials secret to dex chart
- Use static-users-generator tool in Dex
- Use generated user credentials in acceptance tests
- Remove secrets overwrites with hardcoded credentials
- Update readme entries related to use of user credentials

That PR can't be merged until merge of #1235. It temporary uses version of components released in that PR. After the merge of #1235 images of static-users-generator tool and acceptance tests should be changed to newer versions and should be downloaded from `develop` directory.

This PR contains all the changes necessary to test the static-users-generator tool and changes in acceptance tests introduced in PR #1235.

**Related issue(s)**

See also #1235
See also #741


resources/core/charts/cluster-users/README.md
```diff
@@ -2,7 +2,7 @@
 
 ## Overview
 
-The `cluster-users` sub-chart uses role-based access control (RBAC) to define roles and accesses in Kyma. Kyma does not have a user base, but it has an admin defined as **admin@kyma.cx**. The admin's password is **nimda123**. Additionally, admin roles are bound to group of users named **kyma-admins**.
+The `cluster-users` sub-chart uses role-based access control (RBAC) to define roles and accesses in Kyma. Kyma does not have a user base, but it has an admin defined as **admin@kyma.cx**. The admin's password is generated and can be obtained from the secret named `admin-user` in the namespace of Dex release. Additionally, admin roles are bound to group of users named **kyma-admins**.
```
tomekpapiernik:
```suggestion
The `cluster-users` sub-chart uses role-based access control (RBAC) to define roles and accesses in Kyma. Kyma does not have a user base, but it has an admin defined as **admin@kyma.cx**. The admin password is generated by Helm.  Get it from the `admin-user` Secret located in the Namespace of Dex release. Additionally, admin roles are bound to group of users named **kyma-admins**.
```

resources/dex/templates/dex-users-secret.yaml
```diff
@@ -0,0 +1,15 @@
...
+  name:  admin-user
+  namespace: {{ .Release.Namespace }}
+  annotations:
+    "helm.sh/hook": "pre-install"
+  labels:
+    "dex-user-config": "true"
+data:
+  email: YWRtaW5Aa3ltYS5jeA==
+  username: YWRtaW4=
+  password: {{ randAlphaNum 12 | b64enc }}
```
meeque: Is randAlphaNum securely random? Afaics, it ultimately uses https://golang.org/pkg/math/rand/, which is not securely random. Its documentation says:
> For random numbers suitable for security-sensitive work, see the crypto/rand package.

Also, I wonder if 12 characters are enough? Afaics, randAlphaNum draws from an alphabet of 62 symbols. 62^12 is slightly less than 2^72, i.e. we have slightly less than 72 bits of randomness.

How does dex prevent brute-force attacks against passwords? Is there some kind of throttling for username/password authentication? If not, we should aim for at least 256 bits of randomness.

On the other hand: Why the additional round of base64 encoding? It seems that randAlphaNum already uses a subset of the base64 alphabet?

Issue #1218, opened 2018-10-11
Refactor installer after removing legacy support (provision bundles)


strekm:
To do:
- [x] installer image hardening (remove kubectl and try to change base image)
- [x] clean up install / update steps
- [ ] ~remove bash executor if possible~ 

Bash executor must stay because sources download step still uses that.

PR #1248, opened 2018-10-12
Installer image hardening

kfurgol:
**Description**

- bumped base image version in installer
- removed bash and kubectl from installer docker image
- cleaned up update steps

**Related issue(s)**
Refactor installer after removing legacy support (provision bundles) #1218

### DIFF
--- a/components/installer/deploy/installer/Dockerfile
+++ b/components/installer/deploy/installer/Dockerfile
@@ -2,6 +2,9 @@ FROM alpine:3.8
 
 LABEL source="git@github.com:kyma-project/kyma.git"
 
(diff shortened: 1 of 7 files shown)


components/installer/deploy/installer/Dockerfile
```diff
@@ -1,17 +1,7 @@
-FROM alpine:3.7
+FROM alpine:3.8
 
 LABEL source="git@github.com:kyma-project/kyma.git"
 
-RUN apk --no-cache add ca-certificates
-RUN apk add --no-cache curl
```
crabtree: You cannot remove curl, as we still support downloading kyma package from URL

Issue #1154, opened 2018-10-05
Events with same subscription name fail to trigger lambdas in different ns 


sayanh:
**Description**

It is found if there are 2 subscriptions listening to the same event although in different namespaces but with same names, `push` fails to trigger both subscribers.

Logs from `nats-streaming`:
```
[1] 2018/10/04 14:18:09.389627 [DBG] STREAM: [Client:core-push-86dfd6c8f6-f9tld] Removed member from durable queue subscription, subject=ec-default.customer\.created.v1, inbox=_INBOX.D2lNwSRD2STNawHYSNwhZh, queue=lambda-order-service-customer.created-v1:event-bus-push, subid=1
```
In this case, both the subscriptions had this name `lambda-order-demo-order.created-v1`

Subscriptions were:
```
k --kubeconfig ~/Desktop/demo2kubeconfig -n demo get subscription -oyaml                                                                                                                                                               ✓  10620  16:35:13
apiVersion: v1
items:
- apiVersion: eventing.kyma.cx/v1alpha1
  kind: Subscription
  metadata:
    clusterName: ""
    creationTimestamp: 2018-10-05T13:47:46Z
    generation: 1
    labels:
      Function: order-demo
    name: lambda-order-demo-order.created-v1
    namespace: demo
    resourceVersion: "249299"
    selfLink: /apis/eventing.kyma.cx/v1alpha1/namespaces/demo/subscriptions/lambda-order-demo-order.created-v1
    uid: 3ddb566e-c8a5-11e8-aa90-000d3a46ee35
  spec:
    endpoint: http://order-demo.demo:8080/
    event_type: order.created
    event_type_version: v1
    include_subscription_name_header: true
    max_inflight: 400
    push_request_timeout_ms: 2000
    source_id: ec-default
  status:
    conditions:
    - last_transition_time: 2018-10-05T13:47:47Z
      message: ""
      reason: ""
      status: "True"
      type: events-activated
kind: List
metadata:
  resourceVersion: ""
  selfLink: ""
```
```
k --kubeconfig ~/Desktop/demo2kubeconfig -n stage get subscription -oyaml                                                                                                                                                              ✓  10621  16:36:22
apiVersion: v1
items:
- apiVersion: eventing.kyma.cx/v1alpha1
  kind: Subscription
  metadata:
    clusterName: ""
    creationTimestamp: 2018-10-04T14:18:09Z
    generation: 1
    labels:
      Function: order-service
[...]

PR #1252, opened 2018-10-12
Include the namespace in the subscription durable name

marcobebway:
**Description**

- Include the namespace in the subscription durable name
- Add a test to check the fix

**Related issue(s)**
#1154

### DIFF
--- a/components/event-bus/test/integration/publish_push_test.go
+++ b/components/event-bus/test/integration/publish_push_test.go
@@ -20,6 +20,7 @@ import (
 	"github.com/kyma-project/kyma/components/event-bus/generated/push/informers/externalversions/eventing.kyma.cx/v1alpha1"
 	"github.com/kyma-project/kyma/components/event-bus/internal/common"
 	"github.com/kyma-project/kyma/components/event-bus/internal/publish"
+	"github.com/kyma-project/kyma/components/event-bus/internal/push/actors"
 	"github.com/kyma-project/kyma/components/event-bus/internal/push/opts"
 	"github.com/kyma-project/kyma/components/event-bus/test/util"
 	"github.com/nats-io/nats-streaming-server/server"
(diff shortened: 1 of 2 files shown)


components/event-bus/test/integration/publish_push_test.go
```diff
@@ -279,3 +279,68 @@ func Test_pushRequestShouldNotIncludeKymaTopicHeader(t *testing.T) {
...
 }
+
+func Test_sameSubjectSubscribersInDifferentNamespacesShouldReceiveEventsOfThatSubject(t *testing.T) {
+	// create two subscriptions with the same name in two different namespaces to the same subject
+	name := "test-subscription"
+	subscription1 := util.NewSubscription(name, "namespace-1", subscriberServerV1.URL+util.SubServer1EventsPath, eventType, eventTypeVersion, sourceIDV1)
+	subscription2 := util.NewSubscription(name, "namespace-2", subscriberServerV2.URL+util.SubServer2EventsPath, eventType, eventTypeVersion, sourceIDV1)
+
+	// handle the two subscriptions
+	pushApplication := initPushApplication()
```
abbi-gaurav: All the initialization should be done in the `TestMain` including starting the applications.

Please also fix `Test_pushRequestShouldNotIncludeKymaTopicHeader` which has similiar issues

marcobebway:
Verified that the tracing flow is working correctly, check the screenshot below.

subscription-per-env

components/event-bus/test/integration/publish_push_test.go
```diff
@@ -279,3 +278,58 @@ func Test_pushRequestShouldNotIncludeKymaTopicHeader(t *testing.T) {
 		t.Fatalf("request to endpoint should not include the header %s", headerKymaTopic)
 	}
 }
+
+func Test_sameSubjectSubscribersInDifferentNamespacesShouldReceiveEventsOfThatSubject(t *testing.T) {
```
Abd4llA: I think this test is indecisive. The test case we need to verify is the isolation of two subscriptions with the same name in two different namespaces. So either send an event to sub1 and verify that sub 2 didn't receive it, or send two different events to the two subscriptions and verify each received the correct one.

The current test will not fail if the bug was not fixed so it does not expose it.

Abd4llA: You can keep the current test, but the crucial one is the isolation test IMHO.

marcobebway: As per our offline discussion with @Abd4llA, this test is meant to verify that two subscriptions with the same name but in different kubernetes namespaces, can both receive the same events when subscribing to the same subject.

As agreed with @Abd4llA, no more action required regarding this and it is resolved.

PR #1260, opened 2018-10-15
Stability checker - fix path for testing script.

aszecowka:


tools/stability-checker/deploy/chart/stability-checker/values.yaml
```diff
@@ -14,7 +14,7 @@ image:
 storage:
   claimName: "stability-test-scripts-pvc"
 
-pathToTestingScript: "/data/input/testing-kyma.sh"
+pathToTestingScript: "/data/input/testing.sh"
```
mszostok: IMO modification this value here for fixing purpose is more like a side effect. This should be changed in `/install-stability-checker.sh` script where we are mounting given script. There we should define the contract. What's more this solution is temporary and error prone

PR #1268, opened 2018-10-15
Bump stability checker version

aszecowka:


tools/stability-checker/deploy/chart/stability-checker/values.yaml
```diff
@@ -6,7 +6,7 @@ containerRegistry:
   path: eu.gcr.io/kyma-project
 
 image:
-  tag: 4a0904b3
+  tag: b4384a6e
```
mszostok: bump also in Chart.yaml file line `appVersion` or just remove this field

PR #1289, opened 2018-10-16
Fix component overrides

### LINKED ISSUE #1276: Invalid overrides creation when multiple objects exists for the same component.
There's a bug in overrides for components.
When multiple sources for overrides exists for the same component (e.g. two ConfigMaps), only one (last one found) is actually taken as overrides source, the other ones are discarded completely.

Tomasz-Smelcerz-SAP:
**Description**

- All logic moved out of config-reader to be able to unit-test without K8s dependencies
- Logic for joining component overrides from multiple sources fixed
- Tests for joining override values from multiple sources added (for both common and component overrides)

**Related issue(s)**
Fixes #1276

### DIFF
--- a/components/installer/pkg/overrides/config-reader.go
+++ b/components/installer/pkg/overrides/config-reader.go
@@ -57,7 +57,7 @@ func (r reader) readComponentOverrides() ([]component, error) {
 	for _, sec := range secrets {
 		comp := component{
 			name:      sec.Labels["component"],
-			overrides: toStringStringMap(sec.Data),
+			overrides: toInputMap(sec.Data),
 		}
 		components = append(components, comp)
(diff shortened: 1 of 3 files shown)


components/installer/pkg/overrides/config-reader.go
```diff
@@ -111,8 +112,8 @@ func concatLabels(labels ...string) string {
 	return strings.Join(labels, ", ")
 }
 
-func toStringStringMap(input map[string][]byte) map[string]string {
-	var output = make(map[string]string)
+func toStringStringMap(input map[string][]byte) inputMap {
```
Tomasz-Smelcerz-SAP: Rename

Issue #1214, opened 2018-10-10
Improve logs in acceptance test


polskikiel:
```
12:16:28 Run remote-environment tests
12:16:28 --- FAIL: TestRemoteEnvironmentAPIAccess (276.45s)
12:16:28     re_test.go:31: Running Remote Environment Test with config: {DockerImage:eu.gcr.io/kyma-project/develop/acceptance-tests:6e0a7fe2 LinkingTimeout:4m0s UnlinkingTimeout:4m0s KeepTestResources:false Disabled:false}
12:16:28     testsuite.go:75: Creating deployments and services
12:16:28     testsuite.go:77: Creating Istio resources
12:16:28     testsuite.go:79: Creating RemoteEnvironment
12:16:28     re_test.go:40: Waiting for service class
12:16:28     re_test.go:44: Provisioning service instance
12:16:28     testsuite.go:146: waiting for service instance to be ready
12:16:28     testsuite.go:146: waiting for service instance to be ready
12:16:28     testsuite.go:146: waiting for service instance to be ready
12:16:28     testsuite.go:146: waiting for service instance to be ready
12:16:28     re_test.go:47: Binding
12:16:28     re_test.go:51: All pods are running
12:16:28     re_test.go:53: Creating binding usage
12:16:28     re_test.go:58: Expecting call succeeded and env injected
12:16:28     kubernetes.go:219: Logs from pod acc-test-client-w6n2-6b7879db58-pd9dk (created at 10:07:10):
12:16:28         
12:16:28     kubernetes.go:219: Logs from pod acc-test-client-w6n2-7f4868c8c9-7fd8n (created at 10:06:44):
12:16:28         
12:16:28     assertions.go:247: 
                          
    Error Trace:    testsuite.go:237
12:16:28         
                    
            testsuite.go:210
12:16:28         
                    
            re_test.go:59
12:16:28         
    Error:          timeout for tester results (4m0s) exceeded
12:16:28         
    Test:           TestRemoteEnvironmentAPIAccess
12:16:28 FAIL
```
AC:
- shorten logs by improving retries etc
- namely start and ok/fail status for every step

aszecowka:
In this test we also print a lot of information like that:
```
		remote environment gateway, name:
		- server: req headers
		-- X-B3-Spanid => [bae04b7c4d7aeb2d]
		-- Content-Length => [0]
		-- X-B3-Traceid => [bae04b7c4d7aeb2d]
[...]

PR #1300, opened 2018-10-17
Improve logging - make it more compacted

piotrmiskiewicz:
**Description**

- reduce the frequency of client calls to the gateway
- reduce number of lines in the log

**Related issue(s)**
Related: #1214

### DIFF
--- a/tests/acceptance/remote-environment/cmd/fake-gateway/main.go
+++ b/tests/acceptance/remote-environment/cmd/fake-gateway/main.go
@@ -25,11 +25,7 @@ type httpHandler struct {
 }
 
 func (h *httpHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
-	fmt.Fprintf(w, "remote environment gateway, name: %s\n", r.URL.Path[1:])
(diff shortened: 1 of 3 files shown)


tests/acceptance/remote-environment/cmd/fake-gateway/main.go
```diff
@@ -25,11 +25,7 @@ type httpHandler struct {
 }
 
 func (h *httpHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
-	fmt.Fprintf(w, "remote environment gateway, name: %s\n", r.URL.Path[1:])
-	fmt.Fprintln(w, "- server: req headers")
-	for k, v := range r.Header {
-		fmt.Fprintf(w, "-- %s => %s\n", k, v)
-	}
+	fmt.Fprintf(w, "remote environment gateway, request URL: \"%s\" request headers: %v\n", r.URL, r.Header)
```
mszostok: could you give me and example how it will look like ? because `r.Header` does not implemenets the String interface so I do not know how it will be printed

PR #1312, opened 2018-10-17
Update event-bus version

marcobebway:
**Description**

- Update event-bus version

**Related issue(s)**
#1154


montaro:
@marcobebway I can see that this is not the latest tag! The latest is `ce3ebf0c`

marcobebway:
@montaro, No this should not be the latest version. It should be the version that contains my changes done in this pull request. And this version is `2408daf3`. Please make sure to do `git show 2408daf3` and you will see my changes.

Issue #971, opened 2018-09-26
Installer should uninstall components from CR


pbochynski:
**Description**
I installed kyma without ec-default and hmc-default components. Then I started uninstall action and the installer failed on the step uninstalling release ec-default. 

**Expected result**
The installer should not try to uninstall ec-default because it was not installed.

**Actual result**

```
2018/09/26 20:37:11 Removing Kyma resources...
2018/09/26 20:37:11 Uninstalling releases...
2018/09/26 20:37:11 Uninstalling release ec-default
2018/09/26 20:37:11 Uninstall Error:  Details: rpc error: code = Unknown desc = release: "ec-default" not found
2018/09/26 20:37:11 Error. Some releases may have not been removed.
```

**Steps to reproduce**

Install kyma without remote-environments and uninstall it.

AC:
- Uninstall action should uninstall components that are listed in CR

pbochynski:
I think releases are still hardcoded:
https://github.com/kyma-project/kyma/blob/70d5c84f444c1eceae55b91fda3355366b908501/components/installer/pkg/steps/remove-kyma-components.go#L25-L40

PR #1329, opened 2018-10-18
Installer uninstall components from CR

kfurgol:
**Description**

- Uninstall action takes components from CR instead of hardcoded ones.
- Uninstall action can be performed multiple times

**Related issue(s)**
Installer should uninstall components from CR #971

### DIFF
--- a/components/installer/pkg/steps/remove-kyma-components.go
+++ b/components/installer/pkg/steps/remove-kyma-components.go
@@ -4,7 +4,6 @@ import (
 	"log"
 
 	"github.com/kyma-project/kyma/components/installer/pkg/config"
-	rls "k8s.io/helm/pkg/proto/hapi/services"
 )
 
 //RemoveKymaComponents .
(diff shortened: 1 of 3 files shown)


components/installer/pkg/steps/remove-kyma-components.go
```diff
@@ -22,9 +25,15 @@ func (steps InstallationSteps) RemoveKymaComponents() {
 	log.Println("All Kyma resources have been successfully removed")
 }
 
-func (steps InstallationSteps) uninstallReleases() error {
+func (steps InstallationSteps) uninstallReleases(installationData *config.InstallationData) error {
 	log.Println("Uninstalling releases...")
-	releasesToBeDeleted := []string{"ec-default", "hmc-default", "dex", "core", "prometheus-operator", "istio", "cluster-essentials"}
+
+	releasesRes, err := steps.helmClient.ListReleases()
+	if err != nil {
```
crabtree: You can simply convert the response from helm list into map[string]bool where key is the release name (value is not relevant here, can be bool). Then iterate over the installationData.Components and check if the release from component is installer (exists in map). If no, then log this fact, otherwise perform DeleteRelease action.

jakkab: +1. To maximize performance, a string-struct map (map[string]struct{}) should be your first choice.

components/installer/pkg/steps/remove-kyma-components.go
```diff
@@ -2,55 +2,51 @@ package steps
...
-	releasesToBeDeleted := []string{"ec-default", "hmc-default", "dex", "core", "prometheus-operator", "istio", "cluster-essentials"}
+	installedReleases := make(map[string]bool)
 
-	for _, releaseName := range releasesToBeDeleted {
-		log.Println("Uninstalling release", releaseName)
-		uninstallReleaseResponse, uninstallError := steps.helmClient.DeleteRelease(releaseName)
-		if !steps.errorHandlers.CheckError("Uninstall Error: ", uninstallError) {
-			steps.helmClient.PrintRelease(uninstallReleaseResponse.Release)
-		}
+	releasesRes, err := steps.helmClient.ListReleases()
```
jakkab: Since it is for the user to decide which components will be uninstalled, there's no need to fetch Helm releases here.  IMHO the function should try to uninstall each component from the CR, logging some sort of a warning should there be no such release. Release exists = uninstall. Release not found = log warning and continue. This is how I see it.

components/installer/pkg/steps/remove-kyma-components.go
```diff
@@ -2,33 +2,30 @@ package steps
...
-	if err := steps.removeNamespaces(); err != nil {
-		log.Println("Error. Some namespaces may not have been removed")
-		return
-	}
-
 	log.Println("All Kyma resources have been successfully removed")
 }
 
-func (steps InstallationSteps) uninstallReleases() error {
+func (steps InstallationSteps) uninstallReleases(installationData *config.InstallationData) error {
```
jakkab: Please enclose this function's logic in RemoveKymaComponents() func. There's no need to call another function here.

PR #1350, opened 2018-10-19
Fix Websocket connections to UI API Layer

pkosiec:
**Description**

- Add origin verification (Moved @michal-hudy's solution)
- Rewrite token given in `Sec-Websocket-Protocol` to `Authorization` header

To test it, please run console from the PR https://github.com/kyma-project/console/pull/249.

**Related issue(s)**
#1356

### DIFF
--- a/components/ui-api-layer/pkg/origin/origin.go
+++ b/components/ui-api-layer/pkg/origin/origin.go
@@ -1,0 +1,46 @@
+package origin
+
+import (
+	"net/http"
+	"strings"
+)
+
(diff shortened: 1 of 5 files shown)


components/ui-api-layer/pkg/origin/origin.go
```diff
@@ -0,0 +1,46 @@
+package origin
```
michal-hudy: Please add unit tests to that package 👍

michal-hudy: Please add unit tests to that package 👍

Issue #1271, opened 2018-10-15
Ensure the Istio version used in Kyma is compatible with Knative features


abbi-gaurav:
### Description
To ensure we can run Kyma and Knative together, both should be able to use the same Istio version.

### AC:
- [ ] Kyma is using the same K8S version as used by Knative.
- [ ] Knative serving and eventing work with the Istio version.
- [ ] Kyma features work with the Istio version.
- [ ] Istio patching done in Kyma is successfull

Issue #1270, opened 2018-10-15
Ensure Kyma uses the same K8S version as used by Knative


abbi-gaurav:
### Description
The first prerequisite to run Kyma on top of (or along with) Knative is to ensure they are using the same K8S version.

Note:
* Knative serving and eventing should work with the k8s version
* Fix any issues encountered in Kyma with the k8S version

### AC:
- [ ] Kyma is using the same K8S version as used by Knative

abbi-gaurav:
Waiting for the PR to get merged.

Abd4llA:
Based on the PR above and since it won't be merged as the Knative PoC WG is taking over, the verification was done by that PR.

PR #1360, opened 2018-10-22
Add support for Knative mode

Abd4llA:
**Description**
Add a --knative flag to run.sh which:
  - Installs a minikube with kubeadm bootstrapper
  - Installs upstream istio 1.0.2 .
  - Installs knative build, serving and eventing.
  - Uses installer-cr-knative CR to skip legacy components
  - Uses installer-config-knative to define global.feature.knative flag
   and skip legacy config

- Add support for Knative mode

**Related issue(s)**

#1270 #1271


installation/scripts/minikube.sh
```diff
@@ -170,19 +180,36 @@ function start() {
...
-    --extra-config=apiserver.GenericServerRunOptions.CorsAllowedOriginList=".*" \
-    --extra-config=controller-manager.ClusterSigningCertFile="/var/lib/localkube/certs/ca.crt" \
-	--extra-config=controller-manager.ClusterSigningKeyFile="/var/lib/localkube/certs/ca.key" \
-    --extra-config=apiserver.Admission.PluginNames="LimitRanger,ServiceAccount,DefaultStorageClass,MutatingAdmissionWebhook,ValidatingAdmissionWebhook,ResourceQuota" \
-    --kubernetes-version=v$KUBERNETES_VERSION \
-    --vm-driver=$VM_DRIVER \
-    --feature-gates="MountPropagation=false" \
-    -b=localkube
-
+    if [[ $KUBEADM ]]; then
```
Tomasz-Smelcerz-SAP: Lot's of repetition here. It's hard to understand the differences. Please extract all common properties to some variable, and put only the differences in the conditional.

installation/cmd/run.sh
```diff
@@ -39,21 +58,61 @@ do
 done
 set -- "${POSITIONAL[@]}" # restore positional parameters
 
+MINIKUBE_ARGS="--domain \"${DOMAIN}\" --vm-driver \"${VM_DRIVER}\""
+
+if [[ $KNATIVE ]]; then
```
Tomasz-Smelcerz-SAP: This is a very shortsighted solution.
Think about introducing more feature flags
The number of combinations (if/else) grows exponentially:
```
if(knative)
  if(other) {}
    else {}
  fi
else
  if(other) {}
    else {}
  fi
fi
```
If you have three feature flags, the number of combinations is eight, and it can't be handled by if/else anymore.
This has to be refactored so that adding more ffs is possible in the future.

installation/cmd/run.sh
```diff
@@ -26,11 +26,30 @@ do
             shift # past argument
             shift # past value
         ;;
+        --crtpl_path)
```
sjanota: I think it should be `crtpl-path` for consistency

piotrmsc:
@Abd4llA  I think we can close this PR?

Issue #1045, opened 2018-10-01
Installer does not start installation after pod crash


crabtree:
When running local installation when installer pod crashes the installation does not restarts. It hangs on downloading kyma sources. The problem exists because during first start the kyma sources are delivered to the installer using `kubectl cp` and after restart they are not available.

AC:
The local installation process should handle kyma sources in more persistent way or we should switch to kyma-installer image which has sources embedded.

ormos:
Why are we still dynamically copy the installation artefacts into the installer pod? The goal was to have one static image including all the needed artefacts for the installation - with the exception of the container images?
Where can I find a kind of architectural overview and some status diagrams how the installer is working today?

nexus49:
Is there any update on this topic? We have also observing this a number of times during cluster installations.

Tomasz-Smelcerz-SAP:
Local installation is using Kyma-Installer image built on the fly. It means this error will never occur again, Kyma sources are always available in Kyma-installer.

PR #1383, opened 2018-10-22
Installer retry on error

Tomasz-Smelcerz-SAP:
**Description**
Installer does not handle errors, restarts and unexpected-outcomes nicely. It usually enters "Error" or "In Progress" state and remains in that state forever.
In order to get closer to a true K8s controller, we have to introduce control loop. This P/R is making a step towards this goal, fixes  a problem caused by Helm failure at the very beginning of the installation (context deadline exceeded) and allows for Installer Pod restart/crash fail-over

- Installer re-tries operation regardless of the installation status (the status is reported in Installation CR instance)
- Installer continues installation after pod restart/crash. Pending Helm operation (if any) causes failures but does not break control loop. Once Helm finishes, installer proceeds with next component from the list.
- Fix some (not all...) errors found in core chart (the ones that directly break chart (re)installation).
- Change step name from "Downloading Kyma" to "Get Kyma Sources" to avoid user's confusion.

Known problems:
- 'core' chart is broken, can't be reinstalled. Attempts to do it failed, both using the Installer and manually. Installer can't automatically fix a problem in a broken chart. Once 'core' can be re-installed (helm delete core, helm install core), Installer should be able to get past failing 'core' installation.

**Related issue(s)**
See also #1045

### DIFF
--- a/components/installer/cmd/operator/main.go
+++ b/components/installer/cmd/operator/main.go
@@ -85,7 +85,7 @@ func main() {
 	kubeInformerFactory.Start(stop)
 	internalInformerFactory.Start(stop)
 
-	installationController.Run(2, stop)
+	installationController.Run(stop)
 }
 
 func getClientConfig(kubeconfig string) (*rest.Config, error) {
--- a/components/installer/pkg/apis/installer/v1alpha1/types.go
+++ b/components/installer/pkg/apis/installer/v1alpha1/types.go
@@ -26,7 +26,12 @@ func (i *Installation) ShouldInstall() bool {
 }
 
 func (i *Installation) canInstall() bool {
-	return (i.Status.State == StateEmpty || i.Status.State == StateUninstalled || i.Status.State == StateInstalled)
+	//TODO: Now this function always returns true. Remove.
+	return (i.Status.State == StateEmpty ||
+		i.Status.State == StateUninstalled ||
+		i.Status.State == StateInstalled ||
+		i.Status.State == StateError ||
+		i.Status.State == StateInProgress)
 }
 
 // ShouldUninstall returns true when user requested uninstall action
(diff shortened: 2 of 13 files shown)


components/installer/cmd/operator/main.go
```diff
@@ -85,7 +85,7 @@ func main() {
 	kubeInformerFactory.Start(stop)
 	internalInformerFactory.Start(stop)
 
-	installationController.Run(2, stop)
+	installationController.Run(stop)
```
Tomasz-Smelcerz-SAP: We don't need two workers, we actually want a singleton here

components/installer/pkg/apis/installer/v1alpha1/types.go
```diff
@@ -26,7 +26,12 @@ func (i *Installation) ShouldInstall() bool {
 }
 
 func (i *Installation) canInstall() bool {
-	return (i.Status.State == StateEmpty || i.Status.State == StateUninstalled || i.Status.State == StateInstalled)
+	//TODO: Now this function always returns true. Remove.
```
Tomasz-Smelcerz-SAP: This is left for the reviewer. Can be removed now or later.

components/installer/pkg/installation/controller.go
```diff
@@ -70,6 +71,23 @@ func NewController(kubeClientset *kubernetes.Clientset, kubeInformerFactory kube
 	eventBroadcaster.StartRecordingToSink(&typedcorev1.EventSinkImpl{Interface: kubeClientset.CoreV1().Events("")})
 	recorder := eventBroadcaster.NewRecorder(scheme.Scheme, corev1.EventSource{Component: "kymaInstaller"})
 
+	//backOffIntervals := []int{0, 10, 20, 40, 60, 120, 180, 300}
+	backOffIntervals := []uint{0, 1, 2, 4, 6, 12, 18, 30}
```
Tomasz-Smelcerz-SAP: remove

components/installer/pkg/installation/controller.go
```diff
@@ -121,22 +137,23 @@ func (c *Controller) worker() {
 func (c *Controller) processNextWorkItem() bool {
 
 	key, quit := c.queue.Get()
+	log.Println("Queue len: ", c.queue.Len())
```
Tomasz-Smelcerz-SAP: remove

resources/core/templates/istio-injection-label.yaml
```diff
@@ -20,8 +20,9 @@ spec:
         command:
         - kubectl
         - label
+        - --overwrite
```
Tomasz-Smelcerz-SAP: Required when reinstalling `core` chart, this label already exists and kubectl fails without --overwrite flag

Issue #1255, opened 2018-10-12
Helm Broker handles repositoryURL pointing to index.yaml too


PK85:
AC:
- when last segment of the **repository.baseURL** is a **directory** HB tries to find index.yaml and load bundles
- when a last segment of the **repository.baseURL** is a *.yaml file HB uses that index.yaml and load bundles

PR #1387, opened 2018-10-23
Change Helm Broker configuration method

jasiu001:
Related to #1255

### DIFF
--- a/components/helm-broker/internal/bundle/repository_config.go
+++ b/components/helm-broker/internal/bundle/repository_config.go
@@ -8,29 +8,29 @@ import (
 
 // RepositoryConfig provides configuration for HTTP Repository.
 type RepositoryConfig struct {
-	BaseURL string `json:"baseUrl" valid:"required"`
+	URL string `json:"URL" valid:"required"`
 }
 
 // IndexFileName returns name of yaml file with configuration (if not exist return default name)
 func (cfg RepositoryConfig) IndexFileName() string {
 	if cfg.hasConfigFile() {
-		return path.Base(cfg.BaseURL)
+		return path.Base(cfg.URL)
 	}
 
 	return "index.yaml"
 }
 
-// GetBaseURL returns base url to bundles with trailing slash
-func (cfg RepositoryConfig) GetBaseURL() string {
+// BaseURL returns base url to bundles with trailing slash
+func (cfg RepositoryConfig) BaseURL() string {
 	if cfg.hasConfigFile() {
-		return strings.TrimRight(cfg.BaseURL, cfg.IndexFileName())
+		return strings.TrimRight(cfg.URL, cfg.IndexFileName())
 	}
 
-	return strings.TrimRight(cfg.BaseURL, "/") + "/"
+	return strings.TrimRight(cfg.URL, "/") + "/"
 }
 
 func (cfg RepositoryConfig) hasConfigFile() bool {
-	extension := filepath.Ext(path.Base(cfg.BaseURL))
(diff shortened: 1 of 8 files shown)


components/helm-broker/internal/bundle/repository_config.go
```diff
@@ -0,0 +1,36 @@
...
+	"strings"
+)
+
+// RepositoryConfig provides configuration for HTTP Repository.
+type RepositoryConfig struct {
+	BaseURL string `json:"baseUrl" valid:"required"`
+}
+
+// GetIndexFile return name of yaml file with configuration (if not exist return default name)
+func (cfg RepositoryConfig) GetIndexFile() string {
```
mszostok: shouldn't be more like `GetIndexFileName`?  `GetIndexFile` for me is more like getting the file content itself. In Go we also not using the prefix for Getter/Setters. More idiomatic it can be just `IndexFileName` but I'm also good with your approach ;)

components/helm-broker/internal/bundle/repository_config_test.go
```diff
@@ -0,0 +1,39 @@
...
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+
+	"github.com/kyma-project/kyma/components/helm-broker/internal/bundle"
+)
+
+func TestRepositoryConfig_GetIndexFile_defaultValue(t *testing.T) {
```
mszostok: In Go for such testing, we are using the table tests.

```bash
func TestRepositoryConfig(t *testing.T) {
	tests := map[string]struct {
		givenRepoURL string

		expIndexName string
		expBaseURL   string
	}{
		"index not provided in URL": {
			givenRepoURL: "http://example.com/repository/stable/",

			expIndexName: "index.yaml",
			expBaseURL:   "http://example.com/repository/stable/",
		},
		"custom index provided in URL": {
			givenRepoURL:"http://example.com/repository/stable/conf.yaml",

			expIndexName: "conf.yaml",
			expBaseURL: "http://example.com/repository/stable/",
		},
		"index provided in URL same as default one": {
			givenRepoURL:"http://example.com/repository/stable/index.yaml",

			expIndexName: "index.yaml",
			expBaseURL: "http://example.com/repository/stable/",
		},
		"trailing slash is always added for BaseURL": {
			givenRepoURL:"http://example.com/repository/stable",

			expIndexName: "index.yaml",
			expBaseURL: "http://example.com/repository/stable/",
		},
	}

	for tn, tc := range tests {
		t.Run(tn, func(t *testing.T) {
			cfg := bundle.RepositoryConfig{BaseURL: tc.givenRepoURL}

			assert.Equal(t, tc.expBaseURL, cfg.GetBaseURL())
			assert.Equal(t, tc.expIndexName, cfg.GetIndexFile())

		})
	}
}

```

components/helm-broker/internal/bundle/repository_config.go
```diff
@@ -0,0 +1,36 @@
...
+
+import (
+	"path"
+	"path/filepath"
+	"strings"
+)
+
+// RepositoryConfig provides configuration for HTTP Repository.
+type RepositoryConfig struct {
+	BaseURL string `json:"baseUrl" valid:"required"`
```
mszostok: IMO it is not anymore a BaseURL, it can be but not always it is. My suggestion is to name it just `URL`.

jasiu001: I will change it with next commit during upgrade component version along with values.yaml in core.

PR #1026, opened 2018-10-01
Try to fix api-controller consumption

sjanota:
**Description**
Changes proposed in this pull request:

- synchronise caches before starting api-controller workers

**Related issue(s)**

Tries to solve: #661

PR #1393, opened 2018-10-23
Use api-controller with fixed consumption

sjanota:
**Description**

- Use image built in #1026 

**Related issue(s)**


sjanota:
CLA is broken, need to recreate

Issue #365, opened 2018-08-30
Subscriptions updates are not honoured in event bus


abbi-gaurav:
## Description
When updating a subscription spec but not changing the name, the updates are not honored or reflected.

Note: Check the functionality of `subscription validator` to make sure it won't get affected.

Hint:
* Update could be achieved by deleting old and creating new NATS Streaming subscription.
* Validatior should check the subscription resource and update the status flag accordingly.
* One idea to verify the update is to change the endpoint of the subscription

## AC:

- [x] An update to the Subscription K8S Custom Resource is reflected in the NATS Subscription.
- [x] The event activation status is retained.
- [x] There should be integration tests to ensure update is honored.
- [x] Event bus functionality is verified using the event bus example.
- [x] Since current update can result in loss of events under certain scenarios, the same should be documented.

radufa:
The unit and acceptance tests are green.
The PR was manually tested.
The integration tests, as required by AC, should be still written.

abbi-gaurav:
The integration test should verify that for the update
-  NATS Streaming subscription is deleted.
- New NATS Streaming subscription is created.

abbi-gaurav:
As @montaro is on vacation, @marcobebway will take over this.

abbi-gaurav:
As discussed, I will take over this task and explore what are the possible challenges in mocking or using the controller APIs from @montaro.

@montaro will take over #365 since this is commited for MS10 and is the sprint goal.

abbi-gaurav:
The code has been merged.
Left is:
* Update deployment configurations
* Add documentation

marcobebway:
Documentation and deployment is completed.

@abbi-gaurav, I assume you already did the AC item "**Event bus functionality is verified using the event bus example**" as part of your pull request. If so, please make sure to mark this item as checked.

abbi-gaurav:
@marcobebway I verified using the integration tests. The verification using the Kyma runtime is still pending.
Moving it back to **To Do**

marcobebway:
- Verified kyma acceptance tests are green.
- Verified event-subscription example is working correctly.

PR #1405, opened 2018-10-23
update event-bus version and add readme to describe subscription crd updates

marcobebway:
**Description**

- update event-bus version
- add readme to describe subscription crd updates

**Related issue(s)**
#365

### DIFF
--- a/docs/event-bus/docs/050-subscription-updates.md
+++ b/docs/event-bus/docs/050-subscription-updates.md
@@ -1,0 +1,12 @@
+---
+title: Subscription updates
+type: Details
+---
+
+To update the subscription CRD, run this command:
+
+`kubectl edit crd subscriptions.eventing.kyma.cx`
+
+The Event Bus reacts to the changes in subscription CRD, and updates the corresponding NATS-Streaming subscription accordingly.
+
+>**Note**: The current subscription update mechanism recreates a subscription with new specifications. This may result in the loss of messages delivered during the recreation process.
--- a/resources/core/values.yaml
+++ b/resources/core/values.yaml
@@ -54,7 +54,7 @@ global:
     version: 6e0a7fe2
   event_bus:
     dir: develop/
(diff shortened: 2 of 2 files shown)


docs/event-bus/docs/050-subscription-updates.md
```diff
@@ -0,0 +1,12 @@
+---
+title: Subscription updates
+type: Details
+---
+
+The subscription crd can be updated using kubectl:
```
bszwarc: `To update the subscription CRD, run this command: ` 

I would use the similar wording that we have in the CRD description to keep the consistency.

docs/event-bus/docs/050-subscription-updates.md
```diff
@@ -0,0 +1,12 @@
...
+
+`kubectl edit crd subscriptions.eventing.kyma.cx`
+
+Event-bus reacts on subscription crd changes, and it updates the corresponding NATS-Streaming subscription accordingly.
```
bszwarc: `The Event Bus reacts to the changes in subscription CRD, and updates the corresponding NATS-Streaming subscription accordingly.`

docs/event-bus/docs/050-subscription-updates.md
```diff
@@ -0,0 +1,12 @@
...
+
+**Note**: The current subscription update mechanism recreates a subscription with new specifications. This could lead to a scenarios where messages delivered during the recreation can get lost. An improved design that will address this concern will be provided in the future.
```
bszwarc: ` >**Note**: The current subscription update mechanism recreates a subscription with new specifications. This may result in the loss of messages delivered during the recreation process.  `

We typically avoid saying that the concern will be addressed in the future, because we never know when it is going to happen - that's why I would remove this line, leaving just the statement that something like this may happen. If the solution to this problem is added, we can remove the note.

PR #1415, opened 2018-10-24
Fix typo in webhook container start script

piotrmsc:
**Description**
Webhook in istio-patch has a typo in container start arguments. Webhook itself was running but was logging error with saying cannot open file.

- fixed file name in the start argument

**Related issue(s)**


sjanota:
Can we add a test to catch this kind of error in the future?

piotrmsc:
I agree and changes within the component will be in separate PR

Issue #1239, opened 2018-10-11
Helm Broker bundles marked as local and/or showcase


PK85:
AC:
- provide a bundle functionality where bundle could be marked as local and/or showcase by setting ```
metadata.labels ``` in a given SC class
- in bundle **meta.yaml** we can add any label, and **showcase** is one of them
- all by default marked as local

Same as we are doing for REB via RE definition with labels.

We agreed on having the following convention in metadata object of the service class:
```
labels:
  { 
        showcase: true
        local: true
   }
```
Related to: https://github.com/kyma-project/console/issues/127

PR #1444, opened 2018-10-26
Add support for defining labels in Helm Broker bundles

mszostok:
**Description**

- Add support for defining labels in Helm Broker bundles
- 

**Related issue(s)**
- https://github.com/kyma-project/kyma/issues/1239


components/helm-broker/internal/broker/catalog_test.go
```diff
@@ -69,60 +76,40 @@ func TestGetCatalogOnConversionError(t *testing.T) {
 func TestBundleConversion(t *testing.T) {
```
mszostok: Refactor inspired by golden testing, see: https://golang.org/src/go/doc/doc_test.go#L119

components/helm-broker/internal/broker/catalog_test.go
```diff
@@ -69,60 +76,40 @@ func TestGetCatalogOnConversionError(t *testing.T) {
...
-		p1 = s.Plans[1]
-	}
+	require.NoError(t, err)
+
+	normalizedGotSvc := tc.marshal(t, convertedSvc)
+
+	updateGoldenFileIfRequested(t, goldenPath, normalizedGotSvc)
+
+	exp := tc.fixtureMarshaledOsbService(t, goldenPath)
+	assert.JSONEq(t, exp, string(normalizedGotSvc))
```
mszostok: I tried to assert on Go obj 
using sth like that:
```go
func fixtureOsbService(t *testing.T, testdataBasePath string) osb.Service {
	data, err := ioutil.ReadFile(testdataBasePath)
	require.NoError(t, err, "failed reading .golden")

	out := osb.Service{}
	require.NoError(t, json.Unmarshal(data, &out))
	return out
}
```

Unfortunately kinds are different, e.g. `map[string]string` vs `Labels`
where `Labels` is  `type Labels map[string]string`. Then in the assertion, it cannot be compared easily. Because of that, I decided to normalize it to the JSON form.

PR #1449, opened 2018-10-26
Handle 422 error in ui-api

polskikiel:
**Description**

- Implemented 422 error handling in the UI-API 

**Related issue(s)**


components/ui-api-layer/pkg/gqlerror/errors.go
```diff
@@ -183,3 +198,9 @@ func appendArgument(arguments, name, value string) string {
 
 	return result
 }
+
+func buildInvalidErrorMsg(msg string) string {
```
piotrmiskiewicz: it is not a good idea to change the message., let's have an original one even if it is not so pretty

polskikiel: Ok, I removed that. Now I return a raw error.

Issue #1359, opened 2018-10-21
Alertmanager alerts: Many Kubelets cannot be scraped


nexus49:
**Description**
When running kyma on GKE we see the following slack alerts frequently firing: 

![image](https://user-images.githubusercontent.com/52737/47265220-c5de0180-d524-11e8-9a7f-e68241e0f490.png)

**Expected Result**
a) I don't expect to see alerts on a healthy system after the installation.

b) When receiving an alert i'd expect that it is easy to track from what system the alert originated. The link in the alert links to `http://core-alertmanager.kyma-system:9093/#/alerts?receiver=slack` which is not a valid url for the system.

I'm happy to provide access to test systems for investigations.

venturasr:
Related PR https://github.com/kyma-project/kyma/pull/1463

joek:
The issues should be fixed with 0.5 version

PR #1463, opened 2018-10-29
Set as default http-metrics instead of https-metrics for exporter-kubelets

venturasr:
Set as default http-metrics instead of https-metrics for exporter-kubelets

**Description:**
It configures exporter-kubelets to expose metrics on port http-metrics by default. Service Monitor it  has been changed.

**Related issue(s)**
https://github.com/kyma-project/kyma/issues/1359


venturasr:
https://github.com/kyma-project/kyma/issues/1359

PR #1472, opened 2018-10-30
Fix ownerReference adding in BUC

polskikiel:
**Description**

-  BUC adds an ownerReference to SBU just after it ensures that referenced binding exist.

**Related issue(s)**


components/binding-usage-controller/internal/controller/controller.go
```diff
@@ -360,6 +365,25 @@ func (c *ServiceBindingUsageController) addOwnerReference(newUsage *sbuTypes.Ser
 	return nil
 }
 
+func (c *ServiceBindingUsageController) checkDeleteOwnerRef(newUsage *sbuTypes.ServiceBindingUsage) error {
```
piotrmiskiewicz: ensureOwnerRefNotExists

PR #1482, opened 2018-10-31
Improve CPU limits for minio

derberg:
**Description**

- Improve CPU limits. Grafana shows that minio has peaks in CPU consumption that come to around 130. Decided to set the recommended official limits https://github.com/helm/charts/blob/master/stable/minio/values.yaml#L124


installation/resources/installer-config-cluster.yaml.tpl
```diff
@@ -43,6 +43,7 @@ data:
   minio.accessKey: "admin"
   minio.secretKey: "topSecretKey"
   minio.resources.limits.memory: 128Mi
+  minio.resources.limits.cpu: 250m
```
PK85: why globally? please move it to core configuration.

Issue #1422, opened 2018-10-24
Add support for basic auth authentication


lszymik:
**Description**

The Metadata service must support registration of API with basic auth credentials which will be further used by the Proxy service.

The implementation will follow the original proposal with minor changes or improvements possible. The backward compatibility must be ensured. https://github.com/kyma-project/community/blob/master/sig-and-wg/sig-core/proposals/odata-support.md

**Reasons**

OData integration requires the basic auth in development mode.

**Acceptance criteria**
- Update swagger
- Add changes in Metadata service
- Add changes in Proxy service
- Update documentation

PR #1519, opened 2018-11-05
Basic auth in proxy service

akgalwas:
**Description**

- functionality of authentication moved to the separate package 
- proxy package refactored to use the new authentication package
- handling basic auth added

**Related issue(s)**

It is part of implementation related to #1422


janmedrek:
Unit tests are failing:
```
internal/metadata/servicedefservice.go:42: WrongInput format %s reads arg #1, but call has 0 args
```

The cause is not amongst your changes, but still - we may want to fix this :)

components/proxy-service/internal/proxy/unauthorizedresponseretrier_test.go
```diff
@@ -0,0 +1,52 @@
+package proxy
+
+import (
+	"github.com/kyma-project/kyma/components/proxy-service/internal/apperrors"
+	"github.com/stretchr/testify/assert"
+	"net/http"
+	"testing"
+)
+
+func TestForbiddenResponseRetrier_CheckResponse(t *testing.T) {
```
janmedrek: We are missing two cases here:
- should retry if status code is 401/403
- should not retry if it already did and still received 401/403

components/proxy-service/internal/proxy/cache.go
```diff
@@ -0,0 +1,49 @@
...
+
+const cleanupInterval = 60
+
+type CacheEntry struct {
+	Proxy                 *httputil.ReverseProxy
+	AuthorizationStrategy authorization.Strategy
+}
+
+type Cache interface {
+	// Get returns entry fro the cache
```
janmedrek: small typo: fro -> from :)

components/proxy-service/internal/proxy/cache.go
```diff
@@ -0,0 +1,49 @@
+package proxy
+
+import (
+	gocache "github.com/patrickmn/go-cache"
+	"net/http/httputil"
+	"time"
+	"github.com/kyma-project/kyma/components/proxy-service/internal/authorization"
+)
+
+const cleanupInterval = 60
```
janmedrek: Isn't that time a bit too much? Cleaning up cache more often would reduce amount of retries that will be made due to deprecated tokens.
What is your opinion on that?

components/proxy-service/internal/metadata/serviceapi/serviceapiservice.go
```diff
@@ -19,7 +19,13 @@ type API struct {
 // Credentials contains OAuth configuration.
 type Credentials struct {
 	// Oauth is OAuth configuration.
-	Oauth Oauth
+	Oauth *Oauth
+	Basic *Basic
+}
+
+type Basic struct {
```
janmedrek: Please add a comment here (or remove the rest) to keep them consistent across this file.

components/proxy-service/internal/metadata/serviceapi/serviceapiservice.go
```diff
@@ -19,7 +19,13 @@ type API struct {
 // Credentials contains OAuth configuration.
```
janmedrek: It's not only OAuth anymore :)
Also, we could move API / Credentials / Basic / Oauth to some kind of model package (like in metadata service). What is your opinion?

components/proxy-service/internal/metadata/remoteenv/util.go
```diff
@@ -41,3 +40,16 @@ func convertFromK8sType(service v1alpha1.Service) (Service, apperrors.AppError)
 		Events:              events,
 	}, nil
 }
+
+func convertCredentialsFromK8sType(credentials v1alpha1.Credentials) *Credentials {
```
janmedrek: It would be a good idea to have both of these methods unit tested.

components/proxy-service/internal/authorization/oauth/oauthclient.go
```diff
@@ -22,53 +22,58 @@ type oauthResponse struct {
...
 }
 
-func (oc *oauthClient) requestToken(clientID string, clientSecret string, authURL string) (*oauthResponse, apperrors.AppError) {
+func (c *client) InvalidateTokenCache(clientID string) {
+	c.tokenCache.Remove(clientID)
+}
+
+func (c *client) requestToken(clientID string, clientSecret string, authURL string) (*oauthResponse, apperrors.AppError) {
 	transport := &http.Transport{
 		TLSClientConfig: &tls.Config{InsecureSkipVerify: true},
```
janmedrek: Is it a good idea to keep InsecureSkipVerify always set to true? In my opinion this should be passed as a paremeter / env variable but that's rather idea for a follow-up task. What do you think on this?

components/proxy-service/internal/authorization/externaltokenstrategy_test.go
```diff
@@ -0,0 +1,78 @@
...
+		require.NoError(t, err)
+		authHeader := request.Header.Get(httpconsts.HeaderAuthorization)
+		externalTokenHeader := request.Header.Get(httpconsts.HeaderAccessToken)
+
+		assert.Equal(t, "Bearer external", authHeader)
+		assert.Equal(t, "", externalTokenHeader)
+		oauthClientMock.AssertNotCalled(t, "GetToken")
+	})
+
+	t.Run("should use provider strategy in external token header is missing", func(t *testing.T) {
```
janmedrek: Did you mean `should use provider strategy when external token header is missing` ?

components/proxy-service/internal/authorization/factory.go
```diff
@@ -0,0 +1,77 @@
...
+		return newExternalTokenStrategy(basicAuthStrategy)
+	} else {
+		noAuthStrategy := newNoAuthStrategy()
+
+		return newExternalTokenStrategy(noAuthStrategy)
+	}
+}
+
+// Factory configuration options
+type Configuration struct {
```
janmedrek: I would suggest something more precise, like FactoryConfiguration / StrategyFactoryConfiguration :)

components/proxy-service/internal/authorization/oauth/oauthclient.go
akgalwas: There is 'skipVerify' flag used when creating ReverseProxy. I think we could use the same flag and pass it to oauthclient.

components/proxy-service/internal/metadata/remoteenv/util.go
akgalwas: Actually I'm not convinced if adding this test is vital. Those functions are tested by repository_test.Adding a separate test for functions defined in utils.go won't make repository_test.go any shorter. Maybe we should move the contents of util.go to repository.go as they are detail of repository?

components/proxy-service/internal/authorization/oauth/oauthclient.go
akgalwas: The following issue has been added : https://github.com/kyma-project/kyma/issues/1639

PR #1526, opened 2018-11-05
Refactor Metadata Service to support different authentication methods

janmedrek:
**Description**

- Metadata Service now supports basic authentication for registered APIs,
- Preparation for further auth-methods extensions.

**Related issue(s)**
It is part of implementation related to #1422


components/metadata-service/internal/externalapi/servicedetailsvalidation_test.go
```diff
@@ -289,12 +324,35 @@ func TestServiceDetailsValidator_API_OAuth(t *testing.T) {
 		// when
 		err := validator.Validate(serviceDetails)
 
+		// then
+		assert.NoError(t, err)
+	})
+
+	t.Run("should not accept Basic Auth credentials with empty basic", func(t *testing.T) {
```
akgalwas: We should not allow to register service with both authorization methods provided (I was able to register api with both oauth and basic auth)

components/metadata-service/internal/metadata/secrets/service.go
```diff
@@ -0,0 +1,127 @@
...
+const (
+	OauthClientIDKey     = "clientId"
+	OauthClientSecretKey = "clientSecret"
+
+	BasicAuthUsernameKey = "username"
+	BasicAuthPasswordKey = "password"
+)
+
+type Service interface {
+	CreateOauthSecret(remoteEnvironment, name, clientID, clientSecret, serviceID string) apperrors.AppError
```
akgalwas: This design could be simplified. Currently serviceapiservice is responsible for detecting what is the type of credentials and creating secret name. It could be a responsibility of secrets.service.
Proposed solution:
1) creating a new package for data model 
2) moving all definitions from serviceapi/model.go to the new package
3) importing the model package is secrets/service.go
4) Change secrets.Service interface to have three methods (Create, Update, Delete) . Create and update would take two arguments: remote environment name and credentials object (so that it could figure out the type of credentials and secret name)

With such a design serviceapiservice delegates all the details of secretes handling to secrets.service. It gets shorter and also the tests gets easier.

components/metadata-service/internal/metadata/secrets/service.go
```diff
@@ -0,0 +1,282 @@
...
+	}
+
+	name := s.nameResolver.GetResourceName(remoteEnvironment, serviceID)
+
+	err := s.createCredentialsSecret(remoteEnvironment, serviceID, name, credentials)
+	if err != nil {
+		return remoteEnvCredentials, err
+	}
+
+	if s.oauthCredentialsProvided(credentials) {
```
akgalwas: You could add a function converting model.Credentials to remoteEnv.Credentials and reuse it in Update method.

Issue #60, opened 2018-07-30
Helm Broker is not updated correctly by sync-bundles.sh


KarolGrzyb:
Confirm these statements before you submit the issue:

- [x] I have searched open and closed issues for duplicates.
- [x] I have read the contributing guidelines.
---

**Description**
After updating content of kyma/resources/helm-broker-repo/bundles directory I want to update repository by executing sync-bundles.sh script from kyma/resources/helm-broker-repo/development directory.

**Expected result**

Repository and index.yaml visible via web server is updated. No error messages are shown.

**Actual result**

"bundles does not exists" error message is shown, repo is not updated.

PR #1549, opened 2018-11-07
Add Kyma integration test scripts for the Prow job

suleymanakbas91:
**Description**

- Adds Kyma deployment and test script to be used for Kyma integration job on Prow
- Adds installation script of the necessary tooling for Debian to deploy Kyma on Minikube

**Related issue(s)**

See [#60](https://github.com/kyma-project/test-infra/issues/60)


prow/scripts/deploy_and_test_kyma.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
```
aszecowka: in prow directory, our scripts use "-" and "_" as a separator, please make it consistent.

prow/scripts/deploy_and_test_kyma.sh
```diff
@@ -0,0 +1,15 @@
...
+
+###
```
aszecowka: I am a little bit afraid about mixing all prow scripts in one directory. Maybe we should create a separate directory for running kyma-int on vm, and call it "kyma-integration-on-debian" ?

prow/scripts/install_deps_debian.sh
```diff
@@ -0,0 +1,55 @@
...
+sudo apt-get update
+sudo apt-get install -y \
+     apt-transport-https \
+     ca-certificates \
+     curl \
+     gnupg2 \
+     socat \
+     software-properties-common
+
+curl -fsSL https://download.docker.com/linux/debian/gpg | sudo apt-key add -
```
michal-hudy: Instead of installing it every time, maybe we should prepare a custom VM image?

suleymanakbas91: We already have an issue to investigate this [here](https://github.com/kyma-project/test-infra/issues/47). So, this shouldn't be a blocker for this PR.

Issue #1458, opened 2018-10-29
Remove default RE from Kyma


lszymik:
**Description**

Removes the default RE `ec-default` and `hmc-default`.

PR #1584, opened 2018-11-08
 Remove default re (#1458)

Maladie:
**Description**

- Default Remote Environments removed from installation scripts, charts etc.
- Tests adjusted to create RE if necessary
- Documentation updated

**Related issue(s)**
#1458


components/connector-service/docs/api/connectorapi.yaml
```diff
@@ -128,7 +128,7 @@ components:
       properties:
         url:
           type: 'string'
-          example: 'https://cert-service.kyma.cluster.cx/v1/remoteenvironments/ec-default/info?token=1edfc34g'
```
Szymongib: The file `docs/application-connector/docs/assets/connectorapi.yaml` should be the same as this one.

tests/connector-service-tests/test/testkit/k8sClient.go
```diff
@@ -0,0 +1,56 @@
...
+	"k8s.io/client-go/kubernetes"
+	restclient "k8s.io/client-go/rest"
+)
+
+type K8sResourcesClient interface {
+	CreateDummyRemoteEnvironment(name string, accessLabel string) (*v1alpha1.RemoteEnvironment, error)
+	DeleteRemoteEnvironment(name string, options *v1.DeleteOptions) error
+}
+type k8sResourcesClient struct {
+	coreClient              *kubernetes.Clientset
```
Szymongib: We do not need the core Kubernetes client as we only use the RE client.

tests/connector-service-tests/test/testkit/k8sClient.go
```diff
@@ -0,0 +1,56 @@
...
+	return &k8sResourcesClient{
+		coreClient:              coreClientset,
+		remoteEnvironmentClient: remoteEnvironmentClientset,
+		namespace:               namespace,
+	}, nil
+}
+func (c *k8sResourcesClient) CreateDummyRemoteEnvironment(name string, accessLabel string) (*v1alpha1.RemoteEnvironment, error) {
+	dummyRe := &v1alpha1.RemoteEnvironment{
+		TypeMeta:   v1.TypeMeta{Kind: "RemoteEnvironment", APIVersion: v1alpha1.SchemeGroupVersion.String()},
+		ObjectMeta: v1.ObjectMeta{Name: name, Namespace: c.namespace},
```
Szymongib: We probably can get rid of the namespace also, as RE's are not namespaced.

tests/connector-service-tests/test/apitests/connector_test.go
```diff
@@ -263,32 +270,55 @@ func TestCertificateValidation(t *testing.T) {
...
 		require.NoError(t, err)
-		require.Equal(t, http.StatusForbidden, response.StatusCode)
+		require.Equal(t, StatusForbidden, response.StatusCode)
 	})
+
+	k8sResourcesClient.DeleteRemoteEnvironment(remoteEnvName, &v1.DeleteOptions{})
+	k8sResourcesClient.DeleteRemoteEnvironment(forbiddenRemoteEnvName, &v1.DeleteOptions{})
+}
+
+func repeatUntilReIsCreated(tlsClient *Client, gatewayUrlFormat string, remoteEnvName string) (*Response, error) {
```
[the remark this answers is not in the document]
Szymongib: It seems that retries are required but the reason is that the ingress is not created therefore I would suggest changing the name to `repeatUntilIngressIsCreated`

Issue #1281, opened 2018-10-16
Authorization of graphql queries doesn't work.


magicmatatjahu:
**Description**

After creation a `ServiceRole` and `ServiceRoleBinding` for secure a specific query in `/graphql` path, the non-specific users in `ServiceRoleBinding` can query the path for resources.

**Expected result**
 
Users (not specific in `ServiceRoleBinding`) cannot query a specific `/graphql` path for resources.

**Actual result**

Authorizations of the `graphql` queries doesn't work after creation `ServiceRole` and `ServiceRoleBinding`.

**Steps to reproduce**

Create `ServiceRole` and `ServiceRoleBinding` described in [this](https://github.com/kyma-project/kyma/tree/master/resources/core/charts/ui-api#example) example, and then logIn as `user1` in `console` of Kyma and go to `/home/settings/remoteEnvs`.

piotrmsc:
Additionally, after finding the root cause, we should create a test covering this scenario.

sjanota:
For now only authentication on graphql was enabled. There is follow up issue #474 where authorisation will be brought back to live.

PR #1594, opened 2018-11-08
Enable RBAC on graphql

kubadz:
**Description**

- Delete istio webhook app and envoy filter
- Enable RBAC for graphql
- Update the readme in the ui-api chart

**Related issue(s)**

See also #1281


pkosiec:
@kubadz Why do you remove lua scripts in this PR, and you bring them back in #1595? It would be much better to keep the files here and just enable RBAC to see actual changes.

sjanota:
@pkosiec So you suggest to only disable webhook installation instead of completely removing it?

pkosiec:
@sjanota Yes, exactly

sjanota:
@pkosiec So I brought the lua scripts back, as we will use them later on. Everything else stays as it is, because it shouldn't be needed any longer.

Issue #366, opened 2018-08-30
Enable secure access for K8S dashboard


abbi-gaurav:
We are currently bundling the K8S dashboard but not exposing it as no security concept is in place.
After we enabled the JaegerUI, we can adopt the same concept for K8S dashboard and finally link it in console even.

### Jester Input
- A different security context for K8S dashboard should be created.
- The deployment to which sidecar should be applied needs to be figured out.

### AC
- [ ] The access to K8S dashboard is secure.

radufa:
About securing Kubernetes dashboard using "keycloak-proxy": 
https://itnext.io/protect-kubernetes-dashboard-with-openid-connect-104b9e75e39c

abbi-gaurav:
Please sync with @radufa about this task.

abbi-gaurav:
After having the discussion with @piotrmsc, we both share the same concern to not have default access for the cluster-admin privileges.

We have decided to park this PR and have the Istio RBAC implemented later on.
When it will happen will depend upon the capacity of both teams.

radufa:
We have to wait with the usage of Istio RBACs till the results of the investigation for a general solution for "authorization" will be finished (currently done by  @piotrmsc).

montaro:
@abbi-gaurav @radufa This means that until we get to the working istio RBAC auth. working, the dashboard will be exposed insecurely? i.e. The proxy will not be merged?!

abbi-gaurav:
@montaro We will not merge the helm charts. So until there is a clear authorization model in place, K8S Dashboard will not be accessible.

The subsequent work will be built on top of effort done so far.

abbi-gaurav:
Closing this issue and the dashboard access for minikube is handled separately.

PR #1605, opened 2018-11-09
K8s Dashboard proxy Helm chart

montaro:
**Related issue(s)**
#366

### DIFF
--- a/resources/core/charts/k8s-dashboard-proxy/README.md
+++ b/resources/core/charts/k8s-dashboard-proxy/README.md
@@ -1,0 +1,23 @@
+# Kubernetes Dashboard Proxy
+
+## Overview
+
+The `Kubernetes Dashboard Proxy` is a transparent proxy for the `Kubernetes Dashboard`
+
+Read the [Kubernetes Dashboard Proxy](https://github.com/kyma-project/kyma/blob/master/components/k8s-dashboard-proxy/README.md) document to learn more about Kubernetes Dashboard Proxy in Kyma.
+
+This chart installs [Keycloak Proxy](https://github.com/keycloak/keycloak-gatekeeper) and `reverseproxy`, a thin layer Go application as a reverse proxy.
+
+Configure these options for each business requirement:
+
+| Component                 | Configuration  | Description |
+|---------------------------| --------------:| ----------: |
+| **k8s-dashboard-proxy** |
+| |`containerPort` | The proxy listen port. |
+| **reverseproxy** |
+| |`host`| The upstream host. |
+| |`port`| The upstream port. |
+| |`secret_token_path`| K8s Service Account Secret value. |
+| |`k8s_dashboard_URL`| Kubernetes Dashboard. |
+| **keycloak** |
+| |`image`| keycloak Proxy Docker image. |
--- a/resources/core/charts/k8s-dashboard-proxy/README.md
+++ b/resources/core/charts/k8s-dashboard-proxy/README.md
@@ -1,23 +1,23 @@
-# Kubernetes Dashboard Proxy
+# Kubernetes Dashboard proxy
 
 ## Overview
 
-The `Kubernetes Dashboard Proxy` is a transparent proxy for the `Kubernetes Dashboard`
+The Kubernetes Dashboard proxy is a transparent proxy for the Kubernetes Dashboard. Read the [Kubernetes Dashboard proxy](https://github.com/kyma-project/kyma/blob/master/components/k8s-dashboard-proxy/README.md) document to learn more about Kubernetes Dashboard proxy implementation in Kyma.
 
-Read the [Kubernetes Dashboard Proxy](https://github.com/kyma-project/kyma/blob/master/components/k8s-dashboard-proxy/README.md) document to learn more about Kubernetes Dashboard Proxy in Kyma.
+## Details
 
-This chart installs [Keycloak Proxy](https://github.com/keycloak/keycloak-gatekeeper) and `reverseproxy`, a thin layer Go application as a reverse proxy.
+This chart installs [Keycloak proxy](https://github.com/keycloak/keycloak-gatekeeper) and **reverseproxy**, a thin layer Go application.
 
-Configure these options for each business requirement:
+Configure the following options for each business requirement:
 
 | Component                 | Configuration  | Description |
 |---------------------------| --------------:| ----------: |
 | **k8s-dashboard-proxy** |
-| |`containerPort` | The proxy listen port. |
+| |`containerPort` | The port used by the proxy to listen on. |
 | **reverseproxy** |
 | |`host`| The upstream host. |
 | |`port`| The upstream port. |
-| |`secret_token_path`| K8s Service Account Secret value. |
-| |`k8s_dashboard_URL`| Kubernetes Dashboard. |
+| |`secret_token_path`| The path of Kubernetes Service Account Secret. |
+| |`k8s_dashboard_URL`| The Kubernetes Dashboard URL. |
 | **keycloak** |
-| |`image`| keycloak Proxy Docker image. |
+| |`image`| The Keycloak proxy Docker image. |
--- a/resources/core/charts/k8s-dashboard-proxy/templates/deployment.yaml
+++ b/resources/core/charts/k8s-dashboard-proxy/templates/deployment.yaml
@@ -1,0 +1,85 @@
+apiVersion: extensions/v1beta1
+kind: Deployment
+metadata:
+  name: {{ template "name" . }}
+  namespace: {{ .Release.Namespace }}
+spec:
+  replicas: 1
+  template:
+    metadata:
+      name: {{ template "name" . }}
+      namespace: {{ .Release.Namespace }}
+      annotations:
+        sidecar.istio.io/inject: "true"
+      labels:
+        app: {{ template "name" . }}
+        release: {{ .Release.Name }}
+    spec:
+      selector:
+        app: {{ template "name" . }}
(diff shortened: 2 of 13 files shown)


resources/core/charts/k8s-dashboard-proxy/README.md
```diff
@@ -0,0 +1,23 @@
...
+Configure these options for each business requirement:
+
+| Component                 | Configuration  | Description |
+|---------------------------| --------------:| ----------: |
+| **k8s-dashboard-proxy** |
+| |`containerPort` | The proxy listen port. |
+| **reverseproxy** |
+| |`host`| The upstream host. |
+| |`port`| The upstream port. |
+| |`secret_token_path`| K8s Service Account Secret value. |
```
piotrmsc: Is it a path to service account or is the value of SA token ?

resources/core/charts/k8s-dashboard-proxy/templates/deployment.yaml
```diff
@@ -0,0 +1,84 @@
...
+          requests:
+            memory: 128Mi
+        {{ end }}
+        imagePullPolicy: IfNotPresent
+        args:
+          - --discovery-url=https://dex.{{ .Values.global.domainName }}
+          - --skip-openid-provider-tls-verify=true
+          - --client-id=kubernetes-dashboard
+          - --client-secret=94e412ebd7c9440f6947972a1bb35698f007ba2d6401d65a8c2e8490a453d472
+          - --resources=uri=/*|methods=GET,POST,PUT,DELETE
```
piotrmsc: Would be nice to have this secret parametrized to allow overriding on different envs

resources/core/charts/k8s-dashboard-proxy/templates/deployment.yaml
```diff
@@ -0,0 +1,84 @@
...
+          limits:
+            memory: 256Mi
+          requests:
+            memory: 128Mi
+        {{ end }}
+        imagePullPolicy: IfNotPresent
+        args:
+          - --discovery-url=https://dex.{{ .Values.global.domainName }}
+          - --skip-openid-provider-tls-verify=true
+          - --client-id=kubernetes-dashboard
```
piotrmsc: Hmm, I assume you skip this verification due to self-signed TLS cert on minikube, but why It shouldn't be validated on a cluster env?

resources/core/charts/k8s-dashboard-proxy/templates/rbac.yaml
```diff
@@ -0,0 +1,25 @@
+apiVersion: v1
```
piotrmsc: If you provide a SA with cluster-admin privileges here, and access to this dashboard is only limited by authentication against dex, then here is a security blocker. Any user in the kyma, with any privileges will have access to ie. all secrets in the cluster

resources/core/charts/k8s-dashboard-proxy/README.md
```diff
@@ -0,0 +1,23 @@
+# Kubernetes Dashboard Proxy
```
bszwarc:
```suggestion
# Kubernetes Dashboard proxy
```

resources/core/charts/k8s-dashboard-proxy/README.md
```diff
@@ -0,0 +1,23 @@
...
+
+## Overview
+
+The `Kubernetes Dashboard Proxy` is a transparent proxy for the `Kubernetes Dashboard`
```
bszwarc:
```suggestion
The Kubernetes Dashboard proxy is a transparent proxy for the Kubernetes Dashboard.
```

resources/core/charts/k8s-dashboard-proxy/README.md
```diff
@@ -0,0 +1,23 @@
...
+
+Read the [Kubernetes Dashboard Proxy](https://github.com/kyma-project/kyma/blob/master/components/k8s-dashboard-proxy/README.md) document to learn more about Kubernetes Dashboard Proxy in Kyma.
+
+This chart installs [Keycloak Proxy](https://github.com/keycloak/keycloak-gatekeeper) and `reverseproxy`, a thin layer Go application as a reverse proxy.
```
bszwarc:
```suggestion
This chart installs [Keycloak Proxy](https://github.com/keycloak/keycloak-gatekeeper) and `reverseproxy`, a thin layer Go application.
```

resources/core/charts/k8s-dashboard-proxy/README.md
bszwarc: I agree, a bit confusing :)
Also:
K8s > `Kubernetes`

resources/core/charts/k8s-dashboard-proxy/README.md
```diff
@@ -0,0 +1,23 @@
...
+
+Configure these options for each business requirement:
+
+| Component                 | Configuration  | Description |
+|---------------------------| --------------:| ----------: |
+| **k8s-dashboard-proxy** |
+| |`containerPort` | The proxy listen port. |
```
bszwarc:
```suggestion
| |`containerPort` | The port used by the proxy to listen on. |

```
I think this is a bit better - we can also go with `The proxy listening port` but I think the one above is a bit better.

resources/core/charts/k8s-dashboard-proxy/README.md
```diff
@@ -0,0 +1,23 @@
...
+| **reverseproxy** |
+| |`host`| The upstream host. |
+| |`port`| The upstream port. |
+| |`secret_token_path`| K8s Service Account Secret value. |
+| |`k8s_dashboard_URL`| Kubernetes Dashboard. |
```
bszwarc:
```suggestion
| |`k8s_dashboard_URL`| The Kubernetes Dashboard. |
```

resources/core/charts/k8s-dashboard-proxy/README.md
```diff
@@ -0,0 +1,23 @@
...
+| **keycloak** |
+| |`image`| keycloak Proxy Docker image. |
```
bszwarc:
```suggestion
| |`image`| The Keycloak proxy Docker image. |
```

resources/core/charts/k8s-dashboard-proxy/README.md
```diff
@@ -0,0 +1,23 @@
+# Kubernetes Dashboard Proxy
+
+## Overview
+
+The `Kubernetes Dashboard Proxy` is a transparent proxy for the `Kubernetes Dashboard`
+
+Read the [Kubernetes Dashboard Proxy](https://github.com/kyma-project/kyma/blob/master/components/k8s-dashboard-proxy/README.md) document to learn more about Kubernetes Dashboard Proxy in Kyma.
```
bszwarc:
```suggestion
Read the [Kubernetes Dashboard proxy](https://github.com/kyma-project/kyma/blob/master/components/k8s-dashboard-proxy/README.md) document to learn more about Kubernetes Dashboard proxy in Kyma.
```
Additionally, I would add the following subchapter right below the link:

`##Details `

So above we have the overview and a link to further resources, and below the details of chart configuration.

resources/core/charts/k8s-dashboard-proxy/README.md
```diff
@@ -0,0 +1,23 @@
...
+
+This chart installs [Keycloak Proxy](https://github.com/keycloak/keycloak-gatekeeper) and `reverseproxy`, a thin layer Go application as a reverse proxy.
+
+Configure these options for each business requirement:
```
bszwarc:
```suggestion
Configure the following options for each business requirement:
```

Question: What exactly do we mean by a business requirement here?

resources/core/charts/k8s-dashboard-proxy/templates/rbac.yaml
radufa: Yes, "cluster-admin" should be used only on minikube and "kyma-admin" on a real cluster.

abbi-gaurav: After having the discussion with @piotrmsc, we both share the same concern to not have default access for the cluster-admin privileges. 

We have decided to park this PR and have the Istio RBAC implemented later on.
When it will happen will depend upon the capacity of both teams.

abbi-gaurav: @piotrmsc How about this idea?
* Use `kyma-admin` role for the cluster deployment.
* Use `cluster-admin` role for the local/minikube deployment.

In this way, we allow the same privileges as we do with our kubeconfig.

Would it be possible for you and @radufa to discuss this on Monday and reach a conclusion?

Next steps would be to then modify this PR based on the above proposal if this is agreed upon.

piotrmsc: @abbi-gaurav  Have you checked wheter `kyma-admin` will be enough? The problem is not with the binding for the SA but with lack of authorization here so we have to allow this access to only certain group of people, for example with kyma-admin assigned role.

Regarding kubeconfig you download from the console, it works a bit different. Downloaded kubeconfig consists your id token and based on you're assigned RBAC role, authorization proxy checked if you have access to the requested resources so you will not be able to read secrets if your assigned RBAC role does not allow it.

abbi-gaurav: The verification if `kyma-admin` is enough still needs to be done as mentioned in the previous comment.

abbi-gaurav:
Closing this PR as dashboard access for minikube is handled separately.

PR #1617, opened 2018-11-09
Use patch image with mTLS verification

sjanota:
**Description**

**Related issue(s)**


sjanota:
Awaits successful test run

PR #1521, opened 2018-11-05
Split API controller tests

sjanota:
**Description**

Changes proposed in this pull request:

- split api-controller tests into two parts: integration with istio and api-controller component tests
- integration tests with istio are more heavy-weight tests, thus there are less scenarios which checks if api-controller creates CRs understandable by istio
- component tests test api-controller in isolation, which means they verify only direct output of the api-controller (created CRs) and not the effect they have on istio. They are lighter and there is more of them. They are meant to check api-controller internals easier. 

What's the gain? Tests are more reliable as bulk of functionality is tested without interaction with istio and execute three times faster.

**Related issue(s)**

PR #1636, opened 2018-11-13
Split API controller tests 

sjanota:
**Description**

split api-controller tests into two parts: integration with istio and api-controller component tests
integration tests with istio are more heavy-weight tests, thus there are less scenarios which checks if api-controller creates CRs understandable by istio
component tests test api-controller in isolation, which means they verify only direct output of the api-controller (created CRs) and not the effect they have on istio. They are lighter and there is more of them. They are meant to check api-controller internals easier.
What's the gain? Tests are more reliable as bulk of functionality is tested without interaction with istio and execute three times faster.

**Related issue(s)**

Recreated #1521


tests/api-controller-acceptance-tests/apicontroller/component_test.go
```diff
@@ -0,0 +1,397 @@
...
+			_, err := kymaClient.GatewayV1alpha2().Apis(namespace).Create(api)
+			So(err, ShouldNotBeNil)
+		})
+
+		Convey("create API with default jwt configuration to enable authentication", func() {
+			t.Log("create API with default jwt configuration to enable authentication")
+
+			testId := ctx.generateTestId(testIdLength)
+			t.Logf("Running test: %s", testId)
+			api := ctx.apiFor(testId, domainName, apiSecurityDisabled, true)
```
kubadz: Why do you create api with parameter apiSecurityDisabled and then manually set AuthenticationEnabled to true below? Wouldn't that make more sense to just use an apiSecurityEnabled parameter (deleted by you) instead?

tests/api-controller-acceptance-tests/apicontroller/component_test.go
```diff
@@ -0,0 +1,397 @@
...
+
+			authEnabled = false
+			createdApi.Spec.AuthenticationEnabled = &authEnabled
+
+			updatedApi, err := kymaClient.GatewayV1alpha2().Apis(namespace).Update(createdApi)
+			So(err, ShouldBeNil)
+			So(updatedApi, ShouldNotBeNil)
+			So(updatedApi.ResourceVersion, ShouldNotBeEmpty)
+
+			updatedApi, err = ctx.awaitApiCreated(kymaClient, updatedApi, false, true)
```
kubadz: We should rename the function awaitApiCreated if we use it also for updates. For me, that name is misleading, as we do not wait for API creation but for API update.

tests/api-controller-acceptance-tests/apicontroller/component_test.go
```diff
@@ -0,0 +1,397 @@
...
+			So(err, ShouldBeNil)
+			vs, err := istioNetClient.NetworkingV1alpha3().VirtualServices(namespace).Get(lastApi.Status.VirtualServiceStatus.Resource.Name, metav1.GetOptions{})
+			So(err, ShouldBeNil)
+
+			err = kymaClient.GatewayV1alpha2().Apis(namespace).Delete(lastApi.Name, &metav1.DeleteOptions{})
+			So(err, ShouldBeNil)
+
+			time.Sleep(5 * time.Second)
+
+			_, err = kymaClient.GatewayV1alpha2().Apis(namespace).Get(lastApi.Name, metav1.GetOptions{})
```
kubadz: Maybe we should create the function awaitApiDeleted for a consistency with create/update. What do you think?

sjanota: I think it's better how it is right now, because those three checks are strict requirements: api-controller must clean all dependent resources. Extracting them to different method would hide it and make it look like a cleanup rather then expectation.

kubadz: The same can be told about using awaitApiChanged, where you "hide" expectations after creating/updating API. IMO hiding that doesn't make the code unclear. But it is more a personal preference so let's leave it unchanged.

PR #1666, opened 2018-11-14
Add grafana dashboard for heptio ark

mwieczorek:
**Description**

- Add grafana dashboard for heptio ark
- fix ark's service monitor (proper namespace)

**Related issue(s)**

### DIFF
--- a/resources/ark/templates/service-monitor.yaml
+++ b/resources/ark/templates/service-monitor.yaml
@@ -8,7 +8,7 @@ metadata:
     release: "{{ .Release.Name }}"
     heritage: "{{ .Release.Service }}"
   name: {{ template "ark.fullname" . }}
-  namespace: kyma-system
+  namespace: {{ .Values.prometheus.serviceMonitorNamespace }}
 spec:
   endpoints:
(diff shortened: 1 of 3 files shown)


mwieczorek:
image

example dashboard

resources/ark/templates/service-monitor.yaml
```diff
@@ -8,6 +8,7 @@ metadata:
     release: "{{ .Release.Name }}"
     heritage: "{{ .Release.Service }}"
   name: {{ template "ark.fullname" . }}
+  namespace: kyma-system
```
venturasr: `namespace:` should be defined as a template. Also can be removed as `helm `take care of the namespaces.

mwieczorek: ark is installed in a different namespace than Prometheus operator so I need to put the namespace for the service monitor explicitly. Without it the service monitor won't be picked up by the operator.
I extracted the value to the values.yaml

mwieczorek:
I've also made some changes to the dashboard. New layout:

image

Issue #1398, opened 2018-10-23
Restore blocker: REB should not recreate eventactivation when the serviceinstance is recreated


mwieczorek:
**Description**
When we restore a remote environment and related serviceinstances the remote environment broker tries to recreate eventactivation even if the resource already exists

**Expected result**
REB should check if EA exists and not create one if so

**Actual result**
We get an error:

```
OrphanMitigation - Provision call failed: provisioning failed while creating EventActivation on error:
while creating EventActivation with name: "ac031e8c-9aa4-4cb7-8999-0d358726ffaa" in 
namespace: "production": eventactivations.applicationconnector.kyma-project.io "ac031e8c-9aa4-4cb7-8999-0d358726ffaa" 
already exists @ 2018-10-23 11:58:21 +0000 UTC
```

**Steps to reproduce**

- Install kyma
- `kc apply -f https://raw.githubusercontent.com/kyma-project/kyma/master/components/remote-environment-broker/contrib/examples/remote-env.yaml`
- bind remote env to one of the environments, f.e. production
- add serviceinstance (class: promotions)
- create ark backup:
```
ark create backup reb-cs --include-resources remoteenvironment
ark create backup reb-prod --include-namespaces production
```
- install a new cluster with kyma
- restore:
```
ark restore create reb-cs --from-backup reb-cs
ark restore create reb-prod --from-backup reb-prod
```

**Troubleshooting**

polskikiel:
It seems to be the same error as here: #546

jasiu001:
**Order of restoration resources**

I tested twice restore backup operation to check which resources are restored first.

Result of first test:
1) ApplicationMapping
2) Application (2 sec later)
3) EventactivAtions (3 sec later)
4) ServiceInstance (9 sec later)

Result of second attempt
1) Application
2) ApplicationMapping (1sec later)
3) EventactivAtions (27 sec later)
4) ServiceInstance (14 sec later)

In both cases error from https://github.com/kyma-project/kyma/issues/1398#issuecomment-451919035 appears

jasiu001:
**TO DO**
Add all ServiceInstance logs. Add first reason of failed for ServiceInstance.

polskikiel:
Ok, so there was 2 troubles which are already addressed:
1. Not updated RBAC `update` role for EA. #2230 
[...]

PR #1667, opened 2018-11-14
Handle already existing EventActivation

polskikiel:
**Description**

- REB won't break provisioning logic when it's trying to create already existing EventActivation.

**Related issue(s)**

#1398

### DIFF
--- a/components/remote-environment-broker/internal/broker/provision.go
+++ b/components/remote-environment-broker/internal/broker/provision.go
@@ -13,6 +13,7 @@ import (
 	"github.com/pkg/errors"
 	osb "github.com/pmorie/go-open-service-broker-client/v2"
 	"github.com/sirupsen/logrus"
+	apiErrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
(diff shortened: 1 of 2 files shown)


components/remote-environment-broker/internal/broker/provision.go
```diff
@@ -220,10 +221,15 @@ func (svc *ProvisionService) createEaOnSuccessProvision(reName, reID, ns string,
 				SourceID:    reName,
 			},
 		})
-	if err != nil {
+	switch {
+	case err == nil:
+		svc.log.Infof("Created EventActivation: [%s], in namespace: [%s]", reID, ns)
+	case apiErrors.IsAlreadyExists(err):
+		svc.log.Infof("%v: EventActivation `%s` already exists", err, reID)
```
mszostok: you should perform update action in case of "already exists" error. In this scenario, you are not updating the owner reference which was changed. Please also add a business comment why we are performing such action. Describe a scenario when sth like this can happened because this is not a normal happy-path.

components/remote-environment-broker/internal/broker/provision_test.go
```diff
@@ -269,6 +272,90 @@ func TestProvisionErrorOnCreatingEventActivation(t *testing.T) {
 	}
 }
 
+func TestProvisionAlreadyExistsErrorOnCreatingEventActivation(t *testing.T) {
```
mszostok: you test is more like a copy of the `TestProvisionErrorOnCreatingEventActivation` please make that more DRY or even consider to do the table test.

components/remote-environment-broker/internal/broker/provision.go
polskikiel: fixed and added comment

components/remote-environment-broker/internal/broker/provision_test.go
polskikiel: I did table test for all cases.

components/remote-environment-broker/internal/broker/provision.go
```diff
@@ -200,30 +201,37 @@ func (svc *ProvisionService) createEaOnSuccessProvision(reName, reID, ns string,
...
+	_, err = svc.reClient.EventActivations(ns).Create(ea)
+	switch {
+	case err == nil:
+		svc.log.Infof("Created EventActivation: [%s], in namespace: [%s]", reID, ns)
+	case apiErrors.IsAlreadyExists(err):
+		// We perform update action to adjust OwnerReference of the EventActivation after the backup restore.
+		_, err := svc.reClient.EventActivations(ns).Update(ea)
+		if err != nil {
+			return errors.Wrapf(err, "while updating EventActivation with name: %q in namespace: %q", reID, ns)
+		}
```
mszostok: add `svc.log.Infof` log that EA was updated, thanks to that it will be symmetrical with creating path

components/remote-environment-broker/internal/broker/provision_test.go
```diff
@@ -189,7 +191,7 @@ func TestProvisionWhenProvisioningInProgress(t *testing.T) {
 	assert.Equal(t, &expOpKey, actResp.OperationKey)
 }
 
-func TestProvisionErrorOnCreatingEventActivation(t *testing.T) {
+func TestProvisionCreatingEventActivation(t *testing.T) {
 	// GIVEN
```
mszostok: all `given` statements should be inside the `t.Run` function

you cannot reuse assertion between different test cases

components/remote-environment-broker/internal/broker/provision_test.go
```diff
@@ -226,46 +225,73 @@ func TestProvisionErrorOnCreatingEventActivation(t *testing.T) {
...
-		clientset.ApplicationconnectorV1alpha1(),
-		mockInstanceStorage,
-		mockOperationIDProvider, spy.NewLogDummy())
+	clientset := fake.NewSimpleClientset(fixEventActivation())
 
-	asyncFinished := make(chan struct{}, 0)
-	sut.asyncHook = func() {
-		asyncFinished <- struct{}{}
+	tests := map[string]func(){
+		"some error when creating EA": func() {
```
mszostok: "some error" -> "generic error", same for other occurences

Issue #1576, opened 2018-11-08
Metadata Service returns 500 when RE does not exist


Szymongib:
**Description**

Metadata Service returns 500 error when calling the service for Remote Environment that does not exist.

**Acceptance criteria**

The service should return 404 with an appropriate message indicating that RE does not exist.

PR #1686, opened 2018-11-15
Metadata Service returns correct error code when RE does not exist

franpog859:
**Description**

- Metadata Service returns error with "not found" status when calling the Remote Environment service that does not exist
- Error message in this case says that RE does not exist
- There is an option to append error messages without creating new one with new error code `err.Append("Some additional message")` instead of `apperrors.Internal("Some additional message, %s", err.Error())`
- Error checking in the unit tests with the `require` function makes developer sure that single test will stop if error is unexpected

**Related issue(s)**

See also #1576

### DIFF
--- a/components/metadata-service/internal/metadata/servicedefservice_test.go
+++ b/components/metadata-service/internal/metadata/servicedefservice_test.go
@@ -1317,16 +1317,13 @@ func TestServiceDefinitionService_Delete(t *testing.T) {
 
 	t.Run("should return an error if API deletion failed", func(t *testing.T) {
 		// given
-		serviceRepository := new(remoteenvmocks.ServiceRepository)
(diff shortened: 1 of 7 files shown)


components/metadata-service/internal/metadata/servicedefservice_test.go
```diff
@@ -660,11 +718,26 @@ func TestServiceDefinitionService_GetAll(t *testing.T) {
...
 		// when
 		result, err := service.GetAll("re")
-		require.NoError(t, err)
 
 		// then
+		require.NoError(t, err)
 		assert.Len(t, result, 0)
 	})
+
+	t.Run("should return 404 if cannot find Remote Environment", func(t *testing.T) {
```
Szymongib: `t.Run("should return not found error if cannot find Remote Environment", func(t *testing.T) {`

Issue #1279, opened 2018-10-16
Update service-catalog


piotrmiskiewicz:
AC:
- update Service Catalog to newer version than  0.1.34
- check status of kubernetes-incubator/service-catalog#2291
- run test for service catalog, from https://github.com/kyma-incubator/service-catalog-tester.
- if update done adjust a name of the ticket to the given version, and create a new one for the next sprint

PR #1692, opened 2018-11-16
Update service catalog version

jasiu001:
Related to #1279


piotrmiskiewicz:
missing `minReadySeconds` in deployments

PR #1701, opened 2018-11-16
Use containerTemplate in podTemplate for buildpack in Jenkinsfiles

crabtree:
**Description**

Our pipelines for components use buildpack to make the magic. This PR moves this buildpack container from simple `docker run` command to a slave pod runtime.

- Use containerTemplate in podTemplate for buildpack in Jenkinsfiles

**Related issue(s)**

See also #1703

### DIFF
--- a/components/binding-usage-controller/Jenkinsfile
+++ b/components/binding-usage-controller/Jenkinsfile
@@ -1,9 +1,16 @@
 #!/usr/bin/env groovy
-
 def label = "kyma-${UUID.randomUUID().toString()}"
 def application = "binding-usage-controller"
 def dockerPushRoot = "${env.DOCKER_REGISTRY}${params.PUSH_DIR}"
 def dockerImageTag = params.APP_VERSION
+def registry = env.DOCKER_REGISTRY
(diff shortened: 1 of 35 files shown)


components/binding-usage-controller/Jenkinsfile
```diff
@@ -1,9 +1,16 @@
...
 def application = "binding-usage-controller"
 def dockerPushRoot = "${env.DOCKER_REGISTRY}${params.PUSH_DIR}"
 def dockerImageTag = params.APP_VERSION
+def registry = env.DOCKER_REGISTRY
+def buildpack = 'golang-buildpack:0.0.9'
+def appFolder = env.APP_FOLDER
+def repositoryName = 'kyma'
+
+def workDir = ""
+def checkoutDir = ""
```
mszostok: why it cannot by simply inlined here?
you are creating it by concatenating hardcoded value + env variable defined above.
```
checkoutDir = "src/github.com/kyma-project/$repositoryName"
```

I do not see any reason to have it defined in the setup stage

same story for the componentDir variable

crabtree: Done

Issue #1711, opened 2018-11-19
Metadata Service tests failing in stability-check


janmedrek:
**Description**

Metadata Service tests are failing in stability checks:
```
Testing 'metadata-service-tests'
'metadata-service-tests' has Failed status
Fetching logs from 'metadata-service-tests' with options tailLimit=2000 and bytesLimit=500000
Metadata-service tests
Run api tests
2018/11/19 01:42:37 Read configuration: {MetadataServiceUrl:http://metadata-service-external-api.kyma-integration.svc.cluster.local:8081 Namespace:kyma-integration}
2018/11/19 01:42:38 Read configuration: {MetadataServiceUrl:http://metadata-service-external-api.kyma-integration.svc.cluster.local:8081 Namespace:kyma-integration}
2018/11/19 01:42:45 Read configuration: {MetadataServiceUrl:http://metadata-service-external-api.kyma-integration.svc.cluster.local:8081 Namespace:kyma-integration}
PASS
Run kubernetes tests
2018/11/19 01:42:58 Read configuration: {MetadataServiceUrl:http://metadata-service-external-api.kyma-integration.svc.cluster.local:8081 Namespace:kyma-integration}
--- FAIL: TestK8sResources (10.92s)
    --- FAIL: TestK8sResources/when_creating_service_only_with_OAuth_API (2.89s)
    	metadata_api_client.go:208:
    		--------------------------------
    		HTTP/1.1 200 OK
    		Content-Length: 46
    		Content-Type: application/json;charset=UTF-8
    		Date: Mon, 19 Nov 2018 01:43:01 GMT
    		Server: envoy
    		X-Envoy-Decorator-Operation: metadata-service-external-api.kyma-integration.svc.cluster.local:8081/*
    		X-Envoy-Upstream-Service-Time: 223

    		{"id":"f8d58ad1-1314-468f-bef9-98e115ed5ae1"}

    		--------------------------------
        --- FAIL: TestK8sResources/when_creating_service_only_with_OAuth_API/should_add_service_to_remote_environment_custom_resource (0.01s)
        	require.go:157:
        			Error Trace:	k8sresources_check.go:103
        			            				k8sresource_test.go:110
        			Error:      	Not equal:
        			            	expected: "http://oauth.com"
        			            	actual  : ""

        			            	Diff:
        			            	--- Expected
        			            	+++ Actual
        			            	@@ -1 +1 @@
        			            	-http://oauth.com
        			            	+
[...]

PR #1740, opened 2018-11-21
Fix unstable Metadata-Service tests

janmedrek:
**Description**

- Tests now skip inappropriate checks based on auth type
- Tests now wait before executing checks so data can be propagated into the CRs

**Related issue(s)**

#1711

### DIFF
--- a/tests/metadata-service-tests/test/k8stests/k8sresource_test.go
+++ b/tests/metadata-service-tests/test/k8stests/k8sresource_test.go
@@ -13,6 +13,8 @@ import (
 	"k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
+const crPropagationWaitTime = 10
+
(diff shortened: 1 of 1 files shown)


tests/metadata-service-tests/test/k8stests/k8sresource_test.go
```diff
@@ -54,6 +55,9 @@ func TestK8sResources(t *testing.T) {
 
 		expectedLabels := map[string]string{"re": dummyRE.Name, "serviceId": serviceId}
 
+		// wait for changes to propagate in the CR
+		time.Sleep(10 * time.Second)
```
franpog859: I suggest to export `10 * time.Second` to some const like `waitForChangesPropagationInCR`. No comments would be needed and the value would be changeable in easier way.

Issue #1152, opened 2018-10-05
Get Application Connector connection link without a port-forward command


lszymik:
**Description**

To acquire a link to Connector service with a one-time token, a user needs to make a `kubectl port-forward` or use Console UI. 

The alternative method must be created to avoid usage of `port-forward`. 

**Reasons**

We need to avoid `port-forward` as it will not work with Kubeconfig file from Console UI. Moreover, we need to stick to K8S best practices.

Without it, the automation of one-click integration will be difficult.

**Acceptance Criteria**

- Customer resource to trigger a token generation.
- The token field in the status.
- CR will be deleted after 5 minutes.
- Update documentation (replace port-forward from AC docs, describe a new CRD)
- Test with Kubeconfig acquired from Console UI

crabtree:
Implementation follows investigation done in #1461

PR #1742, opened 2018-11-21
Connection-token-handler component source code

crabtree:
**Description**

- Controller for connector service token requests implementation

**Related issue(s)**

See also #1152


components/connector-service-token-request/Dockerfile
```diff
@@ -0,0 +1,16 @@
...
+ARG DOCK_PKG_DIR=/go/src/github.com/kyma-project/kyma/components/connector-service-token-request
+
+RUN mkdir -p $DOCK_PKG_DIR
+
+COPY ./ $DOCK_PKG_DIR
+WORKDIR $DOCK_PKG_DIR
+
+RUN CGO_ENABLED=0 GOOS=linux go build -a -o manager ./cmd/manager
+
+FROM alpine
```
a-thaler: we should use fixed versions. In my understanding that image is like "latest"

crabtree: Good catch! I missed that. Fixed.

components/connection-token-handler/cmd/manager/manager.go
```diff
@@ -0,0 +1,66 @@
...
+	if err != nil {
+		log.Error(err, "unable to set up overall controller manager")
+		os.Exit(1)
+	}
+
+	log.Info("Registering Components.")
+
+	log.Info("setting up scheme")
+	if err := apis.AddToScheme(mgr.GetScheme()); err != nil {
+		log.Error(err, "unable add APIs to scheme")
```
janmedrek:
```suggestion
		log.Error(err, "unable to add APIs to scheme")
```

components/connection-token-handler/sample-cr.yaml
```diff
@@ -0,0 +1,5 @@
+
```
janmedrek: Do we need this sample? If yes - shouldn't it be placed somewhere near the CR and not inside controller's code?

crabtree: No we don't, removed :)

PR #1746, opened 2018-11-21
Add possibility to enable steps based on feature-gate

sjanota:
**Description**

Installation CR contains new field `feature`. This enables given step only if this feature is enabled. Installer loads enabled feature-gates from ConfigMap `installer-feature-gates`.

**Related issue(s)**


sjanota:
Will not be done for know as it would need also ability to disable steps based on features and it gets to complicated. Needs proper design later.

PR #1747, opened 2018-11-21
Add feature gates support to installation scripts

sjanota:
**Description**

Script `run.sh` accepts additional flag `--feature-gates`. Passed feature gates are added to `installer-feature-gates` ConfigMap which is read but installer (#1746)

**Related issue(s)**

Uses #1746


sjanota:
Will not be done for know as it would need also ability to disable steps based on features and it gets to complicated. Needs proper design later.

PR #1748, opened 2018-11-21
Add knative-installer component

sjanota:
**Description**

There is new component: knative-installer. It is a kubernetes job which installs knative along with kyma.

**Related issue(s)**

Based on @Abd4llA's #1360. Changes taken from there:
* components/knative-installer/install.sh
* components/knative-installer/is-ready.sh

### DIFF
--- a/components/knative-installer/README.md
+++ b/components/knative-installer/README.md
@@ -2,8 +2,8 @@
 
 ## Overview
 
-Knative is distributes as single YAML file with resources definitions. `knative-installer` containerizes the process to make it possible to install Knative with Kyma's installer.
+Knative is distributed as a single YAML file containing resource definitions. The Knative installer containerizes the process to allow the installation of Knative using the Kyma installer.
 
 ## Prerequisites
 
-Knative requires istio in `istio-system` namespace. It must be installed before `isito-kyma-patch`.
+Knative requires running Istio in the `istio-system` Namespace. It must be installed before `isito-kyma-patch` is applied.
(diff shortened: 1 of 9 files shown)


components/knative-installer/install.sh
```diff
@@ -0,0 +1,41 @@
+#!/usr/bin/env bash
+######### Knative build & serving ########
+
+echo "Installing Knative build and serving ..."
+
+DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null && pwd )"
+
+curl -L ${SERVING_URL} \
```
pbochynski: I don't like the idea we download the serving sources during installation. Can't we put it in the docker image? This way we can have quicker and repeatable installations.

sjanota: I think it's reasonable as we depend on knative release structure in scripts anyway. There is no point in having it configurable.

sjanota: Fixed

components/knative-installer/README.md
```diff
@@ -0,0 +1,9 @@
+## Knative installer
+
+## Overview
+
+Knative is distributes as single YAML file with resources definitions. `knative-installer` containerizes the process to make it possible to install Knative with Kyma's installer.
```
tomekpapiernik:
```suggestion
Knative is distributed as a single YAML file containing resource definitions. The Knative installer containerizes the process to allow the installation of Knative using the Kyma installer.
```

components/knative-installer/README.md
```diff
@@ -0,0 +1,9 @@
...
+
+## Prerequisites
+
+Knative requires istio in `istio-system` namespace. It must be installed before `isito-kyma-patch`.
```
tomekpapiernik:
```suggestion
Knative requires running Istio in the `istio-system` Namespace. It must be installed before `isito-kyma-patch` is applied.
```

Issue #1676, opened 2018-11-15
ServiceBindingUsage doesn't work


PK85:
**Description**

When ServiceBindingUsage created nothing happen, credentials are not injected. ServiceBindingUsage controller logs show that informers for CSC and SC are failing. I've seen that ServiceCatalog apiserver and controller were restarted in the past, but work as expected. See logs.

**Expected result**

SBU works.

**Actual result**

SBU doesn't work.

**Steps to reproduce**
Restart ServiceCatalog apiserver and  CSC and SC will disappear.

**Troubleshooting**
SBU controller restart solves

PK85:
[buc-cannot-start-sc-was-restarted.txt](https://github.com/kyma-project/kyma/files/2584159/buc-cannot-start-sc-was-restarted.txt)

jasiu001:
**Recap**
I tried reproduce the issue but neither locally neither on azure cluster I could not achieve state from description. 

**Steps to try reproduce on minikube**
- I reset Service Catalog for some period of time 1m, 5m, after starting everything was working fine
- I reset Service Catalog for more than 10m (max idle time 30m), sometimes after starting BindingServiceController could not fetch BindingUsage from cache during 15 attempts, which is default max retires for controller. Information about that wasn't shown in status message of ServiceBindingUsage, I updated it in PR #1761. This issue is not connected with bug description.

**Steps on azure cluster**
- I created 5 instances of redis and 4 lambda functions, I made ServiceBindingUsage between them
- I reset Service catalog for more than one hour, after starting I could create new ServiceBindingUsages and old connections worked correct.

PR #1761, opened 2018-11-22
Update SBU status messages

jasiu001:
Ralated to #1676; Upgrade message SBU status information.

### DIFF
--- a/components/binding-usage-controller/internal/controller/controller.go
+++ b/components/binding-usage-controller/internal/controller/controller.go
@@ -214,15 +214,14 @@ func (c *ServiceBindingUsageController) processNextWorkItem() bool {
 		return true
 	}
 
-	retry := c.maxRetires
+	retry := c.queue.NumRequeues(key)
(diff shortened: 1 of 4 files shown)


components/binding-usage-controller/internal/controller/controller.go
```diff
@@ -207,12 +207,21 @@ func (c *ServiceBindingUsageController) processNextWorkItem() bool {
...
 
-	err := c.syncServiceBindingUsage(key.(string))
+	namespace, name, err := cache.SplitMetaNamespaceKey(key.(string))
+	if err != nil {
+		c.log.Errorf("Error processing %q (splitting meta namespace key failed): %v", key, err)
+		c.queue.Forget(key)
+		return true
+	}
+
+	retry := c.maxRetires
```
polskikiel: `retries := c.queue.NumRequeues(key)`,  as discussed.
Please use `retries` instead of each occurrence `c.queue.NumRequeues(key)` occurence.

PR #1773, opened 2018-11-23
Copy graphql security docs to docs

sjanota:
**Description**

- Copy ui-api graphql security docs into official docs

**Related issue(s)**


sjanota:
Requires more careful approach then just copying. Closing this PR and next one will be opened.

PR #1803, opened 2018-11-26
Create KNative chart

sjanota:
**Description**

- KNAtive release YAMLs are added as charts to resources
- Necessary changes were introduced to those files
- `knative-installer` is no longer used and thus removed
- WG Kyma KNative Integration is added as owners of new chart

**Related issue(s)**


sjanota:
> 1. Be very clear about our dependencies and document them properly "e.g. k8s version, istio version and knative version".

Totally agree. This information is included for k8s and istio, and will be for KNative regardless the option we will choose.

> 2. Add a helper script to install it for minikube for a better developer experience.

While it makes Kyma's "core" (what ever it is) clear it puts on us the burden of maintaining Bash and Powershell scripts. I would rather make it a job so it's OS agnostic and can be developed in more civilised languages. Or we could have a CLI which could do that sort of things, but that's much broader topic.

> But completely maintaining a version of knative CRDs is a dangerous choice IMHO. Adds an overhead for maintaining the chart, the upgrades, and upgrading is gonna be very tricky. 

+💯 for the upgrades part

> 1. Kyma released against a release version of knative. Have that be documented and changed with every release "e.g. Kyma 0.5 with Knative 0.1".
> 2. Kyma master working with knative latest.
> 3. Have Kyma master break the day knative dependencies break, this way the change was introduced yesterday in knative probably and we can figure it out easily, as opposed to hiding behind a copy of knative CRDs and finding out what changed over the last two months in order to upgrade to newer version.

I can't agree. Kyma should be bound to specific version of KNative as we both have products in very early stage of development. We would put on ourselves unnecessary burden of catching up with KNative every time they do a breaking change. I'm a fan of frequent updates of dependencies, but I'd like to do them on my terms. When I have manpower to do them, not when KNative team decided to do release. If we bind ourselves to their latest KNative's release cycle is controlling our development plans.

That being said I'm against having more maintenance (of not our codebase) on our heads. The only reason I see why it's reasonable to make KNative a chart (now) is to be consistent with installer, which has very little value for me. 

What I see as a solution here is:
* Have a job which installs KNative and modifies it if necessary (maybe split in two steps, as with istio)
* When KNative introduces breaking changes in next release use new version and fix those modifications
* When they are ready with helm charts migrate the process to use them instead of YAMLs

Gain:
* Upgrades are easy until they have helm support
* The migration is done once (when they are ready with charts) instead of twice (now and migrating our chart to theirs).

I don't have opinion about storing raw release files in kyma repository. I think I'm leaning towards having them checked in, because: 
1. We need to have the version hardcoded somewhere
2. Files in their release shouldn't change

If we keep them raw upgrade means only downloading new file and that's acceptable "maintenance" for me. On the other hand we left far behind times when JARs were checked into VCS, so why charts should be :)

Abd4llA:
> I would rather make it a job so it's OS agnostic

+1, moving it to a job was a good step. I didn't really have a preference for a script.

> I can't agree. Kyma should be bound to specific version of KNative as we both have products in very early stage of development. We would put on ourselves unnecessary burden of catching up with KNative every time they do a breaking change. I'm a fan of frequent updates of dependencies, but I'd like to do them on my terms. When I have manpower to do them, not when KNative team decided to do release. If we bind ourselves to their latest KNative's release cycle is controlling our development plans.

I agree. I think I was thinking of the ideal world where we are very involved in knative decision making then it'd be probably long expected upfront as you see with Pivotal guys, and we would be part of the decision of introducing breaking changes. So since we're not there, it would be good to pinpoint to a "released" version. And try to regularly update it with every new "release" but in a controlled manner. 

So +1 from me for no "latest".

> What I see as a solution here is:
> 
> Have a job which installs KNative and modifies it if necessary (maybe split in two steps, as with istio)
> When KNative introduces breaking changes in next release use new version and fix those modifications
> When they are ready with helm charts migrate the process to use them instead of YAMLs

+1 . That would be a good solution.

sjanota:
One thing worth noticing: we cannot use raw YAMLs as chart, because CRDs need additional annotation to work with helm.

resources/knative/README.md
```diff
@@ -0,0 +1,14 @@
...
+
+## Overview
+
+This chart packs the [knative](https://github.com/knative/docs) release files.
+
+Included releases:
+ * https://github.com/knative/serving/releases/download/v0.2.1/release-no-mon.yaml
+ * https://github.com/knative/eventing/releases/download/v0.2.0/release.yaml
+
+Our changes:
```
tomekpapiernik:
```suggestion
Kyma-specific changes:
```

resources/knative/README.md
```diff
@@ -0,0 +1,14 @@
...
+ * Every CRD have annotation `"helm.sh/hook": "crd-install"` so Helm installs it before rest of resources.
```
tomekpapiernik:
```suggestion
 * Every CRD has the `helm.sh/hook: crd-install` annotation set. This forces Helm to install the CRDs before other resources.
```

resources/knative/README.md
```diff
@@ -0,0 +1,14 @@
...
+ * There is duplicated CRD `images.caching.internal.knative.dev` in serving release. One copy is removed.
```
tomekpapiernik:
```suggestion
 * The duplicate of the `images.caching.internal.knative.dev` CRD is removed from the serving release.
```

resources/knative/README.md
```diff
@@ -0,0 +1,14 @@
...
+ * If `isLocalEnv` variable is set `knative-ingressgateway` Service's `type` is changed to `NodePort`.
```
tomekpapiernik:
```suggestion
 * If the **isLocalEnv** variable is set to `knative-ingressgateway`, the Service's type automatically changes to `NodePort`.
```

CODEOWNERS
```diff
@@ -145,8 +145,8 @@
 # The `istio-kyma-patch` chart
 /resources/istio-kyma-patch/ @piotrmsc @kubadz @sjanota
 
-# The `knative-installer` chart
-/resources/knative-installer/ @piotrmsc @kubadz @sjanota
+# The `knative` chart
+/resources/knative/ @piotrmsc @kubadz @sjanota @Abd4llA @montaro @marcobebway @radufa
```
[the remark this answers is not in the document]
sjanota: GitHub team should work. Is there a group for knative activities?

resources/knative/Chart.yaml
```diff
@@ -1,5 +1,5 @@
 apiVersion: v1
-name: knative-installer
+name: knative
 version: 1.0.0
 appVersion: 1.0.0
```
Abd4llA: Please use the actual knative version you're using. "0.2.1" ?

PR #162, opened 2018-08-09
Add a section about the UsageKind to the Service Catalog documentation

klaudiagrz:
Confirm these statements before you submit your pull request:

- [x] I have read and submitted the required [Contributor Licence Agreements](https://github.com/kyma-project/community/blob/master/CONTRIBUTING.md#agreements-and-licenses).
- [x] This pull request follows the contributing guidelines, recommended Git workflow, and templates.
- [ ] I have tested my changes.
- [x] I have updated the relevant documentation.
---

**Description**

Changes proposed in this pull request:

- Add a section about the UsageKind 
- Add a diagram showing the UsageKind workflow

**Issue link**

PR #1824, opened 2018-11-27
Prow migration for tool etcd-backup

jasiu001:
Related to #162 test-infra repository


tools/etcd-backup/Makefile
```diff
@@ -0,0 +1,30 @@
+APP_NAME = "etcd-backup"
```
aszecowka: quotes are not necessary

PR #1829, opened 2018-11-27
Set connector-service token timeout to 5 min

crabtree:
**Description**

- Set connector-service token timeout to 5 min

**Related issue(s)**


crabtree:
@franpog859 Good catch. Updated.

PR #1853, opened 2018-11-29
Add makefile for tools/alpine-net

mwieczorek:
**Description**

- Add makefile for tools/alpine-net

**Related issue(s)**

### DIFF
--- a/tools/alpine-net/Makefile
+++ b/tools/alpine-net/Makefile
@@ -1,7 +1,6 @@
 APP_NAME = alpine-net
 IMG = $(DOCKER_PUSH_REPOSITORY)$(DOCKER_PUSH_DIRECTORY)/$(APP_NAME)
 TAG = $(DOCKER_TAG)
-binary=$(APP_NAME)
 
 .PHONY: build-image
 build-image:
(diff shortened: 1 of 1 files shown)


tools/alpine-net/Makefile
```diff
@@ -0,0 +1,27 @@
+APP_NAME = alpine-net
+IMG = $(DOCKER_PUSH_REPOSITORY)$(DOCKER_PUSH_DIRECTORY)/$(APP_NAME)
+TAG = $(DOCKER_TAG)
+binary=$(APP_NAME)
```
suleymanakbas91: `binary` is not needed as there is no building done for this tool.

tools/alpine-net/Makefile
```diff
@@ -0,0 +1,27 @@
...
+.PHONY: ci-pr
+ci-pr: build-image push-image
+
+.PHONY: ci-master
+ci-master: build-image push-image
+
+.PHONY: ci-release
+ci-release: build-image push-image
+
+.PHONY: clean
```
suleymanakbas91: Since it only removes `binary` and we don't need it, this target can be deleted as well.

tools/alpine-net/Makefile
mwieczorek: done

tools/alpine-net/Makefile
mwieczorek: done

Issue #1812, opened 2018-11-26
Migrate application-connector to prow


Disper:
**Description**
As we are moving to Prow, we need to migrate **application-connector** pipeline. 

Migration guide: https://github.com/kyma-project/test-infra/blob/master/docs/prow/migration-guide.md

**Reasons**
Resign from internal CI - Jenkins.

**Components to migrate**
- [x] Connector Service - @Szymongib 
- [x] Metadata Service - @Szymongib 
- [x] Proxy Service - @crabtree 
- [x] Connection Token Handler - @crabtree 
- [x] Remote Environment Controller - @Szymongib

PR #1867, opened 2018-11-29
Add prow related files for proxy-service component

crabtree:
**Description**

- Add prow related files for proxy-service component

**Related issue(s)**

#1812


components/proxy-service/before-commit.sh
```diff
@@ -0,0 +1,93 @@
...
+if [ ${goBuildResult} != 0 ]; then
+	echo -e "${RED}✗ go build${NC}\n$goBuildResult${NC}"
+	exit 1
+else echo -e "${GREEN}√ go build${NC}"
+fi
+
+##
+# DEP STATUS
+##
+echo "? dep status"
```
Szymongib: `dep status` will fail on CI as Gopkg.lock was generated by `dep` with version higher than 0.4.1, which is the version used by buildpack.

PR #1869, opened 2018-11-30
Whitelist a link with domain variable

bszwarc:
**Description**

-Whitelist a link with domain variable

### DIFF
--- a/milv.config.yaml
+++ b/milv.config.yaml
@@ -25,3 +25,6 @@ files:
   - path: "./kyma/docs/service-brokers/docs/011-configuration-helm-broker.md"
     config:
       white-list-external: ["custom.bundles-repository"]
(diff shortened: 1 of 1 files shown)


milv.config.yaml
```diff
@@ -25,3 +25,6 @@ files:
   - path: "./kyma/docs/service-brokers/docs/011-configuration-helm-broker.md"
     config:
       white-list-external: ["custom.bundles-repository"]
+  - path: "./examples/tracing/README.md"
```
michal-hudy: Are you sure it is the proper path?

bszwarc: Thank you for pointing this out - should be in examples.

bszwarc:
Changes should be made to a different repository.

PR #1927, opened 2018-12-05
Prow migration for event-service and k8s-dashboard-proxy components

marcobebway:
**Description**

- update the event-service component to support prow CI
- refactor some parts in the events-service component to fix golint issues
- update the k8s-dashboard-proxy component to support prow CI

**Related issue(s)**
https://github.com/kyma-project/test-infra/issues/234


marcobebway:
@montaro There is a before-commit phase which has a golint step that fails because of a list issues in the go code style. That's why the refactoring has to be done as part of this PR.

PR #1946, opened 2018-12-06
Prow migration for event-bus component

marcobebway:
**Description**

- update event-bus component to support prow CI

**Related issue(s)**
https://github.com/kyma-project/test-infra/issues/234

### DIFF
--- a/components/event-bus/Makefile
+++ b/components/event-bus/Makefile
@@ -1,18 +1,62 @@
-.PHONY: all resolve test test-report validate vet
+TAG = $(DOCKER_TAG)
 
-all: resolve test validate vet
+# event-bus-publish
+APP_NAME_PUBLISH = event-bus-publish
+IMG_PUBLISH = $(DOCKER_PUSH_REPOSITORY)$(DOCKER_PUSH_DIRECTORY)/$(APP_NAME_PUBLISH)
+BINARY_PUBLISH = $(APP_NAME_PUBLISH)
 
+# event-bus-push
+APP_NAME_PUSH = event-bus-push
+IMG_PUSH = $(DOCKER_PUSH_REPOSITORY)$(DOCKER_PUSH_DIRECTORY)/$(APP_NAME_PUSH)
+BINARY_PUSH = $(APP_NAME_PUSH)
+
+# event-bus-sv
+APP_NAME_SV = event-bus-sv
+IMG_SV = $(DOCKER_PUSH_REPOSITORY)$(DOCKER_PUSH_DIRECTORY)/$(APP_NAME_SV)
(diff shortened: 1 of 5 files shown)


components/event-bus/Makefile
```diff
@@ -1,18 +1,62 @@
...
+build:
+	# todo activate before-commit phase and remove the gometalinter
+	#./before-commit.sh ci
+	dep ensure -vendor-only -v
+	curl https://raw.githubusercontent.com/alecthomas/gometalinter/master/scripts/install.sh | sh -s v2.0.8
+	./bin/gometalinter --skip=generated --vendor --deadline=2m --disable-all --enable=vet ./...
+
+.PHONY: build-image
+build-image:
+	./docker-build-event-bus-publish.sh $(APP_NAME_PUBLISH) latest event-bus
```
abbi-gaurav: Should we consider to have makefiles for the sub-components being invoked from the root level Makefile? 

Ref: https://www.oreilly.com/library/view/managing-projects-with/0596006101/ch06.html

components/event-bus/before-commit.sh
```diff
@@ -0,0 +1,126 @@
...
+for vPackage in "${packagesToVet[@]}"; do
+	vetResult=$(go vet ${vPackage})
+	if [ $(echo ${#vetResult}) != 0 ]; then
+		echo -e "${RED}✗ go vet ${vPackage} ${NC}\n$vetResult${NC}"
+		exit 1
+	else echo -e "${GREEN}√ go vet ${vPackage} ${NC}"
+	fi
+done
+
+rm ./cmd/event-bus-publish/event-bus-publish
```
abbi-gaurav: the before-commit.sh might go away in the future. We should not have any customizations in it.

components/event-bus/Makefile
marcobebway: Of course we can do that, but I would advise to do it after we go away from Jenkins as its still using the sub-components makefiles. But for now to make both Prow and Jenkins working together, it's better to keep both configurations isolated.

components/event-bus/before-commit.sh
marcobebway: First of all, The pre-commit file is used by many kyma components, and each component customizes it to meet their needs.

Secondly, For event-bus this is totally a new file, so Why can't we customize it to meet our needs? Even if it will go away in the future, but still it will be working for sometime, so Why not make it work for event-bus in the meantime.

Finally, See lines 36, 37 and 38 those are pretty custom lines specifically to compile and build event-bus apps. Without these lines the pre-commit phase is useless. The same goes for lines 124, 125 and 126 to clean-up the apps' binaries.

PR #1947, opened 2018-12-06
Add ingress gateway client for testing purposes

sjanota:
**Description**

- Create test utility which helps calling Kyma through its external interface

**Related issue(s)**


sjanota:
## Review notes
* Talk with sig-quality

common/ingressgateway/ingressgateway.go
```diff
@@ -0,0 +1,98 @@
...
+	Getenv        func(string) string
+	LookupHost    func(string) ([]string, error)
+	GetMinikubeIP func() (string, error)
+	Dialer        Dialer
+}
+
+// Default returns clientCreator ready for production use.
+func Default() *clientCreator {
+	return &clientCreator{
+		Getenv:        os.Getenv,
```
Tomasz-Smelcerz-SAP: Consider being more specific: This is used only to fetch ingress FQDN, so perhaps: `IngressFQDN func() string` is better than very generic `os.Getenv`

common/ingressgateway/ingressgateway.go
```diff
@@ -0,0 +1,98 @@
...
+// clientCreator abstracts interactions with outside world for easy testing.
+type clientCreator struct {
+	Getenv        func(string) string
+	LookupHost    func(string) ([]string, error)
+	GetMinikubeIP func() (string, error)
+	Dialer        Dialer
+}
+
+// Default returns clientCreator ready for production use.
+func Default() *clientCreator {
```
Tomasz-Smelcerz-SAP: Consider adding second "constructor" here, or make clientCreator exported.
This implementation forces users of this library to execute "minikube ip" as a child process. Some clients might want to handle it differently (having this value in an env variable, for example)

common/ingressgateway/ingressgateway.go
sjanota: I would agree with `IngressFQDN func() string` but the signature is different. It is `func(string) string`. I could change that to:
```go
IngressFQDN: func() { return os.Getenv(ServiceNameEnv) }
```
but in this approach we wouldn't be able to test if code looks up valid environmental variable. Comparing those two I prefer to have it dirty, but better tested.

common/ingressgateway/ingressgateway.go
sjanota: Hmm, not sure about that. Those functions are abstracted only to make it possible to test the stuff. If it wasn't for testing those functions wouldn't be even configurable. I'd like to have one more opinion before changing that.

common/ingressgateway/ingressgateway.go
sjanota: I will apply the suggestion. Also there will be new test to cover getting environment variable.

common/ingressgateway/ingressgateway.go
sjanota: I will hide clientCreator behind interface so it can be passed around, but its fields are not public. Also the name INGRESSGATEWAY_FQDN is not reflecting the possibility to pass IP here, so I'll change it to INGRESSGATEWAY_ADDRESS. Thanks to that client can overcome dependency on minikube by passing `INGRESSGATEWAY_ADDRESS={minikube ip}`.

PR #1957, opened 2018-12-06
tests event-bus Prow build job

montaro:
See also `https://github.com/kyma-project/test-infra/issues/234`


montaro:
@marcobebway Thanks for the review. Jenkins images are published so far.

Issue #1802, opened 2018-11-26
The ServiceCatalog `etcd` crashing with OOMKilled


mszostok:
**Description**

The `etcd` crashing with OOMKilled. The problem is caused by the orphans ServiceInstance and ServiceBindings from Service Catalog. If those resources are in error `ReferencesNonexistentServiceClass` then the Service Catalog in infinite loop try to delete those resources and after some time the etcd crashing. 

**Steps to reproduce**

Register some Broker then create a few ServiceInstance and ServiceBinding from it. After that remove the Broker from the ServiceCatalog and then delete previously created ServiceInstance and ServiceBinding. That resource will end up in `ReferencesNonexistentServiceClass` error and after some time the etcd will crash

This problem was already reported in this issue: https://github.com/kyma-project/kyma/issues/1336#issuecomment-433894093

adamwalach:
Documentation about --snapshot-count parameter:
https://github.com/etcd-io/etcd/blob/ae25c5e1320f731a2ffaafbf756aca8b0a94dfab/Documentation/op-guide/maintenance.md#L11

adamwalach:
Outcomes:
We are able to easily kill the etcd cluster using stress-etcd tool in less then one hour.
The main issue is size of raft log entries kept in memory - they are flushed out when number of entries reaches 'snapshot-count' value. To improve the situation we increased memory limit from 256M to 512 and decreased value of snapshot-count to 10,000.
According to documentation the default value of --snapshot-count has changed from from 10,000 to 100,000 since version 3.2.
When the "snapshot-count" limit was reached, we could see it in the log file:
<img width="1112" alt="screen shot 2018-12-07 at 14 36 56" src="https://user-images.githubusercontent.com/8530211/49650753-9a0fd000-fa2d-11e8-9289-27f9c175a176.png">
After this event occurs memory usage on all nodes is visibly decreased:
<img width="1414" alt="screen shot 2018-12-06 at 12 02 46" src="https://user-images.githubusercontent.com/8530211/49651008-523d7880-fa2e-11e8-9091-158a8939d377.png">

Final test was performed with following settings:
- 3 etcd nodes, ram=512MB, snapshot-count=10000
- 2 stress-etcd workers, key_size=2000

We were not able to break etcd with OOMKilled error, but when we kept the test running for a few hours in the log file we could observe "failed to send out heartbeat on time" entries in the log:
[...]

PR #1966, opened 2018-12-07
Modify default etcd configuration (for service catalog)

adamwalach:
**Description**

- memory limit for non-local cluster increased to 512M (previously 256M)
- --snapshot-count is now set to 10 000 (default value: 100 000)

**Related issue(s)**
* #1802


resources/service-catalog/charts/etcd-stateful/templates/03-statefulset.yaml
```diff
@@ -137,6 +137,7 @@ spec:
                 --initial-cluster-token etcd-cluster-1 \
                 --initial-cluster ${PEERS} \
                 --initial-cluster-state ${STATE} \
+                --snapshot-count=10000 \
```
mszostok: as discussed, please extract it to the variable and add comment with link to our issue on GitHub. Thanks to that this value won't be so magic

PR #1969, opened 2018-12-09
Make Catalog UI to be more resilient on empty creating schemas

magicmatatjahu:
**Description**

- Return empty `InstanceCreateParameterSchema` and/or `BindingCreateParameterSchema` when it has a empty `$ref` or `properties`,
- Update relevant unit tests.

**Related issue(s)**
See also https://github.com/kyma-project/console/issues/356

### DIFF
--- a/components/ui-api-layer/internal/domain/servicecatalog/clusterserviceplan_converter.go
+++ b/components/ui-api-layer/internal/domain/servicecatalog/clusterserviceplan_converter.go
@@ -8,6 +8,7 @@ import (
 	"github.com/kyma-project/kyma/components/ui-api-layer/internal/gqlschema"
 	"github.com/kyma-project/kyma/components/ui-api-layer/internal/resource"
 	"github.com/pkg/errors"
+	"bytes"
 )
 
 type clusterServicePlanConverter struct{}
@@ -32,7 +33,9 @@ func (p *clusterServicePlanConverter) ToGQL(item *v1beta1.ClusterServicePlan) (*
 		if err != nil {
 			return nil, p.wrapConversionError(err, item.Name)
 		}
-		instanceCreateParameterSchema = &unpackedSchema
+		if unpackedSchema != nil && len(unpackedSchema) != 0 {
+			instanceCreateParameterSchema = &unpackedSchema
+		}
 	}
 	var bindingCreateParameterSchema *gqlschema.JSON
 	if item.Spec.ServiceBindingCreateParameterSchema != nil {
@@ -40,7 +43,9 @@ func (p *clusterServicePlanConverter) ToGQL(item *v1beta1.ClusterServicePlan) (*
 		if err != nil {
 			return nil, errors.Wrapf(err, "while unpacking service binding create parameter schema from ClusterServicePlan [%s]", item.Name)
 		}
-		bindingCreateParameterSchema = &unpackedSchema
+		if unpackedSchema != nil && len(unpackedSchema) != 0 {
+			bindingCreateParameterSchema = &unpackedSchema
+		}
 	}
 
 	displayName := resource.ToStringPtr(externalMetadata["displayName"])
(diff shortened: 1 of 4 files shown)


components/ui-api-layer/internal/domain/servicecatalog/clusterserviceplan_converter.go
```diff
@@ -91,12 +97,28 @@ func (p *clusterServicePlanConverter) unpackCreateParameterSchema(raw []byte) (g
 }
 
 func (p *clusterServicePlanConverter) extractCreateSchema(raw []byte) (map[string]interface{}, error) {
+	raw = bytes.Trim(raw, "\x00")
```
pkosiec: Hmm, is this necessary?

components/ui-api-layer/internal/domain/servicecatalog/clusterserviceplan_converter.go
```diff
@@ -91,12 +97,28 @@ func (p *clusterServicePlanConverter) unpackCreateParameterSchema(raw []byte) (g
...
+	if p.isEmptyCreateParameterSchema(extracted) {
+		return nil, nil
+	}
 
 	return extracted, nil
 }
+
+func (p *clusterServicePlanConverter) isEmptyCreateParameterSchema(schema map[string]interface{}) bool {
+	properties := schema["properties"]
+	if properties != nil && len(properties.(map[string]interface{})) != 0 {
```
pkosiec: `properties.(map[string]interface{})`

Always check if the conversion is OK

components/ui-api-layer/internal/domain/servicecatalog/clusterserviceplan_converter.go
```diff
@@ -91,12 +97,28 @@ func (p *clusterServicePlanConverter) unpackCreateParameterSchema(raw []byte) (g
...
+		return false
+	}
+	ref := schema["$ref"]
+	if ref != nil && ref.(string) != "" {
```
pkosiec: `properties.(map[string]interface{})`

Always check if the conversion is OK

components/ui-api-layer/internal/domain/servicecatalog/serviceplan_converter.go
```diff
@@ -92,12 +98,28 @@ func (p *servicePlanConverter) unpackCreateParameterSchema(raw []byte) (gqlschem
...
 		return nil, errors.Wrap(err, "while extracting creation parameter schema")
 	}
+	if p.isEmptyCreateParameterSchema(extracted) {
+		return nil, nil
+	}
 
 	return extracted, nil
 }
+
+func (p *servicePlanConverter) isEmptyCreateParameterSchema(schema map[string]interface{}) bool {
```
pkosiec: Maybe move it to a separate package within `internal` directory, as it is shared.

For example: `jsonschema.isEmpty()`

components/ui-api-layer/internal/domain/servicecatalog/clusterserviceplan_converter.go
magicmatatjahu: Yes, this is necessary, but only for our unit tests, because we use standard implementation of base64, that sometimes add `\x00` chars at the end of bytes, when decode it some data.

components/ui-api-layer/internal/domain/servicecatalog/serviceplan_converter.go
magicmatatjahu: I do not think it's a good idea, because in this case we check specific fields and the package you propose should be 100% generic.

components/ui-api-layer/internal/domain/servicecatalog/clusterserviceplan_converter.go
magicmatatjahu: I fix it 👌

components/ui-api-layer/internal/domain/servicecatalog/clusterserviceplan_converter.go
magicmatatjahu: I fix it 👌

components/ui-api-layer/internal/domain/servicecatalog/serviceplan_converter.go
magicmatatjahu: I can agree to extract this function + functions: `unpackCreateParameterSchema` and `extractCreateSchema` only to folder in this same domain (servicecatalog).

magicmatatjahu: @michal-hudy What do you thinking about this?

components/ui-api-layer/internal/domain/servicecatalog/clusterserviceplan_converter.go
michal-hudy: In fact, the null characters are added because of that: `base64.StdEncoding.DecodedLen(len(raw))`
It returns the max length of the encoded string. I would rather move that line to `unpackCreateParameterSchema`

https://golang.org/pkg/encoding/base64/#Encoding.DecodedLen

We need to trim that null characters, because `json.Unmarshal` is failing.

components/ui-api-layer/internal/domain/servicecatalog/serviceplan_converter.go
pkosiec: @magicmatatjahu no, please, don't create new files in `servicecatalog` domain 😛 There are too many already.

If this is specific, then still, you can create subpackage like `servicecatalog/parameterschema` or `servicecatalog/jsonschema`

PR #1998, opened 2018-12-11
Migrate configurations-generator to prow

Demonsthere:
**Description**

- Add required `before-commit.sh` script
- Add goimports to Gopkg
- Simplify Jenkinsfile to ensure compatibility
- Add makefile

**Related issue(s)**
https://github.com/kyma-project/test-infra/issues/224


components/configurations-generator/Jenkinsfile
```diff
@@ -54,19 +54,19 @@ podTemplate(label: label,
                         }
 
                         stage("install dependencies $application") {
-                            execute("make resolve", workDir, componentDir)
+                            execute("dep ensure --vendor-only -v", workDir, componentDir)
```
strekm: consider combining stages and using before-commit.sh

Demonsthere: From what I have been told, we want to implement the least amount of changes, and this won't remove existing stages

sjanota: I agree with @Demonsthere. Jenkinsfiles are almost extinct, so I'd say every approach that just keep them running is a valid one.

components/configurations-generator/Jenkinsfile
```diff
@@ -54,19 +54,19 @@ podTemplate(label: label,
...
+                            execute("gometalinter --vendor --deadline=2m --disable-all --enable=vet --skip=\$(go list ./... | grep \"/fake\") ./...", workDir, componentDir)
                         }
 
                         stage("build $application") {
                             execute("CGO_ENABLED=0 go build -o bin/app cmd/generator/main.go", workDir, componentDir)
                         }
 
                         stage("test $application") {
-                            execute("make test-report", workDir, componentDir)
+                            execute("2>&1 go test -short -v ./... | go2xunit -fail -output unit-tests.xml", workDir, componentDir)
```
sjanota: Why -short flag is present here? If it means we are skipping some tests we need to understand why and find a way to reenable them.

components/configurations-generator/before-commit.sh
```diff
@@ -0,0 +1,116 @@
...
+	echo -e "${RED}✗ dep status\n$depResult${NC}"
+	exit 1
+else echo -e "${GREEN}√ dep status${NC}"
+fi
+
+##
+# GO TEST
+##
+echo "? go test"
+go test -short ./...
```
sjanota: Same as in Jenkinsfile

components/configurations-generator/before-commit.sh
```diff
@@ -0,0 +1,116 @@
...
+if [ $(echo ${#goImportsResult}) != 0 ]; then
+	echo -e "${RED}✗ goimports and fmt${NC}\n$goImportsResult${NC}"
+	exit 1
+else echo -e "${GREEN}√ goimports and fmt${NC}"
+fi
+
+##
+# GO VET
+##
+packagesToVet=($(go list ./... | grep -v /vendor/ | grep -v api-controller/pkg/clients ))
```
sjanota: `api-controller` ;) You probably don't need this grep anyway. Filter out only vendor directory

components/configurations-generator/Jenkinsfile
sjanota: Here's an example of such test: https://github.com/kyma-project/kyma/blob/master/components/api-controller/pkg/controller/networking/v1/networkingv1_integration_test.go#L17. If yo ucopied this from api-controller there may be no such test in configurations-generator, but please verify anyway.

Demonsthere:
Closing this PR due to some CLA problem

Issue #1868, opened 2018-11-29
Implement Library concept and provide utils for acceptance testing


mszostok:
**Description**

In Kyma project we have an agreement to introduce common libraries. The proposal was discussed and accepted. 

**Now is time to kick-off and first implementation!**

AC:
- [x] the `common` pkg folder is created in [kyma](https://github.com/kyma-project/kyma) repository with strong CODEOWNERS
- [ ] all projects under the [kyma-project/kyma/tests](https://github.com/kyma-project/kyma/tree/master/tests) are analyzed and common logic is identified 
- [ ] common logic from [kyma-project/kyma/tests](https://github.com/kyma-project/kyma/tree/master/tests) is unified by testing utils which follows all restrictions (see below section) to be in kyma libraries
- [ ] testing utils from kyma libraries are used in all projects under [kyma-project/kyma/tests](https://github.com/kyma-project/kyma/tree/master/tests)
- [ ] The first implementation of kyma libraries are announced. The Kyma developers are encouraged to analyze their projects and extract common logic to kyma libraries which can be used by others.

> **NOTE**: Restrictions
Libraries must be well-tested. Tests must be executed with the `-count 100` flag to ensure stability and with the `-race` flag to detect all race conditions. 
+Each library must contain good documentation. It is also highly recommended that each library has the [Example Test](https://blog.golang.org/examples) similar to this [one](https://github.com/mszostok/kyma/blob/libs/common/http/client_example_test.go#L13).

**Reasons**

Please see [this](https://github.com/kyma-project/community/issues/142) proposal for more information

sjanota:
This issue is partially blocking my PR with common ingress gateway client for tests #1947. I'm giving a try to creating `common` package. Expect a PR soon.

jasiu001:
**First Step**
Find and update all test parts which can be replaced by `FuncAtMost()` test in `/acceptance/pkg/repeat/repeat.go`

| test | user |
| ---- | ---- |
| acceptance | |
| api-controller-acceptance-test | @jasiu001  |
| application-operator-tests | |
| application-registry-tests | |
| connector-service-tests | |
| event-bus | |
| gateway-tests | |
| knative-serving-acceptance | |
| kubeless | |
| kubeless-integration | |
| logging | |
| test-logging-monitoring | |
| test-namespace-controller | |
| ui-api-layer-acceptance-tests | |

sjanota:
[...]

PR #2018, opened 2018-12-12
Create common package

sjanota:
**Description**

- package `common` is created at the root of repository
- initial volunteers are added to CODEOWNERS

**Related issue(s)**

#1868


sjanota:
I also wonder if we could name it `pkg` instead of `common`. We would then have import path `github.com/kyma-project/kyma/pkg/library-name` which is more in line with go convention. @mszostok ?

sjanota:
I've added support for Jenkins. Also Makefile is created so we can easily add it to prow later.

mszostok:
@sjanota the naming was already addressed in the proposal, see: https://github.com/kyma-project/community/pull/143/files#diff-3a446b5bee42fa61dac96e5848cfddcdR31

sjanota:
@mszostok Thanks for the reference, I see the reasoning now. Could you have a look and approve?

mszostok:
For me in this common pkg we missing the README.md file. We need to add information about the scope (purpose), the restrictions (https://github.com/kyma-project/community/pull/143/files#diff-3a446b5bee42fa61dac96e5848cfddcdR111) etc. (you get get this information from proposal).

e.g. such section should appear in README
```
## Criteria for adding code here

- Used by multiple components in Kyma.
- Full unit test coverage
- Stable unit tests (executed with the `-count 100` flag)
- Test are execute with `-race` flag to detect all race conditions. 
```

mszostok:
See this [README](https://github.com/kubernetes/utils/blob/master/README.md) for inspiration ;)

PR #2026, opened 2018-12-12
Migrate configurations-generator

Demonsthere:
**Description**
Clean PR
- Update Jenkinsfile
- Create before-commit
- Fix tests

**Related issue(s)**

Fixes https://github.com/kyma-project/kyma/pull/1998
Implements https://github.com/kyma-project/test-infra/issues/224

### DIFF
--- a/components/configurations-generator/Jenkinsfile
+++ b/components/configurations-generator/Jenkinsfile
@@ -54,11 +54,11 @@ podTemplate(label: label,
                         }
 
                         stage("install dependencies $application") {
-                            execute("make resolve", workDir, componentDir)
+                            execute("dep ensure --vendor-only -v", workDir, componentDir)
                         }
 
                         stage("code quality $application") {
-                            execute("make validate", workDir, componentDir)
+                            execute("gometalinter --vendor --deadline=2m --disable-all --enable=vet --skip=\$(go list ./... | grep \"/fake\") ./...", workDir, componentDir)
                         }
 
                         stage("build $application") {
@@ -66,7 +66,7 @@ podTemplate(label: label,
                         }
 
                         stage("test $application") {
-                            execute("make test-report", workDir, componentDir)
+                            execute("2>&1 go test -v ./... | go2xunit -fail -output unit-tests.xml", workDir, componentDir)
                             junit '**/unit-tests.xml'
                         }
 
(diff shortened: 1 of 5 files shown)


components/configurations-generator/Jenkinsfile
```diff
@@ -54,19 +54,19 @@ podTemplate(label: label,
                         }
 
                         stage("install dependencies $application") {
-                            execute("make resolve", workDir, componentDir)
+                            execute("dep ensure --vendor-only -v", workDir, componentDir)
                         }
 
                         stage("code quality $application") {
-                            execute("make validate", workDir, componentDir)
+                            execute("gometalinter --vendor --deadline=2m --disable-all --enable=vet --skip=\$(go list ./... | grep \"/fake\") ./...", workDir, componentDir)
```
kubadz: we don't need to skip packages containing "/fake" as there are no such packages in this component

components/configurations-generator/before-commit.sh
```diff
@@ -0,0 +1,93 @@
...
+##
+# GO BUILD
+##
+buildEnv=""
+if [ "$1" == "$CI_FLAG" ]; then
+	# build binary statically
+	buildEnv="env CGO_ENABLED=0"
+fi
+
+${buildEnv} go build -o generator ./cmd/generator/
```
Tomasz-Smelcerz-SAP: Have you ever run it?
Output file path (relative to project dir) is: ./generator. In Dockerfile you build image from file ./bin/app

components/configurations-generator/before-commit.sh
```diff
@@ -0,0 +1,93 @@
...
+echo "? go test"
+go test ./...
+# Check if tests passed
+if [ $? != 0 ]; then
+	echo -e "${RED}✗ go test\n${NC}"
+	exit 1
+else echo -e "${GREEN}√ go test${NC}"
+fi
+
+goFilesToCheck=$(find . -type f -name "*.go" | egrep -v "\/vendor\/|_*/automock/|_*/testdata/|/pkg\/|_*export_test.go")
```
kubadz: why do you skip go files in the `/pkg` directory? I would say that we should format them, as in most cases they contain code written by us.

components/configurations-generator/before-commit.sh
Demonsthere: Good catch, fixing!

components/configurations-generator/Jenkinsfile
Demonsthere: Fixing!

components/configurations-generator/before-commit.sh
Demonsthere: Good call, fixed.

PR #2055, opened 2018-12-17
Add xip patch component

kfurgol:
**Description**

Changes proposed in this pull request:

- Added xip-patch component

**Related issue(s)**
Adopt kyma installation for xip.io domain #1871

Issue #1871, opened 2018-11-30
Adapt kyma installation for xip.io domain


piotrmsc:
**Description**

As a part of publishing Kyma to GCP Marketplace, we need to have bit different approach for installing kyma from Marketplace. During installation, we have to obtain load balancer ip address and override the domain to consist of {loadbalancerIP}.xip.io. So for example console is available under `https://console.{loadbalancerIP}.xip.io`. With xip.io we do not need to make entries do DNS.

This change should be applied only when installation from Marketplace is triggered(helm feature toggle in the values.yaml).
Additionally, self-signed TLS cert should be generated for given domain.
Create a separate step in the installation phase which will override certs used in istio patch.

AC: 

- kyma can be installed under xip.ip domain on self-signed cert only (via helm feature toggle) 

**Reasons**
Must have to enable kyma on GCP Marketplace

**Attachments**

sjanota:
I don't think patch direction is the right one. On one hand we are leaving bash scripts behind and don't allow creation of new ones (#1785), but istio-kyma-patch is nothing more than a bash script. If we are going in this direction we need to refine this component to make it more stable, tested and bulletproof. And rewrite it from bash.

pbochynski:
I would do it in separate step. It should not be part of istio patch. The new component would check if domain is set - if not it is set to the ingresIP.xip.io and self signed cert is issued.

Tomasz-Smelcerz-SAP:
Additional informations:
- No interactions with external resource providers (IP, DNS), everything is in-cluster.
- IP Address is taken from the cluster (kubectl get ingress ... ?)
- Additional component that is launched after istio and injects IP, DNS and self-signed certs as overrides for subsequent steps
- Additional config for this kind of installation.
- Consider where to put this additional config. kyma-incubator ?
- Create a README (simple-installation.md) that documents the steps. Installation is one step!

How to prove it works (Demo):
- Provision a fresh GKE cluster (or any other with IPs that are public, not minikube!)
- kubectl apply -f simple-installation.yaml
- wait ...
- get IP address from the cluster
- visit: https://console.{loadbalancerIP}.xip.io
- Install

PR #2056, opened 2018-12-17
Add xip patch chart

kfurgol:
**Description**

- Added xip patch chart
- Added xip-installation.yaml (please suggest better place for that file)
- xip-patch component - please check #2055

**Related issue(s)**
Adopt kyma installation for xip.io domain #1871


installation/resources/xip-installation.yaml
```diff
@@ -0,0 +1,265 @@
...
+  name: service-catalog-overrides
+  namespace: kyma-installer
+  labels:
+    installer: overrides
+    component: service-catalog
+data:
+  etcd-stateful.etcd.resources.limits.memory: 512Mi
+
+---
+apiVersion: "installer.kyma-project.io/v1alpha1"
```
sjanota: Why you copied whole installer.yaml, not only installation-cr.yaml?

resources/xip-patch/templates/clusterrole.yaml
```diff
@@ -0,0 +1,8 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: xip-patch
+rules:
+- apiGroups: ["*"]
```
sjanota: Specify api groups here. With `*` this script would have access to all resources named `configmaps` and `scripts`, regardless of their apiGroup. In Knative for exmaple is resource `service` and we don't want the script to have access to it.

resources/xip-patch/templates/clusterrole.yaml
```diff
@@ -0,0 +1,8 @@
...
+  resources: ["configmaps", "services"]
+  verbs: ["*"]
```
sjanota: You don't need write access to `service`. Split it into two rules: one with read-write access to `configmaps` and second with read-only access to `services`.

sjanota:
I've tested it on my own GKE cluster. Everything works fine.

resources/xip-patch/templates/clusterrole.yaml
```diff
@@ -0,0 +1,11 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: xip-patch
```
Demonsthere: The name element in the chart should contain the Release.name too: `{{ .Release.Name }}-xip-patch`

resources/xip-patch/templates/job.yaml
```diff
@@ -0,0 +1,17 @@
...
+spec:
+  template:
+    metadata:
+      name: xip-patch
+    spec:
+      serviceAccountName: xip-patch
+      restartPolicy: OnFailure
+      containers:
+      - name: xip-patch
+        image: eu.gcr.io/kyma-project/pr/xip-patch:PR-2055
```
Demonsthere: I see you have created default values, why are you not using them here?

resources/xip-patch/values.yaml
```diff
@@ -0,0 +1,6 @@
+global:
```
Demonsthere: I would avoid the `global` keyword here, just keep the structure:
```yaml
---
containerRegistry:
  path: eu.gcr.io/kyma-project
xip_patch:
  dir: pr/xip-patch/
  version: PR-2055
```

resources/xip-patch/templates/clusterrole.yaml
```diff
@@ -0,0 +1,11 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
```
sjanota: When thing about it it probably shouldn't be `ClusterRole`, but two separate `Role`s. One of them should allow access to services in `istio-system` and second to config maps in `kyma-installer`. Current scope is too broad.

sjanota:
@jakkab I don't agree. As far as I'm aware the helm convention is to have file per resource (at least that's what we do in Kyma). I don't claim it's wise convention, though ;). I'd say let's stay with file per resource

PR #2057, opened 2018-12-18
Fix sampleapp behaviour

Demonsthere:
…o on the namespace. Refactor nameing

**Description**

- Fix golang naming 
- Fix hardcoded namespace in VirtualService
- Create new isito enabled namespace in chart
- Change default namespace to the custom one

**Related issue(s)**

### DIFF
--- a/resources/core/charts/api-controller/templates/tests/test-namespace.yaml
+++ b/resources/core/charts/api-controller/templates/tests/test-namespace.yaml
@@ -1,0 +1,13 @@
+{{ if .Values.tests.enabled }}
+---
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: sampleapp
+  annotations:
+    "helm.sh/hook": test-success
+  labels:
+    "helm-chart-test": "true"
+    istio-injection: enabled
+---
+{{ end }}
--- a/tests/api-controller-acceptance-tests/apicontroller/component_test.go
+++ b/tests/api-controller-acceptance-tests/apicontroller/component_test.go
@@ -46,23 +46,23 @@ func TestComponentSpec(t *testing.T) {
 		Convey("create API with authentication disabled", func() {
 			t.Log("create API with authentication disabled")
 
-			testId := ctx.generateTestId(testIdLength)
-			t.Logf("Running test: %s", testId)
-			api := ctx.apiFor(testId, domainName, apiSecurityDisabled, true)
+			testID := ctx.generatetestID(testIDLength)
+			t.Logf("Running test: %s", testID)
+			api := ctx.apiFor(testID, domainName, apiSecurityDisabled, true)
 
-			lastApi, err := kymaClient.GatewayV1alpha2().Apis(namespace).Create(api)
-			defer ctx.cleanUpApi(kymaClient, lastApi, t, false)
+			lastAPI, err := kymaClient.GatewayV1alpha2().Apis(namespace).Create(api)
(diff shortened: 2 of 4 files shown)


resources/core/charts/api-controller/templates/tests/test-namespace.yaml
```diff
@@ -0,0 +1,13 @@
+{{ if .Values.tests.enabled }}
```
sjanota: You shouldn't add it here. You should create separate PR which changes in charts. This PR must also contain also bump of tests image. Maybe it is not critical in this particular case as creating new namespace wouldn't break tests, but for the sake of keeping up to the process I would move it to the second PR with image change.

I wish it was simpler :D

tests/api-controller-acceptance-tests/apicontroller/fixture.go
```diff
@@ -3,42 +3,45 @@ package apicontroller
...
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/intstr"
 	"k8s.io/client-go/kubernetes"
-	"time"
 )
 
 const (
-	namespace                                      = "kyma-system"
+	namespace                                      = "sampleapp"
```
sjanota: As the namespace is now expected to exist before test is run I'd suggest passing its name as environmental variable, not hardcoding it here.

tests/api-controller-acceptance-tests/apicontroller/component_test.go
```diff
@@ -233,17 +238,17 @@ func TestComponentSpec(t *testing.T) {
 	})
 }
 
-func (ctx componentTestContext) apiFor(testId, domainName string, secured ApiSecurity, hostWithDomain bool) *kymaApi.Api {
+func (ctx componentTestContext) apiFor(testID string, domainName string, namespace string, secured APISecurity, hostWithDomain bool) *kymaApi.Api {
```
sjanota:
```suggestion
func (ctx componentTestContext) apiFor(testID, domainName, namespace string, secured APISecurity, hostWithDomain bool) *kymaApi.Api {
```

tests/api-controller-acceptance-tests/apicontroller/integration_test.go
```diff
@@ -72,60 +77,60 @@ func TestIntegrationSpec(t *testing.T) {
...
 
-			_, err = kymaInterface.GatewayV1alpha2().Apis(namespace).Get(lastApi.Name, metav1.GetOptions{})
+			_, err = kymaInterface.GatewayV1alpha2().Apis(namespace).Get(lastAPI.Name, metav1.GetOptions{})
 			So(err, ShouldNotBeNil)
 		})
 	})
 }
 
-func (ctx integrationTestContext) apiFor(testId, domainName string, svc *apiv1.Service, secured ApiSecurity, hostWithDomain bool) *kymaApi.Api {
+func (ctx integrationTestContext) apiFor(testID string, domainName string, namespace string, svc *apiv1.Service, secured APISecurity, hostWithDomain bool) *kymaApi.Api {
```
sjanota:
```suggestion
func (ctx integrationTestContext) apiFor(testID, domainName, namespace string, svc *apiv1.Service, secured APISecurity, hostWithDomain bool) *kymaApi.Api {
```

PR #2062, opened 2018-12-18
Update api controller chart 

Demonsthere:
**Description**

Followup to https://github.com/kyma-project/kyma/pull/2057
- Update chart resource with namespace
- Set namespace env to test program

**Related issue(s)**

### DIFF
--- a/resources/core/charts/api-controller/templates/tests/test-api-controller.yaml
+++ b/resources/core/charts/api-controller/templates/tests/test-api-controller.yaml
@@ -1,6 +1,14 @@
 {{ if .Values.tests.enabled }}
 ---
 apiVersion: v1
+kind: Namespace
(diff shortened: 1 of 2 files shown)


resources/core/charts/api-controller/templates/tests/test-api-controller.yaml
```diff
@@ -16,6 +24,9 @@ spec:
     env:
     - name: DOMAIN_NAME
       value: {{ .Values.global.domainName }}
+    env:
```
sjanota: All env variables must be in the same `env` section

Demonsthere: Another error brought to you by too much copy&paste :)

PR #2071, opened 2018-12-19
Rename BUC CRs domain to kyma-project.io

polskikiel:
**Description**

- Adjusted servicecatalog CRDs domain to kyma-project.io
- Regenerated client

**Related issue(s)**

### DIFF
--- a/components/binding-usage-controller/internal/controller/usagekind/protection_controller.go
+++ b/components/binding-usage-controller/internal/controller/usagekind/protection_controller.go
@@ -20,9 +20,10 @@ import (
 )
 
 const (
-	finalizerName = "servicecatalog.kyma.cx/usage-kind-protection"
-	indexKind     = "usedByKind"
+	indexKind = "usedByKind"
 )
+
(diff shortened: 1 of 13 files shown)


components/binding-usage-controller/internal/controller/usagekind/protection_controller.go
```diff
@@ -20,7 +20,7 @@ import (
 )
 
 const (
-	finalizerName = "servicecatalog.kyma.cx/usage-kind-protection"
+	finalizerName = "servicecatalog.kyma-project.io/usage-kind-protection"
```
mszostok: **suggestion**

What do you think about this approach?

```
var finalizerName = fmt.Sprintf("%s/usage-kind-protection", v1alpha1.SchemeGroupVersion.Group)
```

thanks to that in future we will not maintain that, what do u think?

polskikiel: I believe that we won't change the domain name again. But you never know how kyma will look like tomorrow, so I will apply your suggestion 👍

PR #2074, opened 2018-12-19
Rename environments to namespaces

dariadomagala:
**Description**

- Rename environments to namespaces

**Related issue(s)**


tomekpapiernik:
The screenshots still use "Environments" and "Remote Environments". If the interface is already updated then you should update the s-shots as well. 

image

Issue #1610, opened 2018-11-09
is-installed script gives up on installation error 


Tomasz-Smelcerz-SAP:
After #1045 is merged, errors in installation do not break the process, retry will occur. is-installed script is not ready for this and gives up on first error encountered. The script must handle installation errors with reasonable timeout.

kubadz:
There are at least three options to consider:
- set the timeout on the is-installed script
- set the timeout in an environment in which the script is executed (e.g. CI) and stop is-installed script execution only when Kyma is successfully installed
- retry installation a finite number of times

While the first one is the simplest, as it depends only on measuring time, it might be difficult to find an appropriate time which needs to be set on timeout due to the variety of environments in which Kyma is installed (too little time might result in script always returning an error in the environments with less memory, cpu units or with slower internet). That option would need a further investigation on how to measure time in the script.

The second one requires only one change to is-installed script, which is removing the logic responsible for returning error number from the script when the error in Kyma installation occurs. It assumes setting the timeout in the environment in which the script is executed or terminating it manually.

The last option is the most difficult to apply. It comes with a few options related to measuring how many times has Kyma installation already failed and some of them require changes in the installer component:
- let installer count the times it tries to restart the installation, set in the Installation CR how many times it should retry before setting some final error in the Installation CR status
- let installer count the times it tries to restart the installation, add a new field to Installation CR status which will contain the number of retries
- count the number of retries in the is-installed script by parsing the logs from the installer

sjanota:
1. (8) Installer has information and is responsible for making decision about installation success. Make installer allow to set strict number of retries and fail if it is reached. Outside clients would depend on installer behaviour. New status fields will be available. Risks:
  * API version bump
  * How k8s behaves when job reaches its number of max retries

2. (5) Installer is not responsible for the decision, but exposes the information about retries. Let installer count the times it tries to restart the installation, add a new field to Installation CR status which will contain the number of retries. Based on this information script can decide if installation is successful or not.

3. (3) Script calculates the information about retries and is responsible for making the decision. Is-installed accepts `--timeout` flag which tells how long it waits for installation to succeed. In comparison to having timeout on CI it allows to:
  * shorten pipeline time in case of installation problems
  * there is a place where we can print logs

piotrmsc:
[...]

PR #2087, opened 2018-12-20
Make is-installed script not exit on error

kubadz:
**Description**
This PR removes the logic responsible for immediately returning an error code from the is-installed script when the Status in Installation CR becomes `Error`. By default, the script will wait forever until the installation will be successful, but there is also an option added allowing to set the timeout after which the script will return an error code. The script additionally fetches logs from the installer if the timeout occurs while the `Error` status is set on Installation CR.
 
- stop immediately returning an error code from the is-installed script when the Status in Installation CR becomes `Error`
- add timeout option to the is-installed script
- set default timeout in dockerfiles/scripts using is-installed.sh as a part of Kyma installation
- update documentation

**Related issue(s)**

Reasons of introducing those changes are described in #1610


sjanota:
@kubadz Remember to update Prow jobs

installation/scripts/is-installed.sh
```diff
@@ -33,18 +86,29 @@ do
...
-    echo "----------"
-    exit 1
+    echo "kyma installation error, description: ${DESC}"
+    if [ "$TIMEOUT_SET" -eq 0 ]; then
+      echo "to fetch the logs from the installer execute: kubectl logs -n kyma-installer $(kubectl get pods --all-namespaces -l name=kyma-installer --no-headers -o jsonpath='{.items[*].metadata.name}')"
+    fi
+    if [ "$TIMEOUT_SET" -ne 0 ] && [ "$ITERATIONS_LEFT" -le 0 ]; then
+      echo "timeout reached on kyma installation error. Fetching logs from the installer:"
+      echo "----------"
+      echo "$(kubectl logs -n kyma-installer $(kubectl get pods --all-namespaces -l name=kyma-installer --no-headers -o jsonpath='{.items[*].metadata.name}'))"
```
Tomasz-Smelcerz-SAP: Is there a particular reason to use command substitution / subshell here? It's more explicit to just run the command directly.

installation/scripts/is-installed.sh
```diff
@@ -33,18 +86,29 @@ do
...
-      echo "----------"
-      echo "$(kubectl get po --all-namespaces)"
+    if [ "$TIMEOUT_SET" -ne 0 ] && [ "$ITERATIONS_LEFT" -le 0 ]; then
+      echo "timeout reached. Exiting"
+      exit 1
     fi
-    sleep $DELAY
   fi
+  if [ "${VERBOSE}" -eq 1 ]; then
+    echo "$(kubectl get installation/kyma-installation -o yaml)"
```
Tomasz-Smelcerz-SAP: Is there a particular reason to use command substitution / subshell here? It's more explicit to just run the command directly.

installation/scripts/is-installed.sh
kubadz: Probably not. I just left it like it was before. I agree that it should be run directly.

installation/scripts/is-installed.sh
kubadz: The same answer as in your previous comment.

docs/kyma/docs/033-inst-local-installation-scripts.md
```diff
@@ -78,3 +78,7 @@ The `installation/scripts/installer.sh` script creates the default RBAC role, in
 The script applies the Installation custom resource and marks it with the `action=install` label, which triggers the Kyma installation.
 
 >**NOTE:** The Kyma installation runs in the background. Execute the `./installation/scripts/is-installed.sh` script to follow the installation process.
+
+## The is-installed.sh script
+
+The `installation/scripts/is-installed.sh` script follows the process of Kyma installation. It checks the status on Installation CR and returns on `Installed` status. If there was a timeout set it fetches the logs from installer when the time reaches the timeout. If timeout is not set it waits infinitely.
```
tomekpapiernik:
```suggestion
The `installation/scripts/is-installed.sh` script shows the status of Kyma installation in real time. The script checks the status of the Installation custom resource. When it detects that the status changed to `Installed`, the script exits. If you define a timeout period and the status doesn't change to `Installed` within that period, the script fetches the installer logs when it reaches the timeout. If you don't set a timeout period, the script waits for the change of the status until you terminate it. 
```

Issue #607, opened 2018-09-10
Adjust CRDs domain to  'kyma-project.io'


PK85:
CRDs:
- Adjust domain of usagekinds.servicecatalog CRD to 'kyma-project.io'
- Adjust domain of servicebindingusages.servicecatalog CRD to 'kyma-project.io'

AC:
- group name is adjusted in the CRD
- all usage is adjusted
- documentation and examples are adjusted (examples repository as well)

polskikiel:
I decided to separate the renaming process into **3 steps**.
Each step is blocked by the previous one.

1. Rename BUC and regenerate its client.
#2071 (in progress)
2. Rename and align kyma revision in gopkg.toml in ui-api, acceptance, docs in all related repos.
#2104 (blocked)
3. Bump all the images, adjust RBAC roles, adjust CRDs and examples.

PR #2112, opened 2018-12-21
Bump images after BUC CRs domain renaming

polskikiel:
**Description**

- Bump BUC, acceptance, ui-api, ui-api-acceptance, lambda-ui and docs images after BUC CRs domain renaming
- Renamed SBU and UK CRDs group - from kyma.cx to kyma-project.io
- Adjusted examples
- Adjusted RBAC roles

**Related issue(s)**

#607


components/binding-usage-controller/docs/assets/architecture.html
```diff
@@ -2,10 +2,11 @@
 <!DOCTYPE html>
```
mszostok: there is a new rule about the diagrams:

>  Export the diagram as an SVG and save it under the corresponding assets directory.

source: https://github.com/kyma-project/community/blob/master/guidelines/content-guidelines/diagrams.md#tool

Because you already changing that part then IMO is worth to align with new rule and repalce  *html and *png files with *svg

PR #2124, opened 2018-12-24
Set ingressgateway to NodePort

sjanota:
**Description**

- ...
- ...
- ...

**Related issue(s)**


installation/resources/installer-config-cluster.yaml.tpl
```diff
@@ -94,6 +94,10 @@ data:
 
   security.enabled: "true"
 
+  # service type is later changed by isito p
```
kubadz:
```suggestion
   # service type is later changed by isito patch
```

Issue #1883, opened 2018-11-30
UI API Layer module pluggability


pkosiec:
**Description**

Implement module pluggability for UI API Layer, according to the accepted proposal: https://github.com/kyma-project/community/pull/141

**Documentation**
- [x] Create decision log for UI API Layer modularization concept (issue: https://github.com/kyma-project/kyma/issues/1721, proposal: https://github.com/kyma-project/community/pull/141)
- [x] Document ui-api-layer in Console topic in Kyma docs by explaining its purpose and scope, modularity and provide new CRD reference docs (https://github.com/kyma-project/community/blob/master/guidelines/templates/resources/custom-resource.md)

**UI API Layer**
- [x] Create custom resource and generate Informer API for it
- [x] Fork and modify https://github.com/josharian/impl to always return error for every method - PK: impl won't fit there, https://github.com/vektra/mockery will be a better pick
- [x] Implement production-ready module pluggability in UI API Layer based on PoC https://github.com/kyma-project/kyma/pull/1849. Hide it behind feature flag. Sync with @Abd4llA and his proposal https://github.com/kyma-project/community/pull/92 to do the feature flag properly
- [x] Make all domain modules pluggable
- [x] Make necessary modifications in UI API Layer acceptance tests
- [x] Split servicecatalog module to servicecatalog and servicecatalogaddons
- [x] Create BackendModule custom resources in Kyma components

**Console UI**
- [x] Set up a meeting with UI teams to show how to make resilient UIs, when some UI API Layer modules are turned off. Prepare example of modularity in the Service Catalog UI
- [x] Review all Console UI views - which crash when a module is not available. Prepare a list and working cluster. Best would be to do a pull request with such simulation. Remember about UI tests in the demo

**Reasons**

UI API Layer exposes GraphQL API for the Kyma Console. It consists of resolver logic for different domains, like Kubernetes, Service Catalog, Remote Environments, Kubeless, etc. As Kyma needs to be modularized, users should be able to install only the components they need.

The current approach we use in UI API Layer makes it impossible. During the GraphQL server start, we try to synchronize Stores of all Informers. As a result, if a resource of a specific type doesn't exist, UI API Layer won't start properly because its Informer will return an error.

pkosiec:
See GraphQL directives: https://gqlgen.com/reference/directives/

mjasinski5:
Review: @michal-hudy

mjasinski5:
#2084 Review: @michal-hudy

PR #2131, opened 2018-12-27
Introduce UI API Layer module pluggability

pkosiec:
**Description**

- Add GraphQL query for BackendModule custom resource
- Enable/Disable modules when specific BackendModule exists
- Fix client for BackendModule custom resource

Refactoring current modules and acceptance tests will be done in next pull request.

**Related issue(s)**
See #1883


components/ui-api-layer/internal/module/eventhandler.go
```diff
@@ -0,0 +1,104 @@
...
+func newEventHandler(module PluggableModule) *eventHandler {
+	return &eventHandler{module: module}
+}
+
+func (h *eventHandler) OnAdd(obj interface{}) {
+	resource, ok := obj.(*v1alpha1.BackendModule)
+	h.printIncorrectTypeErrorIfShould(ok, obj)
+
+	if !h.isAddOrDeleteEventRelatedToTheModule(resource) {
+		return
```
michal-hudy: What do you thing about adding some log with `warning` level in such case? Same in update event.

pkosiec: Wait a second... Actually, there shouldn't be any warnings as other modules could handle the event. Every pluggable module will be wrapped with this function, and will ignore events which don't belong to them. So it's hard to log warnings only for events that have been missed, for example when there is no module with given name.

PR #2157, opened 2019-01-04
Fix skipping UI API Layer acceptance tests when Dex is disabled

pkosiec:
**Description**

- Fix skipping UI API Layer acceptance tests when Dex is disabled

### DIFF
--- a/tests/ui-api-layer-acceptance-tests/internal/dex/dex.go
+++ b/tests/ui-api-layer-acceptance-tests/internal/dex/dex.go
@@ -5,11 +5,27 @@ import (
 	"log"
 	"os"
 	"strings"
+	"testing"
 
 	"github.com/pkg/errors"
 )
 
-func IsSCIEnabled() bool {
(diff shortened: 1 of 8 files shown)


tests/ui-api-layer-acceptance-tests/internal/dex/dex.go
```diff
@@ -5,11 +5,27 @@ import (
...
+
+	t.Skip(sciEnabledMessage)
+}
+
+func SkipMainIfShould() {
+	log.Println(sciEnabledMessage)
+	os.Exit(0)
+}
+
+func isSCIEnabled() bool {
```
michal-hudy: Hm... maybe we should change it to something similar as in console tests: https://github.com/kyma-project/console/blob/master/tests/ui-tests/utils/dex.js

pkosiec: good idea, but I'd rather remove this piece of code in order to always run our tests 😛

PR #2168, opened 2019-01-04
Split "servicecatalog" module in UI API Layer Acceptance tests

pkosiec:
**Description**

- Split `servicecatalog` module for UI API Layer Acceptance tests
- Fix skipping tests when SCI is enabled
- Check and handle errors in missing places

**Related issue(s)**
#1883


tests/ui-api-layer-acceptance-tests/internal/dex/dex.go
```diff
@@ -21,6 +21,10 @@ func SkipTestIfShould(t *testing.T) {
 }
 
 func SkipMainIfShould() {
+	if !isSCIEnabled() {
```
piotrmiskiewicz: In my opinion such method should have name like: ExitIfSCIEnabled or something like this. In the place where it is used nobody knows what it does.

pkosiec: Right. I will rename it

Issue #1855, opened 2018-11-29
(POC) Authorization concept for GraphQL


piotrmsc:
**Description**

Currently `core-ui` has request authentication and basic authorization based on istio envoy filter. Unfortunately, the authorization works only for normal HTTP requests but does not work for established WS connection. Additionally `core-ui` is caching retrieved data, and authorization for the cache is not targeted at all. 

During this POC we should implementing authorization directly in the `core-ui` instead of trying to secure it via envoy. The goal for this POC is a working **concept** which can be discussed later on with kyma community and be a foundation for authorization change in the next release.

Details : 
- Implement RBAC(SAR) check for requests made to core-ui. For that, we can use the logic from the apiserver-proxy and think about moving it to some library which could be then used in the core-ui as well as in other components. Use mentioned in the comment **directives**.

- Create http handler responsible for authenticating the token in order to fetch information about the user. Next use directive to do the authorization for the given user. Authorization will be based on RBAC, SubjectAccessReview will be used in order to ask if the user is allowed. 
- For the first iteration caching SAR responses is not in the scope.

Creating a common library for RBAC check is **optional**, can be copy-pasted from apiserver-proxy for the purpose of this POC. If time allows, such common library can be created but does not need to be imported and used in the `core-ui` already.

AC:
- Current logic in the envoy filter is replaced with authorization logic in the `core-ui` directly. Mutations are not yet supported
- Proposal created

**Reasons**

Current authorization does not work fully with WS, we do not have a concept for securing cache as well as it is in memory to which envoy filter does not have access. Developing and maintaining the Envoy filter is not the easiest and pleasant way as well. Plus, support for envoy filters in istio is on alpha stage which we already experience.

**Attachments**

pkosiec:
Recommended solution is to use GraphQL directives: https://gqlgen.com/reference/directives/

sjanota:
RBAC allows to grant access to specific resources by their names. It makes no sense for read operations in our scenario, as we cannot force it on subscriptions. There would need to be implemented check when receiving data from informer.

Issue #2147, opened 2019-01-02
Authorization for GraphQL


piotrmsc:
Follow up for  #1855

After #1855 gets done, the goal is to provide final implementation, fulfilling all of the minimal requirements.

Acceptance criteria:
- Envoy filter deleted
- AuthnPolicy deleted
- Plugability of core-ui working (check how SAR behaves for not defined CRDs)
- Authorization and authentication is fully done based on the POC but SAR and JWT authn logic is moved to a common library (if applicable)
- AuthZ&AuthN mechanism tested, integration test covers single resource, for example IDPPresets
- CI passing
- Documentation covered in #277

PR #2213, opened 2019-01-09
(POC) Authorization concept for GraphQL

kubadz:
**Description**
This PR is an implementation of POC for authorizing GraphQL resources in UI API component. The reason for introducing the new concept and requirements which this PR has to meet are described in issue #1855.

This PR is created only for review purposes. Please DO NOT MERGE.

major:
- switch to gorilla mux router to easily and unobtrusively add middleware
- add authentication middleware
- add user info obtained from authentication to context, so it can be used later for authorization
- define and implement checkRBAC directive (using k8s SAR) for checking if a user can perform requested operation on the resource
- use checkRBAC directive on IDPPresets as an example
- use checkRBAC directive on limitRanges to showcase usage of namespaceArg field in checkRBAC
- use checkRBAC directive on serviceBinding to showcase usage of nameArg field in checkRBAC

minor:
- use newer revision of envconfig package to be able to set default value on slice value in configuration

**Related issue(s)**

See also #1855


resources/core/charts/ui-api/templates/deployment.yaml
```diff
@@ -18,15 +18,17 @@ spec:
...
-        sidecar.istio.io/inject: "true"
       labels:
         app: {{ template "name" . }}
         chart: "{{ .Chart.Name }}-{{ .Chart.Version }}"
         release: "{{ .Release.Name }}"
         heritage: "{{ .Release.Service }}"
     spec:
       serviceAccountName: {{ template "fullname" . }}
+      {{ if .Values.global.isLocalEnv }}
+      hostNetwork: true
```
pkosiec: Do we need that?

sjanota: Yes. `dex.kyma.local` is not resolvable by DNS as there is static mapping in hosts `/etc/hosts`. Because of that component is not able to call dex and thus initialise OIDC authenticator.

sjanota:
Following scenarios tested manually (not automated) on minikube and Azure cluster:
* negative case (access denied) on queries (positive is tested automatically)
* updating user role changes user access 

Still missing:
- [ ] negative case on subscriptions (positive is tested automatically)
- [ ] negative case on mutations (positive is tested automatically)

kubadz:
Further work on that topic will be continued in PR #2317 under issue #2147.

PR #2231, opened 2019-01-10
Release 0.6.0 rc2

aszecowka:
Test prow pipeline - Release Candidate 2


resources/core/charts/console/values.yaml
```diff
@@ -3,7 +3,7 @@
 # Declare variables to be passed into your templates.
 replicaCount: 1
 image:
-  tag: 'fb88b3f0'
+  tag: '35238a98'
```
kwiatekus: Why so old image is taken for console? Can we use latest one?
```suggestion
  tag: 'c66360c1'
```

Issue #2072, opened 2018-12-19
Ensure kyma system components have sidecar injected


piotrmsc:
**Description**
(Update)
With https://github.com/kyma-project/kyma/issues/2073 we will introduce default sidecar injection in the kyma. We have to ensure that kyma-system components that do not already have a sidecar will remain without it, components with sidecar must still have it.

**Reasons**

Sidecars are required to make use of service-mesh functionality.

All kyma components which do not have sidecars should remain unchanged after enabling the policy - `sidecar.istio.io/inject: "false"` annotation should be added to such deployments. 

**Attachments**

Acceptance Criteria:

- Kyma components without envoy remain without it
- Kyma examples updated if necessary

sjanota:
I suggest other approach.
1. All Kyma components have explicit  `sidecar.istio.io/inject` annotation either true or false. Thanks to that Kyma operations no longer depends on injection policy.
2. Because policy is no longer a requirement for Kyma to run we can move it from patch to overrides. Patch is greatly simplified thanks to that.
3. When someone decides to install Kyma using their own Istio they have the freedom to choose injection policy. Policy is set to `enabled` by default, so one need to intentionally disable it.

piotrmsc:
Thx for your thoughts ;) 
In the istio patch we have to ensure the policy is enabled, policy enablement, however, can be done via override `{{ .Values.global.proxy.autoInject }}`. All kyma-system components that do not have sidecar should have annotation added. Injection should be enabled by default because we heavily rely on istio functionality. Users get `batteries included` with kyma, and have monitoring, tracing security and other feature enabled by default with the possibility to disable them if they want to.

Tomasz-Smelcerz-SAP:
One safe way to proceed is:

Step 1:
- Identify objects that can possibly create a Pod (istio injection target): Deployments, StatefulSets, ReplicaSets, Pods, etc.
- Add ```sidecar.istio.io/inject: "false"``` to every object that does not have it yet. For now, the policy is disabled, so this step should not change anything.
- To improve the review process, make separate PR's for every Kyma component. Lot's of PR's, but will be small, easy to review by owners.

After Step 1 is done, every component is explicitly labeled with ```sidecar.istio.io/inject``` annotation and Kyma is working as before.

Step 2:
- Separate PR that switches default istio policy from disabled to enabled.

Because every component is already labeled explicitly, this step should not actually change anything for existing components. But new Deployments will have sidecars injected automatically as required.

Step 3:
[...]

PR #2311, opened 2019-01-16
disable istio sidecar injection for components explictly

Tomasz-Smelcerz-SAP:
**Description**

**DO NOT MERGE**

For now, Istio injection is _disabled_ by default.
We plan to switch the default condition to _enabled_.
To ensure everything is working as before, we have to add an annotation to all K8s Pods that would be in the scope of the change.
These are Pods that do not have "sidecar.istio.io/inject" label yet and exist in a namespace labeled with: "istio-injection: enabled"

Pods are created by K8s resources like: Pods (obvious), Deployments, Jobs, StatefulSets, etc - everything that can spawn a Pod.

- Add `sidecar.istio.io/inject: "false"` to the involved resources.

**Related issue(s)**
See #2072

### DIFF
--- a/resources/istio/charts/grafana/templates/create-custom-resources-job.yaml
+++ b/resources/istio/charts/grafana/templates/create-custom-resources-job.yaml
@@ -58,6 +58,8 @@ spec:
   template:
     metadata:
       name: istio-grafana-post-install
+      annotations:
+        sidecar.istio.io/inject: "false"
       labels:
         app: istio-grafana
         release: {{ .Release.Name }}
--- a/resources/istio/charts/security/templates/cleanup-secrets.yaml
+++ b/resources/istio/charts/security/templates/cleanup-secrets.yaml
@@ -80,6 +80,8 @@ spec:
   template:
     metadata:
       name: istio-cleanup-secrets
+      annotations:
+        sidecar.istio.io/inject: "false"
       labels:
         app: {{ template "security.name" . }}
         release: {{ .Release.Name }}
--- a/resources/istio/charts/security/templates/create-custom-resources-job.yaml
+++ b/resources/istio/charts/security/templates/create-custom-resources-job.yaml
@@ -69,6 +69,8 @@ spec:
   template:
     metadata:
       name: istio-security-post-install
+      annotations:
+        sidecar.istio.io/inject: "false"
       labels:
         app: istio-security
         release: {{ .Release.Name }}
--- a/resources/service-catalog/charts/catalog/templates/apiserver-deployment.yaml
+++ b/resources/service-catalog/charts/catalog/templates/apiserver-deployment.yaml
@@ -27,8 +27,9 @@ spec:
         release: "{{ .Release.Name }}"
         releaseRevision: "{{ .Release.Revision }}"
         heritage: "{{ .Release.Service }}"
-      {{ if .Values.apiserver.annotations }}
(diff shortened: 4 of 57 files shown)


resources/istio/charts/grafana/templates/create-custom-resources-job.yaml
```diff
@@ -58,6 +58,8 @@ spec:
   template:
     metadata:
       name: istio-grafana-post-install
+      annotations:
```
Tomasz-Smelcerz-SAP: Is it necessary?

resources/istio/charts/security/templates/cleanup-secrets.yaml
```diff
@@ -80,6 +80,8 @@ spec:
   template:
     metadata:
       name: istio-cleanup-secrets
+      annotations:
```
Tomasz-Smelcerz-SAP: It is likely installed in istio-system namespace, this namespace does not have sidecar injection configured. Do we need this annotation here then?

resources/istio/charts/security/templates/create-custom-resources-job.yaml
```diff
@@ -69,6 +69,8 @@ spec:
   template:
     metadata:
       name: istio-security-post-install
+      annotations:
```
Tomasz-Smelcerz-SAP: Probably not needed.

sjanota:
You shouldn't change anything in istio charts. They do not rely on sidcar injection, but every pod that needs a envoy has it added explicitly.

resources/service-catalog/charts/catalog/templates/apiserver-deployment.yaml
```diff
@@ -27,8 +27,9 @@ spec:
         release: "{{ .Release.Name }}"
         releaseRevision: "{{ .Release.Revision }}"
         heritage: "{{ .Release.Service }}"
-      {{ if .Values.apiserver.annotations }}
       annotations:
+        sidecar.istio.io/inject: "false"
```
Tomasz-Smelcerz-SAP: Consider moving it into values.yaml file.

sjanota: Why?

resources/service-catalog/charts/catalog/templates/controller-manager-deployment.yaml
```diff
@@ -22,6 +22,7 @@ spec:
   template:
     metadata:
       annotations:
+        sidecar.istio.io/inject: "false"
```
Tomasz-Smelcerz-SAP: Consider moving it to values.yaml file

resources/service-catalog/charts/catalog/templates/apiserver-deployment.yaml
Tomasz-Smelcerz-SAP: Just to not forget about it :)
There are few cases that "inject" entries from values.yaml into spec.template.metadata.annotations using Go Template "if" blocks.
A uniform approach should be chosen: Either we ignore these blocks and add `        sidecar.istio.io/inject: "false"` or we use this mechanism so that all annotations are declared in one place in values.yaml.

Tomasz-Smelcerz-SAP:
This PR was replaced with few smaller ones, already merged to master.

PR #2337, opened 2019-01-17
Set sidecar injection to false in service-catalog, cluster-essential, and core charts

kfurgol:
**Description**

For now, Istio injection is disabled by default.
We plan to switch the default condition to enabled.
To ensure everything is working as before, we have to add an annotation to all K8s Pods that would be in the scope of the change.
These are Pods that do not have "sidecar.istio.io/inject" label yet and exist in a namespace labeled with: "istio-injection: enabled"
Pods are created by K8s resources like: Pods (obvious), Deployments, StatefulSets, etc - everything that can spawn a Pod.

- Add `sidecar.istio.io/inject: "false"` to the involved resources.

**Related issue(s)**
#2072

### DIFF
--- a/resources/cluster-essentials/charts/pod-preset/templates/controller-manager/stateful-set.yaml
+++ b/resources/cluster-essentials/charts/pod-preset/templates/controller-manager/stateful-set.yaml
@@ -17,6 +17,8 @@ spec:
   serviceName: {{ template "pod-preset.fullname" . }}-controller
   template:
     metadata:
+      annotations:
+        sidecar.istio.io/inject: "false"
       labels:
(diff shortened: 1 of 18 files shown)


michal-hudy:
You also need too add this annotation to the [`minio`](https://github.com/kyma-project/kyma/blob/master/resources/core/charts/minio) chart.

kfurgol:
@michal-hudy @mszostok it's split into several pull requests for easier reviewing:
https://github.com/kyma-project/kyma/pull/2327
https://github.com/kyma-project/kyma/pull/2331
https://github.com/kyma-project/kyma/pull/2340

PR #2348, opened 2019-01-17
Add plugin for backing up and restore lambda functions

venturasr:


venturasr:
PR https://github.com/kyma-project/kyma/pull/2365 will replace #2348

Issue #1469, opened 2018-10-30
If an arg is passed without a value, scripts exit without any error messages


suleymanakbas91:
**Description**

If an arg that needs a value is passed without specifying a value, then the scripts under `/installation/scripts` exit without any error. I tried with installer.sh and minikube.sh but since the other scripts have the same parsing logic, they might have the issue as well.

Try to run: 
`minikube.sh --domain`
`installer.sh --cr` 

**Expected result**

They should print out proper error messages.

strekm:
Validate input parameters for scripts:
 - run.sh
 - installer.sh
 - minikube.sh

Demonsthere:
I think we could make this more resilient by switching the argument passing to [getotps](https://wiki.bash-hackers.org/howto/getopts_tutorial) which has the required logic build in, and we do not need to reinvent it

PR #2356, opened 2019-01-18
Validate scripts input param fix

kfurgol:
**Description**
Fixes apply to:
  - minikube.sh
  - installer.sh
  - run.sh

- Scripts won't accept param in form with the equal sign (for example --cr=path)
- Added information on how validation function works

**Related issue(s)**
#1469


installation/cmd/run.sh
```diff
@@ -41,6 +47,7 @@ do
             shift
         ;;
         *)    # unknown option
+            echo "Unknown input parameter: ${1}"
```
sjanota: Every argument that is not a flag we treat as positional and we shouldn't log them. I you want to log invalid flags add another case:
```
--*)
  echo "Unknown flag ${1}"
  exit 1
  ;;
```
Apply to other scripts too.

installation/scripts/utils.sh
```diff
@@ -96,15 +96,21 @@ function showFailedResources {
...
+    if [ -z "${1}" ]; then
+        echo "Value parameter is empty"
         echo "Make sure parameter value is neither empty nor start with two hyphens"
         exit 1
     fi
-    if [ "${2:0:2}" == "--" ]; then
-        echo "Invalid parameter value for ${1}:"
-        echo "${2}"
+
+    if [ "${1:0:2}" == "--" ]; then
```
sjanota: I value of parameter starts with -- it's not its value, it's another flag. In such case I wouldn't report "Invalid value", but "Missing value for --my-flag". In such case you can simplify this function to one `if`.

Demonsthere:
Maybe consider using [getopts](https://wiki.bash-hackers.org/howto/getopts_tutorial) which has the logic for checking/setting arguments and required values build in?

sjanota:
I've seen other scripts using `getopts` in test-infra and I can't see benefits, but shorthand flags. I wouldn't change it now in all scripts.

Issue #2363, opened 2019-01-21
Communication between a K8S Service and Istio virtual service is broken


abbi-gaurav:
**Description**
While implementing a Publish served by Knative, it was discovered that a Publish service (a plain K8S service with Istio sidecar) cannot connect to Istio virtual service which represents a Knative eventing channel.

The flow is as below:
```
publish app (k8s deployment + istio side car ) ----> k8s service (channel) ---> istio virtual service (channel) ---> dispatcher service ---> dispatcher pod
```
e.g.
```
publish app --> foo-channel.svc.cluster.local --> foo-channel (Istio Virtual service) --> nats-dispatcher.knative-eventing.svc.cluster.local
```
error is : 

```
Route not found nats-dispatcher.knative-eventing.svc.cluster.local
```

>Note: **The same flow works on vanila K8s + Knative**.
This points to a possible discrepancy with Istio installation in Kyma

Please refer https://github.com/kyma-project/community/blob/master/sig-and-wg/wg-knative/assets/event-types.svg and https://github.com/kyma-project/community/blob/master/sig-and-wg/wg-knative/assets/publish-api.svg for details.

radufa:
Findings:
1. After installation of Kyma/Knative the "dispatcher" Knative eventing service was running without Istio sidecar. See in: https://github.com/knative/eventing/blob/master/config/provisioners/in-memory-channel/in-memory-channel.yaml
```
apiVersion: apps/v1
kind: Deployment
metadata:
  name: in-memory-channel-dispatcher
  namespace: knative-eventing
spec:
  replicas: 1
  selector:
    matchLabels: &labels
      clusterChannelProvisioner: in-memory-channel
      role: dispatcher
  template:
    metadata:
      annotations:
        sidecar.istio.io/inject: "true"
      labels: *labels
spec:
....
```

radufa:
The virtualservice created by KN for the channel is:
```
apiVersion: v1
items:
- apiVersion: networking.istio.io/v1alpha3
  kind: VirtualService
  metadata:
    clusterName: ""
    creationTimestamp: 2019-01-21T08:40:33Z
    generation: 1
    labels:
      channel: knlib-channel
      provisioner: in-memory-channel
    name: knlib-channel-channel
    namespace: default
    ownerReferences:
    - apiVersion: eventing.knative.dev/v1alpha1
      blockOwnerDeletion: true
      controller: true
      kind: Channel
      name: knlib-channel
[...]

Issue #2372, opened 2019-01-21
Test service is not cleaned after serving tests


sjanota:
**Description**

In Kyma there are knative-serving acceptance tests run to check if serving is running properly. There is test service defined in chart tests. Helm does not support removing anything else then pods, which causes service to stay even after tests.
 
One way to fix that would be to move creation and removal of service to code.

**Expected result**
Test service is removed after tests

**Actual result**
Test service is left

**Steps to reproduce**

Install Kyma with Knative and run tests.

**Troubleshooting**

k15r:
fixed in the latest knative-serving-tests

PR #2381, opened 2019-01-22
Do not run knative tests

sjanota:
**Description**

- Comment out serving acceptance tests as they are causing problems with eventing

**Related issue(s)**

Issue: #2363


Disper:
Are those comments temporary?
Will there be any follow-up on that?

sjanota:
Yep, they are low cost solution to unblock eventing integration. Follow up is here: https://github.com/kyma-project/kyma/issues/2372

Demonsthere:
Please ass the followup to the related issues :)

Issue #2353, opened 2019-01-18
We need to make the upgrade process between the Kyma releases independent from client installation overrides


mszostok:
**Description**

We need to make the upgrade process between the Kyma releases independent from client installation overrides.

**Plan to do**
To do that we need to split one `kyma-installer-{cluster/local}.yaml` big file. 
- One file called `kyma-installer-{cluster/local}.yaml` should contain only the kyma installer and kyma-installation CR. 
- Second `kyma-config-{cluster/local}.yaml` the file should contain the ConfigMaps/Secrets overrides required for the Kyma installer. 

The `kyma-config-cluster.yaml` should contain only overrides which need to be specified by user. All static values should be moved to Helm Charts (and `kyma-config-local.yaml` has all overrides in comparison to cluster). Thanks to that installation of Kyma without overrides are suitable for cluster eco-system and not for local setup.

In the installation process of Kyma, both files need to be applied. In the case of the upgrade only the `kyma-installer-{cluster/local}.yaml` should be applied. Thanks to that we will not have any conflicts with the overrides used during the installation process.

***NOTE:* This feature breaks our compatibility** because we will have 2 files required for Kyma installation. What's we will still have the kyma-config-cluster.yaml/kyma-config-local.yaml but with different content inside. 

The changes do not affect the Kyma release 0.6 because we will merge new things only to master branch. The release has dedicated branch realeas-0.6 both in Kyma and test-infra repos. All release jobs pre-rel06-* are configured for realeas-0.6 branch, e.g. https://github.com/kyma-project/test-infra/blob/master/prow/jobs/kyma/kyma-artifacts.yaml#L29-L33. So creating new patches for release 0.6 are still possible.

**Acceptance Criteria:**

- The [release-generate-kyma-installer-artifacts.sh](https://github.com/kyma-project/kyma/blob/master/installation/scripts/release-generate-kyma-installer-artifacts.sh) script generates all new required files
- Overrides from [installer-config-cluster.yaml.tpl](https://github.com/kyma-project/kyma/blob/master/installation/resources/installer-config-cluster.yaml.tpl) are moved directly to Helm Charts values
- Installation of Kyma without overrides are suitable for cluster eco-system and not for local setup.
- All required overrides for local installation are placed in  [installer-config-local.yaml.tpl](https://github.com/kyma-project/kyma/blob/master/installation/resources/installer-config-local.yaml.tpl) file.
- Prow release job is updated
- The release process is updated
- Kyma installation guides are updated
[...]

PR #2355, opened 2019-01-18
Install Kyma on AKS

adamwalach:
**Description**

Changes proposed in this pull request:

- Adds documentation for installation of Kyma on AKS
- Modifies installation resources (bug fix for AKS)  

**Related issue(s)**
* #2233

PR #2394, opened 2019-01-22
Move cluster overrides directly to Helm charts, split installation artifacts

mszostok:
**Description**

- Move cluster overrides directly to Helm charts, split installation artifacts. NOTE:
  - `acceptanceTest.application.disabled: "true"` was removed because it was not used by any chart
  - the `etcd-stateful.replicaCount: "3"` was applied as a global overrides for all chart. In this PR it's fixed and only service-catalog chart has this definition

**Related issue(s)**

- https://github.com/kyma-project/kyma/issues/2353


installation/resources/installer-config-cluster.yaml.tpl
```diff
@@ -68,53 +64,14 @@ data:
...
   namespace: kyma-installer
   labels:
     installer: overrides
     component: istio
 data:
-  global.proxy.includeIPRanges: "10.0.0.1/8"
-
-  security.enabled: "true"
-
   gateways.istio-ingressgateway.loadBalancerIP: "__EXTERNAL_PUBLIC_IP__"
```
adamwalach: We should also add (introduced in https://github.com/kyma-project/kyma/pull/2355):
```
 global.proxy.excludeIPRanges: "__PROXY_EXCLUDE_IP_RANGES__"
```

adamwalach:
```suggestion
   gateways.istio-ingressgateway.loadBalancerIP: "__EXTERNAL_PUBLIC_IP__"
   global.proxy.excludeIPRanges: "__PROXY_EXCLUDE_IP_RANGES__"
```

installation/resources/installer-config-cluster.yaml.tpl
```diff
@@ -68,53 +64,15 @@ data:
...
-kind: ConfigMap
-metadata:
-  name: service-catalog-overrides
-  namespace: kyma-installer
-  labels:
-    installer: overrides
-    component: service-catalog
-data:
-  etcd-stateful.etcd.resources.limits.memory: 512Mi
+   gateways.istio-ingressgateway.loadBalancerIP: "__EXTERNAL_PUBLIC_IP__"
```
mszostok:
```suggestion
  gateways.istio-ingressgateway.loadBalancerIP: "__EXTERNAL_PUBLIC_IP__"
  global.proxy.excludeIPRanges: "__PROXY_EXCLUDE_IP_RANGES__"
```

installation/resources/installer-config-cluster.yaml.tpl
```diff
@@ -29,17 +29,13 @@ metadata:
   labels:
     installer: overrides
 data:
-  global.isLocalEnv: "false"
-  global.knative: "false"
```
sjanota: This override needs to stay. Knative installation procedure relies on its existance.

resources/istio/values.yaml
```diff
@@ -1,5 +1,7 @@
 # Common settings.
 global:
+  isLocalEnv: false
```
sjanota: You cannot modify anything in istio chart as its raw copy from istio official release. Please, revert your changes and bring overrides back.

installation/resources/installer-config-cluster.yaml.tpl
mszostok: this will be covered in separate PR, linked to this issue: https://github.com/kyma-project/kyma/pull/2394

mszostok:
> What about changes in PowerShell scripts? 

Scope of this task: Move cluster overrides directly to Helm charts

I did that. I also modified all place where changes were required. In PowerShell scripts we do not have support for KNATIVE so I didnt modify there anything cause those scripts does not contains any logic for it.

You question is valid but not for this thask. IMO you should escalate and maybe report to some SIG :)

resources/istio/values.yaml
piotrmsc: this value is not needed here, in istio charts we do not use it as they raw from release like Szymon mentioned

installation/resources/installer-config-cluster.yaml.tpl
```diff
@@ -68,54 +64,15 @@ data:
...
   name: istio-overrides
   namespace: kyma-installer
   labels:
     installer: overrides
     component: istio
 data:
-  global.proxy.includeIPRanges: "10.0.0.1/8"
-  global.proxy.excludeIPRanges: "__PROXY_EXCLUDE_IP_RANGES__"
-
-  security.enabled: "true"
```
piotrmsc: **security.enabled: "true"** if it's removed from both overrides then it should be moved to values.yaml in istio

crabtree:
@mszostok In the `run.ps1` script there is some support for knative. I'm not sure how and if it works, but in my opinion, we should keep bash script and ps scripts consistent.

mszostok:
@crabtree In `run.ps1` there is support same as we have in `run.sh`.
But in `installer.ps1`  there is lack of this feature. If some knative developer wants to fix that, then he/she can add comment and I will do that by the way.

> (..) but in my opinion, we should keep bash script and ps scripts consistent.

I agree, but this is knative developers responsibility. 

my pr scope: Move cluster overrides directly to Helm charts 
not: ensure Knative support both for Windows and Unix

:)

resources/cluster-essentials/values.yaml
```diff
@@ -0,0 +1,3 @@
+global:
+  isLocalEnv: false
+  knative: false
```
marcobebway: Please add a newline here.

resources/xip-patch/values.yaml
```diff
@@ -3,3 +3,7 @@ containerRegistry:
 xip_patch:
   dir: develop/
   version: e9e635ea
+
+global:
+  isLocalEnv: false
+  knative: false
```
marcobebway: Please add a newline here.

PR #2429, opened 2019-01-24
Add ServiceMonitor CRD to cluster essentials

### LINKED ISSUE #1840: We should not install prometheus-operator when prometheus is modularized
**Description**

After modularization of prometheus we still have to install prometheus operator. This is needed in order to have the required CRDs for `service-monitors` installed by various kyma components. We should move them to monitoring folder. But we need some aligment with other teams on this.

derberg:
**Description**
This would be a temporary solution to enable local development without heavy Prometheus. Long term I will work with @PK85 on proposal how to do modularization the same way in all charts instead of solving modularization challenges case by case.

- Add ServiceMonitor CRD to cluster essentials
- Disable Prometheus on local installation

Kyma resource consumption on minikube without Prometheus:
```
Allocated resources:
  (Total limits may be over 100 percent, i.e., overcommitted.)
  CPU Requests  CPU Limits   Memory Requests  Memory Limits
  ------------  ----------   ---------------  -------------
  955m (23%)    1800m (45%)  3070024Ki (38%)  6246472Ki (77%)
```

**Related issue(s)**
Resolves #1840


a-thaler:
I don't like that prometheus-operator stays a separate module, should be merged with monitoring module. After talking to @joek it seems to not be a trivial taks to merge them, I will create a follow-up for that.

Issue #1744, opened 2018-11-21
HelmBroker fails to refresh repository urls on Minikube


adamwalach:
**Description**

HelmBroker is unable to refresh repository urls after an update. 
I tested it many times on my local minikube installation and it happens every time. After I modify the "APP_REPOSITORY_URLS" in the core-helm-broker deployment new pod is created and after ~90s I can see following message:
```
{"level":"error","log":{"message":"Waiting for service http://core-helm-broker to be ready timeout 1m30s exceeded.","time":"2018-11-21T09:25:40.788Z"}}
{"level":"error","log":{"message":"Last call error: Get http://core-helm-broker/statusz: dial tcp 10.102.85.252:80: i/o timeout","time":"2018-11-21T09:25:40.788Z"}}
```
As far as I can see this is a final state - there are no further attempts to access this url.
This makes the update impossible to perform.

Kyma installed from master branch
minikube version: v0.28.2 on MacOS

**Expected result**
Succesful refresh and new entries visible in the ServiceCatalog
```
{"level":"info","log":{"message":"Loading bundles from repositories","service":"syncer","time":"2018-11-21T09:28:50.472Z"}}
{"level":"info","log":{"message":"Bundle gcp-broker-provider:0.0.1 saved.","service":"syncer","time":"2018-11-21T09:28:53.436Z"}}
{"level":"info","log":{"message":"Bundle redis:0.0.3 saved.","service":"syncer","time":"2018-11-21T09:28:53.436Z"}}
{"level":"info","log":{"message":"Loading bundles finished","service":"syncer","time":"2018-11-21T09:28:53.436Z"}}
{"level":"info","log":{"action":"catalog","message":"action response","resp:services:count":2,"service":"server","time":"2018-11-21T09:28:53.436Z"}}
```

**Actual result**
Helm broker fails to refresh repositories with a "timeout" error.

```
{"level":"error","log":{"message":"Waiting for service http://core-helm-broker to be ready timeout 1m30s exceeded.","time":"2018-11-21T09:25:40.788Z"}}
{"level":"error","log":{"message":"Last call error: Get http://core-helm-broker/statusz: dial tcp 10.102.85.252:80: i/o timeout","time":"2018-11-21T09:25:40.788Z"}}
```

**Steps to reproduce**
1. ```kubectl -n kyma-system edit deployment core-helm-broker```
2. Modify `APP_REPOSITORY_URLS` and add a new repository
3. Wait till the new pod is created
[...]

PR #2444, opened 2019-01-25
Add workaround for minikube configuration of the Helm Broker

derberg:
**Description**

- Add workaround for minikube configuration of the Helm Broker

**Related issue(s)**
See also #1744

### DIFF
--- a/docs/service-brokers/docs/011-configuration-helm-broker.md
+++ b/docs/service-brokers/docs/011-configuration-helm-broker.md
@@ -51,7 +51,11 @@ Follow these steps to add the repository URL:
 
  ```bash
  URLS=$(kubectl get -n kyma-system deployment/core-helm-broker --output=jsonpath='{.spec.template.spec.containers[0].env[?(@.name=="APP_REPOSITORY_URLS")].value}')
+ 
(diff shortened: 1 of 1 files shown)


docs/service-brokers/docs/011-configuration-helm-broker.md
```diff
@@ -51,7 +51,11 @@ Follow these steps to add the repository URL:
 
  ```bash
  URLS=$(kubectl get -n kyma-system deployment/core-helm-broker --output=jsonpath='{.spec.template.spec.containers[0].env[?(@.name=="APP_REPOSITORY_URLS")].value}')
+ 
  kubectl set env -n kyma-system deployment/core-helm-broker -e APP_REPOSITORY_URLS="$URLS;http://custom.bundles-repository/bundles.yaml"
+ 
+ #At the moment there is a bug in Minikube, therefore for local installation you need to manually edit the broker and bump the relistRequests attribute
```
klaudiagrz:
```suggestion
 # At the moment, there is a bug in Minikube. As a result, for the local installation you need to manually edit the broker and bump the relistRequests attribute.
 
```

PR #2457, opened 2019-01-28
Fix test for multi dex providers

aszecowka:
**Problem**
On a nightly cluster, where stability checker executes all tests defined in a `testing.sh` file, we are going to enable Github authentication in addition to the static user. Thanks to that, any developer assigned to the proper Github team, can easily login to the Kyma console and download Kubeconfig file.
After enabling second authentication method, some tests start failing, because in tests there was an assumption that dex has only one, static connector and always redirection to the static connector is done.

**Solution**
In this PR, we added a new condition in code fetching token, and if /auth endpoint responds with HTTP 200, instead of making redirection, we fetch a link to static authentication by parsing returned HTML.

### DIFF
--- a/resources/core/values.yaml
+++ b/resources/core/values.yaml
@@ -16,8 +16,8 @@ global:
     enabled: false
     secretName: etcd-backup-abs-credentials
   acceptance_tests:
-    dir: pr/
-    version: PR-2457
+    dir: develop/
+    version: 501c4fe5
   alpine_net:
     dir: develop/
     version: ed568f0f
@@ -71,8 +71,8 @@ global:
     dir: develop/
     version: 0b8409ea
   ui_api_layer_acceptance_tests:
-    dir: pr/
-    version: PR-2457
+    dir: develop/
+    version: 0b8409ea
 
 test:
   acceptance:
--- a/resources/core/values.yaml
+++ b/resources/core/values.yaml
@@ -16,8 +16,8 @@ global:
     enabled: false
     secretName: etcd-backup-abs-credentials
   acceptance_tests:
-    dir: develop/
-    version: 501c4fe5
+    dir: pr/
+    version: PR-2457
   alpine_net:
     dir: develop/
     version: ed568f0f
@@ -71,8 +71,8 @@ global:
     dir: develop/
     version: 0b8409ea
   ui_api_layer_acceptance_tests:
-    dir: develop/
-    version: 0b8409ea
+    dir: pr/
+    version: PR-2457
 
 test:
   acceptance:
--- a/tests/acceptance/dex/idtokenprovider.go
+++ b/tests/acceptance/dex/idtokenprovider.go
@@ -2,13 +2,14 @@ package dex
 
 import (
 	"fmt"
-	"github.com/pkg/errors"
 	"io"
 	"io/ioutil"
 	"net/http"
 	"net/url"
 	"regexp"
 	"strings"
+
+	"github.com/pkg/errors"
 
 	"log"
 
--- a/tests/acceptance/dex/idtokenprovider.go
+++ b/tests/acceptance/dex/idtokenprovider.go
@@ -4,15 +4,13 @@ import (
 	"fmt"
 	"io"
 	"io/ioutil"
+	"log"
 	"net/http"
 	"net/url"
 	"regexp"
 	"strings"
 
 	"github.com/pkg/errors"
-
-	"log"
-
 	"golang.org/x/net/html"
 )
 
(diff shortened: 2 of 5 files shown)


resources/core/values.yaml
```diff
@@ -19,8 +19,8 @@ global:
     enabled: false
     secretName: etcd-backup-abs-credentials
   acceptance_tests:
-    dir: develop/
-    version: 501c4fe5
+    dir: pr/
```
aszecowka: only for testing purposes, will be reverted before merging

michal-hudy:
It would be nice to move logic responsible for authentication to `commons`. Please create follow-up task.

tests/acceptance/dex/idtokenprovider.go
```diff
@@ -2,12 +2,18 @@ package dex
 
 import (
 	"fmt"
+	"io"
```
mszostok: please sort imports

tests/acceptance/dex/idtokenprovider.go
```diff
@@ -44,48 +50,55 @@ func (p *dexIdTokenProvider) implicitFlow() (map[string]string, error) {
 	if err != nil {
 		return nil, err
 	}
-	if authorizeResp.StatusCode < 300 || authorizeResp.StatusCode > 399 {
-		return nil, fmt.Errorf("Authorize - response error: '%s' - %s", authorizeResp.Status, readRespBody(authorizeResp))
+	if authorizeResp.StatusCode < 200 || authorizeResp.StatusCode > 399 {
+		return nil, errors.New(fmt.Sprintf("Authorize - response error: '%s' - %s", authorizeResp.Status, readRespBody(authorizeResp)))
```
mszostok: why? in line 67 you left the `fmt.Errorf` and just make the first letter lowercase. Please unify your approach

tests/ui-api-layer-acceptance-tests/internal/graphql/idtokenprovider.go
```diff
@@ -40,66 +43,73 @@ func (p *dexIdTokenProvider) fetchIdToken() (string, error) {
...
 
 	// /approval?req=qruhpy2cqjvv4hcrbuu44mf4v
 	approvalEndpoint := loginResp.Header.Get("location")
 	if strings.Contains(approvalEndpoint, "#.*error") {
-		return nil, errors.New(fmt.Sprintf("Approval - Redirected with error: '%s'", approvalEndpoint))
+		return nil, fmt.Errorf("approval - Redirected with error: '%s'", approvalEndpoint)
 	}
-	approvalResp, err4 := p.httpClient.Get(p.config.DexConfig.BaseUrl + approvalEndpoint)
-	if err4 != nil {
-		return nil, err4
```
mszostok: `err1`, `err2`, `err3`, `err4`
dee

tests/acceptance/dex/idtokenprovider.go
```diff
@@ -44,48 +50,55 @@ func (p *dexIdTokenProvider) implicitFlow() (map[string]string, error) {
...
-	loginEndpoint := authorizeResp.Header.Get("location")
-	if strings.Contains(loginEndpoint, "#.*error") {
-		return nil, fmt.Errorf("Login - Redirected with error: '%s'", loginEndpoint)
+	var loginEndpoint string
+	if authorizeResp.StatusCode == 200 {
+		loginEndpoint, err = getLocalAuthEndpoint(authorizeResp.Body)
+		if err != nil {
+			return nil, errors.Wrapf(err, "while fetching link to static authentication")
+		}
+	} else {
```
mszostok: why u assuming that it is a redirection call? why we are not checking the status code?

tests/acceptance/dex/idtokenprovider.go
aszecowka: fixed

tests/acceptance/dex/idtokenprovider.go
aszecowka: Added following condition:
```
	switch authorizeResp.StatusCode {
	case http.StatusFound:
	case http.StatusOK:
	default:
		return nil, fmt.Errorf("got unexpected response on authorize: %d - %s", authorizeResp.StatusCode, readRespBody(authorizeResp))
	}
```

piotrmsc:
I agree with @michal-hudy  with the idea of extracting this to commons ;)

tests/acceptance/dex/idtokenprovider.go
```diff
@@ -44,48 +48,59 @@ func (p *dexIdTokenProvider) implicitFlow() (map[string]string, error) {
...
+	default:
+		return nil, fmt.Errorf("got unexpected response on authorize: %d - %s", authorizeResp.StatusCode, readRespBody(authorizeResp))
 	}
 
 	// /auth/local?req=qruhpy2cqjvv4hcrbuu44mf4v
-	loginEndpoint := authorizeResp.Header.Get("location")
-	if strings.Contains(loginEndpoint, "#.*error") {
-		return nil, fmt.Errorf("Login - Redirected with error: '%s'", loginEndpoint)
+	var loginEndpoint string
+	if authorizeResp.StatusCode == 200 {
```
piotrmsc: you already check this status code few lines above, why duplicate yourself? if you need this double check, consider comparing with `http.StatusOK` like you do above in the switch

aszecowka: Few lines above I just validate a response, if the status code is unexpected I exit with an error.
Here I will change 200 to http.StatusOK. Is it blocking this PR?

piotrmsc: ofc it's not ;-)

PR #2470, opened 2019-01-29
Add bash trap to delete GC credentials file

### LINKED ISSUE #2469: Add bash trap to delete GC credentials file
**Description**

Add bash trap to delete Google Cloud credentials file made in `tools/gcp-broker-provider/bin` by `gcp-broker.sh`

**Reasons**

After using `gcp-broker.sh` sa-key is left in repository, which may lead to leakage in accidental push.

aerfio:
**Description**

Add bash trap to delete Google Cloud credentials file made in `tools/gcp-broker-provider/bin` by `gcp-broker.sh`

- add bash trap to remove secure data (GC credentials)

**Related issue(s)**
Resolves #2469

### DIFF
--- a/tools/gcp-broker-provider/bin/gcp-broker.sh
+++ b/tools/gcp-broker-provider/bin/gcp-broker.sh
@@ -13,6 +13,16 @@ set -o errexit # exit immediately if a command exits with a non-zero status.
 readonly CURRENT_DIR="$(CDPATH= cd -- "$(dirname -- "$0")" && pwd)"
 readonly GCP_SECRET_NAME="gcp-broker-data"
 
+trap cleanup EXIT SIGINT SIGTERM
+
+function cleanup() {
+    printf  "\n- Cleaning up...\n"
+    echo "Ensure gcloud credentials are removed"
+    rm -f "${GOOGLE_APPLICATION_CREDENTIALS}"
+    echo "OK"
+}
+
+
 function discoveredUnsetVar() {
     local discoveredUnsetVar=false
     for e in WORKING_NAMESPACE; do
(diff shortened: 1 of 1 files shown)


tools/gcp-broker-provider/bin/gcp-broker.sh
```diff
@@ -57,6 +57,12 @@ function configureGCloud() {
     return 0
 }
 
+function deleteGoogleApplicationCredentials () {
```
mszostok: we can make it more general and put at the beginning of the script this:

```bash
trap cleanup EXIT SIGINT SIGTERM

function cleanup() {
    printf  "\n- Cleaning up...\n"

    echo "Ensure gcloud credentials are removed"
    rm "${GOOGLE_APPLICATION_CREDENTIALS}"
    echo "OK"
}
```

for now, you have one specified command for clean-up. To make it more extensible you can just define the clean-up function and as a first step implement cleaning the credentials, in future we can add next steps. 

what do u think?

aerfio: Yeah, good idea 👍

tools/gcp-broker-provider/bin/gcp-broker.sh
```diff
@@ -13,6 +13,16 @@ set -o errexit # exit immediately if a command exits with a non-zero status.
 readonly CURRENT_DIR="$(CDPATH= cd -- "$(dirname -- "$0")" && pwd)"
 readonly GCP_SECRET_NAME="gcp-broker-data"
 
+trap cleanup EXIT SIGINT SIGTERM
+
+function cleanup() {
+    printf  "\n- Cleaning up...\n"
+    echo "Ensure gcloud credentials are removed"
+    rm -f "${GOOGLE_APPLICATION_CREDENTIALS}"
```
michal-hudy: You should also verify if `GOOGLE_APPLICATION_CREDENTIALS` is not empty

Issue #1488, opened 2018-10-31
Improve local development with admin user


derberg:
**Description**

- for local installation keep static dummy password
- would be awesome to display it in the UI or autologin on local

**Reasons**

admin static user is now secured and the password is always unique. Problem is that it is also secured for local and it is complicating dev flow, with each installation you always need to check the new password.

piotrmsc:
Static password cannot stay but we can improve UX in other way.
We should allow users to specify their own password during the installation phase, if not specified then generate it as we do now.
In such case we can add a flag ie. `-p` to startup script which will pass given string to the overrides, then dex patch can use this values from helm. 
Input password parameter should be base64 encoded as later on dex patch bcrypts it.

tomekpapiernik:
Documentation must be updated to indicate that local deployments can have their admin password set by the user.

kfurgol:
@kfurgol  - assigne for bash modifications

PR #2471, opened 2019-01-29
Improve local development with custom password for admin account

kfurgol:
**Description**
For local installation `run.sh` script can be started with `--p your_password` flag

- Added --p flag to run.sh script
- Added global.adminPassword override
- Provided changes in dex chart to use a custom password

**Related issue(s)**
#1488

### DIFF
--- a/installation/cmd/run.sh
+++ b/installation/cmd/run.sh
@@ -40,6 +40,12 @@ do
             KNATIVE="--knative"
             shift
         ;;
+        --p)
+            checkInputParameterValue "$2"
+            ADMIN_PASSWORD="${2}"
+            shift # past argument
+            shift # past value
+        ;;
         --*)
             echo "Unknown flag ${1}"
             exit 1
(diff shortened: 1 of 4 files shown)


installation/cmd/run.sh
```diff
@@ -40,6 +40,12 @@ do
             KNATIVE="--knative"
             shift
         ;;
+        --p)
```
strekm: consider `--password` instead

kfurgol: done

installation/resources/installer-config-local.yaml.tpl
```diff
@@ -34,6 +34,7 @@ data:
   global.domainName: "kyma.local"
   global.etcdBackup.containerName: ""
   global.etcdBackup.enabled: "false"
+  global.adminPassword: ""
```
strekm: do you need that here?

Issue #2222, opened 2019-01-10
Enable Service Catalog standalone installation - phase 1


PK85:
**Description**

Make Service Catalog component e2e (backend, UI, installation process) modular, that means Service Catalog and all dependencies(are modular themselves as well) are installed only. That means to have a minimum setup in Kyma to be able to login and see only Service Catalog in UI and only custom resources from Service Catalog domain.

**Reasons**

1. Next phase of Kyma modularization.  It's the first step towards enablement only selected Kyma components in a standalone fully working solution.
2. It will enable to develop and run Kyma locally in a lightweight version and do not be disturbed by other components.
3. Will make contribution to Kyma easier (see point above). Will allow to install Service Catalog UI to any k8s cluster also with already installed Service Catalog.
4. Will make possible and easier to donate SC UI and SC addons to the kubernetes community in the future.

**Acceptance Criteria**
- Move Service Catalog Addons chart from core to root resources.
- List all components that are required to be installed to make Service Catalog and SC UI working (like dex, ui-api-layer), document that.
- Make instruction how to install Kyma without Service Catalog.
- document next step, to make our test to be modular as well (example: acceptance-test, kubeles test using SBU etc)

**Out of scope**
- UI will display all others component. But they will return 404. Next development phase will fix that.

**Attachments**

adamwalach:
| Subtask | PRs |
| -------- | ---- |
| Move Service Catalog Addons chart from core to root resources. | https://github.com/kyma-project/kyma/pull/2515 |
List all components that are required to be installed to make Service Catalog and SC UI working (like dex, ui-api-layer), document that.| |
Make instruction how to install Kyma without Service Catalog.| |
document next step, to make our test to be modular as well (example: acceptance-test, kubeles test using SBU etc)| |

adamwalach:
Additional acceptance criteria:
- move Helm Broker from core to root resources
- list dependencies to Service Catalog add-ons (we want to consolidate some of the modules)
- tests documentation - tests only per component, e2e should be triggered only during installation on cluster

mszostok:
### General points for testing after modularization

TL;DR; 
- Component specific tests should be moved directly to the related component charts.
- Component tests should be more independent. Currently, e.g. Kubeless tests use the Helm Broker classes. Instead of that, they should register their own broker for that purpose.
- Component specific tests can required that their dependencies are installed. Cause there is no reason to e.g. test helm-broker without preinstalled service catalog.

[...]

PR #2515, opened 2019-01-30
Make service catalog addons as a root chart

adamwalach:
**Description**

- Updated paths for service catalog addons

**Related issue(s)**
#2222


mszostok:
Handle changing docs also in bundles repository

resources/helm-broker/README.md
```diff
@@ -0,0 +1,7 @@
+# Helm Broker
+
+## Overview
+
+Helm Broker provides bundles in the [Service Catalog](../service-catalog/README.md). A bundle is an abstraction layer over a Helm chart which enables you to provide more information about the Helm chart. For example, a bundle can provide plan definitions or binding details. Service Catalog requires this information. Bundles are services available in Service Catalog.
+
+Helm Broker implements the [Service Broker API](https://github.com/openservicebrokerapi/servicebroker/blob/master/spec.md). For more information about the Service Brokers, see the [Service Brokers overview](../../../../docs/service-brokers/docs/001-overview-service-brokers.md) documentation. Find the details about the Helm Broker in the [docs](../../docs/service-brokers/docs) repository.
```
klaudiagrz: Could you please fix the broken Service Brokers overview link? :(

adamwalach: Fixed

resources/service-catalog-addons/requirements.yaml
```diff
@@ -0,0 +1,9 @@
+dependencies:
```
PK85: please add req file to helm-broker as well

resources/helm-broker/values.yaml
```diff
@@ -22,15 +22,21 @@ config:
...
-etcd:
-  # By default etcd subchart will be release with name "core-etcd", and it's to much generic
-  # but currently there is no way to pass to the etcd subchart a parent Release.Name, see https://github.com/kubernetes/helm/issues/2506
-  # so to make it more verbose I've added nameOverride and now the release name will be "core-helm-broker-etcd"
-  nameOverride: "helm-broker-etcd"
\ No newline at end of file
+global:
+  containerRegistry:
+    path: eu.gcr.io/kyma-project
+  istio:
```
PK85: we do not use istio right now, please remove

resources/service-catalog-addons/requirements.yaml
adamwalach: Fixed

resources/helm-broker/values.yaml
adamwalach: Fixed

resources/core/templates/tests/test-core-acceptance.yaml
```diff
@@ -15,7 +15,7 @@ spec:
       image: {{ .Values.global.containerRegistry.path }}/{{ .Values.global.acceptance_tests.dir }}acceptance-tests:{{ .Values.global.acceptance_tests.version }}
       env:
       - name: HELM_BROKER_URL
-        value: http://{{.Release.Name }}-helm-broker.{{ .Release.Namespace }}.svc.cluster.local
+        value: http://helm-broker.kyma-system.svc.cluster.local
```
pkosiec: If this is not equal to the `.Release.Namespace` anymore, maybe we could put it to global values

adamwalach: According to my knowledge there is no need to keep this value as a variable. We have no scenario where we would need to override it. The global variable would have to be broker specific, something like `helm-broker-release-namespace`.

Issue #1716, opened 2018-11-19
Add support for only once provisioning in Helm Broker


mszostok:
**Description**

Helm broker should support only once provisioning for given Bundle.

Add new flag **provisionOnlyOnce** under the [metadata](https://github.com/kyma-project/kyma/blob/master/docs/service-brokers/docs/012-configuration-helm-broker-bundles.md#the-bundle-metayaml-file) information of given Bundle.

If flag **provisionOnlyOnce** is set to `true` then Helm Broker triggers provisioning of given bundle only once (at the first time). All next requests for creating ServiceInstance for this bundle should be blocked. HelmBroker should not trigger installation of that bundle and immediately returns an error that only one provision action is allowed for that ServiceClass.

![screen shot 2018-11-19 at 12 49 52](https://user-images.githubusercontent.com/17568639/48705520-a3113c80-ebf9-11e8-8798-5e8eb089004c.png)

> **NOTE:** Backward compatibility is kept because the new workflow is executed only when a new flag is specified.

**Reasons**
There is a use-case that provisioning given Bundle multiple time can cause an error. For provisioning GCP Broker Provider service class more than once.

**AC**
- Helm Broker supports **provisionOnlyOnce** flag  in a context of given namespace.
- That piece of information needs to be available in ServiceClass as well. That in the future UI could make a class disabled. Implementation: add to labels: "once" at runtime
- Mark Azure and GCP Service Brokers with new flag. Update theirs documentation in an overview section. Release bundles repo  and apply to kyma.
- Documentation for Helm Broker updated, see: https://github.com/kyma-project/kyma/blob/master/docs/service-brokers/docs/012-configuration-helm-broker-bundles.md#the-bundle-metayaml-file

jasiu001:
| Subtask | PRs |
| -------- | ---- |
| Modify Helm broker component in Kyma, add tests. Modify metadata model sending as a request | PJ |
|  Mark Azure and GCP Service Brokers with new flag. Update theirs documentation in an overview section. Release bundles repo and apply to kyma. | |
| Update Helm Broker documentation https://github.com/kyma-project/kyma/blob/master/docs/service-brokers/docs/05-02-helm-broker-bundles.md#the-bundle-metayaml-file | |

jasiu001:
[...]

PR #2528, opened 2019-01-31
Add support for onlyOnceFlag to helm broker component

jasiu001:
Related to #1716

### DIFF
--- a/components/helm-broker/internal/broker/automock/instance_storage.go
+++ b/components/helm-broker/internal/broker/automock/instance_storage.go
@@ -31,6 +31,29 @@ func (_m *InstanceStorage) Get(id internal.InstanceID) (*internal.Instance, erro
 	return r0, r1
 }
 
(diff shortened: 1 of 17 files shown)


piotrmiskiewicz:
The error message is:
```
Error provisioning ServiceInstance of ClusterServiceClass (K8S: "84dad23a-794d-404f-92a1-5e73e91a74af" ExternalName: "azure-service-broker") at ClusterServiceBroker "core-helm-broker": Status: 400; ErrorMessage: during provisioning bundle with once instance while bundle instance already exist; Description: ; ResponseError: <nil>
```

maybe we can change the message value "during provisioning bundle with once instance while bundle instance already exist"

klaudiagrz:
Maybe you can use sth like this as an ErrorMessage:

`This bundle has a provisioningOnlyOnce flag and an instance of this bundle already exists.`

Issue #2279, opened 2019-01-14
Create event-bus-knative helm chart


abbi-gaurav:
To enable Kyma served by Knative eventing, we need to have a separate parent chart that contains all the artifacts needed for that.

This chart is installed based on the feature flag. 

This will require moving default event-bus out of core to kyma/resources.

**AC**
- [ ] `event-bus` chart has been moved out of core to kyma/resources.
- [ ] `event-bus-knative` chart has been created  under kyma/resources
- [ ] `event-service` can discover the `event-bus-publish` app.
- [ ] Chart is installed based on knative feature flag.
- [ ] If the flag is active, the default event-bus is not installed.

PR #2566, opened 2019-02-04
Move event-bus Helm chart to resources

montaro:
Create event-bus-knative Helm chart and make it conditionally installed only if the `--knative` is provided
Prevent the `event-bus` installation in case of `--knative`
collective parent chart
Add publish-knative service
change the global.natsStreaming.clusterID to be a global value
rename the global.eventBus to global.kymaEventBus
revert the application-operator resource
CODEOWNERS

See also #2279


CODEOWNERS
```diff
@@ -93,8 +93,11 @@
 # The `minio` subchart
 /resources/core/charts/minio/ @akucharska @michal-hudy @pkosiec @mjasinski5 @aerfio @magicmatatjahu
 
-# The `event-bus` subchart
-/resources/core/charts/event-bus/ @a-thaler @montaro @abbi-gaurav @marcobebway @radufa @suleymanakbas91
+# The `event-bus` chart
+/resources/charts/event-bus/ @a-thaler @montaro @abbi-gaurav @marcobebway @radufa @suleymanakbas91
```
marcobebway: Should this be **/resources/event-bus/**?

CODEOWNERS
```diff
@@ -93,8 +93,11 @@
...
+
+# The `nats-streaming` chart
+/resources/charts/nats-streaming/ @montaro @abbi-gaurav @marcobebway @radufa @suleymanakbas91
```
marcobebway: Should this be **/resources/nats-streaming/**?

CODEOWNERS
montaro: fixed

CODEOWNERS
montaro: fixed

installation/scripts/installer.sh
```diff
@@ -95,6 +95,7 @@ kubectl apply -f - <<<"$COMBO_YAML"
 
 if [ $KNATIVE ]; then
     kubectl -n kyma-installer patch configmap installation-config-overrides -p '{"data": {"global.knative": "true"}}'
+    kubectl -n kyma-installer patch configmap installation-config-overrides -p '{"data": {"global.kymaEventBus": "false"}}'
```
sjanota: If those two publishers cannot be installed together in one cluster I would rely on one flag: `global.knative`. There should be only one flag to enable whole knative functionality. If we keep adding new flags it will quickly get unmaintainable. I'm not sure how to solve it in chart though. You probably need to wrap every resource in `{{ if not .global.knative }}`.

montaro: Unfortunately, Helm doesn't support the `not` operator, that's why I had to have two :(

installation/resources/installer-config-cluster.yaml.tpl
```diff
@@ -84,3 +84,25 @@ metadata:
...
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: nats-streaming-overrides
+  namespace: kyma-installer
+  labels:
+    installer: overrides
+    component: nats-streaming
+data:
+  global.natsStreaming.clusterID: "kyma-nats-streaming"
```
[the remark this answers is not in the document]
montaro: OK, will be fixed, the idea was to keep the value in one place instead of being copied in two helm charts

installation/scripts/installer.sh
sjanota: Helm does not support not in `requirements.yaml`, but it does in charts. I cannot think of better way to do this, but in my opinion we cannot allow two flags to allow same thing.

installation/scripts/installer.sh
```diff
@@ -96,7 +96,7 @@ kubectl apply -f - <<<"$COMBO_YAML"
 if [ $KNATIVE ]; then
     kubectl -n kyma-installer patch configmap installation-config-overrides -p '{"data": {"global.knative": "true"}}'
     kubectl -n kyma-installer patch configmap installation-config-overrides -p '{"data": {"global.kymaEventBus": "false"}}'
-    kubectl -n kyma-installer patch configmap nats-streaming-overrides -p '{"data": {"global.natsStreaming.clusterID": "kyma-nats-streaming"}}'
+    kubectl -n kyma-installer patch configmap installation-config-overrides -p '{"data": {"global.natsStreaming.clusterID": "knative-nats-streaming"}}'
```
sjanota: Why do you need this override?

installation/scripts/installer.sh
piotrmsc: How about https://github.com/helm/helm/blob/master/docs/chart_template_guide/functions_and_pipelines.md#operators-are-functions

resources/event-bus/charts/publish-knative/templates/deployment.yaml
```diff
@@ -0,0 +1,20 @@
+# dummy pod   
+apiVersion: apps/v1 # for versions before 1.9.0 use apps/v1beta2
+kind: Deployment
```
piotrmsc: Looks like not needed?

installation/scripts/installer.sh
sjanota: Ok, let's leave it as is. Two points:
 * Please join those two flags in one patch
 * You also have to update [installation docs](https://github.com/kyma-project/kyma/blob/master/docs/kyma/docs/04-09-knative.md)

resources/event-bus/charts/publish-knative/templates/deployment.yaml
montaro: It's deliberately added like that as a place holder for the coming real implementation in the coming pull requests.

installation/scripts/installer.sh
montaro: Thanks, @piotrmsc, and @sjanota for your review. I tried now different approaches to avoid the second flag but, unfortunately, none worked. I'll keep them in a single dict as advised.

installation/scripts/installer.sh
montaro: Because, in the case of knative, we need to change the NATS Streaming Cluster id *not* to have the default `kyma-` prefix but to have the `knative-` prefix instead.

resources/event-bus/charts/publish-knative/templates/deployment.yaml
piotrmsc: ok

installation/scripts/installer.sh
piotrmsc: We talked offline about it and joing flags is fine

docs/kyma/docs/04-09-knative.md
```diff
@@ -34,5 +35,6 @@ When you install Kyma locally from sources, add the `--knative` argument to the
 
 Run the following command before triggering the Kyma installation process:
 ```
-kubectl -n kyma-installer patch configmap installation-config-overrides -p '{"data": {"global.knative": "true"}}'
+kubectl -n kyma-installer patch configmap installation-config-overrides -p '{"data": {"global.knative": "true", "global.kymaEventBus": "false"}}'
+kubectl -n kyma-installer patch configmap installation-config-overrides -p '{"data": {"global.natsStreaming.clusterID": "knative-nats-streaming"}}'
```
sjanota: You join this patch into the previous one.

Issue #2113, opened 2018-12-21
Don't ignore errors in UI API Layer


pkosiec:
**Description**

Informer API has a method `AddIndexers`, which we use in services, for example here: https://github.com/kyma-project/kyma/blob/master/components/ui-api-layer/internal/domain/servicecatalog/clusterserviceplan_service.go#L16. It returns error, but we ignore them.

Maybe we could use github.com/kisielk/errcheck to avoid this kind of issues in the future, similar as in binding-usage-controller.

AC:
- [x] Setup github.com/kisielk/errcheck on CI as a part of before-commit.sh
- [x] Handle errors while using `AddIndexers` method and all others detected by the tool

PR #2581, opened 2019-02-05
Don't ignore errors in UI API Layer

tgorgol:
**Description**

- Add errcheck tool to ui-api-layer and ui-api-layer acceptance tests before-commit.sh
- Handle errors detected by errcheck

**Related issue(s)**

#2113 #2587


components/ui-api-layer/internal/domain/application/app_service_test.go
```diff
@@ -500,7 +502,10 @@ func newDummyInformer() cache.SharedIndexInformer {
 func newTestServer(data string, statusCode int) *httptest.Server {
 	return httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		w.WriteHeader(statusCode)
-		fmt.Fprintln(w, data)
+		_, err := fmt.Fprintln(w, data)
+		if err != nil {
+			glog.Warningf("Unable to write response body. Cause: %v", err)
```
pkosiec: log as an error

components/ui-api-layer/internal/domain/application/app_service.go
```diff
@@ -290,8 +293,14 @@ func (svc *applicationService) extractErrorCause(body io.ReadCloser) error {
 }
 
 func (svc *applicationService) drainAndCloseBody(body io.ReadCloser) {
-	_ = iosafety.DrainReader(body)
-	body.Close()
+	err := iosafety.DrainReader(body)
+	if err != nil {
+		glog.Warningf("Unable to drain body reader. Cause: %v", err)
```
pkosiec: maybe it would be good to return these errors. If this is not an option, log them with error level

components/ui-api-layer/main.go
```diff
@@ -75,7 +75,10 @@ func exitOnError(err error, context string) {
 
 func parseFlags(cfg config) {
 	if cfg.Verbose {
-		flag.Set("stderrthreshold", "INFO")
+		err := flag.Set("stderrthreshold", "INFO")
+		if err != nil {
+			glog.Error(errors.Wrap(err, "while parsing flags"))
```
pkosiec: Please consider `glog.Fatal`- but it's totally up to you

tgorgol: I decided that the error occurring during setting that flag wasn't critical because it wasn't really related to parsing user provided flags and that logging the error without exiting the application would suffice.

PR #2583, opened 2019-02-05
Remove duplicated annotations section in dex chart

kfurgol:
**Description**

- Removed duplicated annotations section in dex chart

**Related issue(s)**
#2072

### DIFF
--- a/installation/resources/installer-local.yaml
+++ b/installation/resources/installer-local.yaml
@@ -3,6 +3,8 @@ apiVersion: v1
 kind: Namespace
 metadata:
   name: kyma-installer
+  labels:
+    istio-injection: disabled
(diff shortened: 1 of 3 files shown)


resources/istio-kyma-patch/templates/configmap.yaml
```diff
@@ -77,6 +77,7 @@ data:
   injection-in-namespaces: |
     istio-system disabled
     kube-system disabled
+    kyma-installer disabled
```
piotrmsc: kyma-installer is our ns so we can label it there directly

kfurgol: done

PR #2621, opened 2019-02-06
API controller: add API validation - allow one API per service

jakkab:
**Description**

- api-controller validates if an API already exists for a specified Kubernetes service
- tests


components/api-controller/pkg/controller/v1alpha2/controller.go
```diff
@@ -204,22 +206,59 @@ func (c *Controller) onCreate(api *kymaApi.Api) error {
...
+	}
+
+	targetServiceName := newAPI.Spec.Service.Name
+
+	existingAPIs, err := c.apisLister.List(labels.Everything())
+	if err != nil {
+		log.Errorf("Error while listing APIs %s/%s ver: %s. Root cause: %s", newAPI.Namespace, newAPI.Name, newAPI.ResourceVersion, err)
+		return setStatus(kymaMeta.Error)
+	}
+
```
piotrmsc: You need to validated api for service in the given namespace of the created api

components/api-controller/pkg/apis/gateway.kyma-project.io/v1alpha2/types.go
```diff
@@ -46,31 +46,37 @@ type JwtAuthentication struct {
...
 	AuthenticationStatus kymaMeta.GatewayResourceStatus `json:"authenticationStatus,omitempty"`
 	VirtualServiceStatus kymaMeta.GatewayResourceStatus `json:"virtualServiceStatus,omitempty"`
 }
 
 func (s *ApiStatus) IsEmpty() bool {
-	return s.VirtualServiceStatus.IsEmpty() && s.AuthenticationStatus.IsEmpty()
+	return s.VirtualServiceStatus.IsEmpty() && s.AuthenticationStatus.IsEmpty() && s.ValidationStatus.IsEmpty()
 }
 
 func (s *ApiStatus) IsDone() bool {
```
piotrmsc: according to renaming should be `isSuccessful` here as well :)

piotrmsc:
If you create an API for a service that does not have API attached but you have selected wrong service and make an update to correct one, Virtual service and Policy(if set) should be deleted for the wrong one (the first selected)

piotrmsc:
Please extend our api-controller-acceptnce-tests with assertions for validationStatus.
Please also create an integration test for scenarios
And while chaning that please consider making this line https://github.com/kyma-project/kyma/blob/master/tests/api-controller-acceptance-tests/apicontroller/component_test.go#L252
configurable via env like it is with domain for example

jakkab:
@piotrmsc Virtual Services and Polices are updated as well, no need to do delete anything.

Follow-up will be created.

Issue #277, opened 2018-08-27
Make docs about GraphQL security official


derberg:
Confirm these statements before you submit the issue:

- [x] I have searched open and closed issues for duplicates.
- [x] I have read the contributing guidelines.
---

**Description**

[This](https://github.com/kyma-project/kyma/blob/master/resources/core/charts/ui-api/README.md)
should most probably go [here](https://github.com/kyma-project/kyma/tree/master/docs/authorization-and-authentication/docs) as we need clear security documentation. We would also probably need some general section about this in the overarching kyma docs

 other ideas? I'm all ears

**Reason**

Security applied on graphql layer that serves as catalog ui backend must be visible in official docs set

piotrmsc:
Update: While working on a #2147 mentioned document can be deleted or it's content replaced with new security mechanism described.

Document should:

- Describe how authorization and authentication is done in `core-ui`
- What RBAC roles are required for specific resources
- How to add a new role to secure access to the resource
- **not** mention envoy filter and **istio** RBAC (ServiceRole&Binding) as it will no longer be used

PR #2643, opened 2019-02-07
Add instructions on how to secure graphql

kubadz:
**Description**

- add instructions on how to secure graphql

**Related issue(s)**

See also #277


components/ui-api-layer/docs/security.md
```diff
@@ -0,0 +1,34 @@
...
+
+Reference this table for details on the elements that make up a defined and secured GraphQL action:
+
+| Defined GraphQL action element | Description |
+|:----------|:------|
+| `limitRanges(namespace: String!)` |  Name of the action, which in this case is a query, followed by a string that specifies the Namespace of the queried resources. |
+| `: [LimitRange!]!` | Defines the type of object returned by the query. In this case, it's a list of LimitRanges resources. |
+| `@HasAccess(attributes:` | Defines the GraphQL directive that secures access to the resource. |
+| `resource: "limitranges"` | Defines the type of secured Kubernetes resource, in this case all `limitranges` resources. |
+| `verb: "list"` | Defines the secured interaction type. It is related to action type, which in this case is "query". |
```
sjanota: Add some guidelines on what verbs should be used in which scenario. You may utilise what is already in security docs.

components/ui-api-layer/docs/security.md
```diff
@@ -0,0 +1,34 @@
...
+| `apiGroup: ""` | Defines the apiGroup to which the user must have access to get the result of this query. In this case it is empty because limitRanges is the resource built into Kubernetes, not some Custom Resource created by us. |
+| `apiVersion: "v1alpha1"` | Specifies the apiVersion of the query subject. |
+| `namespaceArg: "Namespace"` | Specifies the name of the argument or field in the parent object from which the resource Namespace is fetched. |
+| `isChildResolver: false` | (Optional) Specifies if the directive is set on a field in a type. Determines if the Namespace should be fetched from the parent object instead of the argumentsnot available in the child object. By default it is set to `false`. |
```
sjanota:
```suggestion
| `isChildResolver: false` | (Optional) Specifies if the directive is set on a field in a type. Determines if the Namespace should be fetched from the parent object instead of the arguments not available in the child object. By default it is set to `false`. |
```

kubadz: no longer relevant

components/ui-api-layer/docs/security.md
kubadz: done

components/ui-api-layer/docs/security.md
```diff
@@ -0,0 +1,36 @@
...
+| `@HasAccess(attributes:` | Defines the GraphQL directive that secures access to the resource. |
+| `resource: "limitranges"` | Defines the type of secured Kubernetes resource, in this case all `limitranges` resources. |
+| `verb: "list"` | Defines the secured interaction type. It is related to action type, which in this case is "query". Use the the table from [this paragraph](https://kyma-project.io/docs/master/components/security#details-graphql) to choose the right verb. |
+| `apiGroup: ""` | Defines the apiGroup to which the user must have access to get the result of this query. In this case it is empty because limitRanges is the resource built into Kubernetes, not some Custom Resource created by us. |
+| `apiVersion: "v1alpha1"` | Specifies the apiVersion of the query subject. |
+| `namespaceArg: "namespace"` | Specifies the name of the argument or field in the parent object from which the resource namespace is fetched. |
+| `isChildResolver: false` | Required to be "true" for fields nested in types which have to be secured. Determines if the namespace argument should be fetched from the parent object. By default it is set to `false`. |
```
tomekpapiernik:
```suggestion
| `isChildResolver: false` | Must be "true" for fields nested in types which have to be secured. Determines if the `namespace` argument should be fetched from the parent object. By default, it is set to `false`. |
```

components/ui-api-layer/docs/security.md
```diff
@@ -0,0 +1,36 @@
...
+
+If the directive is set on a field nested in a type, the value returned from a query returning that type depends on the type modifier set on the field. If the field is set as Non-Null, as described [here](https://graphql.github.io/learn/schema/#lists-and-non-null), the error will be returned in case the user has no rights to access the child resource. If the field is not set as Non-Null, only the child field will be returned as null.
```
tomekpapiernik: Is "type" the name of a modifier? 
If yes, format it as `code`

tomekpapiernik:
```suggestion
If the directive is set on a field nested in a type, the value returned from a query returning that type depends on the type modifier set on the field. If the field is set as [Non-Null](https://graphql.github.io/learn/schema/#lists-and-non-null), the error is returned for users who do have rights to access the child resource. If the field is not set as Non-Null, only the child field will be returned as null.
```

kubadz: Type is not the name of modifier. `Non-Null` is. I will format it.

kubadz: Suggestion applied.

PR #2619, opened 2019-02-06
E2e test restore backup

k15r:
**Description**

Introduce a starting point for e2e testing of the ark backup procedure. This PR includes only the go-part

**Related issue(s)**
#2271

Issue #2271, opened 2019-01-14
E2E Test setup for Backups on GCP


joek:
As a kyma operator I need to trust backups on GCP.

AC:
- Create Kyma Cluster
- Create resources on Kyma Cluster (e.g. functions)
- Backup Cluster
- Recreate Cluster
- Restore Data and check if the restored resources are working link expected.
- Guideline how to hookup own resources to the test

Issue #2639, opened 2019-02-07
Create a test runner for acceptance tests that replace helm test


aszecowka:
Currently, we use helm test for running acceptance tests. Such an approach has many drawbacks. Running integration tests on GKE takes in average 23 minutes. Locally it takes 18 minutes. There is no easy way to run tests selectively. There is no easy way to run only failed tests. There is a possibility to run tests concurrently, but we cannot limit the number of running test pods at the same time. 

Because Helm test has a very limited number of features, we should create a library for running acceptance tests. Such a library eventually should have the following features:
- execute all tests
- execute only a subset of tests. For example, I want to run only tests for Service Catalog. This can be especially useful for local development.
allow executing tests concurrently. Concurrency level (number of tests executed concurrently) should be configurable. Thanks to that IMO we can reduce test execution time twice or even threefold. It means that every developer can save 10-15 minutes on every PR that requires running integration tests.
- allow retrying only failed tests. Flaky tests are an everyday reality. Definitely, we should strive to improve tests and make them stable, but we should accept that they will exist. IMO it will be beneficial to retry only failed tests automatically on Prow integration jobs. Thanks to that, PR author does not have to retrigger tests manually, a new cluster does not have to be provisioned etc.
- allow running given test many times. This can be helpful when a developer adds a new test and he/she needs to validate if a test is stable or he wants to reproduce a problem that sometimes occurs on CI.
- there is a CLI that allows easily create test executions

**Acceptance Criteria**
- In the first iteration we should provide the same functionality as helm test provided:
  - executing all tests
  - providing logs from tests
- inform sig-production about a concept and current status
- in case of success and full transparency merge into master
- check all places where testing.sh is used (for example stability checker)
- CLI support stretch

Library probably should be placed in the kyma-incubator.

PK85:
Things to consider:
- define CRDs, decide how to register them
- helm chart, new module
- how to store logs, find a way how to manage pods termination (logs aspect)
- 

-

aszecowka:
Tasks: 
[ ] bootstrap project according to kyma project structure: https://github.com/kyma-project/community/blob/master/guidelines/internal-guidelines/repository-template/README.md
@aszecowka 
[ ] Define prowjobs
[ ] Define CRDs
[ ] Implement the simplest solution

PR #2646, opened 2019-02-07
E2e test script

k15r:
**Description**

- an initial version of a script that can be used to run end to end tests

**Related issue(s)**
https://github.com/kyma-project/kyma/issues/2271
**Related PR(s)**
#2619


k15r:
@sjanota imho the difference to other test is: 
normally you install a release and it stays in the cluster. The e2e-test currently use helm just out of pure convenience. The do not stay in the cluster as a component of kyma. Their release will be installed, the tests will be executed, the release will be removed. In the future helm will probably get replaced by a different mechanism.

sjanota:
* Our acceptance tests rarely test single component in isolation. Some of them are already e2e tests rather then integration tests. They are bound to a component by test entry point (not documented anywhere, it's my gut feeling). For example kubeless-integration tests test lambda creation. It involves at least kubeless, API controller and istio and verifies complete user scenario.
* They stay in cluster not because they are bound to come component, but because you can easily rerun them if needed without the need to have Kyma sources. Binding is a way of achieving that, not the goal itself. 
* New mechanism to run our tests is currently under development (#2639) and all acceptance tests (no matter integration or e2e) should use it.

sjanota:
FYI @PK85

aszecowka:
why do we have 2 files with tests: testing.sh and e2e-testing.sh?

venturasr:
@aszecowka `e2e-testing.sh` is the script we are using for executing only end to end test. As you suggested in the PR, https://github.com/kyma-project/test-infra/pull/583 . we have created a `kyma-gke-end-to-end-test.sh` to provisioning cluster on for the execution of end to end test.

joek:
> * Our acceptance tests rarely test single component in isolation. Some of them are already e2e tests rather then integration tests. They are bound to a component by test entry point (not documented anywhere, it's my gut feeling). For example kubeless-integration tests test lambda creation. It involves at least kubeless, API controller and istio and verifies complete user scenario.
> * They stay in cluster not because they are bound to come component, but because you can easily rerun them if needed without the need to have Kyma sources. Binding is a way of achieving that, not the goal itself.
> * New mechanism to run our tests is currently under development (#2639) and all acceptance tests (no matter integration or e2e) should use it.

Actually I would suggest, we are going with the solution how it is implemented right now. As soon as we have the new process in place we will migrate to the new standard. I don't know the technical implications but my experience with helm test is not the best. Especially if we have a look on the tests testing more then one component.

sjanota:
@joek @k15r @venturasr I'd like to understand what's so different about those tests that we need to bring whole new infrastructure to run them. Can you point me to already implemented e2e test?

venturasr:
@sjanota 
Current implementation of Backup E2E Test:

    https://github.com/kyma-project/test-infra/pull/584
    https://github.com/kyma-project/test-infra/pull/583
    https://github.com/kyma-project/kyma/pull/2646
    https://github.com/kyma-project/kyma/pull/2619

Test Modularisation plan:

    https://github.com/kyma-project/kyma/issues/2696

installation/scripts/e2e-testing.sh
```diff
@@ -0,0 +1,59 @@
...
+cleanupE2EErr=$?
+
+
+
+echo "----------------------------"
+echo "- E2E Testing Kyma..."
+echo "----------------------------"
+
+
+for testcase in ../../tests/end-to-end/*/deploy/chart/*
```
aszecowka: could you point me what will be tested here? I cannot find this files on the master branch

venturasr: Hi @aszecowka, the whole implementation of the e2e test is in this PR. 
https://github.com/kyma-project/kyma/pull/2619
However we are working on the modularization of the e2e test as it is explained in this issue, https://github.com/kyma-project/kyma/issues/2696

sjanota:
@venturasr The test implemented in #2619 is simply an acceptance test, there's nothing special about it. It operates on it's own resources, have it's own namespace and does not collide with other subsystems. It could be as well run as simple helm test attached to ark chart. There is no need to implement new way of running tests for this scenario.

venturasr:
@sjanota and @aszecowka thank you for your help guys

Issue #2243, opened 2019-01-11
Setup links validation of relative links from docs section against https://kyma-project.io/docs/master/


derberg:
**Description**

- add proper nightly links validation of relative links from `docs` directory against https://kyma-project.io/docs/master/
- make sure other links on https://kyma-project.io/docs/master/ are also tested

**Reasons**

Although we have already a good set of links governance jobs in kyma-project we are missing one last check. The relative links to documentation that renders in kyma-project.io and Content view in the Console are not tested. This brings risk that we have broken links in product documentation.

PR #2666, opened 2019-02-08
 Fix broken links in docs

magicmatatjahu:
**Description**

- Fix broken links in docs

**Related issue(s)**
See also #2243


derberg:
Force merging before of criticality of the issue and no need to block it as for docs there is not need to wait for integration tests

PR #2695, opened 2019-02-11
support minikube version in range

### LINKED ISSUE #1923: Compatibillity with minikube >= 0.28.2 not only = 0.28.2
**Description**

Would be nice to be able to use newer versions of minikube, not only 0.28.2.

**Reasons**

Makes life easier (easier installation, easier reusing minikube for other project, etc.)

k15r:
**Description**

Currently only one specific version of minikube is supported by the script. This PR introduces functionality to specify a range of versions.

Fixes #1923

### DIFF
--- a/installation/scripts/minikube.sh
+++ b/installation/scripts/minikube.sh
@@ -116,7 +116,7 @@ function checkIfMinikubeIsInitialized() {
 }
 
 function checkMinikubeVersion() {
-    local version=$(minikube version | awk '{print  $3}' | sed s/v// )
+    local version=$(minikube version | awk '{print  $3}' | grep -o '[0-9\.]\+' )
     local version_clean=$(echo $version | awk -F '.' '{print $1"."$2;}')
     local supported_version_min=$(echo ${MINIKUBE_VERSION} | awk -F '.' '{print $1"."--$2;}')
     local supported_version_max=$(echo ${MINIKUBE_VERSION} | awk -F '.' '{printf "%d.%d", $1, ++$2;}')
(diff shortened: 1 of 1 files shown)


sjanota:
#1923

sjanota:
@k15r thanks for picking it up. I've linked an issue to that problem. There's one more requirement on the implementation: do not prevent user from using the script on version mismatch, just print warning. Could you delete exiting on mismatch also?

k15r:
@sjanota done

sjanota:
I'm afraid negative scenario is not working. How I tested it:
```
$ ./installation/scripts/minikube.sh

################################################################################
# Minikube setup with driver hyperkit
################################################################################

Minikube is already initialized
Do you want to remove previous minikube cluster [y/N]: ^C
$ minikube version
minikube version: v0.32.0
```

As far as I understand it should show warning before checking if minikube is installed.

installation/scripts/minikube.sh
```diff
@@ -6,7 +6,9 @@ CURRENT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
 RESOURCES_DIR="${CURRENT_DIR}/../resources"
 
 MINIKUBE_DOMAIN=""
-MINIKUBE_VERSION=0.33.0
+# Supported Minikube Versions: MINIKUBE_VERSION_MIN (inclusive) up to MINIKUBE_VERSION_MAX (exclusive)
+MINIKUBE_VERSION_MIN=0.33.0
+MINIKUBE_VERSION_MAX=0.34.0
 KUBERNETES_VERSION=1.11.5
```
piotrmsc: Please do not introduce max version of k8s, we do not have an infra to test that. Instead, you should print a warning if you're using a different version than the one in our resources. Here you can limit the check to minor and skip patch.

k15r: done

PR #2736, opened 2019-02-13
Dex: Change default expire settings

Demonsthere:
**Description**

- IDTokens are set to 8h
- SigningKeys are set to 1 month (assume 30 days -> 720h)

**Related issue(s)**


Demonsthere:
How to test:
1 - Change `idTokens` to a few minutes
2 - Install kyma from fork
3 - Login to console, and download kubeconfig
4 - Use the kubeconfig to call some commands (`kubectl get all --all-namespaces`)
5 - Wait for it
6 - After a few minutes try the command again

The expected outcome is an kubectl based error: `error: the server doesn't have a resource type "all"`, which is a bit misleading, as the resources exist, but we don't have access to it. 
However, calling the server with increased verbosity gives the expected result:
`Response Status: 401 Unauthorized in 1 milliseconds`

Issue #2525, opened 2019-01-31
Implement token authentication in configurations generator


piotrmsc:
**Description**

Currently **configurations generator** relies on istio **authentication policy** for validating tokens. See here (https://github.com/kyma-project/kyma/tree/master/resources/core/charts/configurations-generator)
This would be enough, however, due to the problem with JWKS handling [reported here](https://github.com/istio/istio/issues/10547) after 15 days, it will not be possible to download new kubeconfig from the console without restarting config generator pod.

We should implement token authentication in the application as we did in **apiserver-proxy** and **ui-api-layer**. Example [here](https://github.com/kyma-project/kyma/blob/master/components/apiserver-proxy/internal/authn/oidc.go)

**Reasons**

Token authentication functionality will remain, however, it will not be bugged anymore. Additionally, we will be able finally to change signing keys expiration time in dex, which was increased due to the issue with istio policy

**Attachments**

- [x] Token authentication implemented in configurations generator @Tomasz-Smelcerz-SAP -> #2743
- [x] Authentication policy deleted @Tomasz-Smelcerz-SAP -> #2751
- [ ] Token validator library used from kyma/commons library - if code is worth sharing then  create a common library and use it in the app code

PR #2751, opened 2019-02-14
Configurations generator remove istio policy

Tomasz-Smelcerz-SAP:
**Description**

- Remove Istio Authentication Policy from configurations-generator
- Bump configurations-generator version to the one with support for JWT token authentication
- Use configurations-generator built-in JWT token authentication mechanism

See also #2525


resources/dex/templates/dex-config-map.yaml
```diff
@@ -12,7 +12,7 @@ data:
     issuer: https://dex.{{ .Values.global.domainName }}
     # extending it to 15days because apparently Istio does not handle ID token validation properly
     expiry:
-      signingKeys: "360h"
+      signingKeys: "15m"
     storage:
```
piotrmsc: this is already covered in https://github.com/kyma-project/kyma/pull/2736
where btw keys should expire after 1 month :)

Tomasz-Smelcerz-SAP: Yes, this is for testing purposes only. Will be removed before merge.

Tomasz-Smelcerz-SAP:
Note for reviewer: The following procedure can be used to reproduce the problem on master and/or verify that the problem is fixed on the P/R branch. 

How to reproduce the problem: (Use Kyma installation that DOES NOT contain fixes from this PR):

- Set up dex expiration key time to about 15 mins (resources/dex/templates/dex-config-map.yaml)
- Launch Kyma locally on Minikube
- Once installed, log in to console.kyma.local
- Verify you can download the configuration. Stay logged in!
- Wait until dex rotates the keys (take a look at dex logs)
- Verify you can still download the configuration (without logging out!) - console app still has an old token, should work.
- Logout and log-in again. Console app has a new token, but the Istio authentication policy can't verify it because it doesn't have new dex keys (it keeps old one in cache)
- Verify you can't download configuration anymore.

Issue #2733, opened 2019-02-13
Extend installation CR status with errors which occurred during release install/upgrade


piotrmsc:
**Description**

Installation status should contain information about all errors that occurred during the installation of helm releases. So for example, if installation of istio and core fails, status will contain error messages from those two releases.

**Reasons**
For developers, it will be better to check CR status instead of scrolling between installer logs via `kubectl logs `

**Attachments**

PR #2726, opened 2019-02-13
Save errors to Installation error log

sjanota:
**Description**

Changes proposed in this pull request:

- new field `errorLog` is introduced in Installation status 
- installer saves installation errors into new `errorLog` field

**Related issue(s)**

#2756 
#2733 

**How to test**
1. Introduce an error (e.g. syntax) in one of the charts
2. Run Kyma from this branch
3. `kubectl get installation -o yaml`
4. After some time Installation should contain non-empty `status.errorLog` field

PR #2756, opened 2019-02-15
Add errorLog to installer CRD status

sjanota:
**Description**

* new field `errorLog` is introduced in Installation status
* installer saves installation errors into new `errorLog` field

**Related issue(s)**

#2733
#2726


jakkab:
It seems that the same error is logged multiple times if retry fails. Consider checking if a specific error has already been reported before pushing it to the log. You can even go a step further add another int field - something like "occurrences" - and increment it in case of recurrent errors.

sjanota:
@jakkab @Demonsthere I've made errorLog a set. Could you verify again?

PR #2785, opened 2019-02-18
Backup setup

venturasr:


tools/backup-setup/README.md
```diff
@@ -0,0 +1,39 @@
...
+  # null/unset. 
+  includeClusterResources: null
+  # Individual objects must match this label selector to be included in the backup.
+  labelSelector:
+    matchLabels:
+```
+
+### Apply backup.yaml
+
+`kondemandatx apply -f tools/backup-setup/config/backup.yaml`
```
pbochynski: Nice alias, but I believe most people will use regular `kubectl` ;-)

venturasr: Yes I have updated

venturasr: :-)

tools/backup-setup/config/backup.yaml
```diff
@@ -0,0 +1,19 @@
+apiVersion: ark.heptio.com/v1
+kind: Backup
+metadata:
+  name: kyma-backup
+  namespace: heptio-ark
+spec:
+  includedNamespaces:
+  - production
```
joek: Can you use a wildcard (*) here as this file should backup all namespaces which are not system namespaces.

tools/backup-setup/config/backup.yaml
```diff
@@ -0,0 +1,19 @@
...
+  excludedNamespaces:
+  - default
```
joek: Here are some namespaces missing like istio-system, kyma-integration. Just start a kyma instance and check which namespaces are created and not visible for the user.

tools/backup-setup/config/backup.yaml
```diff
@@ -0,0 +1,19 @@
+apiVersion: ark.heptio.com/v1
+kind: Backup
+metadata:
+  name: kyma-backup
```
joek: This should be named customer-namespaces-backup or something like this. Please create a second file for system namespaces. Some teams e.g. gophers need to backup etcd in kyma-system namespace.

joek:
> Please create backup documentation in docs folder. This one won't be displayed in kyma.

@pbochynski I do not agree on that, as this is developer documentation about how to add things to the backup configuration. We will probably add a "end-user-facing" documentation later.

pbochynski:
> @pbochynski I do not agree on that, as this is developer documentation about how to add things to the backup configuration. We will probably add a "end-user-facing" documentation later.

Ok. Then I change my comment to "please add user documentation". I would like to see the documentation with following content:
-  how to backup my cluster, 
- how to restore my backup
- how to customize backup scope (namespaces, resources)
- how to configure storage for my backup
- how to manage backup retention

I think 1st iteration should cover first three bullet points. Next 2 can come later.

joek:
> @joek This should go to docs as this is documentation and guidelines for Kyma user (developer) on how to configure backup. At least this is how it looks like from the readme details. Problem with `later` approach is that it means never ;( we have Ark as a component but still zero official documentation about backups

@derberg I would prefer to document the parts PB suggested as this is the part a user of kyma should do. The parts currently documented are developer documents like the components readme.

tools/backup-setup/config/backup.yaml
venturasr: Yes we can, but actually the official documentation says:
```
# Array of namespaces to include in the backup. If unspecified, all namespaces are included.
# Optional. 
```

derberg:
@joek but you just agreed to @pbochynski recommendation to fulfill first 3 bullet points in first iteration. So one of those is `* how to customize backup scope (namespaces, resources)` which is partially covered in the readme right? and remember `end-user-facing` docs are facing developers ;)

joek:
@derberg That is correct but the process and purpose of this specific file mentioned here is different. We will readjust all the documentation anyway as it turns out some of the information PB likes to have is already there but not at the right place.

Issue #2563, opened 2019-02-04
Enable xip integration as default behaviour


piotrmsc:
**Description**
For cluster environments, we should enable our xip.io integration as a default mechanism in kyma.
However, it should work in the following manner:

- if the domain and TLS certs are provided do not modify anything
- if TLS certs and domain are **not** provided then generate new TLS certs and deploy kyma with xip domain
- if the domain is provided but TLS certs not then generate self-signed TLS certs for that domain
- if cert is provided but the domain is not, then return an error with message indicating incomplete setup
- minikube setup remains the same - no changes here

By enabling xip as default behavior we can merge **xip-patch** with **istio-kyma-patch**. 

Because the application connector is using nginx ingerss controller, we need to create a job that will do the same thing for nginx after installation of app gateway - create a domain and TLS cert but under the same conditions as mentioned above.

**Reasons**

It will simplify kyma installation on the cluster envs. It will be easier to try out kyma out of the box.

**Attachments**

Acceptance criteria:

- [ ] xip handling enabled by default, working under conditions mentioned in task description @jakkab 
- [ ] the same job is created for application gateway as a post install job - the copy paste is acceptable as this is **temporary** untill istio 1.1 will be released and nginx ingress controller will be deleted in kyma (@Tomasz-Smelcerz-SAP )
- [ ] Documentation is updated, DNS setup sections are refactored, mentioning xip is default but you can set it up and refer to existing documents
- [ ] Tests are passing on a cluster 
- [ ] minikube setup remains the same - no changes here
- [ ] @kfurgol notify breaking pixels about this change :)

Tomasz-Smelcerz-SAP:
Application connector can't be easily handled by just a job - bigger refactoring is necessary.
The problem is: The domain and certs are input parameters for application connector chart. These values are then used as arguments for processes running in pods created with Deployments/StatefulSets.
In existing setup it's not a problem because IP Address, Domain and certs are prepared before application connector chart is installed.
In xip scenario the value for domain and corresponding certs depends on IP Address. But IP Address is known only after a LoadBalancer Service is created, which is a part of application connector chart. It means we have a chicken-or-egg problem.
To break this dependency chain, we have to extract the Service (LoadBalancer) from application connector chart and install it as a separate component earlier in the installation pipeline.
Then we have access to IP Address for nginx ingress, we can configure proper Domain and Certs for the remaining application connector chart.

PR #2793, opened 2019-02-19
Enable xip.io integration by default

jakkab:
**Description**

- Adjust xip-patch.sh as per AC #2563 

**Related issue(s)**
Issue: #2563

### DIFF
--- a/components/xip-patch/xip-patch.sh
+++ b/components/xip-patch/xip-patch.sh
@@ -42,16 +42,16 @@ function generateXipDomain() {
         exit 1
     fi
 
-    XIP_DOMAIN="${EXTERNAL_PUBLIC_IP}.xip.io"
+    PUBLIC_DOMAIN="${EXTERNAL_PUBLIC_IP}.xip.io"
 
-    XIP_DOMAIN_YAML=$(cat << EOF
+    DOMAIN_YAML=$(cat << EOF
 ---
 data:
-  global.domainName: "${XIP_DOMAIN}"
+  global.domainName: "${PUBLIC_DOMAIN}"
 EOF
 )
 
-    kubectl patch configmap installation-config-overrides --patch "${XIP_DOMAIN_YAML}" -n kyma-installer
+    kubectl patch configmap installation-config-overrides --patch "${DOMAIN_YAML}" -n kyma-installer
 
 }
 
--- a/components/xip-patch/xip-patch.sh
+++ b/components/xip-patch/xip-patch.sh
@@ -2,70 +2,89 @@
 
 set -o errexit
 
-if [ -z "${INGRESSGATEWAY_SERVICE_NAME}" ]; then
-  INGRESSGATEWAY_SERVICE_NAME=istio-ingressgateway
-fi
+# # # # # # # # # # # # # # # # #
+# VARs coming from environment: #
+#                               #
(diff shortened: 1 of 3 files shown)


components/xip-patch/xip-patch.sh
```diff
@@ -2,77 +2,112 @@
...
+# EXTERNAL_PUBLIC_IP            #
+# INGRESSGATEWAY_SERVICE_NAME   #
+# PUBLIC_DOMAIN                 #
+# TLS_CERT                      #
+# TLS_KEY                       #
+# # # # # # # # # # # # # # # # #
 
-  while [ ${SECONDS} -lt ${END_TIME} ];do
-      echo "Trying to get loadbalancer IP address"
+function generateXipDomain() {
```
Tomasz-Smelcerz-SAP: Function must take explicit input parameters instead of "global" env names, it has to be reusable.
Explanation: We'll need xip "patch" functionality for at least two cases now: istio-ingress and application-connector-ingress. This function should handle both cases.

components/xip-patch/xip-patch.sh
```diff
@@ -2,77 +2,112 @@
...
-fi
+        SECONDS=0
+        END_TIME=$((SECONDS+60))
 
-XIP_PATCH_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+        while [ ${SECONDS} -lt ${END_TIME} ];do
+            echo "Trying to get loadbalancer IP address"
 
-PUBLIC_DOMAIN="${EXTERNAL_PUBLIC_IP}.xip.io"
+            EXTERNAL_PUBLIC_IP=$(kubectl get service -n istio-system ${INGRESSGATEWAY_SERVICE_NAME} -o jsonpath="{.status.loadBalancer.ingress[0].ip}")
```
Tomasz-Smelcerz-SAP: namespace must not be hardcoded.

components/xip-patch/xip-patch.sh
```diff
@@ -2,77 +2,112 @@
...
 
-CERT_PATH="${XIP_PATCH_DIR}/cert.pem"
-KEY_PATH="${XIP_PATCH_DIR}/key.pem"
+            if [ -n "${EXTERNAL_PUBLIC_IP}" ]; then
+                echo "External public IP address is ${EXTERNAL_PUBLIC_IP}"
```
Tomasz-Smelcerz-SAP: Make it more generic, like: "Load Balancer IP Address is ...."

components/xip-patch/xip-patch.sh
```diff
@@ -2,77 +2,112 @@
...
   global.domainName: "${PUBLIC_DOMAIN}"
 EOF
 )
 
-TLS_CERT_YAML=$(cat << EOF
+    kubectl patch configmap installation-config-overrides --patch "${DOMAIN_YAML}" -n kyma-installer
+
+}
+
+function generateCerts() {
```
Tomasz-Smelcerz-SAP: Make this function generic, too. Extract generic functions to a separate script that can be source(d).
There will be another script containing logic for application-connector xip setup.
If we manage to somehow reuse THIS script for application-connector (unlikely) we can always merge everything into a single script.

PR #2833, opened 2019-02-21
Conditional application management

parostatkiem:
**Description**
#### To see the actual effect, switch your Console repo to the branch from [this PR](https://github.com/kyma-project/console/pull/593)

- add `applicationManagementDisabled` bool to ApplicationConnector's `values.yaml`

**Related issue(s)**


resources/cluster-essentials/templates/microfrontend.crd.yaml
```diff
@@ -1,34 +1,34 @@
...
+  name: 'microfrontends.ui.kyma-project.io'
 spec:
   group: ui.kyma-project.io
   version: v1alpha1
   scope: Namespaced
   names:
-    plural: "microfrontends"
-    singular: "microfrontend"
+    plural: 'microfrontends'
+    singular: 'microfrontend'
```
dariadomagala: Applications is a cluster MicroFrontend, not a regular MicroFrontend, so wrong file. Additionally, we probably should add docs for that if we change the resource definition. 
But before that, please as Philipp if that should be an 'official' feature, since for now it's only relevant for Applications. If not, then maybe we don't need to edit the CRDs and add docs at all.

resources/application-connector/templates/applications-microfrontend.yaml
```diff
@@ -15,9 +15,13 @@ spec:
...
     - label: Application Details
       navigationPath: apps/details
       viewUrl: '/home/settings/apps'
     - label: Application Details
       navigationPath: apps/details/:name
-      viewUrl: '/home/settings/apps/:name'
\ No newline at end of file
+      viewUrl: '/home/settings/apps/:name'
+      settings:
+        readOnly: {{ js .Values.global.applicationManagementDisabled }}
```
kwiatekus: AFAIK, context is inherited by child nodes, so the setting on list view level should be sufficient

resources/application-connector/values.yaml
```diff
@@ -1,4 +1,5 @@
 global:
+  applicationManagementDisabled: false
```
akgalwas: I would suggest to rename this variable to applicationManagementEnabled with default true. There are other flags for faros (e.g. flag for disabling connector-service(please see please see https://github.com/kyma-project/kyma/blob/master/resources/application-connector/requirements.yaml)) that use positive logic. It would make more sense to have it consistent as it makes reasoning easier. Note: you could use the similar form of the conditions as in requirements.yaml (applicationmanagement.enabled).

kwiatekus: done

PR #2834, opened 2019-02-21
Update images for kubeless runtimes

rakesh-garimella:
**Description**
Update kubeless nodejs images


resources/core/values.yaml
```diff
@@ -72,6 +72,19 @@ global:
   ui_api_layer_acceptance_tests:
     dir: develop/
     version: "0b5f866d"
+  kubeless_images:
+    runtimes:
```
sayanh: "phase": "runtime"(singular) just like "phase": "installation"

PR #2865, opened 2019-02-25
Fix knative installation option in PowerShell script

JoergBastian:
**Description**
The PowerShell script for kyma installation on Windows fails when the `-knative` switch is set.
I adjusted the installation script based on the procedure in the ,sh script.


radufa:
Start the installation from Kyma installation directory in an "Admin PowerShell" window:
```
cmd/run.ps1 -knative
```

Issue #2315, opened 2019-01-16
Create a procedure for rotation of Root CA certificate in central Connector Service


lszymik:
**Description**

The Root CA certificate needs to be rotated from time to time.

**Reasons**

It is due to security.

**Acceptance criteria**

- Procedure for rotation of Root CA created

janmedrek:
To ensure that both certificates can validate incoming requests we need to disable NGINX SSL chain autocompletion, PR #2875 handles that.

PR #2875, opened 2019-02-26
Disable NGINX ssl chain autocompletion

janmedrek:
**Description**

- Nginx SSL chain autocompletion disabled

**Related issue(s)**
#2315


resources/application-connector/charts/nginx-ingress/values.yaml
```diff
@@ -66,7 +66,9 @@ controller:
   ## E.g. to specify the default SSL certificate you can use
   ## extraArgs:
   ##   default-ssl-certificate: "<namespace>/<secret_name>"
-  extraArgs: {}
+  extraArgs: {
+    enable-ssl-chain-completion: "false"
+  }
```
Szymongib: I think we do not need brackets here

Issue #2603, opened 2019-02-05
Monitoring for nats streaming


abbi-gaurav:
The NFRs for event bus components have been discussed internally. They will be used as a  reference.

Refer: 
https://github.com/kyma-project/kyma/issues/1879
https://github.com/kyma-project/kyma/issues/1817
https://github.com/kyma-project/kyma/issues/1818
https://github.com/kyma-project/kyma/issues/376

marcobebway:
Initial pull request to enable monitoring for nats-streaming on grafan UI is here: https://github.com/kyma-project/kyma/pull/3077.

marcobebway:
Initial pull request to enable monitoring for nats-streaming on grafana is here: https://github.com/kyma-project/kyma/pull/3077.

PR #1739, opened 2018-11-21
Added metrics.k8s.io to API groups of the kyma-admin role

tomekpapiernik:
**Description**

Changes proposed in this pull request:

- Added `metrics.k8s.io` to API groups of the `kyma-admin` role (allows admin to run kubectl top)
- Updated `minikube.sh` to enable heapster addon

**Related issue(s)**
See #1499

PR #2893, opened 2019-02-27
Use metrics-server instead of retired heapster

sslavic:
**Description**

While working on #2603 I've noticed that https://github.com/kyma-project/kyma/pull/1739 got broken, `kubectl top` would no longer work because enabling of heapster minikube addon was failing. It seems at some version of minikube, enabling of addons was made possible only after minikube has already been started. Trying to enable addon before minikube has been started - with 0.33.x minikube it would just log `minikube is not currently running so the service cannot be accessed` error, while with 0.34.x minikube configuration initialization would return non-zero exit code and fail installation script. Also, heapster project is [retired](https://github.com/kubernetes-retired/heapster#heapster) and part of its functionality `kubectl top` depends on is replaced with metrics-server.

- Adjust Kyma minikube installation to enable minikube addons after minikube has been started
- Replace use of retired heapster with metrics-server minikube addon

### DIFF
--- a/installation/scripts/minikube.ps1
+++ b/installation/scripts/minikube.ps1
@@ -31,8 +31,10 @@ function CheckIfMinikubeIsInitialized() {
 function InitializeMinikubeConfig () {
     $cmd = "minikube config unset ingress"
     Invoke-Expression -Command $cmd
+}
 
-    $cmd = "minikube addons enable heapster"
+function ConfigureMinikubeAddons () {
+    $cmd = "minikube addons enable metrics-server"
     Invoke-Expression -Command $cmd
 }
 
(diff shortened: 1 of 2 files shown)


PavelPavlov46:
My notebook:
Summary: Change works on win10 with minikube 0.34

Details:
**wmic os get version**
> Version
> 10.0.17134

**minikube version**
> minikube version: v0.34.1

Test:
**git clone https://github.com/sslavic/kyma.git
cd kyma
git checkout -b sslavic-metrics-server master**
> Switched to a new branch 'sslavic-metrics-server'
**git pull https://github.com/sslavic/kyma.git metrics-server**
> Updating d4937a42..9d7d8050

**.\installation\cmd\run.ps1**

**kubectl top node**
> NAME       CPU(cores)   CPU%      MEMORY(bytes)   MEMORY%
> minikube   1083m        27%       3355Mi          42%

Issue #2517, opened 2019-01-30
Application Operator should handle upgrade scenario


akgalwas:
**Description**

 In order to support upgrade process the Application Operator needs to upgrade installed Application releases .

**Reasons**
Let's consider the following steps:
1. User installed Kyma.
2. User created and configured new Application.
3. There was a bug found in Application Proxy and some updates to Event Service. New release has been prepared containing the fix (there is a new image version for Application Proxy, Event Service and Application Operator)
4. User upgraded Kyma to the new release. 

In the described scenario Application Operator will be upgraded but all the installed instances of Event Service and Application Proxy will be left intact (so that the recent fixes won't be installed). 
Application Operator is responsible for managing lifecycle of Event Service and Application Proxy so that it should handle described case.

PR #2895, opened 2019-02-27
Add upgrade functionality to the operator

janmedrek:
**Description**

- Application Operator handling upgrade of existing deployments

**Related issue(s)**
#2517


components/application-operator/pkg/kymahelm/application/releasemanager.go
```diff
@@ -48,6 +56,37 @@ func (r *releaseManager) InstallChart(application *v1alpha1.Application) (hapi_4
...
+	return nil
+}
+
+func (r *releaseManager) upgradeChart(application *v1alpha1.Application) (hapi_4.Status_Code, string, error) {
+	overrides, err := r.prepareOverrides(application)
+	if err != nil {
+		return hapi_4.Status_FAILED, "", errors.Wrapf(err, "Error parsing overrides for %s Application", application.Name)
+	}
+
+	upgradeResponse, err := r.helmClient.UpdateReleaseFromChart(applicationChartDirectory, application.Name, overrides)
```
Szymongib: It would be good to log that we started the updated operation for a specific release and also log if the update fails.

components/application-operator/pkg/kymahelm/application/releasemanager.go
```diff
@@ -48,6 +56,37 @@ func (r *releaseManager) InstallChart(application *v1alpha1.Application) (hapi_4
...
+	appList, err := r.appClient.List(v1.ListOptions{})
+	if err != nil {
+		return errors.Wrapf(err, "Error fetching application list")
+	}
+
+	for _, app := range appList.Items {
+		status, description, err := r.upgradeChart(&app)
+		if err != nil {
+			setCurrentStatus(&app, status.String(), description)
+			r.appClient.Update(&app)
```
Szymongib: I think it would be good to use retry on conflict mechanism and log some warning if we fail.

PR #2901, opened 2019-02-27
Introduce naming of the gophers component

polskikiel:
**Description**

- Moved watch pods and stability-checker to prow
- Renamed `binding-usage-controller` to `service-binding-usage-controller`
- Renamed and moved etcd-backup and etcd-tls-setup from `/tools` to `/components`

**Related issue(s)**


klaudiagrz:
Please check if there will be any broken links in the documentation as a result of this change.

components/service-binding-usage-controller/README.md
```diff
@@ -1,8 +1,8 @@
-# Binding Usage Controller
```
jasiu001: What about convention of component's name  in `resources/*/values.yaml` files. For example for this component version could be found in `resources/service-catalog-addons/values.yaml` under the key `binding_usage_controller`. Should the key change?

polskikiel: discussed, we will change the key and chart name.

resources/service-catalog/charts/etcd-stateful/templates/05-backup-job.yaml
```diff
@@ -23,7 +23,7 @@ spec:
           serviceAccountName: {{ template "etcd-fullname" . }}-backup
           containers:
           - name: etcd-backup
-            image: "{{ .Values.global.containerRegistry.path }}/{{ .Values.global.etcd_backup.dir }}etcd-backup:{{ .Values.global.etcd_backup.version }}"
+            image: "{{ .Values.global.containerRegistry.path }}/{{ .Values.global.etcd_backup_job.dir }}etcd-backup:{{ .Values.global.etcd_backup.version }}"
```
jasiu001: missed version part, should be `.Values.global.etcd_backup_job.version` is `.Values.global.etcd_backup.version`

polskikiel: good catch, fixed

polskikiel:
I would not change the images names because the task is not defining this and we use those names only internally in our charts. Those names also didn't change that much. 

I have also changed the name `binding-usage-controller` to `service-binding-usage-controller` in the following files: 
- resources/service-catalog-addons/charts/service-binding-usage-controller/Chart.yaml
- resources/service-catalog-addons/requirements.yaml

This files I won't change: 
- resources/service-catalog-addons/charts/service-binding-usage-controller/templates/function-usage-kind.yaml - `buc` name is used here only in the comment
- resources/service-catalog-addons/charts/service-binding-usage-controller/templates/deployment.yaml - `buc` name is used here as a image name.

Issue #2705, opened 2019-02-12
UI tests randomly failing due to net::ERR_NETWORK_CHANGED 


kwiatekus:
**Description**

Kyma ui tests are failing from time to time  as a result of `net::ERR_NETWORK_CHANGED` error happening in the chromium headless browser. 

This is happening randomly for both : kyma-integration and kyma-gke-integration jobs (presubmit and postsubmit)

Here is an exemplary scenario:

- The test script navigates to `Applications` node in the console ui and uses the ui components to create new Application. 
- A request is sent (POST) from the browser to the kyma backend service to create a resource, but the request fails with `net::ERR_NETWORK_CHANGED`. 
- As a result, new *application* is not created and the consecutive assertion fails 

https://storage.googleapis.com/kyma-prow-logs/logs/post-master-kyma-integration/1094873367795732480/build-log.txt
[build-log.txt](https://github.com/kyma-project/kyma/files/2854660/build-log.txt)

OR 

-  js bundles cannot be downloaded by the browser due to `net::ERR_NETWORK_CHANGED` making the application not testeable

https://storage.googleapis.com/kyma-prow-logs/logs/post-master-kyma-gke-integration/1094873367799926784/build-log.txt 
[build-log2.txt](https://github.com/kyma-project/kyma/files/2854709/build-log2.txt)

We don't know why `net::ERR_NETWORK_CHANGED` error happens in the first place.
It's most likely resulting from networking issues in the kyma cluster under test

**Expected result**

`net::ERR_NETWORK_CHANGED` is eliminated

**Steps to reproduce**

Its random, and there fore hard to isolate and reproduce.:

kwiatekus:
Creating DNS record for gke cluster -> https://github.com/kyma-project/test-infra/blob/master/prow/scripts/cluster-integration/helpers/create-dns-record.sh

sjanota:
I've found out that Chrome is throwing this error when IP of the source machine (test in this case) has changed. To check if this is the case we need to verify if containers IP gets changed frequently.

kwiatekus:
This sounds similar:
https://stackoverflow.com/questions/44854296/neterr-network-changed-on-chrome-on-https-socket-connection

kwiatekus:
We could disable 'hostNetwork' configuration and assure hostAliases i.e : 
https://kubernetes.io/docs/concepts/services-networking/add-entries-to-pod-etc-hosts-with-host-aliases/

PR #2909, opened 2019-02-28
Replace hostNetwork setting with hostAliases for ui-acceptance test minikube executions

kwiatekus:
**Description**

-  Applied `hostAliases` instead of  `hostNetwork` configuration for ui test pod when executed on  minikube

**Related issue(s)**
https://github.com/kyma-project/kyma/issues/2705


installation/scripts/installer.sh
```diff
@@ -97,6 +97,24 @@ if [ $KNATIVE ]; then
     kubectl -n kyma-installer patch configmap installation-config-overrides -p '{"data": {"global.knative": "true", "global.kymaEventBus": "false", "global.natsStreaming.clusterID": "knative-nats-streaming"}}'
 fi
 
+if [ $LOCAL ]; then
```
strekm: please move override to the `installer-config-local.yaml.tpl` and sed replacing its value with `${MINIKUBE_IP}`

PR #2912, opened 2019-02-28
Define monitoring configuration alerting defaults

sslavic:
**Description**

Monitoring component and its dependency prometheus operator are disabled by default for Kyma local installation.
While working on https://github.com/kyma-project/kyma/issues/2603 I've observed that enabling monitoring component for local Kyma installation is overly complicated, since working defaults for multiple configuration properties are not defined in the installation files. This PR makes it possible to enable monitoring by simply uncommenting it and its prometheus operator dependency in `installer-cr.yaml.tpl` and then running the installation script.

- Define local monitoring configuration defaults

### DIFF
--- a/resources/monitoring/charts/alertmanager/values.yaml
+++ b/resources/monitoring/charts/alertmanager/values.yaml
@@ -169,3 +169,14 @@ webhookConfig:
     bearerToken: ''
     bearerTokenFile: ''
     proxyUrl: ""
+
+# disable slack and victorops alerting by default
+global:
+  alertTools:
+    credentials:
+      slack:
+        apiurl: ""
+        channel: ""
+      victorOps:
+        routingkey: ""
+        apikey: ""


crabtree:
Shouldn't those values be defined in the charts as defaults?

@mszostok what do you think?

sslavic:
@crabtree I understood that these installer-config-* files play the role of no longer existing top level Kyma chart's `values.yaml` and configure defaults for these "global" configuration properties.

mszostok:
We discussed that with @crabtree f2f and he is right.

Directly in chart you should specify the default values for cluster installation if such are required but also you should specify values that are used in template. 

So in `/resources/monitoring/charts/alertmanager/values.yaml`, u should have entry like this one:
```
global:
  alertTools:
    credentials:
      slack:
        apiurl: ""
        channel: ""
      victorOps:
        routingkey: ""
        apikey: ""
```

cause they are the defaults, and the ConfigMap should be used for the overrides as the name suggest.


Right now installing your chart is broken if u will not provide values from outside. This shouldn't happen. When I want to install ur chart via `helm install ..` then it should work too (e.g. the slack by default should be disabled)

sslavic:
@mszostok updated as suggested. Maybe idea was to force one to configure the chart when installing in a cluster by not to providing defaults for alerting; now monitoring can be installed without alerts enabled. @sayanh can you please verify is this now changed behaviour desired?

mszostok:
> "now monitoring can be installed without alerts enabled"
yes, and this behavior is not the correct one then we should use the proper mechanism [required functions](https://github.com/helm/helm/blob/master/docs/charts_tips_and_tricks.md#using-the-required-function) and not using hack like this one. Thanks to that API will be more transparent to the user

Issue #1989, opened 2018-12-11
kubectl exec results in an error with console using kubeconfig


piotrmsc:
**Description**
When you download kubeconfig from the console and use it in the kubectl with exec, kubectl return an error with 403. 

Affected components: apiserver-proxy, RBAC

**Expected result**

User can execute `kubectl exec` command

**Actual result**

Error returned:
`Use 'kubectl describe pod/test-dan-re-proxy-service-7f66f5745d-pdb2v -n kyma-integration' to see all of the containers in this pod.
Error from server:` 

Giving bigger permissions (I have used cluster-admin role) does not solve 403 issue.
I'm not sure if it comes when trying to upgrade connection to WS

**Steps to reproduce**

Fetch kubeconfig from the console and execute kubectl exec command

**Troubleshooting**

djumalin:
Hi, does anyone have any updates, estimations on this issue?

piotrmsc:
Hi! no fixed dates yet but we will dig into this soon!

tomekpapiernik:
K O P Y T K O

tomekpapiernik:
kopytko 2: the hoof strikes back

tomekpapiernik:
I love my PO <3

sjanota:
Here's what we've found so far:
1. kubectl uses SPDY for some commands (like exec or port-forward)
2. apiserver-proxy does not support SPDY
3. Envoy does not support SPDY 

The plan is to:
1. Implement SPDY proxy in apiserver-proxy
2. Expose apiserver-proxy directly, without istio on secured port. Old path (unsecured) must also work as console is using it to access some resources.
3. When xip.io integration is enabled we need to generate new key and cert and propagate it to both apiserver-proxy and configrations-generator.
4. configurations-generator must be aware of new domain where apiserver-proxy is exposed.

PR #2916, opened 2019-02-28
Add apiserver-proxy kubectl exec test

kfurgol:
**Description**

- Added check on whether `kubectl exec` is working after applying kubeconfig downloaded from console

**Related issue(s)**
#1989


tests/apiserver-proxy-tests/apiserver-proxy/internal/client.go
```diff
@@ -0,0 +1,18 @@
+package graphql
```
sjanota: Folder is called `internal`, file `client.go` and package `graphql`. Please make up your mind ;)

tests/apiserver-proxy-tests/apiserver-proxy/internal/config.go
```diff
@@ -0,0 +1,80 @@
...
+	"github.com/vrischmann/envconfig"
+)
+
+type User string
+
+const (
+	AdminUser    User = "admin"
+	ReadOnlyUser User = "read-only"
+	NoRightsUser User = "no-rights"
+	NoUser       User = "no-user"
```
sjanota: Only `AdminUser` is used, and is passed as unused argument. Remove those constants.

tests/apiserver-proxy-tests/apiserver-proxy/internal/config.go
```diff
@@ -0,0 +1,80 @@
...
+	ID          string
+	RedirectUri string
+}
+
+type userCredentials struct {
+	Username string
+	Password string
+}
+
+func LoadConfig(userContext User) (config, error) {
```
sjanota: userContext is not used

sjanota: You return unexported type from exported function. Caller won't be able to do anything meaningful with it. Make `config` public.

tests/apiserver-proxy-tests/apiserver-proxy/internal/config.go
```diff
@@ -0,0 +1,80 @@
...
+	AdminUser    User = "admin"
+	ReadOnlyUser User = "read-only"
+	NoRightsUser User = "no-rights"
+	NoUser       User = "no-user"
+)
+
+type envConfig struct {
+	Domain        string `envconfig:"default=kyma.local"`
+	AdminEmail    string
+	AdminPassword string
```
sjanota: I'd rather call it UserEmail and UserPassword. It doesn't have to necessarily be admin

tests/apiserver-proxy-tests/apiserver-proxy/internal/idtokenprovider.go
```diff
@@ -0,0 +1,166 @@
...
+		return nil, err
+	}
+	switch authorizeResp.StatusCode {
+	case http.StatusFound:
+	case http.StatusOK:
+	default:
+		return nil, fmt.Errorf("got unexpected response on authorize: %d - %s", authorizeResp.StatusCode, readRespBody(authorizeResp))
+	}
+
+	// /auth/local?req=qruhpy2cqjvv4hcrbuu44mf4v
```
sjanota: ?

tests/apiserver-proxy-tests/apiserver-proxy/internal/idtokenprovider.go
```diff
@@ -0,0 +1,166 @@
...
+		keyAndValue := strings.Split(param, "=")
+		result[keyAndValue[0]] = keyAndValue[1]
+	}
+
+	return result, nil
+}
+
+func readRespBody(resp *http.Response) string {
+	defer func() {
+		err := resp.Body.Close()
```
sjanota: You should always read the body and close it, not only in case of error. `readRespBody` is now called only if an error has occurred. According to docs:

> It is the caller's responsibility to
> close Body. The default HTTP client's Transport may 
> reuse HTTP/1.x "keep-alive" TCP connections if the Body 
> not read to completion and closed.

tests/apiserver-proxy-tests/before-commit.sh
```diff
@@ -0,0 +1,57 @@
...
+if [ ${goBuildResult} != 0 ]; then
+	echo -e "${RED}✗ go build${NC}\n$goBuildResult${NC}"
+	exit 1
+else echo -e "${GREEN}√ go build${NC}"
+fi
+
+##
+# GO VET
+##
+packagesToVet=($(go list ./... | grep -v /vendor/ | grep -v api-controller/pkg/clients ))
```
sjanota: You're not using api-controller/pkg/clients, no need to grep it out.

tests/apiserver-proxy-tests/apiserver-proxy/main.go
```diff
@@ -0,0 +1,21 @@
...
+	"fmt"
+
+	"github.com/kyma-project/kyma/tests/apiserver-proxy-tests/apiserver-proxy/internal"
+)
+
+func main() {
+
+	config, err := graphql.LoadConfig(graphql.AdminUser)
+	if err != nil {
+		fmt.Print(err)
```
sjanota: log.Fatal, otherwise exit status would be 0

tests/apiserver-proxy-tests/fetch-token/internal/config.go
```diff
@@ -0,0 +1,71 @@
...
+)
+
+type envConfig struct {
+	Domain       string `envconfig:"default=kyma.local"`
+	UserEmail    string
+	UserPassword string
+}
+
+type Config struct {
+	GraphQLEndpoint  string
```
kubadz: not needed

tests/apiserver-proxy-tests/fetch-token/internal/idtokenprovider.go
```diff
@@ -53,11 +53,14 @@ func (p *dexIdTokenProvider) implicitFlow() (map[string]string, error) {
 	if err != nil {
 		return nil, err
 	}
+	authorizeRespBody := readRespBody(authorizeResp)
+	defer closeRespBody(authorizeResp)
```
sjanota: Always close body right after checking the error. Not very critical in this case, but it is a good practice.

kubadz: @sjanota that was done to always read the body to completion before closing it - not only in the case of an error - as you suggested.

kubadz: You are right. We should defer closing the body before reading it.

PR #2917, opened 2019-02-28
Add helper script to xip-patch

Demonsthere:
**Description**

- Add helper script

**Related issue(s)**

Needed in #2563


components/xip-patch/certhelper.sh
```diff
@@ -0,0 +1,34 @@
...
+    fi
+done
+
+if [ "${discoverUnsetVar}" = true ] ; then
+    exit 1
+fi
+
+if [[ -z "${DOMAIN}" ]]; then
+	echo "---> DOMAIN not SET. Creating..."
+	INGRESS_IP=$(getLoadBalancerIP "${LB_SERVICE_NAME}" "${LB_SERVICE_NS}")
```
Tomasz-Smelcerz-SAP: This can result in an error. Handle the error.

sjanota: Would `set -e` be enough?

Demonsthere: It is, look at line 1 ;)

PR #2928, opened 2019-03-01
Introduce Asset Store integration tests

pkosiec:
**Description**

Changes proposed in this pull request:
- Introduce Asset Store integration tests
- Fix Asset Upload Service crash on bad request without form

**Related issue(s)**
#2656

PR #2932, opened 2019-03-01
Enable Asset Store integration tests

### LINKED ISSUE #2656: Provide basic integration/acceptance/e2e tests for Asset Store
**Description**

- have test that tests Asset and Bucket at the same time
- webhook services are out of scope for now
- test is runnable also as a part of the CI upgradability plan so we know that after upgrade asset store still works
- we basically 
  - create a Bucket, 
  - create an Asset for newly created bucket
  - asset refers to Kyma release or you can wait for https://github.com/kyma-project/kyma/issues/2653 and upload some test data there (advantage is that upload service will be also tested with this scenario)
  - the asset is uploaded to Minio and we make sure it is really there where the status points

Stretch goal:

- best case scenario would be if that would be tasted in Minio Gateway mode -> only if it doesn't exceed initial estimation (investigate what to do to accomplish this task)

**Reasons**

- we need basic e2e coverage that checks core functionalities happy paths

pkosiec:
**Description**

- Enable Asset Store integration tests
- Bump Asset Upload Service image
- Bump UI API Layer acceptance tests

Changes merged in PR: https://github.com/kyma-project/kyma/pull/2928

**Related issue(s)**
Resolves #2656


resources/assetstore/charts/assetstore-controller-manager/templates/statefulset.yaml
```diff
@@ -58,6 +64,8 @@ spec:
               secretKeyRef:
                 name: assetstore-minio
                 key: secretkey
+          - name: APP_LOADER_VERIFY_SSL
+            value: 'false'
```
michal-hudy: Configure it only for `kyma.local`

installation/scripts/installer.sh
```diff
@@ -91,9 +91,9 @@ if [ ${ADMIN_PASSWORD} ]; then
     COMBO_YAML=$(sed 's/global\.adminPassword: .*/global.adminPassword: '"${ADMIN_PASSWORD}"'/g' <<<"$COMBO_YAML")
 fi
 
-if [ $LOCAL ]; then
+if [ ${LOCAL} ]; then
     MINIKUBE_IP=$(minikube ip)
-    COMBO_YAML=$(sed 's/test\.acceptance\.ui\.minikubeIP: .*/test\.acceptance\.ui\.minikubeIP: '"${MINIKUBE_IP}"'/g' <<<"$COMBO_YAML")
+    COMBO_YAML=$(sed 's/minikubeIP: .*/minikubeIP: '"${MINIKUBE_IP}"'/g' <<<"$COMBO_YAML")
```
sjanota: I'd add a dot just in case to make sure it is replacing only `minikubeIP:` and not `thisisnotminikubeIP:`  
```suggestion
    COMBO_YAML=$(sed 's/\.minikubeIP: .*/.minikubeIP: '"${MINIKUBE_IP}"'/g' <<<"$COMBO_YAML")
```

PR #2941, opened 2019-03-03
Update GKE documentation with better examples

pbochynski:


derberg:
already solved with different PR

Issue #598, opened 2018-09-10
Extend documentation for lambda usage


joek:
Issue:

Juergen Albertsen [16:25]
Maybe not update but extend the documentation. It's quite sparse at the moment.

I have some Node JS knowledge an expected the params to be at `request.params`
Since it isn't, perhaps we should point it out 

Referred documentation: https://github.com/kyma-project/kyma/blob/master/docs/serverless/docs/035-programming-model.md

 

E.g. query params can be accessed as 
console.log(event.extensions.request.query.p);
for https://test-stage.playground.cluster.kyma.cx/?p\=123```

May be add more information about request object:
 { 'event-type': undefined,
'event-id': undefined,
'event-time': undefined,
'event-namespace': undefined,
data: {},
extensions:
{ request:
IncomingMessage {
_readableState: [Object],
readable: true,
domain: null,
_events: {},
_eventsCount: 0,
_maxListeners: undefined,
socket: [Object],
connection: [Object],
httpVersionMajor: 1,
httpVersionMinor: 1,
httpVersion: '1.1',
complete: false,
headers: [Object],
rawHeaders: [Object],
trailers: {},
rawTrailers: [],
upgrade: false,
url: '/',
method: 'GET',
statusCode: null,
statusMessage: null,
client: [Object],
_consuming: false,
_dumped: false,
next: [Function: next],
baseUrl: '',
originalUrl: '/',
_parsedUrl: [Object],
params: [Object],
query: {},
res: [Object],
_startAt: [Object],
_startTime: 2018-06-26T15:22:55.146Z,
_remoteAddress: '::ffff:127.0.0.1',
body: {},
route: [Object] },
response:
ServerResponse {
domain: null,
_events: [Object],
_eventsCount: 2,
_maxListeners: undefined,
output: [],
outputEncodings: [],
outputCallbacks: [],
outputSize: 0,
writable: true,
_last: false,
upgrading: false,
chunkedEncoding: false,
shouldKeepAlive: true,
useChunkedEncodingByDefault: true,
sendDate: true,
_removedHeader: {},
_contentLength: null,
_hasBody: true,
_trailer: '',
finished: false,
_headerSent: false,
socket: [Object],
connection: [Object],
_header: null,
_headers: [Object],
_headerNames: [Object],
_onPendingData: [Function: updateOutgoingData],
req: [Object],
locals: {},
_startAt: undefined,
_startTime: undefined,
writeHead: [Function: writeHead],
__onFinished: [Object] } } }

PR #2957, opened 2019-03-04
Add test for http request to the prometheus api before backup and aft…

venturasr:
**Related issue(s)**
Issue #598

### DIFF
--- a/tests/end-to-end/backup-restore-test/backupe2e/prometheus.go
+++ b/tests/end-to-end/backup-restore-test/backupe2e/prometheus.go
@@ -155,41 +155,47 @@ func (qresp *queryResponse) UnmarshalJSON(data []byte) error {
 
 	m := f.(map[string]interface{})
 
-	for key1, _ := range m {
+	for key1, val1 := range m {
 		switch key1 {
 		case "status":
-			status := m[key1].(string)
-			qresp.Status = status
+			qresp.Status = val1.(string)
 		case "data":
-			data := m[key1]
+
+			data := val1.(map[string]interface{})
 			respData := &responseData{}
-			for key2, d := range data.(map[string]interface{}) {
+			for key2, val2 := range data {
+
 				switch key2 {
 				case "resultType":
-					respData.ResultType = d.(string)
+					respData.ResultType = val2.(string)
 				case "result":
-					reslt := d
+
+					reslt := val2.([]interface{})
 					result := make([]dataResult, 2)
-					for idx, r := range  reslt.([]interface {})  {
+					for idx, r := range reslt {
 						datResult := &dataResult{}
-						for o,v  := range r.(map[string]interface{}) {
+						for o, v := range r.(map[string]interface{}) {
 							switch o {
 							case "metric":
 								datResult.Metric = v.(map[string]interface{})
 							case "value":
-								datResult.Value = v.([]interface {})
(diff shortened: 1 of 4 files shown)


tests/end-to-end/backup-restore-test/backupe2e/prometheus.go
```diff
@@ -0,0 +1,222 @@
...
+
+	if err != nil {
+		log.Fatalf("Unable to get a reponse from the api. \n http response was '%s' (%d) and not OK (200). Error:\n%s\n", resp.Status, resp.StatusCode, err)
+	}
+
+	qresp.UnmarshalJSON(body)
+
+}
+
+func whatIsthisThing(something interface{}) (float64, string, error) {
```
k15r: whatIsThisThing

tests/end-to-end/backup-restore-test/backupe2e/prometheus.go
```diff
@@ -0,0 +1,222 @@
...
+	if pointInTime != "" {
+		values.Set("time", pointInTime)
+	}
+
+	uri := domain + ":" + port + api
+	client := &http.Client{
+		Timeout: 15 * time.Second,
+	}
+	url := uri + values.Encode()
+	log.Println(url)
```
k15r: all lines that you create using log will be shown even when the test does not fail. imho the proper way for such output would be to use the testing object. (which is not yet accessible, so maybe we should somehow pass it to the  end2end tests)

tests/end-to-end/backup-restore-test/backupe2e/prometheus.go
venturasr: Changed to`whatIsThisThing`

tests/end-to-end/backup-restore-test/restore_cluster_backup_test.go
```diff
@@ -51,9 +51,17 @@ func TestBackupAndRestoreCluster(t *testing.T) {
 
 	for idx, backupTest := range backupTests {
 		testUUID := uuid.New()
+
+		name := string("")
+		if t := reflect.TypeOf(backupTest); t.Kind() == reflect.Ptr {
+			name = "*" + t.Elem().Name()
```
k15r: thats not a valid dns name

regex used for validation is ```'[a-z0-9]([-a-z0-9]*[a-z0-9])?'```

tests/end-to-end/backup-restore-test/backupe2e/prometheus.go
```diff
@@ -0,0 +1,222 @@
...
+	}
+
+}
+
+func (pt *prometheusTest) TestResources(namespace string) {
+	qresp := &queryResponse{}
+	qresp.connectToPrometheusApi(pt.domain, pt.port, pt.api, pt.metricQuery, pt.pointInTime.formmattedValue)
+	pt.afterRestore = *qresp
+
+	if len(qresp.Data.Result) > 0 && len(qresp.Data.Result[0].Value) == 2 {
```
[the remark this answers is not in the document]
venturasr: go-cmp looks nice, but why, only for comparing two strings. Importing a library only to do that.

venturasr:
Replaced by https://github.com/kyma-project/kyma/pull/3049

Issue #2929, opened 2019-03-01
Implement k8s clients for Cluster/-Microfrontends CRDs


kwiatekus:
The test for backup and restore requires a k8s client to create test data (MFs & CMFs)

PR #2959, opened 2019-03-04
Generate Apis for microfrontends and clustermicrofrontends

kwiatekus:
**Description**

- Generate apis for MFs and CMFs resources

**Related issue(s)**
https://github.com/kyma-project/kyma/issues/2929


components/microfrontend/hack/generate-groups.sh
```diff
@@ -0,0 +1,89 @@
+#!/usr/bin/env bash
```
pkosiec: It is not needed as you use `./vendor/k8s.io/code-generator/generate-groups.sh`

components/microfrontend/pkg/apis/ui/v1alpha1/types.go
```diff
@@ -0,0 +1,75 @@
...
+func (rem *MicroFrontend) GetObjectKind() schema.ObjectKind {
+	return &MicroFrontend{}
+}
+
+func (rem *ClusterMicroFrontend) GetObjectKind() schema.ObjectKind {
+	return &ClusterMicroFrontend{}
+}
+
+type MicroFrontendSpec struct {
+	Version         string           `json:"version"`
```
pkosiec: As a part of spec is shared between MF and CMF, I would suggest to create a type `CommonMicroFrontendSpec`. It would be inlined in `MicroFrontendSpec` and `ClusterMicroFrontendSpec`.

components/microfrontend/hack/generate-groups.sh
kwiatekus: done

components/microfrontend/README.md
```diff
@@ -0,0 +1,41 @@
...
+Before each commit, use the `before-commit.sh` script, which tests your changes.
+
+## Code generation
+
+Structs related to CustomResourceDefinitions are defined in `pkg/apis/ui/v1alpha1/types.go` and registered in `pkg/apis/ui/v1alpha1/`. After making any changes there, please run:
+```bash
+./hack/update-codegen.sh
+```
+
+## Release
```
pkosiec: I don't think this is valid. I would suggest removing it.

components/microfrontend/pkg/apis/ui/v1alpha1/types.go
```diff
@@ -0,0 +1,72 @@
...
+	Category        string           `json:"category"`
+	ViewBaseUrl     string           `json:"viewBaseUrl"`
+	NavigationNodes []NavigationNode `json:"navigationNodes"`
+}
+
+type ClusterMicroFrontendSpec struct {
+	Placement         string `json:"placement"`
+	MicroFrontendSpec `json:",inline"`
```
pkosiec: Please don't use the spec type for MicroFrontend CR directly. Create another `CommonMicroFrontendSpec` and put all common properties there. The `MicroFrontendSpec` will look like:

 ```golang
type MicroFrontendSpec struct {
    CommonMicroFrontendSpec `json:",inline"`
}
```

and the `ClusterMicroFrontendSpec`:

```golang
type ClusterMicroFrontendSpec struct {
	Placement         string `json:"placement"`
	CommonMicroFrontendSpec `json:",inline"
}
```

That's the way the Service Catalog guys do it. Why? It will be easier for extending these two types without making breaking changes. The `ClusterMicroFrontend` CR won't need to contain everything that the `MicroFrontend` CR contains in spec.

components/microfrontend/pkg/apis/ui/v1alpha1/types.go
```diff
@@ -0,0 +1,72 @@
...
+}
+
+func (rem *ClusterMicroFrontend) GetObjectKind() schema.ObjectKind {
+	return &ClusterMicroFrontend{}
+}
+
+type MicroFrontendSpec struct {
+	Version         string           `json:"version"`
+	Category        string           `json:"category"`
+	ViewBaseUrl     string           `json:"viewBaseUrl"`
```
pkosiec: In Go there is a convention to name abbreviations in uppercase form. In this case:
`ViewBaseURL`. Of course keep the `json` representation as it is (`viewBaseUrl`).

components/microfrontend/pkg/apis/ui/v1alpha1/types.go
```diff
@@ -0,0 +1,72 @@
...
+
+type ClusterMicroFrontendSpec struct {
+	Placement         string `json:"placement"`
+	MicroFrontendSpec `json:",inline"`
+}
+
+type NavigationNode struct {
+	Label            string `json:"label"`
+	NavigationPath   string `json:"navigationPath"`
+	ViewUrl          string `json:"viewUrl"`
```
pkosiec: In Go there is a convention to name abbreviations in uppercase form. In this case:
`ViewURL`. Of course keep the `json` representation as it is (`viewUrl`).

components/microfrontend/README.md
pkosiec: Please remove it also from IDP presets directory. It's from far times before repo consolidation.

common/README.md
```diff
@@ -16,6 +16,10 @@ To add a new library to this package, it must meet these criteria:
 - It must be well covered with unit tests. Many components may end up depending on the library you add.
 - It must have open source-grade documentation. The library will be widely used and must be easy to use without the need to contact the author.
 
+## Naming conventions
+
+If the common package is a kubernetes client for a custom resource please add a `-client` suffix to the folder name.
```
bszwarc:
```suggestion
If the common package is a Kubernetes client for a custom resource, add a `-client` suffix to the folder name.
```

PR #2965, opened 2019-03-04
Add monitoring setup for minio and fix other custom monitoring

derberg:
**Description**

- Enable monitoring for minio
  screen shot 2019-03-04 at 21 15 11
- Custom Grafana dashboard for Minio
- Fix for Service Monitors from different components

### DIFF
--- a/resources/monitoring/Chart.yaml
+++ b/resources/monitoring/Chart.yaml
@@ -1,4 +1,7 @@
 apiVersion: v1
+#This value is used as serviceMonitorSelector matching label.
+#Whenever you change this value please make sure to update Service Monitor kind of all components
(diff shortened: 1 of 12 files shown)


resources/monitoring/Chart.yaml
```diff
@@ -1,4 +1,7 @@
 apiVersion: v1
+#This value is used as serviceMonitorSelector matching label.
```
mszostok: IMO in future, they should change that behavior and do not connect it with the chart name. Could u ping them?

derberg: right, but I'm just fixing bug here. Normally I agree that should be in values.yaml
owners of the component are included in review

Issue #2153, opened 2019-01-03
Upgrade Knative to newest compatible version


sjanota:
**Description**

Currently, Kyma is installed with Knative 0.2.1. We should move to version 0.3.x as soon as it is available. One particular reason is: https://github.com/knative/serving/pull/2434 as having this will greatly simplify the installation process. In particular, it will allow using dynamic IPs with xip.io which is not possible now.

**AC**
- [x] K8S is upgraded to the version that is required for Knative 0.4 to work with Kyma
- [ ]  Knative for Kym is upgraded to 0.4 version.

abbi-gaurav:
Currently, there is an issue with K8S 1.11.x (core DNS) that breaks Knatve functionality 0.3. The option is to update K8S version to 1.12. To have a consistent behavior, we need to wait until K8S 1.12 is available on GKE.

Refer https://github.com/knative/docs/pull/873

radufa:
Knative/Eventing 0.4 was changed to support CE for all provisioners.

The changes for NATSS provisioner were done here:
https://github.com/knative/eventing/pull/794/

PR #2972, opened 2019-03-05
Update the cluster user tests to check for patterns

abbi-gaurav:
**Description**
While doing a recent K8S upgrade required for Knative upgrade, it is observed that the
cluster-users-test is failing for a pattern match
`|FAIL| Expected: no got: no - no RBAC policy matched`
The reason being with K8S 1.12 the output of can-i
changed from no to no - reason.
This PR accommodates this possibility to have an extra `elif` check
**Related issue(s)**
See also #2153


abbi-gaurav:
Verified locally that the tests pass.

Issue #2852, opened 2019-02-22
CSRF token support in Application Gateway


pbochynski:
**Description**

Some applications use CSRF tokens to protect their resources. Application Gateway should fetch & cache such token and use it for requests.
Add a configuration that will enable CRSF token handling and provides necessary parameters (token fetch request params, token header name, etc)

**Reasons**

Kyma developer should not handle CRSF tokens in own code. 

**Attachments**

- https://github.com/OWASP/CheatSheetSeries/blob/master/cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.md

piotrmsc:
1. Fetching a token : 
- Set ** 'X-csrf-token': 'fetch'** request header 
- Store token and cookie from the token response headers ('x-csrf-token', 'set-cookie')

2. Call upstream service with a token
based on step 1, set following request headers : 
- **'X-csrf-token': '$TOKEN'** 
- ** 'cookie' : $COOKIE' 

Handling CSRF token is optional, enabled while registering API via CR.

Szymongib:
Consider adding a new authorization type `BasicWithCSRF` by extending the existing type `Basic`. Note that the `AuthenticationUrl` should be provided.

jakkab:
@Szymongib I have spoken with @piotrmsc and apparently the implementation needs to be more generic: we should be able to extend each authentication strategy (basic, oauth, certificate) with CSRF token.

lszymik:
Hello @strekm , @jakkab , @akgalwas Do we have more detailed documentation somewhere. I am looking for an explanation of how it works in the Gateway service.

Disper:
@lszymik It's not yet documented.

PR #2998, opened 2019-03-05
Csrf token support

jakkab:
**Description**

- Add CSRF token support

Details:
CSFR Token support is configured per application.
CSFR support is enabled by adding "csrfInfo" attribute to application instance (see #2996)
If configured, the Gateway sends an "X-csrf-token" header and related cookies along with every request to the target API.
The header value and cookies are pre-fetched before the actual request to the target API.
CSRF Token and cookies are cached to speed up subsequent requests.

**Related issue(s)**
#2852


components/application-gateway/Gopkg.toml
```diff
@@ -57,4 +57,5 @@ required = [
 
 [[constraint]]
   name = "github.com/kyma-project/kyma"
-  revision = "d37318b3a84115bd9ee64d3c3130ef657b8a81a5"
+  source = "github.com/jakkab/kyma"
```
Tomasz-Smelcerz-SAP: Don't forget to remove

components/application-gateway/internal/metadata/applications/repository.go
```diff
@@ -147,5 +148,6 @@ func convertCredentialsFromK8sType(credentials v1alpha1.Credentials) *Credential
 		Type:       credentials.Type,
 		SecretName: credentials.SecretName,
 		Url:        credentials.AuthenticationUrl,
+		CSRFUrl:    credentials.CSRFAuthorizationURL,
```
Tomasz-Smelcerz-SAP: It think it should be: TokenEndpointURL?

components/application-gateway/internal/metadata/model/model.go
```diff
@@ -36,6 +36,8 @@ type Credentials struct {
 	BasicAuth *BasicAuth
 	// CertificateGen is CertificateGen configuration.
 	CertificateGen *CertificateGen
+	// CSRFTokenURL (optional) to fetch CSRF token
+	CSRFTokenURL string
```
Tomasz-Smelcerz-SAP: I would rather introduce a CSRFToken type to align with existing convention:
```
CSRFToken: *CSRFToken

...

type CSRFToken struct {
    CSRFTokenEndpointURL string
}
```

components/application-gateway/internal/authorization/certificategenstrategy.go
```diff
@@ -20,7 +19,7 @@ func newCertificateGenStrategy(certificate, privateKey []byte) certificateGenStr
 	}
 }
 
-func (b certificateGenStrategy) AddAuthorization(r *http.Request, proxy *httputil.ReverseProxy) apperrors.AppError {
+func (b certificateGenStrategy) AddAuthorization(r *http.Request, setter TransportSetter) apperrors.AppError {
```
kfurgol: Consider renaming `setter` to something more accurate

Tomasz-Smelcerz-SAP: As we know, naming things is one of two things really hard in computer science. Please suggest a better name ;)

components/application-gateway/internal/authorization/csrf/csrfclient_test.go
```diff
@@ -0,0 +1,182 @@
...
+		assert.Equal(t, cachedTestToken, response.CSRFToken)
+		assert.Equal(t, cachedTestCookieName, response.Cookies[0].Name)
+	})
+
+	t.Run("Should fetch the token from endpoint and add it to cache if it is not there", func(t *testing.T) {
+
+		// given
+		fakeCache := NewTokenCache()
+
+		c := client{timeoutDuration, fakeCache, testRequestToken}
```
akgalwas: I would suggest different approach to testing this component. In the current approach requestToken function is not tested (it contains the most important part of logic). I think that better approach would be to use httputil.NewServer to start test server so that requestToken will be executed.

components/application-gateway/internal/authorization/csrf/csrf.go
```diff
@@ -0,0 +1,73 @@
+package csrf
```
akgalwas: I'm just wondering if the directory structure is correct. "CSRF" directory is under "authorization". Looking at the proxy code we treat CSRF as something else, not an authorization method. Maybe there should be "internal/csrf" folder?

components/application-gateway/internal/authorization/csrf/csrf_strategy_test.go
```diff
@@ -0,0 +1,105 @@
+package csrf_test
```
akgalwas: Code contained is csrf.go file is tested in csrf_test.go and csrf_strategy_test.go. I think contents of csrf_strategy_test.go could be moved to csrf_test.go.

components/application-gateway/internal/httpconsts/httpconsts.go
```diff
@@ -1,9 +1,17 @@
...
 const (
-	HeaderContentType   = "Content-Type"
-	HeaderAuthorization = "Authorization"
-	HeaderAccessToken   = "Access-Token"
+	HeaderContentType     = "Content-Type"
+	HeaderAuthorization   = "Authorization"
+	HeaderAccessToken     = "Access-Token"
+	HeaderCSRFToken       = "X-csrf-token"
+	HeaderCSRFTokenVal    = "fetch"
+	HeaderSetCookie       = "set-cookie"
```
akgalwas: HeaderSetCookie and HeaderCookie constants not used.

components/application-gateway/internal/authorization/certificategenstrategy.go
akgalwas: Oh yeah, naming is the most difficult stuff! But I'm going to ask other question. What is the idea behind introducing TransportSender interface? I can't see an issue with setting transport directly on proxy.  After this change AddAuthorization method is not aware how the request will be made (which is neither good or bad in my opinion).

components/application-gateway/internal/authorization/csrf/csrf_strategy_test.go
jakkab: csrf_test.go and csrf_strategy_test.go are in different packages. The csrf_strategy_test.go file was extracted to csrf_test package to avoid circular dependencies. Additionally, we can test the logic from outside its own package.

components/application-gateway/internal/httpconsts/httpconsts.go
jakkab: removed

components/application-gateway/internal/authorization/certificategenstrategy.go
jakkab: @akgalwas AddAuthorization is now used in two different scenarios:
- regular calls to upstream service;
- requests sent to the service endpoint that exposes CSRF tokens (secured with the same auth method as regular calls).

If case of certificateGen strategy, we would need to create another instance of proxy to add authorization to requests. @Tomasz-Smelcerz-SAP and I discussed the topic with @Szymongib, he should know all the details.

components/application-gateway/internal/authorization/csrf/csrfclient_test.go
jakkab: That sounds reasonable.

components/application-gateway/internal/authorization/csrf/csrf.go
jakkab: As authorization is required to fetch CSRF tokens, we decided to place our code under authorization dir.

components/application-gateway/internal/authorization/csrf/csrf_strategy_test.go
akgalwas: You could solve the circular dependency by moving csrfclient to a separate package. OAuthStrategy is a good example : it is in authorization package and uses client defined in oauth package.

components/application-gateway/internal/authorization/csrf/csrf.go
akgalwas: Well, there is dependency between CSRF tokens and authorisation but code in proxy package considers it as something else. CSRF is a method of protecting from certain type of attack rather than separate method of authorization.

Tomasz-Smelcerz-SAP: @akgalwas Good point, package is moved.

components/application-gateway/internal/authorization/csrf/csrf_strategy_test.go
Tomasz-Smelcerz-SAP: @akgalwas moving client won't help here.
Circular dependency happens between the mock, the test file from _csrf_ package and public _*csrf.Response_ type.
It goes as follows: Mock requires *csrf.Response type since it's the part of a method signature being mocked. This type is defined in _csrf_ package. If the test itself is in the same package (_csrf_), there's an inevitable circular dependency. The best way to solve is to put test in a _different_ package, like we did here: _csrf_test_.
This does not happen in oauth package, since the Mock there does not depend on any type from "parent" package.

components/application-gateway/internal/authorization/certificategenstrategy.go
Tomasz-Smelcerz-SAP: @akgalwas about reasons for the TransportSetter function
Take a look at how AddAuthorization was implemented before: It took a reverseProxy as an argument, and then, only in one particular implementation (certificategenstrategy.go), it was setting a Transport object on the Proxy. Later on the _Client_ used by the Proxy can make API requests using this Transport.
In this PR we have another _Client_: a client to fetch CSRF tokens. This client is not related to proxy, but it has to use the same authorization method as the Proxy Client (it is an Acceptance Criteria for the task).
It turns out that we have to re-use the same Authorization strategy for another client, but this time we don't have any "Proxy" object to pass to "AddAuthorization"...
We've had two choices:
- Provide a dummy "Proxy" object  For CSRF client just to satisfy AddAuthorization signature and setup Transport on CSRF Client properly
- Refactor AddAuthorization interface to be more specific. AddAuthorization actually never required access to the entire _Proxy_ object, all it was doing was setting a Transport. Applying the "Principle of Least Responsibility" here leads us to the interface that allows just that: setting a transport

We have chosen the second option, after consulting this choice with @Szymongib 
I hope that explains our design choices. The name is still bad, though :)

Issue #2601, opened 2019-02-05
Enable monitoring for event bus services


abbi-gaurav:
* event-service
* publish-knative

Refer 
* https://github.com/kyma-project/community/blob/master/guidelines/add-monitoring-to-component-docs.md
* https://github.com/kyma-project/kyma/issues/1916

**AC**
- [x] Functional Monitoring dashboards have been created for event-service and publish-knative
- [ ] Standard alerts for the APIs have been created based on the guidelines.

montaro:
@abbi-gaurav @marcobebway  Where exactly can the monitoring get injected for the `event-service`, I can see its image is injected in the `application-operator`, should the monitoring enabled for the `application-operator` instead?

abbi-gaurav:
@montaro please check with automatix.

PR #2999, opened 2019-03-05
Enable publish-knative monitoring and event-service

montaro:
See also #2601

### DIFF
--- a/components/application-operator/charts/application/templates/deployment.yaml
+++ b/components/application-operator/charts/application/templates/deployment.yaml
@@ -73,6 +73,8 @@ spec:
         sidecar.istio.io/inject: "true"
       labels:
         app: {{ .Release.Name }}-event-service
+        kyma-grafana: {{ .Values.eventService.deployment.labels.grafana }}
+        kyma-alerts: {{ .Values.eventService.deployment.labels.alerts }}
         release: {{ .Release.Name }}-event-service
     spec:
       containers:
--- a/resources/event-bus/charts/publish-knative/templates/_helpers.tpl
+++ b/resources/event-bus/charts/publish-knative/templates/_helpers.tpl
@@ -1,12 +1,4 @@
 {{- define "publish-knative.name" -}}
-{{- printf "publish-knative" -}}
-{{- end -}}
-
-{{- define "publish-knative.fullname" -}}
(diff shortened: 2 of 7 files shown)


resources/event-bus/charts/publish-knative/templates/_helpers.tpl
```diff
@@ -1,15 +1,11 @@
 {{- define "publish-knative.name" -}}
-{{- printf "publish-knative" -}}
+{{- printf "%s-%s" .Release.Name "publish" | trunc 63 | trimSuffix "-" -}}
 {{- end -}}
 
 {{- define "publish-knative.fullname" -}}
```
marcobebway: Do we still need the **publish-knative.fullname**?

montaro: No, will remove it

components/application-operator/charts/application/templates/deployment.yaml
```diff
@@ -73,6 +73,8 @@ spec:
         sidecar.istio.io/inject: "true"
       labels:
         app: {{ .Release.Name }}-event-service
+        kyma-grafana: enabled
+        kyma-alerts: enabled
```
Szymongib: This changes will take effect only after you bump image of `application-operator`. You might want to extract the to separate PR and bump image of operator on this one.

janmedrek: You are setting those to always be `enabled`, what would happen when we use this on the cluster that has no monitoring/alerting? Parametrisation is a good practice and should be applied here.

Please use the `values.yaml` file to parametrise those, there is already `eventService.deployment` field there, you could do something like:
```
labels:
  app: ...
  kyma-grafana: {{ .Values.eventService.deployment.labels.grafana }}
  kyma-alerts: {{ .Values.eventService.deployment.labels.alerts }}
```

and then in `values.yaml`:
```
eventService:
  deployment:
    ...
    labels:
      grafana: "enabled"
      alerts: "enabled"
    ...
```

montaro: I'll update the `application-operator version` in a following PR after this PR gets merged/image built

Issue #2301, opened 2019-01-15
Integrate subscription controller with Knative eventing


abbi-gaurav:
* Updation of Kyma Subscription to `EventActvation = true` 
  * Create Knative Subscription
  * Update of a Kyma subscription from the user should 
     * identify if there is a delta
     * delete Knative subscription
     * create new Knative subscription
* Updation of Kyma Subscription to `EventActvation = false` deletes Knative Subscription
* When Kyma Subscription is deleted, Knative Subscription is deleted.

**AC**
* [x] The subscription controller reacts to Kyma Subscription update and delete based on `EventActivation` flag.
* [ ] There are tests for the functionality.
* [ ] Audit logs are emitted for the Kyma subscription lifecycle events.

radufa:
If a Knative subscription exists, kn-lib will return now "k8serrors.IsAlreadyExists" if the controller want to create the same subscription again:
https://github.com/kyma-project/kyma/pull/2849

abbi-gaurav:
The functionality was merged but the tests are not yet written.

abbi-gaurav:
@suleymanakbas91 @radufa , deleting a Kyma subscription is not deleting the Knative subscription but only the channel.

Is there a fix already planned for this?

here are the logs
```
{"level":"info","ts":1552321793.4075408,"logger":"subscription-controller","caller":"subscription/reconcile.go:150","msg":"Knative Subscription is created","Subscription":"example-subscription-test--event--bus"}
{"level":"info","ts":1552321793.4076169,"logger":"subscription-controller","caller":"subscription/reconcile.go:73","msg":"Subscription reconciled"}
{"level":"info","ts":1552321793.4077253,"logger":"subscription-controller","caller":"subscription/reconcile.go:44","msg":"Reconcile: ","request":"test-event-bus/example-subscription"}
{"level":"info","ts":1552321793.4077961,"logger":"subscription-controller","caller":"subscription/reconcile.go:64","msg":"Reconciling Subscription","UID":"e61ed5d3-441a-11e9-8c51-42010a8400b0"}
{"level":"info","ts":1552321793.4224305,"logger":"subscription-controller","caller":"subscription/reconcile.go:73","msg":"Subscription reconciled"}
{"level":"info","ts":1552322278.552447,"logger":"subscription-controller","caller":"subscription/reconcile.go:44","msg":"Reconcile: ","request":"test-event-bus/example-subscription"}
[...]

PR #3007, opened 2019-03-06
Modify subscription controller chart

suleymanakbas91:
**Description**

- Modify subscription controller chart
- Delete test resources as we have a chart now

**Related issue(s)**

See #2301


resources/event-bus/charts/subscription-controller-knative/templates/service-account.yaml
```diff
@@ -96,3 +96,27 @@ roleRef:
...
+---
+### Events #####################################################
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: {{ .Release.Name }}-events-controller
+rules:
+  - apiGroups: [""]
+    resources: ["events"]
+    verbs: ["get", "watch", "list", "create", "update", "patch", "delete"]
```
radufa: perhaps "create", "update", "patch" will be enough

radufa:
Please restrict the access rights as much as possible, I suppose that the following rights are enough 
- for KN channels:
verbs: ["get", "create", "update","delete"]

 - for KN subscription:
verbs: ["get", "create", "delete"]

resources/event-bus/charts/subscription-controller-knative/templates/service-account.yaml
suleymanakbas91: Done

Issue #2299, opened 2019-01-15
Integrate EventActivation controller with Kyma Subscription


abbi-gaurav:
* Creation of Subscription
  * Check if EventActivation present
  * Update the status of Kyma Subscriptions (set the EventActivation status to true).

* Creation of EventActivation
  * Update the status of Kyma Subscriptions (set the EventActivation status to true).

* Deletion of EventActivation
  * Update the status of Kyma Subscriptions (set the EventActivation status to false).

**AC**
- [x] EventActivation controller updates the status of Kyma Subscription  on its creation
- [x] EventActivation controller reacts on the EventActivation resource create and delete and updates the Kyma Subscription.
- [ ] ~~There are tests written to verify the controller behavior.~~

abbi-gaurav:
While testing the [e2e scenario](https://github.com/kyma-incubator/examples/tree/master/kyma-local-e2e-scenario), discovered that the event-activation controller only reacted after restarting the subscription-controller pod.

I believe this needs to be investigated.

abbi-gaurav:
The missing tests will be written as a part of #2301

radufa:
See also https://github.com/kyma-project/kyma/pull/3118

radufa:
The following manual tests were done:

1. create kyma-subscritption in "default" namespace
2. create eventactivation for this subscription in "default" namespace
Subscription controller reacts:
   3. it creates a knative-subscription in "kyma-system" namespace
   4. it creates a knative channel in "kyma-system" namespace

1 delete the kyma-subscription in "default" namespace
Subscription controller reacts:
   2. deletes the knative-subscription 
   3. deletes the channel subscription (cause the channel has an empty list of subscribers)

It was tested also if the first two steps are done in a different order:
1. create an eventactivation in "default" namespace
2. create a kyma-subscription in the "default" namespace

PR #3008, opened 2019-03-06
Append kyma subscription namespace to knative subscription name

radufa:
**Related issue(s)**

See also #2299

### DIFF
--- a/components/event-bus/cmd/event-bus-publish-knative/handlers/publish.go
+++ b/components/event-bus/cmd/event-bus-publish-knative/handlers/publish.go
@@ -17,7 +17,7 @@ import (
 )
 
 var (
-	defaultChannelNamespace = "kyma-system"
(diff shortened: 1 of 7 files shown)


components/event-bus/cmd/event-bus-publish-knative/handlers/publish.go
```diff
@@ -17,7 +17,7 @@ import (
 )
 
 var (
-	defaultChannelNamespace = "kyma-system"
+	DefaultChannelNamespace = "kyma-system"
```
marcobebway: As the **DefaultChannelNamespace** became public, I suggest to move it somewhere under **internal/knative**.

radufa: Done

Issue #2602, opened 2019-02-05
Enable monitoring for event bus controllers


abbi-gaurav:
Refer 
* https://github.com/kyma-project/community/blob/master/guidelines/add-monitoring-to-component-docs.md
* https://github.com/kyma-project/kyma/issues/1800

**AC**
- [ ] Monitoring is enabled for subscription-controller (knative)
- [ ] Standard alerts for the controllers have been created based on the guidelines.

montaro:
Subscription Controller pod organically has the Pod metrics monitors
![image.png](https://images.zenhubusercontent.com/5b865e0b83bd076dd89c9034/acbb27ed-4f10-46ae-9895-d555aa589215)
Is that enough? because there are no RED metrics for it? @abbi-gaurav

PR #3046, opened 2019-03-07
WIP Enable monitoring for subscription controller

montaro:
See also #2602


montaro:
will be fixed in https://github.com/kyma-project/kyma/pull/3076

PR #3081, opened 2019-03-10
Limit maximum event publishing request size

sslavic:
**Description**

- Limit maximum event publishing request size in event-bus and event-service

### DIFF
--- a/components/event-bus/cmd/event-bus-publish-knative/validators/publish_validator.go
+++ b/components/event-bus/cmd/event-bus-publish-knative/validators/publish_validator.go
@@ -9,6 +9,10 @@ import (
 	api "github.com/kyma-project/kyma/components/event-bus/api/publish"
 )
 
+const (
+	requestBodyTooLargeErrorMessage = "http: request body too large"
+)
+
 func ValidateRequest(r *http.Request) (*api.PublishRequest, *api.Error) {
 	// validate the http method
 	if r.Method != http.MethodPost {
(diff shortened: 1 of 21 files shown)


components/event-bus/cmd/event-bus-publish-knative/validators/publish_validator.go
```diff
@@ -27,6 +27,11 @@ func ValidateRequest(r *http.Request) (*api.PublishRequest, *api.Error) {
 	_ = r.Body.Close()
 	if err != nil {
 		log.Printf("failed to read request body: %v", err)
+
+		if err.Error() == "http: request body too large" {
```
radufa: Could you please put "http: request body too large" in a const, so that if Golang changes the implementation of  MaxBytesReader: https://golang.org/src/net/http/request.go
we should easier react.

components/event-bus/cmd/event-bus-publish/handlers/handlers.go
```diff
@@ -73,7 +81,13 @@ func handlePublishRequest(w http.ResponseWriter, r *http.Request, publisher *con
 
 	if err != nil {
 		log.Printf("PublishHandler :: handlePublishRequest :: Unexpected error while reading request body. Error: %v", err)
-		publish.SendJSONError(w, api.ErrorResponseInternalServer())
+		var apiError *api.Error
+		if err.Error() == "http: request body too large" {
```
radufa: See previous comment

PR #3099, opened 2019-03-11
Update images for release 0.8.0-rc1

k15r:
**Description**

- Update images used in release

**Related issue(s)**

### DIFF
--- a/installation/resources/watch-pods.yaml
+++ b/installation/resources/watch-pods.yaml
@@ -48,7 +48,7 @@ metadata:
 spec:
   containers:
   - name: watch-pods
-    image: eu.gcr.io/kyma-project/develop/watch-pods:c3c08515
+    image: eu.gcr.io/kyma-project/watch-pods:0.8.0-rc1
     imagePullPolicy: Always
     env:
     - name: ARGS
--- a/resources/ark/values.yaml
+++ b/resources/ark/values.yaml
@@ -71,5 +71,5 @@ global:
   containerRegistry:
     path: eu.gcr.io/kyma-project
   ark_plugin:
-    dir: develop/
-    version: "52070221"
+    dir:
+    version: 0.8.0-rc1
--- a/resources/ark/values.yaml
+++ b/resources/ark/values.yaml
@@ -7,7 +7,7 @@ image:
 kubectl:
   image:
     repository: eu.gcr.io/kyma-project/etcd-tls-setup
-    tag: 0.3.367
+    tag: 0.8.0-rc1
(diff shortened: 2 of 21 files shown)


resources/ark/values.yaml
```diff
@@ -71,5 +71,5 @@ global:
   containerRegistry:
```
aszecowka: here I think we should also bump eu.gcr.io/kyma-project/etcd-tls-setup

resources/event-bus/values.yaml
```diff
@@ -5,11 +5,11 @@ global:
   containerRegistry:
     path: eu.gcr.io/kyma-project
   event_bus:
-    dir: develop/
-    version: 9e7cfa58
+    dir:
```
montaro: @suleymanakbas91 What should this `event-bus` point to?
we don't have images in the root directory, no?

sayanh: @montaro For release builds the images are not in `develop/`. Please refer https://status.build.kyma-project.io/view/gcs/kyma-prow-logs/pr-logs/pull/kyma-project_kyma/3099/pre-rel08-kyma-components-event-bus/1105128449070075904

suleymanakbas91: @montaro here is how we use these values inside a chart: https://github.com/kyma-project/kyma/blob/master/resources/event-bus/charts/publish-knative/templates/deployment.yaml#L29. Of course, we have images in the root directory: https://console.cloud.google.com/gcr/images/kyma-project/EU/event-bus-publish-knative

installation/resources/watch-pods.yaml
```diff
@@ -48,7 +48,7 @@ metadata:
 spec:
   containers:
   - name: watch-pods
-    image: eu.gcr.io/kyma-project/develop/watch-pods:c3c08515
```
suleymanakbas91: `watch-pod` has been recently moved to test-infra. So, it doesn't have `0.8.0-rc1` tag. Either revert this or release it from test-infra.

sayanh: done

resources/ark/values.yaml
sayanh: done

resources/event-bus/values.yaml
montaro: @suleymanakbas91 this is what I really mean 
https://console.cloud.google.com/gcr/images/kyma-project/EU/event-bus?gcrImageListsize=30

PR #3105, opened 2019-03-11
Switch app operator to new domain var

Demonsthere:
**Description**

- Switch app operator to new domain var

**Related issue(s)**

Part of #2563

### DIFF
--- a/components/application-operator/charts/application/templates/ingress.yaml
+++ b/components/application-operator/charts/application/templates/ingress.yaml
@@ -24,7 +24,7 @@ metadata:
       }
 spec:
   rules:
-  - host: gateway.{{ .Values.global.domainName }}
+  - host: gateway.{{ .Values.global.applicationConnector.domainName }}
     http:
       paths:
       - path: /{{ .Release.Name }}/v1/events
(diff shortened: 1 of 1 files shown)


components/application-operator/charts/application/templates/ingress.yaml
```diff
@@ -24,7 +24,7 @@ metadata:
       }
 spec:
   rules:
-  - host: gateway.{{ .Values.global.domainName }}
+  - host: gateway.{{ .Values.global.applicationConnector.domainName }}
```
Szymongib: Those overrides are provided by Application Operator and should not be changed unless you are changing the code which provides the overrides.

components/application-operator/charts/application/templates/ingress.yaml
```diff
@@ -38,5 +38,5 @@ spec:
 
   tls:
   - hosts:
-    - gateway.{{ .Values.global.domainName }}
+    - gateway.{{ .Values.global.applicationConnector.domainName }}
```
Szymongib: Same as above

Demonsthere:
This change seems to not be needed, as the values used here just share the same name, but are not the same. Closing

PR #3118, opened 2019-03-12
Fix subscription controller to delete the Knative Subscriptions

suleymanakbas91:
**Description**

- Fix the problem where Knative Subscriptions are not being deleted

**Related issue(s)**

See #2301

### DIFF
--- a/components/event-bus/internal/knative/subscription/controller/subscription/reconcile.go
+++ b/components/event-bus/internal/knative/subscription/controller/subscription/reconcile.go
@@ -101,7 +101,7 @@ func (r *reconciler) reconcile(ctx context.Context, subscription *eventingv1alph
 		// The object is being deleted
 		if util.ContainsString(&subscription.ObjectMeta.Finalizers, finalizerName) {
 			// our finalizer is present, so lets handle our external dependency
-			if err := r.deleteExternalDependency(ctx, knativeSubsName, knativeSubsURI, knativeChannelName, knativeSubsNamespace); err != nil {
+			if err := r.deleteExternalDependency(ctx, knativeSubsName, knativeChannelName, knativeSubsNamespace); err != nil {
 				// if fail to delete the external dependency here, return with error
 				// so that it can be retried
 				return false, err
@@ -199,7 +199,7 @@ func (r *reconciler) reconcile(ctx context.Context, subscription *eventingv1alph
 	return false, nil
 }
 
-func (r *reconciler) deleteExternalDependency(ctx context.Context, knativeSubsName string, knativeSubsURI string, channelName string, namespace string) error {
+func (r *reconciler) deleteExternalDependency(ctx context.Context, knativeSubsName string, channelName string, namespace string) error {
 	log.Info("Deleting the external dependencies")
 
 	// In case Knative Subscription exists, delete it.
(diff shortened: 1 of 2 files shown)


montaro:
Please edit the title to be more descriptive

components/event-bus/internal/knative/subscription/controller/subscription/reconcile.go
```diff
@@ -205,11 +199,11 @@ func (r *reconciler) reconcile(ctx context.Context, subscription *eventingv1alph
 	return false, nil
 }
 
-func (r *reconciler) deleteExternalDependency(ctx context.Context, subscription *eventingv1alpha1.Subscription, channelName string, channelNamespace string) error {
+func (r *reconciler) deleteExternalDependency(ctx context.Context, knativeSubsName string, knativeSubsURI string, channelName string, namespace string) error {
```
montaro: please split the param list over two code lines

components/event-bus/internal/knative/subscription/controller/subscription/reconcile.go
```diff
@@ -221,13 +215,13 @@ func (r *reconciler) deleteExternalDependency(ctx context.Context, subscription
 	}
 
 	// Check if Channel has any other Subscription, if not, delete it.
-	knativeChannel, err := r.knativeLib.GetChannel(channelName, channelNamespace)
+	knativeChannel, err := r.knativeLib.GetChannel(channelName, namespace)
 	if err != nil && !errors.IsNotFound(err) {
 		return err
 	} else if err == nil && knativeChannel != nil {
```
montaro: Is the case where the `knativeChannel == nil` handled?

components/event-bus/internal/knative/subscription/controller/subscription/reconcile.go
```diff
@@ -205,11 +199,11 @@ func (r *reconciler) reconcile(ctx context.Context, subscription *eventingv1alph
...
-func (r *reconciler) deleteExternalDependency(ctx context.Context, subscription *eventingv1alpha1.Subscription, channelName string, channelNamespace string) error {
+func (r *reconciler) deleteExternalDependency(ctx context.Context, knativeSubsName string, knativeSubsURI string, channelName string, namespace string) error {
 	log.Info("Deleting the external dependencies")
 
 	// In case Knative Subscription exists, delete it.
-	knativeSubs, err := r.knativeLib.GetSubscription(subscription.Name, channelNamespace)
+	knativeSubs, err := r.knativeLib.GetSubscription(knativeSubsName, namespace)
 	if err != nil && !errors.IsNotFound(err) {
 		return err
 	} else if err == nil && knativeSubs != nil {
```
montaro: Is the case where the `knativeSubs == nil` handled?

components/event-bus/internal/knative/subscription/controller/subscription/reconcile.go
radufa: The parameter "knativeSubsURI string" is not needed, cause if a knSubscriptoin exists, the the knSubscriberURI can be obtained from it:
```
 knSubscriberURI := knativeSubs.Spec.Subscriber.DNSName
```

components/event-bus/internal/knative/subscription/controller/subscription/reconcile.go
radufa: If "err == nil" then knativeSubs exists and it's always "!= nil".
so that the additional check "knativeSubs != nil" can be removed.

components/event-bus/internal/knative/subscription/controller/subscription/reconcile.go
radufa: Same for channels.
If "err == nil" then the channel object exists and it is returned.
The check "knativeChannel != nil" is not necessary

suleymanakbas91: if Knative Channel is nil, we don't need to do anything. There is nothing to delete.

components/event-bus/internal/knative/subscription/controller/subscription/reconcile.go
suleymanakbas91: if Knative Subscription is nil, we don't need to do anything. There is nothing to delete.

components/event-bus/internal/knative/subscription/controller/subscription/reconcile.go
suleymanakbas91: Done

suleymanakbas91: I already deleted one param, so it should be ok now

PR #3149, opened 2019-03-13
Add support for --kube-context/--context in testing.sh

sjanota:
**Description**

- testing.sh accepts `KUBE_CONTEXT` environmental variable which allows to set which context will be used by helm and kubectl

**Related issue(s)**

### DIFF
--- a/installation/scripts/testing.sh
+++ b/installation/scripts/testing.sh
@@ -9,7 +9,7 @@ echo "-------------------------------"
 
 if [ -n "$KUBE_CONTEXT" ]; then
     echo "Using context: $KUBE_CONTEXT"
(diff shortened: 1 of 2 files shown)


installation/scripts/testing.sh
```diff
@@ -34,56 +39,56 @@ echo "----------------------------"
 
 echo "- Testing Core components..."
 # timeout set to 10 minutes
-helm test core --timeout 600
+helm $KUBE_CONTEXT_ARG test core --timeout 600
```
Tomasz-Smelcerz-SAP: Let's stay on safe side of the code ;) Use double-quotes and curly braces like this: "${KUBE_CONFIG_ARG}"

Issue #3104, opened 2019-03-11
Rename UI API Layer & Asset Store and move Changelog Generator


pkosiec:
**Description**
Rename: 
- ui-api-layer
- ui-api-layer-acceptance-test
- Rename `assetstore` to `asset-store`

Remember about proper order of PRs. After changing `assetstore` to `asset-store`, update imports in CMS controller.
Do not rename charts.

**Reasons**

See: https://github.com/kyma-project/community/pull/191
See: https://github.com/kyma-project/community/pull/192

PR #3164, opened 2019-03-13
Rename ui-api chart to console-backend-service

pkosiec:
**Description**

- Rename ui-api chart to console-backend-module

**Related issue(s)**
#3104


derberg:
I used my power to merge with one failing plan because we are 100% the plan is failing for other reasons that are still not fixed and we need to unblock others that wait for merge of this change.

PR #3180, opened 2019-03-14
Use console-backend-service in console for pods and hide 'expose API' action for system namespaces

dariadomagala:
**Description**

- Use console-backend-service in console for pods
- Hide 'expose API' action for system namespaces
- Code cleanup

**Related issue(s)**


dariadomagala:
fixed by #3164

Issue #2860, opened 2019-02-25
Test Grafana Backup


joek:
AC:
- Dashboards are still there after the restore

venturasr:
Related PR: #3211

PR #3211, opened 2019-03-18
Test Grafana Backup and Restore

venturasr:
- Verify if the grafana pod is running.
- Login in the grafana ui.
- Gets Dex's credentials from Kubernetes Secret.
- Authenticates through Dex form.
- Execute http request to the grafana api to get the list of folders and dashboards.
- Execute a http request to every dashboard through the grafana api, expecting a 200 http status code.

**Related issue(s)**
#2860 Test Grafana Backup


tests/end-to-end/backup-restore-test/backupe2e/grafana.go
```diff
@@ -0,0 +1,337 @@
...
+			So(err, ShouldBeNil)
+		}
+	}
+
+	err = t.waitForPodGrafana(2 * time.Minute)
+	So(err, ShouldBeError)
+
+}
+
+func (t *grafanaTest) TestResources(namespace string) {
```
k15r: where does it chech that the config is the same as before the backup/restore?

tests/end-to-end/backup-restore-test/backupe2e/grafana.go
```diff
@@ -0,0 +1,372 @@
...
+			apiRequestsStatus[title] = dashboard.StatusCode
+		case "Etcd":
+			apiRequestsStatus[title] = dashboard.StatusCode
+		case "Connector-Service":
+			apiRequestsStatus[title] = dashboard.StatusCode
+		case "Istio Mesh Dashboard":
+			apiRequestsStatus[title] = dashboard.StatusCode
+		case "Istio Service Dashboard":
+
+		}
```
k15r: you could do the following:
before backup: query all dashboard, and store list in test object, try to access all of those dashboards
after backup: dont query the list of dashboard, but just use the list from the previous run, try to access all of them again

tests/end-to-end/backup-restore-test/backupe2e/grafana.go
```diff
@@ -0,0 +1,392 @@
...
+	coreClient, err := kubernetes.NewForConfig(config)
+	if err != nil {
+		return &grafanaTest{}, err
+	}
+
+	return &grafanaTest{
+		coreClient:  coreClient,
+		grafanaName: "grafana",
+		uuid:        uuid.New().String(),
+		before:      false,
```
k15r: thats a bit counter intuitive.
in the before case t.before should be true.
I would also like to have it renamed do beforeBackup (or similar)
Then after it was executed before the backup you could set it to false

venturasr: That is right, should be `beforeBackup :      true` to get into the if in the first run. In the way how it is now it looks more as an `afterBackup` var.

resources/ark/README.md
```diff
@@ -24,12 +24,14 @@ To add components to the backup pipeline, implement the following go interface:
...
     CreateResources(namespace string)
     TestResources(namespace string)
+    DeleteResources()
 }
 ```
 The functions work as follows:
 
 - The `TestResources` function validates if the test data works as expected. 
 - The `CreateResources` function installs the required test data before the backup process starts.
+- The `DeleteResources` function deletes those resources that are  part of the cluster before the execution of the test and need to be deleted to test the restore process.
```
bszwarc:
```suggestion
- The `DeleteResources` function deletes the resources that are a part of the cluster before executing the test. The resources need to be deleted to test the restore process.
```

PR #3239, opened 2019-03-19
Update Gopkg and add more unit tests

radufa:
- Add Kyma Subscription Ready status (true or false)

**Description**

- Assure that Gopkg.toml and Gopkg.lock are in sync ("dep check" returns no errors)
- Add more tests for both controllers ("event activation" and "subscription")
- Set the Ready status for Kyma subscription

**Related issue(s)**

See #2301


components/event-bus/internal/knative/util/kn-eventing.go
```diff
@@ -55,10 +55,31 @@ if err := k.CreateSubscription("my-sub", namespace, channelName, &uri); err != n
 return
 */
 
+// Use it cause it should be mocked
+type KnativeLibIntf interface {
```
marcobebway: Please change name of the `KnativeLibIntf` to something else which does not include the `Intf` suffix.

components/event-bus/internal/knative/util/kn-eventing.go
```diff
@@ -55,10 +55,31 @@ if err := k.CreateSubscription("my-sub", namespace, channelName, &uri); err != n
...
+	GetChannel(name string, namespace string) (*evapisv1alpha1.Channel, error)
```
marcobebway: Can we pass all functions(s) arguments by reference to avoid copying?

components/event-bus/internal/knative/util/kn-eventing.go
radufa: Done

components/event-bus/internal/knative/util/kn-eventing.go
radufa: The new interface exposes the same method as the type KnativeLib. It is needed to made possible to mock this interface for unit tests.
The only "string" parameters are "name" and "namespace" (like in the type KnativeLib)

Issue #1589, opened 2018-11-08
Acceptance tests for Proxy Service


janmedrek:
**Description**

Currently, we have no acceptance tests for our gateway-service, these should be created and integrated with the rest of application-connector tests.

**Acceptance Criteria**
- Scenarios for acceptance tests prepared, mind:
  - Basic Authentication
  - No authentication methods
- Above scenarios implemented
- Charts updated for new tests

Szymongib:
Follow up issue created: #3256

PR #3241, opened 2019-03-19
Run Application tests as part of Application Operator Tests

Szymongib:
**Description**

- Run Application acceptance tests as a part of Application Operator Tests

**Related issue(s)**
#1589


tests/application-operator-tests/test/testkit/config.go
```diff
@@ -19,7 +19,7 @@ const (
 type TestConfig struct {
 	Namespace           string
 	TillerHost          string
-	ProvisioningTimeout int
+	InstallationTimeout int
```
janmedrek: This field is always representing the time in seconds, changing it to `InstallationTimeoutSeconds` would improve readability. What do you think?

tests/application-operator-tests/test/testkit/wait.go
```diff
@@ -1,42 +1,44 @@
...
 }
 
-func (ts *TestSuite) shouldLastFor(conditionalFunc func() bool, message string, timeout time.Duration) {
+func ShouldLastFor(interval, timeout time.Duration, conditionalFunc func() bool) error {
 	done := time.After(timeout)
 
 	for {
 		if !conditionalFunc() {
-			ts.t.Errorf(message)
+			return errors.New("unexpected condition occurred when should not")
```
janmedrek: This error message is a bit blurry, I would suggest using just the `unexpected condition occurred` - don't know whether it makes it any better though. Do you have any ideas?

tests/application-operator-tests/test/testkit/wait.go
```diff
@@ -1,42 +1,44 @@
...
 		}
 
 		select {
 		case <-done:
 			if !conditionalFunc() {
-				ts.t.Errorf(message)
+				return errors.New("unexpected condition occurred when should not")
```
janmedrek: Same as above

tests/application-operator-tests/test/testkit/config.go
janmedrek: Same goes for TestSuite structs in other files, it would be best to change `installationTimeoutEnvName` value to something like `INSTALLATION_TIMEOUT_SECONDS` but if that is problematic (due to chart changes and upgradeability) let's just leave those like they are now.

PR #3156, opened 2019-03-13
Upgrade to Istio 1.1

piotrmsc:
**Description**

Changes proposed in this pull request:

- Added new istio-init 1.1 official release chart 
- Added new istio 1.1 official release chart
- Adjusted configurations in components 

**Related issue(s)**

#1875

PR #3301, opened 2019-03-22
Update istio-kyma-patch to require that policy checks are enabled

kubadz:
**Description**

- add require_policy_checks_enabled step to istio-kyma-patch

**Related issue(s)**

See also #3156


components/istio-kyma-patch/patch.sh
```diff
@@ -135,6 +160,7 @@ function check_requirements() {
 require_istio_system
 require_istio_version
 require_mtls_disabled
+configure_policy_checks_setting
```
sjanota: 1. After `check_requirements`
2. I think `configure_policy_checks` would be enough

Issue #3017, opened 2019-03-06
Update Kyma installation to install Knative by default


abbi-gaurav:
> **NOTE**
- **keep the non-knative deploy as a feature flag**

- **Ensure upgradability can be supported**

**[2019-03-25]**
What is expected of upgradability?
* Post 0.9 Kyma will get installed with Knative as default.
* Imagine a customer having 0.8 Kyma release running. Now when they upgrade to 0.9, they should not see any issues. e.g. with a Kyma subscription created with 0.8. It should still work after upgrading to 0.9.

How to simulate the upgrade from 0.8 to 0.9 (or your branch or wherever your changes), you need to check with gophers.

>**Note:** The Upgrade should be verified on a cluster. 

**Pending Items**
- Fix upgradability (delete volume during upgrade).
  - > **Note**: This action is only for any lower release to 0.9. For 0.9 to higher, no volume will be deleted.
- Fix GKE installation issue
- Fix configmaps, secrets

**AC**
- [ ] The default installation of Kyma should install Knative serving, Knative eventing and components build on top of Knative by default

montaro:
@abbi-gaurav Please, provide some information about 

- What's meant by *ensure upgradability can be supported*
- By having knative components installed by default, is this exclusively means having what's installed when we use `--knative` now, installed by default, or is there any more requirements to be installed/added in the default installation?

abbi-gaurav:
@montaro here's what it means

* Post 0.9 Kyma will get installed with Knative as default.
* Imagine a customer having 0.8 Kyma release running. Now when they upgrade to 0.9, they should not see any issues. e.g. with a Kyma subscription created with 0.8. It should still work after upgrading to 0.9.

How to simulate the upgrade from 0.8 to 0.9 (or your branch or wherever your changes), you need to check with gophers.

montaro:
> @montaro here's what it means
> 
>     * Post 0.9 Kyma will get installed with Knative as default.
Yes. I mean, whatever is installed now with `--knative` should be installed by default, or is there any charts need to be installed in addition?

abbi-gaurav:
Here is the documentation link that could be helpful https://kyma-project.io/docs/master/root/kyma/#installation-update-kyma

radufa:
Tested on Linux and Windows. A small, already corrected issue was found on Windows for "-disable-knative" installation

abbi-gaurav:
Currently, PR #3403 is stuck on the CI test failures.

```
[...]

Issue #3251, opened 2019-03-19
Events and Publish API does not provide proper error response when event is rejected


abbi-gaurav:
**Description**
Post Knative adoption, only those events for which there is a channel is present are accepted.

The channel has been created either as 
* a part of creating a Kyma subscription (at present) or 
* in the future to create a channel based on certain triggers.

The error returned when publishing an event in such a case is **Internal error** with no proper insights as to why this happened.

**AC**
- [x] The response sent should be HTTP 200
- [x] There are additional properties that provide a message indicating that the event is ignored.

abbi-gaurav:
After discussing with @lszymik, it was decided to send back an HTTP 200 response with an additional header indicating the event is ignored.

This will then not break the flow for all existing systems that are connected to Kyma / SAP CP XF

abbi-gaurav:
Updated to use the event property after having a discussion in the team and same is conveyed to @montaro

PR #3403, opened 2019-03-28
Make knative default

abbi-gaurav:
**Description**
- [x] get upgrade working
- [x] verify eventing flow with upgrade
- [x] remove all feature flags
- [x] remove or clean up any hardcodings
- [x] ensure install works on minikube
- [x] ensure install works on cluster
- [x] ensure clean upgrade works
- [x] Replace PR images with develop images

**Related issue(s)**
See also #3017


abbi-gaurav:
I see the following tests failing while making knative as default.

```bash
suite.go:164: --- FAIL: TestGatewayEvents (0.04s)
suite.go:164: --- FAIL: TestGatewayEvents/should_send_events_via_Event_Service (0.04s)
```

I suspect the reason being that with Knative, we only accept events when there is a subscription created (for consuming events) and in this test, it might be the case that event is sent without any consumer.

installation/cmd/run.ps1
```diff
@@ -10,14 +9,9 @@ $CURRENT_DIR = Split-Path $MyInvocation.MyCommand.Path
 $SCRIPTS_DIR = "${CURRENT_DIR}\..\scripts"
 $DOMAIN = "kyma.local"
 $INSTALLER_EXTRA_ARGS = ""
-$MINIKUBE_EXTRA_ARGS = ""
+$MINIKUBE_EXTRA_ARGS = "-memory 10240 -disk_size 30g"
```
sjanota: I'd simplify it and add those args directly to `minikube start` [here](https://github.com/kyma-project/kyma/blob/master/installation/scripts/minikube.ps1#L42).

installation/cmd/run.sh
```diff
@@ -58,13 +54,7 @@ do
...
 set -- "${POSITIONAL[@]}" # restore positional parameters
 
-MINIKUBE_EXTRA_ARGS=""
-
-if [ -n "$KNATIVE" ]; then
-
-    MINIKUBE_EXTRA_ARGS="${MINIKUBE_EXTRA_ARGS} --memory 10240 --disk-size 30g"
-
-fi
+MINIKUBE_EXTRA_ARGS="${MINIKUBE_EXTRA_ARGS} --memory 10240 --disk-size 30g"
```
sjanota: Same as in ps version: you can move those flags to `minikube.sh`

installation/scripts/installer.ps1
```diff
@@ -1,7 +1,6 @@
 param (
     [string]$CR_PATH = $null,
     [switch]$LOCAL = $false,
```
radufa: Remove "," from
```
[switch]$LOCAL = $false,
```

abbi-gaurav:
> I see the following tests failing while making knative as default.
> 
> ```shell
> suite.go:164: --- FAIL: TestGatewayEvents (0.04s)
> suite.go:164: --- FAIL: TestGatewayEvents/should_send_events_via_Event_Service (0.04s)
> ```
> 
> I suspect the reason being that with Knative, we only accept events when there is a subscription created (for consuming events) and in this test, it might be the case that event is sent without any consumer.

The issue with tests will be fixed by #3251

installation/cmd/run.ps1
```diff
@@ -10,16 +9,10 @@ $CURRENT_DIR = Split-Path $MyInvocation.MyCommand.Path
 $SCRIPTS_DIR = "${CURRENT_DIR}\..\scripts"
 $DOMAIN = "kyma.local"
 $INSTALLER_EXTRA_ARGS = ""
```
sjanota:
```suggestion
```

sjanota: Also remove it from call to `installer.ps1` on line 34

abbi-gaurav:
> I'm just waiting for the tests to pass. Other then that LGTM

Thanks, @sjanota, I applied the suggestion. Currently, the logging tests are failing. There is a ticket raised to fix that #3435

abbi-gaurav:
> LGTM from the ServicCatalog perspective
> 
> You removed
> 
> ```
>  -memory 10240 -disk_size 30g
> ```
> and right now Kyma will have enabled knative by default. What about the memory and disk size? our installation will not need higher resource definition?
> 
> Should you consider to add flag for disabling knative? Installing it for local development sometimes could be not needed (plus I do not want to allocate higher memory and disk size if I'm not using such functionality)

@mszostok, going further, kyma eventing will only be built on top of knative eventing. So we do not have the option to optionally enable or disable knative. I removed this after verifying that the current resource allocation was working fine. The reason for `10g` memory  is unknown to me as Knative minikube installlation also uses `8g`

installation/cmd/run.ps1
```diff
@@ -1,25 +1,17 @@
...
 $CREATE_CR_EXTRA_ARGS = ""
 
-if ($KNATIVE -eq $true) {
-    $MINIKUBE_EXTRA_ARGS = "${MINIKUBE_EXTRA_ARGS} -memory 10240 -disk_size 30g"
-    $INSTALLER_EXTRA_ARGS = "${INSTALLER_EXTRA_ARGS} -knative"
-}
-
 if ($SKIP_MINIKUBE_START -eq $false) {
-    Invoke-Expression -Command "${SCRIPTS_DIR}\minikube.ps1 -vm_driver ${VM_DRIVER} -domain ${DOMAIN} ${MINIKUBE_EXTRA_ARGS}"
+    Invoke-Expression -Command "${SCRIPTS_DIR}\minikube.ps1 -vm_driver ${VM_DRIVER} -domain ${DOMAIN}
```
radufa: Missing " at the end

abbi-gaurav: I will create a second PR for this. To unblock other tasks, I will merge it once the tests succeed.

Issue #3270, opened 2019-03-20
Add support for webhooks to CMS


michal-hudy:
**Description**

The current version of CMS controller doesn't support asset webhooks. It should be a generic solution that will simply map asset type to proper webhooks. Currently, in the Asset Store we have 2 types of webhooks - validation and mutation - and also we are working on a new type for extracting metadata from assets. The third (for metadata) is important, because it is needed for migration of UI's to CMS.

While working on this task please also change the spec of DocsTopic to:
```yaml 
displayName: Sample Topic
description: Sample description
sources:
- type: assyncapi # enum with supported types
  url: https://example.com/package.zip
  mode: package
  filter: *.md
```

Also, adjust labels to match: https://github.com/kyma-project/console/issues/666#issuecomment-472363697

AC:
 - The generic solution that will map asset type to webhooks
 - Mapping shouldn't be hardcoded, it should be configurable
 - Assets are created together with proper webhooks
 - Assets are created with proper labels (defined in kyma-project/console#666)
 - Spec of DocsTopic is adjusted to the new format

Please sync with @akgalwas while changing spec of DocsTopic, because he is working on integration with CMS and Asset Store in Application Connector.

**Reasons**

To make migration of UI's to CMS possible we need to add missing functionality to CMS and Asset Store - extracting metadata from files. Without it, the migration will be complicated.

derberg:
All good, one change, `type` should have single `string` validation, no need for enum because we will lose some flexibility that we gain with webhooks configuration through configMap

derberg:
looks like we need to add `name` as a required filed for the source element because we will now allow multiple sources of the same type (for example markdown) we need a clear way to identify them and make it aslo nice for user to query for assets of a given name. So remember also to put an additional label  on the created Asset resource

PR #3410, opened 2019-03-28
Add support for handling multiple assets having same source type in docsTopic - image bump

m00g3n:
**Description**

- Add support for handling multiple assets having same source type in docsTopic

**Related issue(s)**

See also #3270


michal-hudy:
Please clean up this PR. It should contain only new CRD definitions and bump of images.

PR #3419, opened 2019-03-28
Make dex static users login available with stop and restart Kyma without reinstalling scenario

PK85:
**Description**

- After minikube stop/start dex works as expected
- ``` 
   kubectl -n kyma-system logs $(kubectl get pod -l app=dex -n kyma-system -o jsonpath= . 
   {.items..metadata.name}) dex-users-configurator
   ```
  will say:
   ```
   error:
   The connection to the server 10.96.0.1:443 was refused - did you specify the right host or port?
   api-server not available via kubectl - waiting 5s...
   api-server available via kubectl
   successfully created user: admin
   successfully created user: developer-user
   successfully created user: no-rights-user
   successfully created user: read-only-user
   ```
   or
   ```
   error: the server doesn't have a resource type "secrets"  api-server not available via kubectl - waiting 
   5s...
   api-server available via kubectl
   successfully created user: admin
   successfully created user: developer-user
   successfully created user: no-rights-user
   successfully created user: read-only-user
   ```

### DIFF
--- a/tools/static-users-generator/scripts/config_replace.sh
+++ b/tools/static-users-generator/scripts/config_replace.sh
@@ -12,6 +12,19 @@ PLACEHOLDER="#__STATIC_PASSWORDS__"
 
 NEWLINE="%"
 
+# check if kubectl works as expected. Minikube stop and then minikube start cases temporary unavailability of the api-server)
+while :
+do
+  if [[ $(kubectl get secrets -l dex-user-config=true --all-namespaces -o json) ]]
+    then
+      echo "api-server available via kubectl"
+      break
+    else
+      echo "api-server not available via kubectl - waiting 5s..."
+      sleep 5
+    fi
+done
+
 NUM=0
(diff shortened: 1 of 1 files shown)


tools/static-users-generator/scripts/config_replace.sh
```diff
@@ -12,6 +12,19 @@ PLACEHOLDER="#__STATIC_PASSWORDS__"
...
+# check if kubectl works as expected. Minikube stop and then minikube start cases temporary unavailability of the api-server)
+while :
+do
+  if [[ $(kubectl get secrets -l dex-user-config=true --all-namespaces -o json) ]]
+    then
+      echo "api-server available via kubectl"
+      break
+    else
+      echo "api-server not available via kubectl - waiting 5s..."
+      sleep 5
```
sjanota: Could you limit the number of retries, so the pod crashes in case of more serious errors?

Issue #2950, opened 2019-03-04
E2E test for upgrading functions


joek:
**Description**
We have to validate if upgrades are working as expected and do not introduce breaking changes.

**Reasons**
Because it is important

AC:
- Create function
- Check if function is still running as expected after upgrade

PR #3426, opened 2019-03-29
Upgrade function test

venturasr:
- Add upgrade test for lambda functions

Related issue:
#2950


tests/end-to-end/upgrade/pkg/tests/function/function.go
```diff
@@ -0,0 +1,188 @@
...
+}
+
+func (f *FunctionUpgradeTest) CreateResources(stop <-chan struct{}, log logrus.FieldLogger, namespace string) error {
+	log.Println("FunctionUpgradeTest creating resources")
+	_, err := f.createFunction(log, namespace)
+	if err != nil {
+		return err
+	}
+
+	<-stop
```
piotrmiskiewicz: the stop channel is used to stop the long process, for example creating function process is being executed for 10 minutes and the framework decides to stop it, so do not read from the channel (the same for TestResources

tests/end-to-end/upgrade/pkg/tests/function/function.go
```diff
@@ -0,0 +1,188 @@
...
+)
+
+type FunctionUpgradeTest struct {
+	functionName, uuid string
+	kubelessClient     *kubeless.Clientset
+	coreClient         *kubernetes.Clientset
+}
+
+func NewFunctionUpgradeTest() (*FunctionUpgradeTest) {
+	kubeconfig := os.Getenv("KUBECONFIG")
```
piotrmiskiewicz: kubeconfig is creasted in the main.go. You can follow the patter which is used there and pass to the constructor proper clientset (interface)

tests/end-to-end/upgrade/pkg/tests/function/function.go
```diff
@@ -0,0 +1,188 @@
...
+}
+
+func (f *FunctionUpgradeTest) getFunctionPodStatus(namespace string, waitmax time.Duration) error {
+
+	timeout := time.After(waitmax)
+	tick := time.Tick(2 * time.Second)
+	for {
+		select {
+		case <-timeout:
+			//log.Println("timeout")
```
piotrmiskiewicz: uncomment or remove, do not keep such lines

tests/end-to-end/upgrade/pkg/tests/function/function.go
```diff
@@ -0,0 +1,176 @@
...
+		case <-tick:
+			pods, err := f.coreClient.CoreV1().Pods(namespace).List(metav1.ListOptions{LabelSelector: "function=" + f.functionName})
+			if err != nil {
+				return err
+			}
+			if len(pods.Items) == 0 {
+				break
+			}
+
+			if len(pods.Items) > 1 {
```
polskikiel: I got an error: `Deployed 1 pod, got 2` here.
Maybe we shouldn't throw error here?

tests/end-to-end/upgrade/pkg/tests/function/function.go
```diff
@@ -0,0 +1,176 @@
...
+func (f *FunctionUpgradeTest) TestResources(stop <-chan struct{}, log logrus.FieldLogger, namespace string) error {
+	log.Println("FunctionUpgradeTest testing resources")
+	err := f.getFunctionPodStatus(namespace, 10*time.Minute)
+	if err != nil {
+		return err
+	}
+
+	host := fmt.Sprintf("http://%s.%s:8080", namespace, namespace)
+
+	value, err := f.getFunctionOutput(host, 2*time.Minute)
```
polskikiel: I couldn't pass this step on cluster with error: `Could not get function output:\n Post http://hello.functionupgradetest:8080`

tests/end-to-end/upgrade/pkg/tests/function/function.go
```diff
@@ -0,0 +1,188 @@
...
+	tick := time.Tick(2 * time.Second)
+	for {
+		select {
+		case <-timeout:
+			//log.Println("timeout")
+			pods, err := f.coreClient.CoreV1().Pods(namespace).List(metav1.ListOptions{LabelSelector: "function=" + f.functionName})
+			if err != nil {
+				return err
+			}
+			return fmt.Errorf("Pod did not start within given time  %v: %+v", waitmax, pods)
```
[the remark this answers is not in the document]
venturasr: Changed to lowercase

tests/end-to-end/upgrade/pkg/tests/function/function.go
```diff
@@ -0,0 +1,277 @@
...
+}
+
+func int32Ptr(i int32) *int32 { return &i }
+
+// NewLambdaFunctionUpgradeTest returns new instance of the FunctionUpgradeTest
+func NewLambdaFunctionUpgradeTest(config *restclient.Config) *LambdaFunctionUpgradeTest {
+
+	kubelessClient, err := kubeless.NewForConfig(config)
+	if err != nil {
+		return &LambdaFunctionUpgradeTest{}
```
k15r: this hides an error, is it really what we want?

tests/end-to-end/upgrade/pkg/tests/function/function.go
```diff
@@ -0,0 +1,277 @@
...
+	}
+
+	coreClient, err := kubernetes.NewForConfig(config)
+	if err != nil {
+		return &LambdaFunctionUpgradeTest{}
```
k15r: see above

tests/end-to-end/upgrade/pkg/tests/function/function.go
```diff
@@ -0,0 +1,277 @@
...
+	}
+
+	apiClient, err := kyma.NewForConfig(config)
+	if err != nil {
+		return &LambdaFunctionUpgradeTest{}
```
k15r: once again

tests/end-to-end/upgrade/pkg/tests/function/function.go
```diff
@@ -0,0 +1,277 @@
...
+		functionName:   functionName,
+		uuid:           uuid.New().String(),
+		nSpace:         nSpace,
+		hostName:       hostName,
+		apiClient:      apiClient,
+	}
+}
+
+// CreateResources creates resources needed for e2e upgrade test
+func (f *LambdaFunctionUpgradeTest) CreateResources(stop <-chan struct{}, log logrus.FieldLogger, namespace string) error {
```
k15r: due to the way the upgrade tests are executed this `CreateResources` is the only function that will be executed before the upgrade attempt. imho we should check that these newly created resources are valid (we could just call the TestResources function here as well)

venturasr: As I understand the upgrade happens and then the resources are created and tested.
The way how the backup test works is better because we expect the same resource before and after.

polskikiel: You can add here explicit checks for resources which can be not in a ready state and just return an error if they weren't. 
Resources are created before upgrade and tested after - it's described in the readme `usage` section. You can also run this test without doing an upgrade to check if your test case configuration is correct.

tests/end-to-end/upgrade/pkg/tests/function/function.go
```diff
@@ -0,0 +1,275 @@
...
+	}
+
+	err = f.createAPI()
+	if err != nil {
+		log.Printf("create api %v", err)
+		return err
+	}
+
+	// Ensure resources works
+	f.TestResources(stop, log, namespace)
```
polskikiel: err = f.TestResources(stop, log, namespace)

tests/end-to-end/upgrade/pkg/tests/function/function.go
```diff
@@ -0,0 +1,275 @@
...
+	if err != nil {
+		log.Printf("First call to TestResources() failed %v", err)
```
polskikiel: instead of `log.Printf("First call to TestResources() failed %v", err)`, you can just return `errors.Wrapf(err, "first call to TestResources() failed ")`. Same for other occurences.

tests/end-to-end/upgrade/pkg/tests/function/function.go
```diff
@@ -0,0 +1,275 @@
...
+	functionName, uuid string
+	kubelessClient     kubeless.Interface
+	coreClient         kubernetes.Interface
+	apiClient          kyma.Interface
+	nSpace             string
+	hostName           string
+	stop               <-chan struct{}
+}
+
+func int32Ptr(i int32) *int32 { return &i }
```
polskikiel: you can move that function to the bottom of this file

tests/end-to-end/upgrade/pkg/tests/function/function.go
```diff
@@ -0,0 +1,275 @@
...
+	err := f.getFunctionPodStatus(10 * time.Minute)
+	if err != nil {
+		return err
+	}
+
+	host := fmt.Sprintf("https://%s", f.hostName)
+
+	value, err := f.getFunctionOutput(host, 1*time.Minute, log)
+	if err != nil {
+		return err
```
polskikiel: return errors.Wrap(err, "context") to add context to the error and improve error tracking. same for others

PR #3445, opened 2019-03-29
Add TLS filter for helm broker repositories

jasiu001:
**Introduction**
A k8s config map can be used to configure Helm Broker to use a external Bundles Repository. This is configured by means of an http/https url.

**Description**
This PR protects remote cluster against using unsecure url. For remote cluster only https schema will be allowed. For local cluster each schema is permitted.


docs/helm-broker/docs/05-04-configure-hb.md
```diff
@@ -14,6 +14,11 @@ By default, the Helm Broker fetches bundles listed in the `index.yaml` file from
     kubectl create configmap my-helm-repos-urls -n kyma-system --from-literal=URLs=https://github.com/kyma-project/bundles/releases/download/0.3.0/index-testing.yaml
     ```
     >**NOTE:** If you want to fetch bundles from many HTTP servers, use `\n` to separate the URLs.
+    
+    >**CAUTION:** On remote cluster you can use only servers with enabled TLS.
```
klaudiagrz: As discussed, remove both CAUTIONS from this document, and create a separate section from them. My suggestion is:

```
## Configuration rules

These are the rules you must follow when you configure the Helm Broker. Otherwise, the Helm Broker will not display your bundles.
* If you use your bundle in two different repositories simultaneously, the Helm Broker detects a conflict and does not display this bundle at all. You can see the details of the conflict in the Helm Broker application logs. If you need a given bundle in two or more repositories, do not use them at the same time.
* On your non-local clusters, you can use only servers with TLS enabled. All incorrect or unsecured URLs will be omitted. Find the information about the rejected URLs in the Helm Broker logs. You can use unsecured URLs only on your local cluster. To use URLs without TLS enabled, set the **APP_DEVELOP_MODE** environment variable in the [deploy.yaml](https://github.com/kyma-project/kyma/blob/master/resources/helm-broker/templates/deploy.yaml) file to `true`. 
```

PR #3457, opened 2019-03-30
Add upgrade tests for cms and assetstore domain

### LINKED ISSUE #3328: Test upgradability of AssetStore and Headless CMS
**Description**

- reuse tests that we have for backup testing
- make sure we have proper tests integrated with upgradability prow job

**Reasons**

Make sure these components are upgradable, on each PR

magicmatatjahu:
**Description**

- Add upgrade tests for cms and assetstore domains

**Related issue(s)**
See also #3328


tests/end-to-end/upgrade/pkg/tests/asset-store/asset.go
```diff
@@ -0,0 +1,129 @@
...
+
+type asset struct {
+	resCli      *resource.Resource
+	name        string
+	namespace   string
+}
+
+func newAssetClient(dynamicCli dynamic.Interface, namespace string) *asset {
+	return &asset{
+		resCli: resource.New(dynamicCli, schema.GroupVersionResource{
```
mszostok: `resource.New` it's not saying that you creating an abstraction over the dynamicCli. It's more like "create new resource".

Please change that to be more verbose, e.g. `"k8s.io/client-go/dynamic"` says that it's Go Dynamic client for k8s platform. Your `/pkg/resource/` says what?

What's more, I'm not convinced to have this pkg at all. For example: 

`Get(name string) (*unstructured.Unstructured, error)` you have the same interface as dynamic client but you only hiding the not found error. This solution is not generic and shouldn't be placed directly in pkg. Such assertion should belong to the test case because in tests you can also need information that `Get` returned not found to ensure that obj is not available after deletion (e.g. finalizers were removed so obj was release from etcd)

`Delete(name string) error ` you have the same interface as dynamic client but you only using error wrap and creating new error in case of not found. K8s not found error already have all this information and its message is sufficiently readable - I do not see any reason for changing that. And error.Wrap should be used to add context for error but in place where it's executed in tests - without that, you will not find the proper line in code.

Issue #2896, opened 2019-02-27
Cleanup kyma RBAC roles 


piotrmsc:
**Description**

We have introduced kyma developer role in https://github.com/kyma-project/kyma/issues/2557.
However, we still need to clean up a few things and improve our documentation.

What needs to be done :
- namespace-controller does not create kyma-admin and kyma-reader roles in the namespace like it does [here](https://github.com/kyma-project/kyma/blob/master/components/namespace-controller/internal/controller/controller.go#L113)
- check if apigroups can contain a wildcard in the names like "*.kyma-project.io" if so, then replace it for those roles - checked, cannot be simplified with wildcard.
- update our documentation - create a matrix in the security overview section, describing which roles are present in the kyma, what privileges they provide. Additionally, delete section regarding **kyma-admin** and **kyma-reader** roles being created in the namespaces by the NS controller.
- apply the aggregate rule for cluster kyma-admin and kyma-reader if possible

**Reasons**
Namespace controller creates roles which are not needed anymore, admin will be able to create bindings in the console directly like with kyma-developer. Moreover, created roles by the NS controller give too high privileges -  does not reflect kyma-admin and viewer cluster roles.
Documentation will fully cover authorization roles used in kyma, currently it is mixed.

**Attachments**

kazydek:
We need to have one place where all Kyma access roles as collected and well-described.

Currently, some of them are here: https://kyma-project.io/docs/master/components/security#tutorials-manage-static-users-in-dex-bind-a-user-to-a-role-or-a-clusterrole
And some of them are here: https://kyma-project.io/docs/master/components/security/#overview-overview

I leave that up to you whether you prefer to move them to the overview doc or create a separate doc for that (but would definitely opt for that ;) ). Preferably that should be a table with the access matrix describing what permissions the given role provides to the user. It should be also clearly stated if the role is cluster-wide or a namespace-wide.

PR #3473, opened 2019-04-01
Refactor kyma roles

Tomasz-Smelcerz-SAP:
**Description**

Refactor three main kyma cluster roles (`kyma-admin`, `kyma-view`, `kyma-edit`) so that each role only aggregates other roles.

Purpose: improving maintainability of roles.
Note: This PR should not introduce any new permissions nor modify existing ones, it's just refactoring.

**Related issue(s)**
See also #2896


Demonsthere:
Good job, but as you expected the number of roles exploded, which makes it a lot more difficult to manage. How about we split the definition into multiple files, separated by ApiGroups (istio, k8s, kyma) to reduce the size of a single file?

resources/core/charts/cluster-users/templates/clusterroles-k8s-io.yaml
```diff
@@ -0,0 +1,128 @@
...
+    - "list"
+    - "patch"
+    - "update"
+    - "watch"
+
+---
+apiVersion: rbac.authorization.k8s.io/v1beta1
+kind: ClusterRole
+metadata:
+  name: kyma-crd-list
```
Demonsthere: This might me more useable as `kyma-crd-view`, with `list, get` verbs. List alone seems lacking

resources/core/charts/cluster-users/templates/clusterroles-kyma-project-io.yaml
```diff
@@ -0,0 +1,88 @@
...
+  name: kyma-view-kyma
+  labels:
+    app: kyma
+    chart: {{ .Chart.Name }}-{{ .Chart.Version }}
+    rbac.authorization.kyma-project.io/aggregate-to-kyma-view-kyma: "true"
+  annotations:
+    "helm.sh/hook-weight": "0"
+rules:
+- apiGroups:
+    - "applicationconnector.kyma-project.io"
```
Demonsthere: How about set all kyma apiGroups as a helm variable? This will allow to modify the list in one place when a new one is introduced

resources/core/charts/cluster-users/templates/clusterroles-k8s-io.yaml
Tomasz-Smelcerz-SAP: I don't want to change existing setup - this is only refactoring PR. Changes/fixes/improvements will be introduced in a separate PR.

resources/core/charts/cluster-users/templates/clusterroles-kyma-project-io.yaml
Tomasz-Smelcerz-SAP: Good point - I'll extract common groups to a helm value.

Issue #3031, opened 2019-03-07
Verify pre and post upgrade state of API gateway


piotrmsc:
**Description**

Part of https://github.com/kyma-project/test-infra/issues/582
Write an e2e test ensuring api-controller works before and after upgrade.

Test scenario:
- [pre-upgrade] expose lambda with security enabled 
- [post-upgrade] make a call with token to lambda. 
- [post-upgrade] create and expose new lambda and make a call to it to verify api-controller works after upgrade

**Reasons**

We have to ensure component is running smoothly before and after upgrade to new version during kyma upgrade.

**Attachments**

PR #3478, opened 2019-04-01
Implement API upgrade test

sjanota:
**Description**

- Implement API upgrade test

**Related issue(s)**
Issue: #3031 
Depends on: #3476


tests/end-to-end/upgrade/main.go
```diff
@@ -99,6 +114,18 @@ func main() {
 	mfCli, err := mfClient.NewForConfig(k8sConfig)
 	fatalOnError(err, "while creating Microfrontends clientset")
 
+	dexConfig := dex.Config{}
```
jasiu001: The `main.go` file should be responsible for register all test and run them. `dexConfig` is individual part for specific test and it should be sets inside concrete test. If everyone will add configuration of their test inside the main file it will be unreadable.

sjanota: No it is not. Main is responsible also for wiring dependency injection and dex.Config is somehting that will be used (or at least should) across many tests.

Issue #3434, opened 2019-03-29
Application operator test failing on GKE postsubmit jobs


jesusreal:
**Description**
Application operator tests are failing for gke postsubmit jobs with the following error:
```
Testing 'application-operator-tests'
'application-operator-tests' has Failed status
Fetching logs from 'application-operator-tests' with options tailLimit=2000 and bytesLimit=500000
Application Operator tests
Run controller tests
=== RUN TestApplicationOperator
2019/03/29 09:31:27 Read configuration: {Namespace:kyma-integration TillerHost:tiller-deploy.kube-system.svc.cluster.local:44134 InstallationTimeoutSeconds:180}
=== RUN TestApplicationOperator/Application_Operator_-_Application_lifecycle_test
--- FAIL: TestApplicationOperator (33.34s)
--- FAIL: TestApplicationOperator/Application_Operator_-_Application_lifecycle_test (33.31s)
controller_test.go:13: Creating Application without access label
controller_test.go:16: Waiting for Helm release to install
controller_test.go:19: Checking if k8s resource deployed
controller_test.go:22: Checking access label
controller_test.go:25: Deleting Application
controller_test.go:28: Waiting for Helm release to delete
controller_test.go:31: Checking if k8s resources removed
require.go:248:
Error Trace:	suite.go:115
suite.go:103
k8sresources_check.go:69
suite.go:103
controller_test.go:32
Error: An error is expected but got nil.
```

**Expected result**
Application operator tests run successfully

**Actual result**
There is a failing test for application operator

**Steps to reproduce**

**Troubleshooting**

Prow logs:
https://status.build.kyma-project.io/view/gcs/kyma-prow-logs/logs/post-master-kyma-gke-integration/1111546535927091200
https://status.build.kyma-project.io/view/gcs/kyma-prow-logs/logs/post-master-kyma-gke-central-connector/1111546535927091201
https://status.build.kyma-project.io/view/gcs/kyma-prow-logs/logs/post-master-kyma-gke-upgrade/1111546535931285504

sayanh:
```
Fetching logs from 'application-operator-tests' with options tailLimit=2000 and bytesLimit=500000
Application Operator tests
Run controller tests
=== RUN TestApplicationOperator
2019/04/01 09:11:57 Read configuration: {Namespace:kyma-integration TillerHost:tiller-deploy.kube-system.svc.cluster.local:44134 InstallationTimeoutSeconds:180}
[...]

Issue #3430, opened 2019-03-29
Unstable Application Operator tests


franpog859:
**Description**

Prow Job `pre-master-kyma-integration` with number `1111547668477251585` failed on [this pull request](https://github.com/kyma-project/kyma/pull/3332).
`application-operator-tests` has failed.

Please find out if this is a bug or some external problem.

There are the logs I found:

```
ERROR: (gcloud.compute.scp) [/usr/bin/scp] exited with return code [1].
...
418 | - Testing Application Connector
419 | RUNNING: application-operator-tests
420 | FAILED: application-operator-tests, run `kubectl logs application-operator-tests --namespace kyma-integration` for more info
423 | Error: 1 test(s) failed
424 | All test pods should be terminated. Checking...
425 | Flag --show-all has been deprecated, will be removed in an upcoming release
426 | OK
427 | Flag --show-all has been deprecated, will be removed in an upcoming release
428 | Testing 'application-operator-tests'
429 | 'application-operator-tests' has Failed status
430 | Fetching logs from 'application-operator-tests' with options tailLimit=2000 and bytesLimit=500000
431 | Application Operator tests
432 | Run controller tests
433 | 2019/03/29 09:08:15 Read configuration: {Namespace:kyma-integration TillerHost:tiller-deploy.kube-system.svc.cluster.local:44134 InstallationTimeoutSeconds:180}
...
456 | === RUN TestApplicationRelease
457 | 2019/03/29 09:09:07 Read configuration: {Namespace:kyma-integration TillerHost:tiller-deploy.kube-system.svc.cluster.local:44134 InstallationTimeoutSeconds:180}
458 | === RUN TestApplicationRelease/Application_tests_should_succeed
459 | --- FAIL: TestApplicationRelease (162.64s)
460 | suite.go:70: Creating operator-app-test-ltxv Application
461 | --- FAIL: TestApplicationRelease/Application_tests_should_succeed (162.62s)
462 | application_test.go:16: Waiting for application to be deployed...
463 | application_test.go:19: Running Application test...
464 | suite.go:98: Running application tests
465 | suite.go:113: msg:"RUNNING: test-operator-app-test-ltxv-event-service-acceptance" status:RUNNING
466 | suite.go:113: msg:"PASSED: test-operator-app-test-ltxv-event-service-acceptance" status:SUCCESS
467 | suite.go:113: msg:"RUNNING: operator-app-test-ltxv-application-gateway-tests" status:RUNNING
[...]

PR #3507, opened 2019-04-02
Improve Application Operator Tests

Szymongib:
**Description**

- Increase CPU of Istio Pilot on the local installation
- Improve tests of Application Operator

**Related issue(s)**
#3430 #3434

### DIFF
--- a/tests/application-gateway-tests/test/executor/proxy/suite.go
+++ b/tests/application-gateway-tests/test/executor/proxy/suite.go
@@ -22,14 +22,15 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/kubernetes"
 	restclient "k8s.io/client-go/rest"
+	"k8s.io/client-go/util/retry"
 )
 
 const (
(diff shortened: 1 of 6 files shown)


tests/application-gateway-tests/test/executor/proxy/suite.go
```diff
@@ -215,6 +215,10 @@ func (ts *TestSuite) AddDenierLabel(t *testing.T, apiId string) {
 
 	pod.Labels[serviceName] = "true"
 
-	_, err = ts.podClient.Update(pod)
+	err = retry.RetryOnConflict(retry.DefaultBackoff, func() error {
```
aszecowka: I am not sure if this is correct usage of RetryOnConflict, please have a look here:
https://github.com/pingcap/tidb-operator/issues/81
Generally, we should fetch pod one more time, and again apply our changes.

Szymongib: Good point! I do not think it impacts current implementation but we should definitely correct it in the future.

Issue #2237, opened 2019-01-10
Integrate basic log ui to console


a-thaler:
We want a nice UI integrated to the console to reduce the barrier of debuggability. It should be a really quick thing to see recent logs of a pod.

AC:
- Deployment
   - new UI component will be located in console repository
   - UI component will be bundled with the logging component
- console Integration
   - console handles nicely situation where logging component is not installed
   - Navigation has a new node to show log view for the namespace 
- log frame
   - pre-filters namespace/pod
   - shows latest log or logs recent time frames
   - supports filtering by regexp

jesusreal:
PRs in a nutshell:
Kyma: https://github.com/kyma-project/kyma/pull/2514
Console: https://github.com/kyma-project/console/pull/543
Test-infra (I): https://github.com/kyma-project/test-infra/pull/587
Test-infra (after folder renamings):

pekura:
Console code merged to master.

a-thaler:
The UI will be ready and integrated for 0.8 release but not deployed by default in consequence of the blocking bug.

jesusreal:
Remaining action items:
- Enable logging microfrontend once blocking ticket is merged
- Add logging UI tests to UI tests execution in https://github.com/kyma-project/console/blob/master/tests/package.json

a-thaler:
Looks nice, thank you :)

Issue #3396, opened 2019-03-28
Support groups claims in dex static connector


piotrmsc:
**Description**
Currently, our RBAC roles are bound to static users via email, which is fine but tedious. In order to simulate production use case and enable proper e2e tests we need `gorup` claim in JWT.

- We have to check if the latest version of dex support it, if so, then upgrade dex in kyma.
- If it still does not support it (previously They did not want that in the static connector) then we need to fork it and add our implementation. Fork should be placed in kyma-incubator (if not already done in #2203). 
- Add "kyma-admins" group to admin user in `dex-users-secrets.yaml` and "kyma-developers" group to developer static user.
- Remove the clusterrolebinding  to "admin@kyma.cx" user and replace it with binding to group "kyma-admins". Note we do **not** create a binding for cluster role "kyma-developer"  here. This is done by the admin to a specific namespace by role binding in the console.

You can use our old implementation as a reference from here https://github.com/dexidp/dex/issues/1080

**Reasons**

Currently, Dex version used in Kyma does not support groups in JWT claims in static connector. With groups claim in JWT we can easily test integration scenarios for example in apiserver-proxy simulating production use of roles assigned to groups instead of user emails. 

**Attachments**

piotrmsc:
We will not implement this, currently we are aiming to get rid of our fork of dex. Ticket should be reopened if we decide that we a) still need that b) we will maintain the fork

PR #3509, opened 2019-04-02
Enable console logging ui with authorization

hisarbalik:
https://github.com/kyma-project/kyma/issues/2237


resources/logging/templates/loki/istio-rbac.yaml
```diff
@@ -29,23 +29,8 @@ spec:
   subjects:
   - user: "cluster.local/ns/kyma-system/sa/logging"
   - properties:
-      request.auth.claims[groups]: "kyma-admins"
+      request.auth.audiences: "kyma-client"
```
piotrmsc: Please a "follow up" to enable groups here once #3396 is done

PR #3544, opened 2019-04-03
Remove section to install knative by default

abbi-gaurav:
See also #3017


bszwarc:
Within this PR, could you please remove the link from https://github.com/kyma-project/kyma/blob/master/docs/kyma/docs/04-01-overview.md pointing to this installation guide?

PR #3565, opened 2019-04-04
Enable logging installer

hisarbalik:


sayanh:
Why are we not enabling logging tests here? 
https://github.com/kyma-project/kyma/blob/96514373713b045fadf64d71033e66ae31fbe1f9/installation/scripts/testing.sh#L62

hisarbalik:
#3652

Issue #3491, opened 2019-04-01
Add Kyma domain metadata to the Knative channel


abbi-gaurav:
**Description**
To enable easy look up and search, the Kyma domain metadata:

Source id, event type and event type version 

should be added as labels to the channel custom resource.

Prefix `kyma-` in front of label name.

> Note: Truncate the values if longer than 63 characters. This is maxed value length permitted by Kubernetes.

**AC**
- [ ] Kyma domain metadata (Source id, event type, and event type version) is added to the channel custom resource as separate labels.

abbi-gaurav:
@montaro, please link the second PR to update the event bus

PR #3587, opened 2019-04-04
Update event-bus image

montaro:
See also #3491

### DIFF
--- a/resources/event-bus/values.yaml
+++ b/resources/event-bus/values.yaml
@@ -7,7 +7,7 @@ global:
     path: eu.gcr.io/kyma-project
   event_bus:
     dir: develop/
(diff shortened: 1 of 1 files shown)


resources/event-bus/values.yaml
```diff
@@ -7,7 +7,7 @@ global:
     path: eu.gcr.io/kyma-project
   event_bus:
     dir: develop/
-    version: c157195c
+    version: a4acd811
```
marcobebway: This version does not include the changes done for https://github.com/kyma-project/kyma/issues/3491.

montaro:
Updated to the latest version after merging https://github.com/kyma-project/kyma/pull/3592

Issue #3460, opened 2019-04-01
Enhance Connector Service API to make certificate rotation easier.


akgalwas:
**Description**

In order to make the implementation of certificate renewal easier Connector Service should provide information for creating CSR (subject fields and algorithm) in management/info endpoint.

**Reasons**

Currently information for creating CSR is available in signingRequests/info. The external application needs to store this data during the initial connection configuration to use it later. Adding the mentioned data to management/info would make the renewal process more convenient.

PR #3591, opened 2019-04-05
Provide certificate info in managementInfo endpoint

franpog859:
**Description**

- Provide certificate info in managementInfo endpoint

**Related issue(s)**

See the issue #3460 
See the bump #3598 
See the documentation [#3600](https://github.com/kyma-project/kyma/pull/3600)

### DIFF
--- a/components/connector-service/internal/externalapi/managementinfohandler_test.go
+++ b/components/connector-service/internal/externalapi/managementinfohandler_test.go
@@ -4,16 +4,15 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
-	"github.com/kyma-project/kyma/components/connector-service/internal/certificates"
 	"io/ioutil"
 	"net/http"
 	"net/http/httptest"
 	"testing"
 
-	"github.com/kyma-project/kyma/components/connector-service/internal/httperrors"
-
 	"github.com/kyma-project/kyma/components/connector-service/internal/apperrors"
+	"github.com/kyma-project/kyma/components/connector-service/internal/certificates"
 	"github.com/kyma-project/kyma/components/connector-service/internal/clientcontext"
+	"github.com/kyma-project/kyma/components/connector-service/internal/httperrors"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 )
@@ -225,7 +224,7 @@ func TestManagementInfoHandler_GetManagementInfo(t *testing.T) {
 
 		commonName := clusterContext.GetCommonName()
 		expectedCertInfo := certInfo{
-			Subject:      fmt.Sprintf("OU=%s,O=%s,L=%s,ST=%s,C=%s,CN=%s", organizationalUnit, organization, locality, province, country, commonName), //TODO: CommonName will be different!
+			Subject:      fmt.Sprintf("OU=%s,O=%s,L=%s,ST=%s,C=%s,CN=%s", organizationalUnit, organization, locality, province, country, commonName),
 			Extensions:   "",
 			KeyAlgorithm: "rsa2048",
 		}
--- a/components/connector-service/internal/externalapi/managementinfohandler_test.go
+++ b/components/connector-service/internal/externalapi/managementinfohandler_test.go
@@ -17,18 +17,23 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
+const (
+	applicationKey = "application"
+	tenantKey      = "tenant"
+	groupKey       = "group"
+
+	tenant = "test-tenant"
+	group  = "test-group"
+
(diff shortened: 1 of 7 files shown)


components/connector-service/internal/externalapi/managementinfohandler_test.go
```diff
@@ -146,4 +155,103 @@ func TestManagementInfoHandler_GetManagementInfo(t *testing.T) {
 		require.NoError(t, err)
 		assert.Equal(t, http.StatusInternalServerError, rr.Code)
 	})
+
+	t.Run("should successfully get certificate info for application", func(t *testing.T) {
```
akgalwas: The new test exercises exactly the same scenario as  "should successfully get management info urls for application", the only difference is in the checks done.I would suggest to enhance the existing test rather than adding a new one.

components/connector-service/internal/externalapi/managementinfohandler_test.go
```diff
@@ -146,4 +155,103 @@ func TestManagementInfoHandler_GetManagementInfo(t *testing.T) {
...
+		assert.Equal(t, http.StatusOK, rr.Code)
+
+		certificateInfo := infoResponse.CertificateInfo
+
+		assert.Equal(t, expectedCertInfo.Subject, certificateInfo.Subject)
+		assert.Equal(t, expectedCertInfo.Extensions, certificateInfo.Extensions)
+		assert.Equal(t, expectedCertInfo.KeyAlgorithm, certificateInfo.KeyAlgorithm)
+	})
+
+	t.Run("should successfully get certificate info for runtime", func(t *testing.T) {
```
akgalwas: The same as above.

components/connector-service/internal/externalapi/managementinfohandler_test.go
```diff
@@ -146,4 +155,103 @@ func TestManagementInfoHandler_GetManagementInfo(t *testing.T) {
...
+				EventsURL:   "https://events.base.path/application/v1/events",
+			},
+		}
+
+		connectorClientExtractor := func(ctx context.Context) (clientcontext.ClientContextService, apperrors.AppError) {
+			return *extApplicationCtx, nil
+		}
+
+		commonName := extApplicationCtx.GetCommonName()
+		expectedCertInfo := certInfo{
```
akgalwas: Declaring a variable with expected values is redundant here since you do assert calls for each field separately.

components/connector-service/internal/externalapi/managementinfohandler_test.go
```diff
@@ -146,4 +155,103 @@ func TestManagementInfoHandler_GetManagementInfo(t *testing.T) {
...
+			Subject:      fmt.Sprintf("OU=%s,O=%s,L=%s,ST=%s,C=%s,CN=%s", organizationalUnit, organization, locality, province, country, commonName),
```
akgalwas: Please, move it to the code where other variables (e.g. application, tenant) are initialized.

tests/connector-service-tests/test/apitests/connector_test.go
```diff
@@ -554,6 +566,33 @@ func certificateRotationSuite(t *testing.T, tokenRequest *http.Request, skipVeri
 		require.Nil(t, errorResponse)
 	})
 
+	t.Run("should renew client certificate using information from management info endpoint", func(t *testing.T) {
```
akgalwas: This code is almost identical to "should renew client certificate using information from management info endpoint". Please consider adding some function to make code shorter.

components/connector-service/internal/externalapi/managementinfohandler_test.go
franpog859: Done!

components/connector-service/internal/externalapi/managementinfohandler_test.go
franpog859: Done!

components/connector-service/internal/externalapi/managementinfohandler_test.go
franpog859: Done!

components/connector-service/internal/externalapi/managementinfohandler_test.go
franpog859: Done!

tests/connector-service-tests/test/apitests/connector_test.go
franpog859: I merged both tests into one. It uses management info response to fetch the certificate data. As far as I know it will be the default flow. Value is checked if it equals the one from csrInfo response. I think this is the cleanest way to perform such a test

PR #3619, opened 2019-04-07
Prefix Knative channel name with a lowercase alphabet character to fix the istio virtual service name issue

marcobebway:
**Description**

- Prefix Knative channel name with a lowercase alphabet character to fix the istio virtual service name issue.


marcobebway:
Yes @abbi-gaurav, the **computeHash** function is using the golang **encoding/base32** concoder.

Issue #3498, opened 2019-04-02
Serving test disabled


sjanota:
**Description**
There is a test which verifies if Knative serving has been installed properly. The problem with it is that knative Service was created in helm chart instead of test. Because of that the test failed when run second time. 

Knative is comming as default in Kyma so the test needs to be fixed and enabled.

**Expected result**
Knative serving installation is tested automatically

**Actual result**
Knative serving is not tested automatically at all

**Steps to reproduce**

**Troubleshooting**
I've created a PR once, but never had time to finish it or verify if it helps: https://github.com/kyma-project/kyma/pull/2373/files

PR #3661, opened 2019-04-09
Enable knative serving test

k15r:
**Description**

- enable knative serving test

**Related issue(s)**

#3498


resources/knative-serving/values.yaml
```diff
@@ -10,8 +10,8 @@ global:
   containerRegistry:
     path: eu.gcr.io/kyma-project
   test_knative_serving_acceptance:
-    dir: develop/
-    version: 4a2b51ec
+    dir: pr/
+    version: PR-3657
```
sayanh: An image with a PR tag should not be merged to master.

k15r: i know, i will update the PR as soon as i have that. I wanted to create a draft PR, but forgot during creation and GH does not allow you to do this afterwards.

but blocking the merge is a good idea, so i cannot forget it. THX 😄

resources/knative-serving/templates/tests/test-serving.yaml
```diff
@@ -1,22 +1,83 @@
...
+---
+{{- if .Capabilities.APIVersions.Has "testing.kyma-project.io/v1alpha1" }}
+apiVersion: "testing.kyma-project.io/v1alpha1"
+kind: TestDefinition
+metadata:
+  name: test-knative-serving-acceptance
+template:
+  metadata:
+    annotations:
+      sidecar.istio.io/inject: "true"
```
aszecowka: sidecar injection rules should be the same in both cases

resources/knative-serving/templates/tests/test-serving.yaml
```diff
@@ -1,22 +1,83 @@
...
+  spec:
```
aszecowka: Containers shoulf be defined under: spec.template.spec, here you have an example:
```
apiVersion: testing.kyma-project.io/v1alpha1
kind: TestDefinition
metadata:
  name: test-sth
spec:
  template:
    spec:
      shareProcessNamespace: true
      restartPolicy: Never
      containers:
        - name: test
          image: alpine:latest
```

resources/knative-serving/templates/tests/test-serving.yaml
k15r: good catch

resources/knative-serving/templates/tests/test-serving.yaml
k15r: ah, yes that make more sense

Issue #3682, opened 2019-04-11
Application-operator should only fetch helm releases that apply to applications


crabtree:
**Description**

Currently when application-operator observes installed applications it fetches all helm releases with all possible statuses. That leads to unnecessary memory consumption and may end up with OOM.

**Expected result**

Application-operator should fetch only helm releases that apply to applications, or at least are installed in kyma-integration namespace. 

**Actual result**

See description.

PR #3686, opened 2019-04-11
Fix application-operator increased memory usage

crabtree:
**Description**

- Specify namespace when listing helm releases
- Do not fetch releases with status SUPERSEDED

**Related issue(s)**

See also #3682

### DIFF
--- a/components/application-operator/pkg/controller/reconciler.go
+++ b/components/application-operator/pkg/controller/reconciler.go
@@ -10,6 +10,7 @@ import (
 	log "github.com/sirupsen/logrus"
 	k8sErrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/runtime"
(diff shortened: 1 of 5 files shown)


components/application-operator/pkg/controller/reconciler.go
```diff
@@ -77,35 +78,35 @@ func (r *applicationReconciler) handleErrorWhileGettingInstance(err error, reque
 	return reconcile.Result{}, logAndError(err, "Error getting %s Application", request.Name)
 }
 
-func (r *applicationReconciler) enforceDesiredState(application *v1alpha1.Application) error {
+func (r *applicationReconciler) enforceDesiredState(application *v1alpha1.Application) (func(application *v1alpha1.Application), error) {
```
Szymongib: I would suggest creating a type alias for this function `func(application *v1alpha1.Application)`

Issue #3652, opened 2019-04-09
Enable logging by default


a-thaler:
**Description**
Currently, the logging solution based on Loki is not enabled by default, even for a cluster installation.
That is reasoned mainly by failing tests caused by the newest kubernetes version provided on gke, shortly before kyma release 0.9.0. 

Investigations revealed that with kubernetes-1.14 the pod log directory is changed, see https://github.com/kubernetes/kubernetes/issues/73503. That change seems to be effective already on GKE cluster with version 1.12.6-gke.X.
Promtail is not supporting that pattern yet and logs will not be collected, see https://github.com/grafana/loki/issues/393.

**Goal**
- Enable logging module for clusters by default (#3565)
- Have a way to support new log folder pattern of Kubernetes 1.14 by a config change (https://github.com/kyma-project/kyma/pull/3565)
- Still support old pattern as default as long as Kyma is not supporting kubernetes 1.14
- Update documentation about installation on different k8s versions (https://github.com/kyma-project/kyma/pull/3685)
- Enable tests and verify that they work (https://github.com/kyma-project/kyma/pull/3705)

hisarbalik:
#3565 #3685 #3705

hisarbalik:
#3565

a-thaler:
Enablement for new log pattern solved by installer configMap attribute which needs to be set explicitly, see #3565
Logging and tests enabled and docu updated.

PR #3705, opened 2019-04-12
Enable logging helm chart test

hisarbalik:
Logging helm chart test enabled after updating test with latest helm chart changes


hisarbalik:
#3652

Issue #3300, opened 2019-03-22
application-gateway 502 gateway error on retries with POST payload.


Tomasz-Smelcerz-SAP:
**Description**
application-gateway incorrectly processes retries with request body payload.
Instead of returning actual status code and response body of the _second_ call (retry) to the end-user, it returns 502 status code without body.

**Steps to reproduce**
To reproduce the bug, use an app that always returns 403 status code along with some dummy response body.
The response body is not related to the bug, but it clearly shows the difference in behavior.

Deploy the following:

```
apiVersion: extensions/v1beta1
kind: Deployment
metadata:
  name: app-403-with-body
spec:
  template:
    metadata:
      labels:
        name: app-403-with-body
        app: app-403-with-body
    spec:
      containers:
      - name: app-403-with-body
        image: node:10.15.3-alpine
        imagePullPolicy: IfNotPresent
        command: ["node"]
        args:
          - "-e"
          - |
            const http = require('http');
            const port = 8080;

            const requestHandler = (request, response) => {
              console.log(request.url);
              response.statusCode = 403;
              response.end('There\'s an error!');
            }

            const server = http.createServer(requestHandler)

            server.listen(port, (err) => {
              if (err) {
                return console.log('something bad happened', err);
              }

              console.log(`server is listening on ${port}`);
            })
---
apiVersion: v1
kind: Service
metadata:
  name: app-403-with-body-service
spec:
  selector:
    app: app-403-with-body
  ports:
  - protocol: TCP
    port: 8080
    targetPort: 8080
```
Calling such service directly works as expected (notice the POST data!)
```
/ # curl -i -X POST "http://app-403-with-body-service.default:8080/foo" -d "abcd"
HTTP/1.1 403 Forbidden
Date: Fri, 22 Mar 2019 09:59:43 GMT
Connection: keep-alive
Content-Length: 17

There's an error!/ #
```

Now create gateway for such application.
Example JSON payload:
```
{
    "name": "my-service-403",
    "provider": "myCompany403",
    "Identifier": "identifier",
    "description": "This is some test service",
    "api": {
        "targetUrl": "http://app-403-with-body-service.default:8080",
        "spec": {},
        "credentials": {
            "basic": {
                "username": "user1",
                "password": "pass1"
            }
        }
    }
}
```

Calling the service via gateway _without_ POST request body works (notice "There's an error!" response):
```
[...]

PR #3709, opened 2019-04-12
 Fix Application Gateway 502 response from retry with body

franpog859:
**Description**

-  Fix Application Gateway 502 response from retry with body

**Related issue(s)**

See the issue #3300 
See the bump #3710

### DIFF
--- a/components/application-gateway/internal/proxy/proxy.go
+++ b/components/application-gateway/internal/proxy/proxy.go
@@ -156,19 +156,29 @@ func (p *proxy) addModifyResponseHandler(r *http.Request, id string, cacheEntry 
 
 func (p *proxy) createModifyResponseFunction(id string, r *http.Request) (func(*http.Response) error, apperrors.AppError) {
 	// Handle the case when credentials has been changed or OAuth token has expired
-	body1, body2, err := drainBody(r.Body)
+	secondRequestBody, err := createSecondRequestBody(r)
 	if err != nil {
-		return nil, apperrors.Internal("failed to drain request body, %s", err)
+		return nil, err
 	}
 
-	r.Body = body1
-
 	modifyResponseFunction := func(response *http.Response) error {
-		retrier := newUnauthorizedResponseRetrier(id, r, body2, p.proxyTimeout, p.createCacheEntry)
+		retrier := newUnauthorizedResponseRetrier(id, r, secondRequestBody, p.proxyTimeout, p.createCacheEntry)
 		return retrier.RetryIfFailedToAuthorize(response)
 	}
 
 	return modifyResponseFunction, nil
+}
+
+func createSecondRequestBody(r *http.Request) (io.ReadCloser, apperrors.AppError) {
+	if r.Body != nil {
+		bodyCopy, secondRequestBody, err := drainBody(r.Body)
+		if err != nil {
+			return nil, apperrors.Internal("failed to drain request body, %s", err)
+		}
+		r.Body = bodyCopy
+		return secondRequestBody, nil
+	}
+	return nil, nil
 }
 
 func drainBody(b io.ReadCloser) (r1, r2 io.ReadCloser, err error) {
--- a/components/application-gateway/internal/proxy/proxy.go
+++ b/components/application-gateway/internal/proxy/proxy.go
@@ -1,8 +1,12 @@
 package proxy
 
 import (
+	"bytes"
 	"context"
 	"encoding/json"
+	"io"
+	"io/ioutil"
+	"log"
 	"net/http"
 	"time"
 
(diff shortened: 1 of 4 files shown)


components/application-gateway/internal/proxy/proxy.go
```diff
@@ -138,17 +144,55 @@ func (p *proxy) addAuthorization(r *http.Request, cacheEntry *CacheEntry) apperr
...
 
+	modifyResponseFunction := func(response *http.Response) error {
+		retrier := newUnauthorizedResponseRetrier(id, r, secondRequestBody, p.proxyTimeout, p.createCacheEntry)
 		return retrier.RetryIfFailedToAuthorize(response)
 	}
+
+	return modifyResponseFunction, nil
+}
+
+func createSecondRequestBody(r *http.Request) (io.ReadCloser, apperrors.AppError) {
```
Szymongib: I would suggest the name `copyRequestBody`.

components/application-gateway/internal/proxy/proxy.go
```diff
@@ -138,17 +144,55 @@ func (p *proxy) addAuthorization(r *http.Request, cacheEntry *CacheEntry) apperr
...
+	if r.Body != nil {
```
Szymongib: The more Golang approach would be:
```go
if r.Body == nil {
    return nil, nil
}
```

components/application-gateway/internal/proxy/unauthorizedresponseretrier.go
```diff
@@ -12,15 +13,16 @@ import (
 type retrier struct {
 	id                       string
 	request                  *http.Request
+	body                     io.ReadCloser
```
Szymongib: I would suggest a different name like `requestBody` or `requestBodyCopy`.

components/application-gateway/internal/proxy/unauthorizedresponseretrier_test.go
```diff
@@ -55,7 +55,10 @@ func TestForbiddenResponseRetrier_CheckResponse(t *testing.T) {
 		req, err := http.NewRequest(http.MethodGet, "/orders/123", nil)
 		require.NoError(t, err)
 
-		rr := newUnauthorizedResponseRetrier("id1", req, 10, updateCacheEntryFunction)
+		secondRequestBody, err := createSecondRequestBody(req)
```
Szymongib: The request body is `nil` here, we do not need to copy it. The same is for other cases.

components/application-gateway/internal/proxy/unauthorizedresponseretrier_test.go
```diff
@@ -20,7 +20,7 @@ func TestForbiddenResponseRetrier_CheckResponse(t *testing.T) {
 		updateCacheEntryFunction := func(id string) (*CacheEntry, apperrors.AppError) {
 			return nil, nil
 		}
-		rr := newUnauthorizedResponseRetrier("", &http.Request{}, 10, updateCacheEntryFunction)
+		rr := newUnauthorizedResponseRetrier("", &http.Request{}, http.Response{}.Body, 10, updateCacheEntryFunction)
```
Szymongib: `http.Response{}.Body` it does not look good and it might be quite misleading. In this case, the request body will be `nil`.

components/application-gateway/internal/proxy/unauthorizedresponseretrier.go
```diff
@@ -53,15 +55,14 @@ func (rr *retrier) retry() (*http.Response, error) {
 	request, cancel := rr.prepareRequest()
 	defer cancel()
 
-	var err error
+	request.Body = rr.body
```
Szymongib: I suggest adding a test case corresponding to the issue that we are solving so it will not occur again in case of some future changes. The best place for that would be probably `proxy_test.go`

components/application-gateway/internal/proxy/proxy.go
franpog859: Done!

components/application-gateway/internal/proxy/proxy.go
franpog859: Done! Yeah, you're right!

components/application-gateway/internal/proxy/unauthorizedresponseretrier.go
franpog859: Done!

components/application-gateway/internal/proxy/unauthorizedresponseretrier_test.go
franpog859: Done!

components/application-gateway/internal/proxy/unauthorizedresponseretrier_test.go
franpog859: Done!

components/application-gateway/internal/proxy/unauthorizedresponseretrier.go
franpog859: Done!

Issue #3201, opened 2019-03-15
Updating objects  through Console Backend Service sometimes fail because object is outdated


michal-hudy:
**Description**

Updating objects  through Console Backend Service sometimes failing because resource version is outdated.

**Expected result**

Updating process shouldn't fail when requested through GraphQL in such case as Component Backend Serice is building such object.

**Actual result**

Sometimes fails, also in tests.

**Steps to reproduce**

https://status.build.kyma-project.io/view/gcs/kyma-prow-logs/logs/post-master-kyma-gke-integration/1106552453140058115

**Troubleshooting**

PR #3712, opened 2019-04-15
Add retries to CBS k8s update mutations tests

tgorgol:
**Description**

- Add retries to CBS k8s update mutations tests

**Related issue(s)**

#3201


tests/console-backend-service/internal/domain/k8s/k8s.go
```diff
@@ -17,6 +24,19 @@ func fixNamespace(name string) *v1.Namespace {
 	}
 }
 
+func updateWithRetries(fn func() error, retries int) error {
```
pkosiec: Please move it to `pkg`. And also rename it to `retry` or something like that.

tests/console-backend-service/internal/domain/k8s/k8s.go
```diff
@@ -17,6 +24,19 @@ func fixNamespace(name string) *v1.Namespace {
...
+	var err error
+	for i := 0; i <= retries; i++ {
+		err = fn()
+		if err == nil {
+			return nil
+		}
+		log.Println(err)
+	}
+
+	return errors.Wrapf(err, fmt.Sprintf("update failed after %d retries", retries))
```
pkosiec: Add generic info that retry failed, as it will be moved to `pkg`

tests/console-backend-service/internal/domain/k8s/k8s.go
```diff
@@ -2,13 +2,20 @@ package k8s
...
 
+	"github.com/pkg/errors"
 	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
 type json map[string]interface{}
 
+const (
+	updateRetries = 5
```
pkosiec: Maybe it should be also declared in `pkg` as constant and used where the `retry` function is used

tests/console-backend-service/pkg/retrier/retry.go
```diff
@@ -0,0 +1,25 @@
...
+	"log"
+
+	"github.com/pkg/errors"
+)
+
+const (
+	UpdateRetries = 5
+)
+
+func Retry(fn func() error, retries int) error {
```
pkosiec: I would consider `Do` or `Run` func name here, if this is retrier :smile:

PR #3722, opened 2019-04-15
Evaluating self subject rules

kwiatekus:
**Description**

- Edit cluster role rules to fetch user permissions list
- Edit MicroFrontends CRDs to define which permissions are required for the user to display the MicroFrontend

**Related issue(s)**
Fixes console#758


resources/core/charts/console-backend-service/templates/cluster-role.yaml
```diff
@@ -14,6 +14,9 @@ rules:
   - apiGroups: [""]
     resources: ["pods","services", "configmaps", "secrets"]
     verbs: ["get", "list", "watch", "update", "delete"]
+  - apiGroups: [""]
+    resources: ["users"]
```
piotrmsc: This will give permissions to console backend to perform group impersonation
```suggestion
    resources: ["users", "groups"]
```

Issue #3032, opened 2019-03-07
Verify pre and post upgrade state of namespace controller


piotrmsc:
**Description**

Part of https://github.com/kyma-project/test-infra/issues/582
Write an e2e test ensuring namespace-controller works before and after upgrade.

Test scenario:

[pre-upgrade] create namespace with label env=ture
[post-upgrade] create new namespace with label env=true
[post-upgrade] check namespace has resource quotas
Reasons

We have to ensure component is running smoothly before and after upgrade to new version during kyma upgrade.

**Attachments**

PR #3785, opened 2019-04-18
Refactor Namespace backup tests

sjanota:
**Description**

- Make `NamespaceControllerTest `public - can be used in upgrade tests
Add new constructor with dependencies injected. that's how it's done in upgrade tests
- Extract `TestResourcesError` and `CreateResourcesError` as upgrade implementation needs to return error

**Related issue(s)**
Issue: #3032


tests/end-to-end/backup-restore-test/backupe2e/namespacecontroller.go
```diff
@@ -59,29 +74,18 @@ func (n namespaceControllerTest) labelTestNamespace(namespaceName string) error
...
-				messages += fmt.Sprintf("%+v\n", err)
-				continue
-			}
-
-			return nil
-
-		case <-timeout:
-			return fmt.Errorf("unable to fetch resourcequota:\n %v", messages)
+func (n NamespaceControllerTest) waitForResourceQuota(namespaceName string) error {
+	return retry.Do(func() error {
```
rakesh-garimella: what would be the timeout here  ?

sjanota: retry.Do has default timeout 100 ms with doubling backoff and does 10 attempts which gives around 50 s timeout

sjanota: I forgot that delay here is set to 500 ms. It means that whole timeout will be around 250 s

Issue #3192, opened 2019-03-15
Application Broker does not remove Services when namespace with Mapping is removed


aszecowka:
On a weekly cluster, where Stability Checker is executed, there are 121 Services in `kyma-integration` namespace. Most of them have following name pattern: 
- `ab-ns-for-svc-test-ns-2rctp` or 
- `ab-ns-for-test-acc-ns-broker-zcb9`.
Services with name pattern `ab-ns-for-test-acc-ns-broker` are created as a result of the `TestServiceCatalogContainsABServiceClasses` (`kyma/tests/acceptance/servicecatalog/servicecatalog_test.go`). 
At the end of the test, a testing namespace is removed. Removing namespace cause removing Application Mapping, and on that event, Mapping Controller (`kyma/components/application-broker/internal/mapping/controller`) is triggered. 
There is a method:
`ensureNsBrokerNotRegisteredIfNoMappingsOrSync`
which at the beginning check if broker in the given namespace exists, if not, we exit from the function with no error. 
But, when removing namespace, we don't have any gurantees that mapping will be removed before broker. So in our test, we remove namespace, then k8s removes at the beginning broker, then mapping and our controller forgets to remove Services.

PK85:
one of the solution: provice ownerReference in k8s service to ServiceBroker instance, when removed problem is gone.

PR #3789, opened 2019-04-18
Add OwnerReferences to Service for ServiceBroker

jasiu001:
Related to #3192 + additional scenario which creates Application, ApplicationMapping and all connected resources, removes them and check everything is cleaned up.


polskikiel:
@jasiu001 Please fix tests and I will approve :>

PR #3871, opened 2019-04-25
Update microfrontend client package types

dariadomagala:
**Description**

- Update microfrontend client package types to set `showInNavigation` default to true 
- Add new types

**Related issue(s)**


common/microfrontend-client/pkg/apis/ui/v1alpha1/types.go
```diff
@@ -55,6 +57,34 @@ type NavigationNode struct {
 	NavigationPath   string `json:"navigationPath"`
 	ViewURL          string `json:"viewUrl"`
 	ShowInNavigation bool   `json:"showInNavigation"`
+	Order            int    `json:"order"`
+	Settings         `json:"settings"`
+}
+
+type Settings struct {
+	ReadOnly bool `json:"readOnly"`
```
kwiatekus: Settings should't be so strongly typed.
please use `byte[]` and than convert into `gqlschema.JSON`

Issue #3824, opened 2019-04-23
Send different payload in the post-upgrade test for Kyma eventing


abbi-gaurav:
**Description**
To ensure that post upgrade is not receiving earlier events, we should send a different payload in the post-upgrade tests.

**AC**
- [ ] Modify upgrade tests to send a different payload in the post-upgrade test.

PR #3900, opened 2019-04-29
Send different payload in the post-upgrade test for Kyma eventing

montaro:
**Description**

- Send different payload in the post-upgrade test for Kyma eventing

**Related issue(s)**
See also #3824

### DIFF
--- a/tests/end-to-end/upgrade/pkg/tests/event-bus/event_bus.go
+++ b/tests/end-to-end/upgrade/pkg/tests/event-bus/event_bus.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"fmt"
 	"io/ioutil"
+	"math/rand"
 	"net/http"
 	"net/http/httputil"
 	"strings"
@@ -35,6 +36,8 @@ const (
 	publishEventEndpointURL  = "http://event-bus-publish.kyma-system:8080/v1/events"
 	publishStatusEndpointURL = "http://event-bus-publish.kyma-system:8080/v1/status/ready"
 )
+
+var randomInt int
 
 // UpgradeTest tests the Event Bus business logic after Kyma upgrade phase
 type UpgradeTest struct {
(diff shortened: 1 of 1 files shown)


tests/end-to-end/upgrade/pkg/tests/event-bus/event_bus.go
```diff
@@ -266,7 +271,9 @@ func (f *eventBusFlow) publishTestEvent() error {
 
 func (f *eventBusFlow) publish(publishEventURL string) (*api.PublishResponse, error) {
 	payload := fmt.Sprintf(
-		`{"source-id": "%s","event-type":"%s","event-type-version":"v1","event-time":"2018-11-02T22:08:41+00:00","data":"test-event-1"}`, srcID, eventType)
+		`{"source-id": "%s","event-type":"%s","event-type-version":"v1","event-time":"2018-11-02T22:08:41+00:00","data":%s}`,
+		srcID, eventType, composePayloadData("test-event", randomInt))
+	f.log.Debugf("publish payload:%s", payload)
```
suleymanakbas91: This log is redundant as the next one already logs the payload.

tests/end-to-end/upgrade/pkg/tests/event-bus/event_bus.go
```diff
@@ -244,6 +247,8 @@ func (f *eventBusFlow) checkSubscriptionReady() error {
 }
 
 func (f *eventBusFlow) publishTestEvent() error {
+	randomInt = rand.Intn(100)
+	f.log.Debugf("Publish random int: %v", randomInt)
```
abbi-gaurav: It would have been better to 
* not use some kind of state variable instead to inject the value for pre and post upgrade as a function parameter.
* use clear strings such as `pre-upgrade-paylaod + rand.int()` and `post-upgrade + rand.int()` to have better debugging support.

PR #3904, opened 2019-04-30
Improve testing script

polskikiel:
**Description**

Improved testing script:
- Increased the frequency of calling for TestSuite status.
- `Waiting...` log is still displayed only once per minute

### DIFF
--- a/installation/scripts/testing.sh
+++ b/installation/scripts/testing.sh
@@ -32,6 +32,7 @@ EOF
 startTime=$(date +%s)
 
 testExitCode=0
+previousPrintTime=-1
 
 while true
 do
@@ -56,6 +57,7 @@ do
         testExitCode=1
         break
     fi
+
     sec=$((currTime-startTime))
     min=$((sec/60))
     if (( min > 60 )); then
@@ -63,8 +65,11 @@ do
         testExitCode=1
         break
     fi
-    echo "ClusterTestSuite not finished. Waiting..."
-    sleep 60
+    if (( $previousPrintTime != $min )); then
+        echo "ClusterTestSuite not finished. Waiting..."
+        previousPrintTime=${min}
+    fi
+    sleep 3
 done
 
 echo "Test summary"


installation/scripts/testing.sh
```diff
@@ -26,12 +26,15 @@ metadata:
   name: ${suiteName}
 spec:
   maxRetries: 1
-  concurrency: 1
+  concurrency: 2
```
aszecowka: no, no, no - we have to know which tests can be run concurrently and only after that we can enable concurrency

installation/scripts/testing.sh
```diff
@@ -26,12 +26,15 @@ metadata:
...
 EOF
 
 startTime=$(date +%s)
 
 testExitCode=0
+minCache=-1
```
aszecowka: minCache is a very enigmatic name for me

installation/scripts/testing.sh
```diff
@@ -40,6 +43,23 @@ do
     statusFailed=$(${kc} get cts ${suiteName}  -ojsonpath="{.status.conditions[?(@.type=='Failed')]}")
     statusError=$(${kc} get cts  ${suiteName} -ojsonpath="{.status.conditions[?(@.type=='Error')]}" )
 
+    succeededTestsNames=$(${kc} get cts ${suiteName} -ojsonpath="{.status.results[?(@.status=='Succeeded')].name}")
+    failedTestsNames=$(${kc} get cts ${suiteName} -ojsonpath="{.status.results[?(@.status=='Failed')].name}")
+
+    names=(${succeededTestsNames})
+    len="${#names[@]}"
+    if ((${succeededTests} < ${len})); then
```
aszecowka: I think that it won't work if 2 tests finished at the same time. You will print only the name of the last one.

installation/scripts/testing.sh
```diff
@@ -56,15 +76,28 @@ do
...
     sec=$((currTime-startTime))
     min=$((sec/60))
     if (( min > 60 )); then
         echo "Timeout for test suite '${suiteName}' occurred."
         testExitCode=1
         break
     fi
-    echo "ClusterTestSuite not finished. Waiting..."
-    sleep 60
+    if (( $minCache != $min )); then
```
aszecowka: I am afraid that this code becomes too complicated. Maybe you can focus right now on improving script by more frequent checks if a suite is finished?

aszecowka:
It looks that there can be an error in logic because I got the following output:
```
Test 'cluster-users-test' succeeded
Test 'cluster-users-test' succeeded
```

installation/scripts/testing.sh
polskikiel: discussed

installation/scripts/testing.sh
polskikiel: changed to `previousPrintTime`

Issue #3841, opened 2019-04-24
Replace hostNetwork with hostAliases in Console Backend Service


Szymongib:
**Description**

Istio Sidecar is not injected to `console-backend-service` on Minikube because it uses `spec.hostNetwork`. It can be replaced with `spec.hostAliases` to stop preventing sidecar injection. 

**Reasons**

To secure communication between `console-backend-service` and `connector-service` Istio sidecar is required (issue #2982).

derberg:
@Szymongib could you elaborate a bit more here, how local Minikube issue correlates with https://github.com/kyma-project/kyma/issues/2982 that I guess is a cluster/production related issue

derberg:
Some details here https://kyma-community.slack.com/archives/CD172NFCJ/p1557122032068500?thread_ts=1556738591.005700&cid=CD172NFCJ

PR #3931, opened 2019-05-06
Use hostAliases in Console Backend Service chart

pkosiec:
**Description**

- Use hostAliases in Console Backend Service chart

**Related issue(s)**
Resolves https://github.com/kyma-project/kyma/issues/3841


resources/core/charts/console-backend-service/templates/service.yaml
```diff
@@ -1,8 +1,7 @@
 apiVersion: v1
 kind: Service
 metadata:
-  annotations:
-    "auth.istio.io/{{ .Values.service.externalPort }}": NONE    
+  annotations: {}
```
mszostok: nit: if not needed, the just do not specify it :)

resources/core/charts/console-backend-service/templates/policy.yaml
```diff
@@ -0,0 +1,13 @@
+{{ if .Values.global.isLocalEnv }}
+apiVersion: authentication.istio.io/v1alpha1
```
mszostok: can we also add here a business comment on why this is needed at all? 

Right now is a magic settings without any background :/

PR #3884, opened 2019-04-26
display metrics for ignored published events in grafana [Resource]

montaro:
See also #3627 and https://github.com/kyma-project/kyma/pull/3820

PR #3820, opened 2019-04-23
Display metrics for ignored published events in Grafana [Component]

montaro:
See also #3627

Issue #3627, opened 2019-04-08
Display metrics for ignored published events in Grafana


abbi-gaurav:
> Note: Add the  templating variables for application, event type, and event type version

**AC**
- [x] There is a dashboard available OOTB with Kyma installation that displays ignored published events in a Grafana dashboard.

montaro:
![image](https://user-images.githubusercontent.com/531945/56654924-9f468200-6691-11e9-8969-5f0e0bfde22f.png)
@abbi-gaurav Please check

PR #3951, opened 2019-05-07
Fix the rate chart of the ignored published messages grafana dashboard

montaro:
See also #3627 , https://github.com/kyma-project/kyma/pull/3820 and https://github.com/kyma-project/kyma/pull/3884


montaro:
image

Issue #2617, opened 2019-02-06
Support for additional headers and query parameters in App Registry and Proxy service


lszymik:
**Description**

There are the various systems which require additional headers. A good example is a support for `X-ApplicationKey` header.

The headers might be sent with calls to authorization endpoint and/or might be sent with calls to registered API.

**Reasons**

There are APIs which are protected with additional headers, not always according to standards.

**Acceptance criteria**

- The API for adding headers and query parameters to auth endpoint and API added.
- The functionality implemented and tests provided.
- The documentation updated.

**Recommendation**

```
...
"api": {
    "targetUrl": "https://honeybadgers-testing.hybris.com/rest/v2",
    "credentials": {
      "basic": {
        "username": "user",
        "password": "pass"
      },
      "headers" : {
          "X-Application-Key" : "1234"
      },
      "queryParameters" : {
          "q1" : "value"
      }
    },
    "headers" : {
          "X-Application-Key" : "1234"
      },
      "queryParameters" : {
          "q2" : "value"
      }
      ...
```

Highlights:
- If any query parameter or a header is also sent as a call to the proxy, then it has a higher priority
- Also in case of `Authorization` headers - client credentials flow

brendandobbs:
@lszymik , will it also be possible to use headers and query parameters without providing credentials? In some cases the headers with be providing parameters for authentication so the crendentials are not needed.

lszymik:
@brendandobbs , yes we will also have support for headers and query parameters without credentials.

PR #3991, opened 2019-05-09
Extend API with headers and query params property 

crabtree:
**Description**

- Extend API and credentials models with headers property
- Extend API and credentials models with query parameters
- Update swagger file

**Related issue(s)**

See also #2617

### DIFF
--- a/docs/application-connector/08-04-register-secured-api.md
+++ b/docs/application-connector/08-04-register-secured-api.md
@@ -138,3 +138,26 @@ This is an example of the `api` section of the request body for an API secured w
             },
         }
 ```
+
+## Specify custom headers and query parameters for authentication requests
+
+You can specify additional headers and query parameters that will be injected during authentication requests.
+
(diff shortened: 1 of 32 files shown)


docs/application-connector/08-04-register-secured-api.md
```diff
@@ -138,3 +138,26 @@ This is an example of the `api` section of the request body for an API secured w
             },
         }
 ```
+
+## Specify custom headers and query parameters for authentication requests
+
+You can specify additional headers and query parameters that will be injected during authentication requests.
+
+This is an example of the `api` section of the request body for an API secured with Basic Authentication enriched with header `custom-header` with value `foo`, and query parameter `param` with value `bar`.
```
klaudiagrz:
```suggestion
This is an example of the **api** section of the request body for an API secured with Basic Authentication. It is enriched with the **custom-header** header with the `foo` value, and the **param** query parameter with the `bar` value.
```

Issue #3573, opened 2019-04-04
Pre-integrate Kiali and find potential challenges and benefits


a-thaler:
## Description
The operation of and development with a service mesh can be a complex topic. With Kyma Istio is used as underlying technology to build up and manage the service mesh. Currently an operator and developer can have a hard time debugging the mesh to figure out faulty configurations, misbehaving services, understanding the topology.

Kiali (https://www.kiali.io/) promise to bring light into the topic of
- visualizing the topology
- computation of mesh health
- distributed tracing
- configuration validation

So let's give it a try and figure out the concrete benefits and the impact of operatiing it.

## Reasons
The debuggability and observability of the Kyma service mesh is limited to prometheus metrics mainly, having prepared istio dashboards. However, advanced visualization and maybe configuration fault detection can be a very valuable addon.

## Goal
Create a Kyma PR having kiali integrated so that
- others can test it easily as well
- after merge it is disabled
- we have a first base of a production ready integration

Document here in the ticket
- the observed problems and implications to a potential integration of the tool
- the benefits of the tool
- the potential problems for a productive usage which needs to be addressed, like:
  - security
  - maturity
  - ....
- a final conclusion on if it should be integrated and/or potential restrictions to it

igor-karpukhin:
Implemented in the following PR: https://github.com/kyma-project/kyma/pull/4019

Kiali allows us to visualize the traffic flow, including HTTP(S)/GRPC/TCP which is useful to debug the connectivity problems. It also allows us to see the big picture of the system, the diagram of connections and the amount of traffic. 
 Kiali is disabled by default, you have to manually enable it (see `installer-config-local.yaml`). It also allows to edit Istio configuration files, which is not good, so I had to disable this functionality. So far Kiali only supports authorization with the secret, which is now generated during the Istio installation procedure. It's not that flexible for now, so we'd like to see OAuth2 or something like that.

PR #4000, opened 2019-05-09
Feature/kiali support

igor-karpukhin:
**Description**

Kyma installer now has an ability to install Kiali for Istio.

- Added Kiali support

**Related issue(s)**
https://github.com/kyma-project/kyma/issues/3573


installation/resources/installer-cr.yaml.tpl
```diff
@@ -62,5 +62,5 @@ spec:
     #  namespace: "kyma-system"
     #- name: "jaeger"
     #  namespace: "kyma-system"
-    #- name: "monitoring"
-    #  namespace: "kyma-system"
+    - name: "monitoring"
```
a-thaler: you are changing the default installation behavior of kyma. For an optional feature. I think that should be solved by documentation instead

installation/resources/installer.yaml
```diff
@@ -233,7 +233,7 @@ spec:
       serviceAccountName: kyma-installer
       containers:
       - name: kyma-installer-container
-        image: eu.gcr.io/kyma-project/develop/installer:0fdc80dd
+        image: eu.gcr.io/kyma-project/kyma-installer:1.0.0
```
a-thaler: Strange changes here, please rebase your branch to see the real changes

resources/istio/values.yaml
```diff
@@ -88,7 +88,7 @@ tracing:
 # addon kiali tracing configuration
 #
 kiali:
-  enabled: false
+  enabled: true
```
a-thaler: It should not be enabled by default

resources/istio/charts/kiali/values.yaml
```diff
@@ -26,12 +26,16 @@ dashboard:
   usernameKey: username
   passphraseKey: passphrase
 
+credentials:
+  username: admin
+  passphrase: admin
```
a-thaler: Why there is a hardcoded password, can' t it be removed?

a-thaler:
Where are the result of the PoC? Whar do you see as benefit of Kiali, what problems need to be solved before promoting it as an official feature?

PR #4030, opened 2019-05-13
Release 1.1 RC 2

Szymongib:
**Description**

- Create RC 2

**Related issue(s)**


resources/core/values.yaml
```diff
@@ -24,65 +24,65 @@ global:
     secretName: etcd-backup-abs-credentials
   acceptance_tests:
     dir:
-    version: 1.1.0-rc1
+    version: 1.1.0-rc2
   ui_acceptance_tests:
     dir:
     version: "83aa639b"
```
strekm: is this on purpose or should it be `1.1.0-rc2`?

resources/helm-broker/values.yaml
```diff
@@ -32,10 +32,10 @@ global:
...
     dir:
-    version: 1.1.0-rc1
+    version: 1.1.0-rc2
   alpine_net:
     dir:
-    version: 1.1.0-rc1
+    version: 1.1.0-rc2
   add_ons_ui:
     dir: ""
     version: 23b3d5a1
```
strekm: this one also looks suspicious. please verify

resources/core/values.yaml
Szymongib: It is on purpose, `ui_acceptance_tests` is part of a console, which is not being released.

resources/helm-broker/values.yaml
Szymongib: `add_ons_ui` also is not being released.

PR #3926, opened 2019-05-04
Add dashboard to provide insights for a given kyma subscription

montaro:
See also #3916

Issue #3916, opened 2019-05-02
Add dashboard to provide insights for a given Kyma Subscription


abbi-gaurav:
**Description**
There should be a template for namespace and Kyma subscription name.

Following status and metrics should be provided
- [x] Event Activation status
- [x] Kyma Subscription Status
- [x] Knative Subscription Status
- [x] Associated channel status

**AC**
- [x] The dashboard is available as a part of Kyma deployment.

abbi-gaurav:
Waiting for RC2 cluster to verify the changes. There are some issues observed, input from @montaro is that the same is fixed in RC2.

PR #4041, opened 2019-05-13
Fix Event Bus Subscription Dashboard charts colors

montaro:
Fix Event Bus Subscription Chart view

See also #3916 #3926 #4001 #4022


montaro:
image

Issue #3932, opened 2019-05-06
Simplify cluster installation process


piotrmsc:
**Description**

Currently, in order to install kyma on a cluster you have to execute commands which basically replace placeholders in configuration yaml with empty string values. Additionally, you have a different command for k8s < 1.12.6 and  > 1.12.6 for promtail. We should stick to the newer k8s versions.
Take a look for example [here](https://kyma-project.io/docs/root/kyma/#prepare-the-installation-configuration-file) step 4.
To simplify this process, all placeholders can be moved to chart values as empty strings. 
Our installation documentation should describe how to create a configmap with an override for example with own domain and TLS certificate instead of describing how to export variable and execute cat& sed commands.

Additionally, we can label installation CR with action install directly in the yaml file. This way no additional steps is required to trigger installation.

**Reasons**

Kyma should be easy to install without any sed&cat commands, just pure kubectly apply.

**Attachments**

Tomasz-Smelcerz-SAP:
See also: https://github.com/kyma-project/test-infra/issues/1024

PR #4097, opened 2019-05-15
Add the 'action=install' label to crs

jakkab:
**Description**

- remove cr labelling from the `installer.sh` script
- add default `action=install` label to cr templates
- refactor the `tiller-tls.sh` script

**Related issue(s)**
#3932

### DIFF
--- a/installation/scripts/installer.sh
+++ b/installation/scripts/installer.sh
@@ -76,8 +76,7 @@ if [ $CR_PATH ]; then
 
 fi
 
-
-echo -e "\nApplying installation combo yaml"
+echo -e "\nCreating installation combo yaml"
 COMBO_YAML=$(bash ${CURRENT_DIR}/concat-yamls.sh ${INSTALLER} ${INSTALLER_CONFIG} ${AZURE_BROKER_CONFIG} ${CR_PATH})
 
 rm -rf ${AZURE_BROKER_CONFIG}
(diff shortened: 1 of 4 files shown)


jakkab:
pre-master-kyma-gke-upgrade-nosed

installation/scripts/installer.sh
```diff
@@ -92,13 +91,13 @@ if [ ${LOCAL} ]; then
...
-
 echo -e "\nConfiguring sub-components"
 bash ${CURRENT_DIR}/configure-components.sh
 
-echo -e "\nGetting Helm secrets"
-bash ${CURRENT_DIR}/tiller-tls.sh
+echo -e "\nApplying combo yaml"
+kubectl apply -f - <<< "$COMBO_YAML"
+
+echo -e "\nDownloading Helm certificates"
```
Tomasz-Smelcerz-SAP:
```suggestion
echo -e "\Setting up Helm certificates"
```

installation/scripts/installer.sh
```diff
@@ -92,13 +91,13 @@ if [ ${LOCAL} ]; then
...
 fi
 
-kubectl apply -f - <<<"$COMBO_YAML"
-
 echo -e "\nConfiguring sub-components"
 bash ${CURRENT_DIR}/configure-components.sh
 
-echo -e "\nGetting Helm secrets"
-bash ${CURRENT_DIR}/tiller-tls.sh
+echo -e "\nApplying combo yaml"
```
Tomasz-Smelcerz-SAP: I would put this section at the end (after Helm certificates), and log just: "Starting Kyma Installation" or something like that.

installation/scripts/installer.sh
Demonsthere: This is wrong, we are in fact downloading the certs from the kubernetes secrets to loca files, we do not set them :)

Issue #4103, opened 2019-05-15
Only few traces are shown in Jaeger


a-thaler:
**Description**
The current setting of istio-pilot is defining the trace sampling rate with a 1.0 where 100 is a maximum. With that only 1 out of 100 requests will be send to jaeger for trace recording. So showing traces for simple demo setups is nearly impossible.

**Expected result**
A moderate trace rate should be send to jaeger in order to support debugging. Minimum should be prominent documentation on how to setup a demo/development cluster having a high sampling rate.

See traceSampling setting in Kyma 0.8 at https://github.com/kyma-project/kyma/blob/release-0.8/resources/istio/values.yaml where it was at 100.

**Actual result**
Only few (1 out of 100) traces get send to jaeger. See current traceSampling setting at https://github.com/kyma-project/kyma/blob/master/resources/istio/charts/pilot/values.yaml

**Steps to reproduce**
Create a lambda exposed via HTTP and call it. With a high probability, you will not find the trace in jaeger.

**Troubleshooting**
```
kubectl -n istio-system edit deploy istio-pilot
```
and set traceSampling to for example 100.0. Be aware of potential performance/instability impacts on istio when using such high number.

jcawley5:
Can this please be tagged to be included in the documentation?

a-thaler:
Yes, you are right, will trigger documentation of it

a-thaler:
FYI: The setting is also part of values.yaml in kyma/resources/istio/charts/pilot chart, with that it can be configured already at installation time with overwrite: traceSampling

bszwarc:
Done.

a-thaler:
We still want to do some performance investigations and optimize the current setting if possible.

derberg:
I think we need to make sure there is a clear docu that mentions this value https://github.com/kyma-project/kyma/blob/58c85663e587766791675385f2e2fc0091f08d22/resources/istio/charts/pilot/values.yaml#L12 and explains how you can override it before installation. How to do it in runtime (https://kyma-project.io/docs/master/components/tracing/#troubleshooting-basic-troubleshooting) is not enough, user would like to also know how to configure it before installation, what are the configuration parameters here, that they are available as overrides.

derberg:
[...]

PR #4104, opened 2019-05-15
adjusting istio-pilot traceSampling rate to 100

a-thaler:
**Description**
Adjusted trace sampling rate of istio-pilot to 100% in order to send all traces to jaeger.

**Related issue(s)**
https://github.com/kyma-project/kyma/issues/4103


a-thaler:
I will revoke the PR for now until we tested the setting in regards to istio stability. If a 100 setting is not possible we need to provide documentation as well.

Issue #3856, opened 2019-04-25
Enable sending custom extension headers in Kyma eventing


abbi-gaurav:
**Description**
* Kyma eventing by default should forward all the headers with the CloudEvents Extension prefix.

The prefix is : `ce-`

Refer : https://github.com/cloudevents/spec/blob/master/http-transport-binding.md#31-binary-content-mode

In case the headers have the same name as the one used by kyma eventing e.g. `CE-Event-type`, the precedence should be given to the one derived from the payload.

e.g. a header of the form  `ce-my-company-passportid` should be delivered to the consumer or the lambda.

Note: Headers are case-insensitive

**AC**
- [x] Headers sent with `ce-` prefix while publishing an event are delivered to Knative eventing and lambda.

abbi-gaurav:
Sending `CE-` headers from an application to Kyma is not getting delivered to Lambda.

e.g. I sent `Ce-Corr-Id:123` but I do not see it in lambda logs

### event sent
```
echo '{"event-type" : "order.created", "event-type-version" : "v1", "event-time" : "2018-05-02T22:08:41+00:00", "data" : {"orderCode" : "expect-ce"}}'|curl -X POST -d @- https://gateway.${KYMA_DOMAIN}/${APP_NAME}/v1/events --header "Ce-Corr-Id: 123" --cert generated.pem -k

{"event-id":"f1832d5b-ec98-4036-96bc-cd1fd012f135","status":"published","reason":"Message successfully published to the channel"}
```

### Lambda logs
I do not see the header `Ce-Corr-Id`
```
{ orderCode: 'expect-ce' }
{ host: 'test.workshop:8080',
  'user-agent': 'Go-http-client/1.1',
  'content-length': '25',
  'ce-event-id': 'f1832d5b-ec98-4036-96bc-cd1fd012f135',
  'ce-event-time': '2018-05-02T22:08:41+00:00',
  'ce-event-type': 'order.created',
  'ce-event-type-version': 'v1',
  'ce-knativehistory': 'kcugjcrgxkoijubwl6mqm5247x6yfsaag.kyma-system.channels.cluster.local',
  'ce-source-id': 'test-ce-extension-headers',
  'content-type': 'application/json',
  'x-b3-flags': '0',
  'x-request-id': '35a4464d-beb4-4f47-af28-7bdcb48b5926',
  'accept-encoding': 'gzip',
  'x-forwarded-proto': 'http',
[...]

PR #4106, opened 2019-05-15
Enable sending custom extension headers in Kyma eventing

montaro:
See also #3856

### DIFF
--- a/components/event-bus/internal/knative/util/kn-eventing.go
+++ b/components/event-bus/internal/knative/util/kn-eventing.go
@@ -65,7 +65,7 @@ type KnativeAccessLib interface {
 	DeleteSubscription(name string, namespace string) error
 	GetSubscription(name string, namespace string) (*evapisv1alpha1.Subscription, error)
 	UpdateSubscription(sub *evapisv1alpha1.Subscription) (*evapisv1alpha1.Subscription, error)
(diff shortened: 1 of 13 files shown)


components/event-bus/internal/knative/util/kn-eventing.go
```diff
@@ -301,10 +301,7 @@ func makeHttpRequest(channel *evapisv1alpha1.Channel, headers *map[string]string
 		return nil, err
 	}
 	req.Header.Set("Content-Type", "application/json")
-
-	for k, v := range *headers {
-		req.Header.Set(k, v)
-	}
+	req.Header = *headers
```
marcobebway: Please move this line before `req.Header.Set("Content-Type", "application/json")`, to not lose setting the content-type header.

montaro: Thanks for the catch!
Fixed

sayanh:
@montaro can you please add a test which takes care of:
> In case the headers have the same name as the one used by kyma eventing e.g. CE-Event-type, the precedence should be given to the one derived from the payload.

sayanh:
https://github.com/kyma-project/kyma/pull/4106#issuecomment-494412272
It's fine if we add it in a separate PR

Issue #3843, opened 2019-04-24
Bind only selected Service Classes to the namespace


pbochynski:
**Description**
The application can register many service classes (APIs & Events). It should be possible for the administrator to select only those classes that should or can be enabled in the given namespace. 
There should be still a possibility to register all service classes to the namespace (current behaviour). In this case, it is expected that also service classes added after binding is done eventually will be available in the namespace.

**Reasons**
Service classes are tied to credentials/authorizations, hence it will be beneficial to offer more granular options to enable APIs/credentials to developers in a given namespace. 

**AC**
- UI/UX designs provided
- keep possibility to register all service classes to the namespace (current behaviour)
- select only those classes that should or can be enabled in the given namespace
  - ApplicationMapping CRD needs be extended with serviceClassIds attribute
  - UI and console-backend-service needs to list the serviceClassIds from Application CR and then allow to pass some of them to ApplicationMapping CR
- service classes added after binding is done eventually will be available in the namespace.
- all services option marked then updated of the application is always taken into account
- if 2 services added, application update causes only deletetion

**UI/UX views**

<img width="1084" alt="Screen Shot 2019-05-10 at 10 53 33" src="https://user-images.githubusercontent.com/9609118/57515035-04e17200-7312-11e9-8fa7-6fed5f1ee742.png">
<img width="1081" alt="Screen Shot 2019-05-10 at 10 55 26" src="https://user-images.githubusercontent.com/9609118/57515098-25a9c780-7312-11e9-96a2-1f41063e479d.png">
<img width="1079" alt="Screen Shot 2019-05-10 at 10 56 06" src="https://user-images.githubusercontent.com/9609118/57515189-52f67580-7312-11e9-89c9-2f8d82c65974.png">
<img width="1076" alt="Screen Shot 2019-05-10 at 10 57 37" src="https://user-images.githubusercontent.com/9609118/57515268-7e796000-7312-11e9-95a0-1926d61f40d8.png">
<img width="1079" alt="Screen Shot 2019-05-10 at 10 58 34" src="https://user-images.githubusercontent.com/9609118/57515322-a10b7900-7312-11e9-9155-15f515a4dc3d.png">

piotrmiskiewicz:
Define ApplicationMapping struct: https://github.com/kyma-project/kyma/pull/4114

jasiu001:
**New features in console backend service**

To handle new functionality by graphQL, two queries were modified and one mutation was added.

**enableApplication**

```
[...]

PR #4148, opened 2019-05-20
Application broker enables selected services

piotrmiskiewicz:
**Description**

- process a service list from an ApplicationMapping resource to filter `GetCatalog` response
- main_test which tests the application-broker setup
 
**Related issue(s)**
Related: #3843


components/application-broker/internal/access/mapping.go
```diff
@@ -30,3 +38,59 @@ func (c *ApplicationMappingService) IsApplicationEnabled(namespace, name string)
...
+	return false
+}
+
+type allServicesEnabled struct {
+}
+
+// IsServiceEnabled implements ServiceEnabledChecker, always returns true
+func (c *allServicesEnabled) IsServiceEnabled(svc internal.Service) bool {
+	return true
+}
```
jasiu001: Is it not easier create one `ApplicationServiceChecker` with extra boolean field instead three different struct? :)

piotrmiskiewicz: I'll do it as private. 3 simple implementation allows to avoid `if` statements inside the IsServiceEnabled implementations

Issue #4066, opened 2019-05-14
Move istio chart changes from istio-kyma-patch


Demonsthere:
**Description**

Move parts of the istio-kyma-patch that modify the istio charts into the charts. If possible by modifying the values, if not, then the manifests. 

**Reasons**

The istio charts are advanced enough to support required configuration changes, so we shouldn't patch them, which may introduce errors. 

**Attachments**

Demonsthere:
Elements that have been moved:
- Zipkin configuration
- HostPorts for minikube deployment

Default sidecar injection can be also moved, however causes unexpected behaviour and crashes in istio, so the move was aborted.

PR #4151, opened 2019-05-20
Reduce istio-kyma-patch logic

Demonsthere:
**Description**

- Move logic from istio-kyma-patch to the istio charts

**Related issue(s)**

Implements #4066


installation/resources/installer-config-local.yaml.tpl
```diff
@@ -73,6 +73,8 @@ data:
 
   mixer.policy.autoscaleEnabled: "false"
   mixer.telemetry.autoscaleEnabled: "false"
+  
+  global.tracer.zipkin.address: zipkin.kyma-system:9411
 ---
```
piotrmsc: Zipkin address can be moved to values.yaml in the charts directly and removed from config files

PR #4179, opened 2019-05-22
Expand istio limits on cluster settings

Demonsthere:
**Description**

We have seen on nightly clusters that there was not enough system resources (cpu and mem) for creating even a simple test lambda pod. A fair amount of those resources was used by istio. 

- Introduce the same cpu and mem limitations we have on minikube in a cluster setup

**Related issue(s)**

### DIFF
--- a/installation/resources/installer-config-cluster.yaml.tpl
+++ b/installation/resources/installer-config-cluster.yaml.tpl
@@ -67,6 +67,22 @@ data:
   gateways.istio-ingressgateway.loadBalancerIP: "__EXTERNAL_PUBLIC_IP__"
   global.proxy.excludeIPRanges: "__PROXY_EXCLUDE_IP_RANGES__"
   global.tracer.zipkin.address: zipkin.kyma-system:9411
+
+  pilot.resources.limits.memory: 1024Mi
+  pilot.resources.limits.cpu: 500m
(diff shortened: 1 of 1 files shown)


installation/resources/installer-config-cluster.yaml.tpl
```diff
@@ -67,6 +67,22 @@ data:
   gateways.istio-ingressgateway.loadBalancerIP: "__EXTERNAL_PUBLIC_IP__"
   global.proxy.excludeIPRanges: "__PROXY_EXCLUDE_IP_RANGES__"
   global.tracer.zipkin.address: zipkin.kyma-system:9411
+
```
piotrmsc: >  A fair amount of those resources was used by istio.

What exactly do you mean? which components and how much do they use? 
Pilot by default has  memory request set to 2gig, have you tested it on 1 gig limit? 
The same with policy, generally I'm against setting the same limit as for minikube

piotrmsc: Second thing, cluster values should be in the charts. We aim to get rid of this cluster config file.

Demonsthere:
Seems the problem is due to insufficient machine type in AKS [Standard_DS2_v2](https://docs.microsoft.com/en-us/azure/virtual-machines/linux/sizes-general). We can close this

Issue #4075, opened 2019-05-14
Increase natss memory.limit to 256M


sayanh:
**Description**

nats-streaming-0 gets OOMkilled

Although kyma cluster set up needs to be stress tested, it is decided to increase it to 256M for now.

**Expected result**

Change memory limit for nats-streaming sts to 256M.

PR #4228, opened 2019-05-27
Add knative-eventing memory requests config

montaro:
**Description**
Kyma local installation gets stuck while installing the event-bus component complaining of insufficient memory

~Increase Minikube memory to 10GB~
Configure knative-eventing controller and webhook memory limits 

**Related issue(s)**
See also #4075


mszostok:
@montaro Hi, I'm not totally against your PR but for me is a real red light. 

Right now 8GB is quite a lot. We should have really strong arguments why we want to increase that to **10GB**. It is because we added new components? It's that because the event bus has new configuration? Couldn't we fix that problem in a different way?

And if it's just a hot-fix to unblock the installation then we should have some follow-up which describes the problem.

Right now it's to easy to increase the memory without any strong argumentation. In such way we will end-up with minikube (**local instalaltion**) which will require 15GB.

PK85:
@montaro,

Please provide a memory footprint for all kyma pods to see if it is an issue of No. of components or maybe just one of them eats everything.

Cheers, PK

montaro:
@PK85 Thanks for the review!
I'll try to get it here. I was monitoring the pods metrics and didn't see any anomaly or some pod specifically eating the memory up.

montaro:
@mszostok Thanks for the review!
I believe the approach of merging and having a follow-up ticket is a good approach to me.

montaro:
Installing Kyma from `master` while NATS pod is `32M` and minikube memory is 8Gs, Installation is stuck in the `application-connector` component and here are the screenshots of the status:
image
image
image

montaro:
Looks like `istio-pilot` is consuming too much memory as well
image

piotrmsc:
istio pilot is not consuming too much resources, moreover, we have lowered istio requests and limits here https://github.com/kyma-project/kyma/pull/3950/files so resources previously assigned to istio pilot & policy are now available on node for other, non istio components

montaro:
@piotrmsc So we believe now that the upgrade to `knative-eventing v0.5.0` was responsible to that. I will close this PR as a duplicate to https://github.com/kyma-project/kyma/pull/4264
Thanks!

PR #4236, opened 2019-05-27
Add 'name' prop to Tabs in docs

aerfio:
**Description**

- in order to have working linking between tabs in website we need to add special identificator for every tab

**Related issue(s)**

See also https://github.com/kyma-project/website/issues/301


mmitoraj:
Aren't there more <div tabs> that require the `name`?

aerfio:
@mmitoraj nope, `name` attribute will be needed only in case there are headers inside tabs, this will be documented, please watch https://github.com/kyma-project/website/issues/301

PR #4266, opened 2019-05-28
Enable apicontroller backup restore tests

sayanh:
**Description**

- 

**Related issue(s)**


Tomasz-Smelcerz-SAP:
The first issue I found is related to the api-controller test. The loop that awaits for Function Pod to be ready has wrong delay configured. This was easy to fix. Additional logging on every retry is added for better visiblity, it can be removed once the test is fixed.

Issue #2694, opened 2019-02-11
API-controller: introduce a blacklist of K8s services


jakkab:
**Description**

We should have a blacklist of K8s services.

**Reasons**

API-controller should not create APIs for some services (e.g. `kubernetes`).

piotrmsc:
Provide list of blacklisted services in env in format : "service.namespace" ie. "kubernetes.default"

piotrmsc:
@strekm Where I can read about it? https://kyma-project.io/docs/components/api-gateway/#overview-overview does not mention what is blacklisted by default

PR #4411, opened 2019-06-06
Validate API against blacklisted services

strekm:
**Description**

-Validate API against blacklisted services

**Related issue(s)**See #2694

### DIFF
--- a/components/api-controller/cmd/controller/main.go
+++ b/components/api-controller/cmd/controller/main.go
@@ -147,12 +147,9 @@ func getCORSConfig() istioNetworkingV1.CORSConfig {
 }
 
 func getBlacklistedServices() []string {
-	separator := ","
 	list := os.Getenv("BLACKLISTED_SERVICES")
 
-	blacklistedServices := removeEmptyStrings(strings.Split(list, separator))
-
-	return blacklistedServices
+	return removeEmptyStrings(strings.Split(list, ","))
 }
 
 func removeEmptyStrings(strings []string) []string {
--- a/components/api-controller/cmd/controller/main.go
+++ b/components/api-controller/cmd/controller/main.go
@@ -60,7 +60,9 @@ func main() {
 
 	internalInformerFactory := kymaInformers.NewSharedInformerFactory(kymaClientSet, time.Second*30)
 
-	v1alpha2Controller := v1alpha2.NewController(kymaClientSet, istioNetworkingV1Interface, serviceV1Interface, authenticationV2Interface, internalInformerFactory, domainName, getBlacklistedServices())
+	list := os.Getenv("BLACKLISTED_SERVICES")
+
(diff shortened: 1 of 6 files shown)


components/api-controller/pkg/controller/v1alpha2/controller.go
```diff
@@ -256,6 +265,18 @@ func (c *Controller) validateAPI(newAPI *kymaApi.Api, apiStatusHelper *ApiStatus
 	return setStatus(kymaMeta.Successful)
 }
 
+func (c *Controller) validateVirtualService(newAPI *kymaApi.Api) error {
+	for _, svc := range c.blacklistedServices {
+		separator := "."
+		props := strings.Split(svc, separator)
+
+		if newAPI.Spec.Service.Name == props[0] && newAPI.GetNamespace() == props[1] {
+			return fmt.Errorf("creating VirtualService for %s.%s is forbidden", newAPI.Spec.Service.Name, newAPI.GetNamespace())
```
Demonsthere: Maybe a better error would be `Service X has been blacklisted by BLACKLISTED_SERVICES env`

components/api-controller/pkg/controller/v1alpha2/controller.go
```diff
@@ -256,6 +265,18 @@ func (c *Controller) validateAPI(newAPI *kymaApi.Api, apiStatusHelper *ApiStatus
 	return setStatus(kymaMeta.Successful)
 }
 
+func (c *Controller) validateVirtualService(newAPI *kymaApi.Api) error {
```
jakkab: I suggest something like
```
func (c *Controller) validateVirtualService(newAPI *kymaApi.Api) error {

	newApiServiceFullName := fmt.Sprintf("%s.%s", newAPI.Spec.Service.Name, newAPI.GetNamespace())
	
	for _, svc := range c.blacklistedServices {
		
		if svc == newApiServiceFullName {
			return fmt.Errorf("service %s has been blacklisted by BLACKLISTED_SERVICES env", newApiServiceFullName)
		}
	}
	return nil
}
``` 
This way we can avoid multiple conditions and parsing strings.

components/api-controller/cmd/controller/main.go
```diff
@@ -144,6 +146,10 @@ func getCORSConfig() istioNetworkingV1.CORSConfig {
 	}
 }
 
+func readBlacklistedServices(list string) []string {
+	return removeEmptyStrings(strings.Split(list, ","))
+}
+
 func removeEmptyStrings(strings []string) []string {
```
Tomasz-Smelcerz-SAP: This function can as well trim whitespaces. It should be trimmed anyway.

components/api-controller/cmd/controller/main.go
```diff
@@ -144,6 +146,10 @@ func getCORSConfig() istioNetworkingV1.CORSConfig {
 	}
 }
 
+func readBlacklistedServices(list string) []string {
+	return removeEmptyStrings(strings.Split(list, ","))
```
Tomasz-Smelcerz-SAP: this line: removeEmptyStrings(strings.Split(list, ",")) can be re-used with three other places.

PR #4436, opened 2019-06-07
Fix and improve metrics for ServiceCatalog

polskikiel:
**Description**

- Fixed metrics for ServiceCatalog

**Related issue(s)**


resources/service-catalog/charts/catalog-crds/templates/metrics.yaml
```diff
@@ -0,0 +1,53 @@
...
+    port: secure
+    scheme: https
+    tlsConfig:
+      insecureSkipVerify: true
+  namespaceSelector:
+    matchNames:
+    - "{{ .Release.Namespace }}"
+  selector:
+    matchLabels:
+      app: {{ template "fullname" . }}-controller-manager
```
hisarbalik: Service definition above should match this selectors? Or do you have some where else another service definition match this monitoring?

resources/service-catalog/charts/catalog-crds/templates/metrics.yaml
```diff
@@ -0,0 +1,53 @@
...
+    prometheus: monitoring
+    app: {{ template "fullname" . }}-controller-manager
+    chart: "{{ .Chart.Name }}-{{ .Chart.Version }}"
+    release: "{{ .Release.Name }}"
+    heritage: "{{ .Release.Service }}"
+  name: {{ template "fullname" . }}-controller-manager
+spec:
+  endpoints:
+  - interval: 30s
+    port: secure
```
hisarbalik: is the port for service definition above?

hisarbalik: if yes you should change port name

resources/service-catalog/charts/catalog-crds/templates/metrics.yaml
polskikiel: I have another service definition which matches this selector in this file:
 ```kyma/resources/service-catalog/charts/catalog-crds/templates/controller-manager-service.yaml```

resources/service-catalog/charts/catalog-crds/templates/metrics.yaml
polskikiel: It's a port from the file: ```kyma/resources/service-catalog/charts/catalog-crds/templates/controller-manager-service.yaml```. I think this service definition is now unused so I will delete it 👍

PR #4466, opened 2019-06-11
K8S version dependent Promtail configurer

hisarbalik:
Promtail uses k8s container management API to collect logs, this API is not same in every k8s version.

Logging helm chart will detect used k8s version and configure promtail accordingly


hisarbalik:
#4471

resources/logging/templates/promtail/daemonset.yaml
```diff
@@ -8,71 +9,75 @@ metadata:
...
       priorityClassName: {{ .Values.promtail.priorityClassName }}
-    {{- end }}
+            {{- end }}
       containers:
         - name: promtail
           image: "{{ .Values.promtail.image.repository }}:{{ .Values.promtail.image.tag }}"
           imagePullPolicy: {{ .Values.promtail.image.pullPolicy }}
           args:
-            - "-config.file=/etc/promtail/{{ .Values.global.logging.promtail.config.name }}"
+            {{- if (semverCompare ">=1.14.0" .Capabilities.KubeVersion.GitVersion) }}
```
piotrmsc: On the top you have already initialized a variable holding version why not reuse it?

PR #4490, opened 2019-06-11
api-controller bump & api-controller-acceptance-tests bump

strekm:
**Description**

- api-controller bump

**Related issue(s)**
see #2694

### DIFF
--- a/resources/core/charts/api-controller/values.yaml
+++ b/resources/core/charts/api-controller/values.yaml
@@ -10,3 +10,5 @@ cors:
   allow_origin: "*"
   allow_methods: "GET,POST,PUT,DELETE" # no possibility to use "*"
   allow_headers: "*"
+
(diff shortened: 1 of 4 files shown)


kubadz:
I tested this manually with some additional cases not covered in tests, and everything works fine. 

I would give an approve but I have one question: 
are you planning to set blacklisted service mentioned in issue (`kubernetes.default`) in a proper ENV in api-controller deployment as a part of this PR, or you will create another PR with that change?

resources/core/charts/api-controller/values.yaml
```diff
@@ -9,4 +9,6 @@ mtls:
 cors:
   allow_origin: "*"
   allow_methods: "GET,POST,PUT,DELETE" # no possibility to use "*"
-  allow_headers: "*"
\ No newline at end of file
+  allow_headers: "*"
+
+blacklisted_services: "kubernetes.default,istio-citadel.istio-system,istio-galley.istio-system,istio-ingressgateway.istio-system,istio-pilot.istio-system,istio-policy.istio-system,istio-sidecar-injector.istio-system,istio-telemetry.istio-system,apiserver-proxy.kyma-system"
```
piotrmsc: There is also apiserver-proxy-ssl

PR #4555, opened 2019-06-17
Upgrade Loki version to 0.1.0

igor-karpukhin:
Upgrade Loki version to 0.1.0


resources/logging/templates/loki/statefulset.yaml
```diff
@@ -0,0 +1,109 @@
...
+            - name: config
+              mountPath: /etc/loki
+            - name: storage
+              mountPath: "/data"
+              subPath: {{ .Values.loki.persistence.subPath }}
+          ports:
+            - name: http
+              containerPort: {{ .Values.loki.config.server.http_listen_port }}
+              protocol: TCP
+                    {{/*
```
hisarbalik: please add here liveness and readiness check with commands options

resources/logging/templates/loki/deployment.yaml
```diff
@@ -1,3 +1,4 @@
+{{/*
```
hisarbalik: this file can be removed completely, it's replaced with statefulset.yaml

resources/logging/templates/loki/statefulset.yaml
igor-karpukhin: Fixed

resources/logging/templates/loki/deployment.yaml
igor-karpukhin: Fixed

PR #4596, opened 2019-06-19
Allow concurrency in testing.sh

sjanota:
**Description**

- `testing.sh` supports `--concurrency` flag to provide the number of concurrent tests.
- Every TestDefinition has `disableConcurrency` set to `true` so they are run concurrently only when they are ready

**Related issue(s)**

### DIFF
--- a/installation/scripts/testing.sh
+++ b/installation/scripts/testing.sh
@@ -16,7 +16,7 @@ do
         -*)
             echo "Unknown flag ${1}"
             exit 1
-        ;;
+            ;;
         *) # unknown option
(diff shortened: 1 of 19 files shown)


installation/scripts/testing.sh
```diff
@@ -1,6 +1,31 @@
...
+    case ${key} in
+        --concurrency|-c)
+            CONCURRENCY="$2"
+            shift
+            shift
+            ;;
+        -*)
+            echo "Unknown flag ${1}"
+            exit 1
+        ;;
```
nachtmaar: Wrong indention

Issue #4388, opened 2019-06-05
Align event delivery with CloudEvents specification


abbi-gaurav:
**Description**

As a developer, I should be able to deploy a serverless (microservice or lambda) for event trigger and expect the event payload and metadata delivered in alignment with cloud event specification. This will enable me to use cloud events SDK for implementing my event triggers.

As requested by Ralf, we should align with v0.3 of the specification.

https://github.com/cloudevents/spec/blob/v0.3/spec.md
https://github.com/cloudevents/spec/blob/v0.3/http-transport-binding.md

**AC**
- [x] The metadata/context attributes and payload when delivered to a lambda conforms to cloud event specification.
- [x] `event.data` should still contain the payload as it is today.
- [x] Add an example that displays conformance to CE specification.

montaro:
Example of a valid CE event:
```
curl -i \
-H "Content-Type: application/json" \
-X POST http://event-bus-publish.kyma-system:8080/v1/events \
-d '{"source": "external-application", "type": "hello", "eventtypeversion": "v1", "time": "2018-11-02T22:08:41+00:00", "data": { "order-number": 123 }}'
```
Example of Lambda received event headers (CE event context): 
```
{ host: 'hello-with-data.stage:8080',
  'user-agent': 'Go-http-client/1.1',
  'content-length': '20',
  'ce-eventtypeversion': 'v1',
  'ce-id': '9d68c24b-8dc1-411e-b50e-dddd0f8cddba',
  'ce-knativehistory': 'koz27gvp6t5isyg57ei3jtgfdb63s6edn.kyma-system.channels.cluster.local',
  'ce-source': 'external-application',
  'ce-time': '2018-11-02T22:08:41+00:00',
  'ce-type': 'hello',
  'content-type': 'application/json',
  'x-b3-flags': '0',
  'x-request-id': '0e902a83-f6b0-4451-8cfc-6e901808dce1',
  'accept-encoding': 'gzip',
  'x-forwarded-proto': 'http',
  'x-forwarded-client-cert': 'By=spiffe://cluster.local/ns/stage/sa/default;Hash=bcc1f140c29e95c9bec1aee73ca9799b86b222fc3f1f058d70ac6d779941cf9d;Subject="";URI=spiffe://cluster.local/ns/knative-eventing/sa/natss-dispatcher',
  'x-b3-traceid': '4bd8755de6c60f0c',
  'x-b3-spanid': '2b220bff8fbeb349',
[...]

PR #4585, opened 2019-06-18
Align event delivery with cloud events specification

montaro:
See also #4388

PR #4605, opened 2019-06-21
Align event delivery with cloud events specification

montaro:
See also #4388


montaro:
Duplicate https://github.com/kyma-project/kyma/pull/4585

Issue #4461, opened 2019-06-11
Prevent SBU deletion while Service Catalog is migrated


piotrmiskiewicz:
The migration process must not remove SBU

PR #4627, opened 2019-06-24
Add job to ServiceCatalog resource to prevent SBU deletion

jasiu001:
Related to #4461 

For now second stage, restore `ownerReference` does not work because of lack of migration tool, should work after merge this [PR](https://github.com/kyma-incubator/service-catalog/pull/31)


components/service-binding-usage-controller/cmd/migration/main.go
```diff
@@ -0,0 +1,157 @@
...
+	return false
+}
+
+func (sm *sbuManager) findServiceBinding(name, namespace string) (*v1beta1.ServiceBinding, error) {
+	if sm.sbList == nil {
+		list, err := sm.scClient.ServicecatalogV1beta1().ServiceBindings(v1.NamespaceAll).List(v1.ListOptions{})
+		if err != nil {
+			return nil, errors.Wrap(err, "while fetching ServiceBinding list")
+		}
+		sm.sbList = list
```
mszostok: tbh I'm not convinced to have such caching, we will not have too much SBU, so maybe is better to always get related ServiceBinding and have always fresh data?

IMO we can run into some issues with simple such solution, if you want to do sth like that then maybe is better to create informer which has dynamic caching so we will have always fresh data

resources/service-catalog/charts/catalog-crds/templates/pre-upgrade-job.yaml
```diff
@@ -0,0 +1,124 @@
...
+            do
+              if [ "${item}" == "${APISERVER}" ]; then
+                indicator=$((indicator+1))
+              fi
+              if [ "${item}" == "${WEBHOOK}" ]; then
+                indicator=$((indicator-1))
+              fi
+            done
+
+            # if 1 apiserver exist (run migration), if -1 webhook exist,
```
mszostok: that should be definitely aligned with @polskikiel cause he also is doing the same thing (check if migration should be executed or not)

resources/service-catalog/charts/catalog-crds/values.yaml
```diff
@@ -115,3 +115,6 @@ servicePlanDefaultsEnabled: false
 ## by example :
 ## securityContext: { runAsUser: 1001 }
 securityContext: {}
+## image for migration process prevent SBU deletion
+migrationImagePath: eu.gcr.io/kyma-project/develop/service-catalog-upgrade
```
mszostok: this img has to general name. By reading that `service-catalog-upgrade` i think that is connected only with the service catalog which is not true :) please rename that

Issue #4180, opened 2019-05-22
Subscription properties max_inflight and push_request_timeout_ms are not applied post Knative eventing migration


abbi-gaurav:
**Description**
As a developer, I expect available configuration for delivery behavior (Kyma Subscription) is applied.

Post migrating to Knative eventing, the functionality and behavior to dispatch the events to serverless are controlled by the Knative eventing internals.

The two fields `max_inflight` and `push_request_timeout_ms` were added when Kyma eventing was having its own dispatch mechanism. At present, they are not in sync with the Knative eventing semantics.

[2019-06-11]
Decided with the team to remove the configuration. The configuration first will be done at the Knative level and then bring it to Kyma eventing.

>NOTE: Contact Hasselhoff to update the lambda UI.

**AC**
- [x] `max_inflight` and `push_request_timeout_ms` are removed from the Kyma Subscription
- [x] Kyma Subscription should still work in a backward compatible manner.
- [x] Add required fields constraint for the mandatory fields

marcobebway:
Verified all PRs on a cluster using the E2E [test](https://github.com/kyma-incubator/examples/tree/master/kyma-local-e2e-scenario).

![trace-list](https://user-images.githubusercontent.com/24314423/60348643-4ca28280-99c0-11e9-9338-deb9894f8565.png)
![trace-details](https://user-images.githubusercontent.com/24314423/60348652-52986380-99c0-11e9-939f-db1641b26a11.png)

PR #4643, opened 2019-06-26
Update kyma subscription components

marcobebway:
**Description**

- Update the Kyma subscription by removing both the **max_inflight** and **push_request_timeout_ms** properties.
- Update the related tests.

**Related issue(s)**
#4180

### DIFF
--- a/tests/gateway-tests/test/testkit/subscriptions_client.go
+++ b/tests/gateway-tests/test/testkit/subscriptions_client.go
@@ -50,8 +50,6 @@ func (sc *client) Create(namespace, application, eventType string) error {
 		SubscriptionSpec: v1alpha1.SubscriptionSpec{
 			Endpoint:                      "https://some.test.endpoint",
 			IncludeSubscriptionNameHeader: true,
-			MaxInflight:                   400,
-			PushRequestTimeoutMS:          2000,
 			EventType:                     eventType,
 			EventTypeVersion:              "v1",
 			SourceID:                      application,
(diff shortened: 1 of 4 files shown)


tests/gateway-tests/test/testkit/subscriptions_client.go
```diff
@@ -50,8 +50,6 @@ func (sc *client) Create(namespace, application, eventType string) error {
 		SubscriptionSpec: v1alpha1.SubscriptionSpec{
 			Endpoint:                      "https://some.test.endpoint",
 			IncludeSubscriptionNameHeader: true,
-			MaxInflight:                   400,
```
sayanh: IMO, we need both versions of the test as the old clients can come with these fields(MaxInflight...).

resources/core/charts/kubeless/charts/lambdas-ui/templates/configmap.yaml
```diff
@@ -21,10 +21,8 @@ data:
           {size:{ name: 'XL', memory: '1024Mi', cpu: '2.0', minReplicas: 10, maxReplicas: 20, description: '' }}
         ],
       subscriptionConfig:{
-        pushRequestTimeoutMS: 60000,
```
sayanh: To prevent, unnecessary debug hell, I propose this change to go with lambda-ui image bump PR.

resources/cluster-essentials/templates/eventing-subscription.crd.yaml
```diff
@@ -25,14 +25,6 @@ spec:
               maxLength: 512
             include_subscription_name_header:
               type: boolean
-            max_inflight:
```
sayanh: Same as [here](https://github.com/kyma-project/kyma/pull/4643#discussion_r298074341)

tests/gateway-tests/test/testkit/subscriptions_client.go
marcobebway: Here the v1alpha1.SubscriptionSpec does not have these properties anymore, so we have to remove them.

Issue #4655, opened 2019-06-27
Api-controller acceptance tests namespace cleanup broken


Tomasz-Smelcerz-SAP:
**Description**
The test first tries to create the namespace, then **defers** namespace delete.
If the namespace already exists, namespace creation fails and the code never reaches the line with deferred namespace deletion.
Once the namespace is not cleaned up properly in the  cluster, it will be never cleaned up, because namespace creation fails with "namespace already exists" error.
This happened in a weekly cluster. It can happen again, it's enough to terminate test pod during the run.

**Expected result**
Namespace cleanup is executed even if namespace creation failed.

**Actual result**

Test fails on namespace creation because the namespace already exists. Namespace is not cleaned up. The same happens on next test execution.

**Steps to reproduce**

Manually create a namespace: api-controller-tests, then execute the test.

**Troubleshooting**
Remove the namespace manually. Not always possible on a cluster (lack of permissions).

piotrmsc:
Unfortunately, pr does not solve the problem, I have left my comment in the merged PR

PR #4656, opened 2019-06-27
Always perform namespace cleanup

Tomasz-Smelcerz-SAP:
**Description**
Test namespace should always be deleted in the test, even if namespace creation failed.
This guards the test from falling into permanent error state, where namespace already exists and it cannot be created and is not deleted because of the error.

- Ensure namespace cleanup is always executed

**Related issue(s)**
See also #4655

### DIFF
--- a/tests/api-controller-acceptance-tests/apicontroller/main_test.go
+++ b/tests/api-controller-acceptance-tests/apicontroller/main_test.go
@@ -1,15 +1,16 @@
 package apicontroller
 
 import (
+	"os"
+	"os/signal"
+	"testing"
+
 	log "github.com/sirupsen/logrus"
 	"k8s.io/api/core/v1"
 	meta_v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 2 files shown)


tests/api-controller-acceptance-tests/apicontroller/main_test.go
```diff
@@ -85,8 +86,8 @@ func loadKubeConfigOrDie() *rest.Config {
 
 func testWithNamespace(m *testing.M) int {
 	catchInterrupt()
-	createNamespace()
 	defer deleteNamespace()
+	createNamespace()
```
piotrmsc: `createNamespace` function is invoking `log.Fatalf()` which logs an error message and invokes ox.Exit(1) under the hood.  `os.Exit` work in given way : 

> The program terminates immediately; deferred functions are not run.

So this change (PR) does not solve the root cause. To solve the issue `createNamespace` should *not* invoke `log.Fatalf()` but `log.Errorf()` inestad and this error should be  handled in the main

Issue #4470, opened 2019-06-11
Define ClusterAddonsConfiguration and AddonsConfiguration CRD


piotrmiskiewicz:
AC:
 - define CRDs (examples)
 - provide nice description about status, how does it looks like in all scenarios (for example conflicted bundles), define all types of errors
 - describe how Helm Broker works with more than one AddonsConfiguration in the same namespace
 - define golang structs and generate clientset
 - official CRDs and Helm Broker docs updated

Status capability design and implementation:
  - design status  (for example types of errors etc)
  - repositoryURLs ok/error (add index.yaml in case URL point to root)
  - bundles details per repository - ok/error
  - lastfetched
  - decide and document how to handle following situations:
    - prereq: URL registered globally and then same bundle registerd in the given namespace
    - prereq: URL registered in a namespace and then same bundle registerd again in the same namespace

mszostok:
## (Cluster)AddonsConfiguration CRD

### How Helm Broker resolve conflicts

**Statement**

Needs to be clearly stated that preventing conflicts in Helm Broker is implemented only for easier debugging and founding repository URL which contains conflicted bundles. We will not eliminate global problems with conflicting IDs between ClusterServiceClasses/ServiceClasses. There can be still a situation where few different brokers will register ClusterServiceClasses/ServiceClasses with the same ID. In such case, those classes are visible in Service Catalog but provisioning action is blocked with _"Found more that one class"_ reason.

### Implemented strategies

This section describes rules which are used to resolve conflicts when bundle with the same ID is registered multiple times.

Rules:
1. When the AddonsConfiguration and ClusterAddonsConfiguration register bundle with the same ID, then both are visible in the Service Catalog and they do not conflict with each other. 

2. The AddonsConfiguration is marked as failed when there are more that one bundle with the same ID specified under the **repositoryURLs** parameter. In the status entry, you can find all information about conflicted bundles. 
  > **NOTE:** Same applies for ClusterAddonsConfiguration.

3. When many AddonsConfigurations wants to register bundle with the same ID, then first which will be processed by Helm Broker will be registered successfully (unless the 2nd rule is broken). All others will be marked as failed. In the status entry, you can find all information about conflicted bundles.
  > **NOTE:** Same applies for ClusterAddonsConfiguration.

mszostok:
### TODO

This features are considered as stretch and can be implemented in future.

[...]

PR #4657, opened 2019-06-27
Add schema for (Cluster)AddonsConfiguration CRDs

mszostok:
**Description**

- Add schema, clients, crd generators, and docs for (Cluster)AddonsConfiguration CRDs.

**Related issue(s)**

https://github.com/kyma-project/kyma/issues/4470


components/helm-broker/docs/crd.md
```diff
@@ -0,0 +1,93 @@
+---
+title: AddonsConfiguration
```
mszostok: this should be duplicated for the ClusterAddonsConfiguration. The CRD parameters will be the same

components/helm-broker/docs/controller-logic.md
```diff
@@ -0,0 +1,52 @@
+## Controller logic
```
mszostok: this doc describes how controller logic should be implemented (created for developer who will implement that).

we have to rewrite the content so that it is also helpful to the end-user. End-user should know how the Helm Broker will process created Configuration (adding, updating and deleting them)

mszostok:
documentation was moved to this PR https://github.com/kyma-project/kyma/pull/4701

PR #4726, opened 2019-07-02
Fix testing.sh concurrency regexp

sjanota:
**Description**

- ...
- ...
- ...

**Related issue(s)**

### DIFF
--- a/installation/scripts/testing.sh
+++ b/installation/scripts/testing.sh
@@ -11,7 +11,7 @@ function validateConcurrency() {
     exit 1
   fi
 
-  if ! [[ "$1" =~ ^[0-9]+$ ]]; then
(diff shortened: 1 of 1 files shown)


installation/scripts/testing.sh
```diff
@@ -11,7 +11,7 @@ function validateConcurrency() {
     exit 1
   fi
 
-  if ! [[ "$1" =~ '^[0-9]+$' ]]; then
+  if ! [[ "$1" =~ ^[0-9]+$ ]]; then
```
jakkab: Excuse me for being a pain in the neck, but as far as I can tell, this regexp matches leading zeros.

sjanota: Yes, it does... Gimme a sec. I'll teach him good manners.

Issue #4518, opened 2019-06-12
Application status is flickering


akrausesap:
**Description**

Dear Colleagues,
the status of an Application Object is flickering on the ui independent of the status of the underlying pods. This is observed in the 1.2-rc1 release.

Regards

Andreas

![notserving](https://user-images.githubusercontent.com/40229965/59351269-49e83200-8d1e-11e9-8455-5837a8433237.png)

![pods](https://user-images.githubusercontent.com/40229965/59351283-5076a980-8d1e-11e9-98bc-6f2fa54142d6.png)

**Expected result**

No Flickering

**Actual result**

Flickering

**Steps to reproduce**
Create an application and wait...

**Troubleshooting**
N/A

PR #4843, opened 2019-07-09
Fix for flickering application status

janmedrek:
**Description**

- Introduced proper health endpoint to the certificate validator

**Related issue(s)**
#4518

### DIFF
--- a/components/application-operator/charts/application/templates/validator.yaml
+++ b/components/application-operator/charts/application/templates/validator.yaml
@@ -27,12 +27,25 @@ spec:
           args:
             - "/applicationconnectivityvalidator"
             - "--proxyPort={{ .Values.applicationConnectivityValidator.args.proxyPort }}"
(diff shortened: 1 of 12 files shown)


components/application-operator/charts/application/templates/validator.yaml
```diff
@@ -27,19 +27,36 @@ spec:
...
             - containerPort: {{ .Values.applicationConnectivityValidator.args.proxyPort }}
               name: http-proxy
+            - containerPort: {{ .Values.applicationConnectivityValidator.args.externalAPIPort }}
+              name: http-api-port
 ---
 apiVersion: v1
 kind: Service
 metadata:
+  annotations:
+    "auth.istio.io/{{ .Values.applicationConnectivityValidator.args.externalAPIPort }}": NONE
```
Szymongib: I think this annotation is no longer used by Istio, therefore we should remove it.

PR #4875, opened 2019-07-11
Compass runtime agent chart

akgalwas:
**Description**

- Chart for installing Compass Runtime Agent added

**Related issue(s)**
https://github.com/kyma-incubator/compass/issues/113


CODEOWNERS
```diff
@@ -199,6 +199,9 @@
 # The `compass` chart
 /resources/compass/ @PK85 @aszecowka @crabtree @pkosiec @kfurgol @tgorgol
 
+# The `compass-runtime-agent` chart
+/resources/compass/ @akgalwas @janmedrek @Szymongib @franpog859 @Maladie
```
PK85: I'm not sure if it should be a separate chart. Cause with one chart, we will can 
 use subcharts and overrides. We need to discuss that.

crabtree:
```suggestion
/resources/compass-runtime-agent/ @akgalwas @janmedrek @Szymongib @franpog859 @Maladie
```

Issue #4761, opened 2019-07-03
Update Kiali to 1.1.0


a-thaler:
**Description**
Update Kiali to the latest stable version, even if the bundles istio version is not in sync yet. That will assure that it is more stable and stays compatible, furthermore the feature aditions are quite big.

**Reasons**
Using a stable version

Part of epic https://github.com/kyma-project/kyma/issues/4168

Be ware of that the istio main chart gets updated to a newer version as well and that can cause conflicts: https://github.com/kyma-project/kyma/issues/4694

a-thaler:
I created a follow up https://github.com/kyma-project/kyma/issues/4918 as the exposure is not working fine and it should be fixed by having a dedicated module

Furthermore the relevant attributes for istio file editing should be extracted in that ticket

PR #4896, opened 2019-07-14
Kiali upgrade to version 1.1.0

hisarbalik:
#4761


resources/istio/charts/kiali/values.yaml
```diff
@@ -57,3 +57,7 @@ credentials:
   # Override the automatically detected Jaeger URL, useful when Jaeger service has no ExternalIPs
   # jaegerURL:
 prometheusAddr: http://monitoring-prometheus.kyma-system:9090
+global:
```
Demonsthere: I would suggest moving the overrides for the values to the main `istio/values.yaml`, we have all overrides there

hisarbalik: this added here because of local minikube installation, somehow does not work without this value in chart values.yam file

Issue #4297, opened 2019-05-29
event-bus tests can be run in parallel with others


sjanota:
**Description**

It should be possible to run event-bus tests in parallel with other Kyma tests. We need to verify if it is possible now and change the test if necessary. Then its TestDefinition should be modified to enable concurrency.

**Reasons**

One way to make the Kyma test suite faster is to run tests in parallel. 

**Acceptance Criteria**
- [ ] concurrency is enabled in event-bus test and it is stable on CI

sayanh:
[Integration](https://status.build.kyma-project.io/view/gcs/kyma-prow-logs/pr-logs/pull/kyma-project_kyma/4923/pre-master-kyma-gke-integration/1152287108103147520) and [upgrade](https://status.build.kyma-project.io/view/gcs/kyma-prow-logs/pr-logs/pull/kyma-project_kyma/4923/pre-master-kyma-gke-upgrade/1152177764313337856) jobs are working fine after parallelizing the tests. But there is a dependency on this [PR](https://github.com/kyma-project/kyma/pull/4893) and it is better to continue on this after it is merged.

PR #4923, opened 2019-07-17
Enable parallel execution of event-bus tests

sayanh:
**Description**

- Enable parallel execution of event-bus tests

**Related issue(s)**

See #4297


resources/event-bus/templates/tests/test.yaml
```diff
@@ -27,6 +27,7 @@ spec:
             - --publish-event-uri=http://event-bus-publish:8080/v1/events
```
anishj0shi: it needs to be rebased with master to have the latest changes pulled in

resources/event-bus/templates/tests/test.yaml
```diff
@@ -27,6 +27,7 @@ spec:
             - --publish-event-uri=http://event-publish-service:8080/v1/events
             - --publish-status-uri=http://event-publish-service:8080/v1/status/ready
             - --subscriber-image={{ .Values.global.containerRegistry.path }}/{{ .Values.global.event_bus_tests.dir }}event-bus-e2e-subscriber:{{ .Values.global.event_bus_tests.version }}
```
anishj0shi: I know this has been there for quiet some time, but shouldn't we reference the image name "event-bus-e2e-subscriber" in values.yaml ? something like this https://github.com/kyma-project/kyma/blob/0e88f482af87dd1e89ad449fec2b6fdfe3eca748/resources/event-bus/charts/event-publish-service/templates/deployment.yaml#L33

sayanh: IMO, I don't see the benefit. In order to retain consistency with the other charts, I would change https://github.com/kyma-project/kyma/blob/0e88f482af87dd1e89ad449fec2b6fdfe3eca748/resources/event-bus/charts/event-publish-service/templates/deployment.yaml#L33

PR #4929, opened 2019-07-17
Fix arch diag for event bus

sayanh:
**Description**
Turns out, that the natss controller does not create an istio virtual service for the dispatcher. It just creates a k8s svc. Hence the fix is created.

- Fix arch diag for event bus
- Fix steps for events publish process

More details are: [here](https://github.com/knative/eventing/blob/master/pkg/provisioners/provisioner_util.go#L41)


docs/event-bus/02-01-event-bus.md
```diff
@@ -31,7 +31,7 @@ For example, whenever the `order-created` Event comes in, the Event Bus stores i
 
 3. The Application Connector makes a REST API call to **publish-knative** and sends the enriched Event.
 4. **publish-knative** makes the HTTP payload compatible with Knative and sends the Event to the relevant **knative-channel** service URL which is inferred based on **source id**, **event type**, and **event type version** parameters.
-5. Istio Virtual Service forwards the Event to the **nats-dispatcher** service served by the **nats-dispatcher** Pod.
+5. K8s service forwards the Event to the **nats-dispatcher** service served by the **nats-dispatcher** Pod.
```
bszwarc:
```suggestion
5. Kubernetes service forwards the Event to the **nats-dispatcher** service served by the **nats-dispatcher** Pod.
```

PR #4934, opened 2019-07-18
Added JSON schemas for monitoring components

igor-karpukhin:
Added JSON schemas for monitoring components: global, alert-manager, prometheus, grafana


hisarbalik:
Can you please add default values too in the schema definition and also examples if possible

Issue #4760, opened 2019-07-03
Authentication integration for Kiali


a-thaler:
**Description**
Add an integrated authentication strategy to the kiali dashboard so that a single-sign-on is possible based on the dex configuration.
For that the same approach as for the jaeger dashboard integration can be used by disabling the bundled authentication and putting a keycloak proxy in front assuring that a OpenIDConnect token will be present for a request.

**Reasons**
Usability, the user expects the same authentication method as for all other components

Part of https://github.com/kyma-project/kyma/issues/4168

a-thaler:
I added the keycloak-proxy in the same way as in jaeger chart.
I disabled authentication on kiali itself

I did the changes in https://github.com/kyma-project/kyma/pull/4939

Issue #4918, opened 2019-07-16
Extract Kiali into dedicated module


a-thaler:
**Description**
Move the "kiali" sub-chart of istio module to the root level and have it enabled by default, but not for Kyma Lite. It should be installed after monitoring and tracing.

Items:
- Move the chart
- Add it to the installer-cr after monitoring and tracing module, disabled for local installation
- Add integration test checking availability of UI
- Add upgrade test checking availability of UI before and after upgrade
- Define customer relevant values for installation in regards to istio-file editing

**Reasons**
Kiali is currently bundled with the istio chart. However, it has its own release cycle and it has dependencies to istio being in conflict at installation time (uses virtualservice etc). So it will simplify the setup if it will be installed after istio. It should be installed even after monitoring.

a-thaler:
I extracted the chart
I disabled grafana and jaeger integration
I configured dex authentication by default, but support static user authentication as well
I enabled the chart for regular clusters, not for local setup

PR #4939, opened 2019-07-18
Added Kiali chart

a-thaler:
**Description**

- moved kiali into own chart
- enabled kcproxy for authentication

Fixes https://github.com/kyma-project/kyma/issues/4760 and https://github.com/kyma-project/kyma/issues/4918


resources/kiali/values.yaml
```diff
@@ -0,0 +1,62 @@
...
+    passphrase: ""
+  
+global:
+  isLocalEnv: false
+  containerRegistry:
+    path: eu.gcr.io/kyma-project
+  alpine_net:
+    name: alpine-net
+    dir: develop
+    version: 149967d0
```
Demonsthere: ad quotes to the version string or in some cases when the version has only number helm will have problem with interpreting it

resources/kiali/templates/virtualservice.yaml
```diff
@@ -1,14 +1,16 @@
...
   labels:
     app: {{ template "kiali.name" . }}
-    chart: {{ .Chart.Name }}-{{ .Chart.Version }}
+    chart: {{ template "kiali.chart" . }}
+    heritage: {{ .Release.Service }}
+    release: {{ .Release.Name }}
 spec:
   gateways:
-  - {{ .Values.global.istio.gateway.name }}
+  - kyma-gateway
```
Demonsthere: Better to use a variable here as we have the name of the gateway set as variable also

resources/kiali/templates/tests/test-kiali-connection.yaml
```diff
@@ -0,0 +1,28 @@
...
+    metadata:
+      annotations:
+        sidecar.istio.io/inject: "false"
+    spec:
+      restartPolicy: Never
+      containers:
+      - name: "{{ template "kiali.fullname" . }}-test"
+        image: {{ .Values.global.containerRegistry.path }}/{{ .Values.global.alpine_net.dir }}/{{ .Values.global.alpine_net.name }}:{{ .Values.global.alpine_net.version }}
+        imagePullPolicy: "{{ .Values.global.imagePullPolicy }}"
+        command: ['curl']
```
Demonsthere: If the test only curl kiali to see if it alive, better remove the test altogether and change this to a readiness check

resources/kiali/templates/deployment.yaml
```diff
@@ -0,0 +1,111 @@
...
+  template:
+    metadata:
+      name: {{ template "kiali.name" . }}
+      labels:
+        app: {{ template "kiali.name" . }}
+        chart: {{ template "kiali.chart" . }}
+        heritage: {{ .Release.Service }}
+        release: {{ .Release.Name }}
+      annotations:
+        sidecar.istio.io/inject: "false"
```
Demonsthere: Any reason we need to disable istio sidecar?

a-thaler: I think all that observability tools should not be inside the service-mesh, no application should have a need to talk to it.
it calls prometheus only which is also outside the mesh. Have no better argument. Do you have an argument why it should be inside?

resources/kiali/templates/tests/test-kiali-connection.yaml
a-thaler: I mainly want to assure that the full exposure is working as here most of the problems occur. 
I mainly adopted the test from the kiali chart itself and I think we should extend it in future to check that basic functionality is availble.

resources/kiali/values.yaml
a-thaler: done

resources/kiali/templates/virtualservice.yaml
a-thaler: done

a-thaler:
I re-added the kiali subchart to istio for smooth upgrades of istio. However, I still don't like the idea, a gitignore sounds much better. no-code is the best code.
@Demonsthere please check again

resources/kiali/README.md
```diff
@@ -0,0 +1,35 @@
...
+[Kiali](http://kiali.io) is a spyglass for the istio service mesh and provides answers to the questions:
+What microservices are part of my Istio service mesh? How are they connected? How are they performing?
+
+## Introduction
+
+This chart installs Kiali on a Kyma cluster.
+
+## Prerequisites
+
+- The `monitoring` chart must be installed in order to have Kiali functional
```
bszwarc:
```suggestion

-  Install the Monitoring chart to make Kiali work.
```

resources/kiali/README.md
```diff
@@ -0,0 +1,35 @@
...
+- The `jaeger` chart should be installed to have the Jaeger integration of Kiali functional
```
bszwarc:
```suggestion
- Install the Jaeger chart to integrate Jaeger with Kiali. 
```
Can we simply say `integrate` here or `to make Jaeger integration with Kiali work` sounds better? I would suggest the shorter version, but I don't want to change the meaning :)

resources/kiali/README.md
```diff
@@ -0,0 +1,35 @@
...
+- The `logging` chart should be installed to have the Grafana integration of Kiali functional
```
bszwarc:
```suggestion
- Install the Logging chart to integrate Grafana with Kiali.
```
Also the same question as above.

resources/kiali/README.md
```diff
@@ -0,0 +1,35 @@
...
+
+
+## Installing the Chart
+
+To install the chart with the release name `kiali` in namespace kyma-system call:
```
bszwarc:
```suggestion
To install the chart with the release name `kiali` in the `kyma-system` Namespace, run:
```

resources/kiali/README.md
```diff
@@ -0,0 +1,35 @@
...
+
+    ```
+    $ helm install kiali --name kiali --namespace kyma-system
+    ```
+
+## Uninstalling the Chart
```
bszwarc:
```suggestion
### Unistallation
```

resources/kiali/README.md
```diff
@@ -0,0 +1,35 @@
...
+
+To uninstall/delete the `kiali` release but continue to track the release:
```
bszwarc:
```suggestion
To uninstall/delete the `kiali` release but continue tracking the release, run:
```
If we can still track the release maybe we can simply have `unistall` here without the `delete`?

resources/kiali/README.md
```diff
@@ -0,0 +1,35 @@
...
+    ```
+    $ helm delete kiali
+    ```
+
+To uninstall/delete the `kiali` release completely and make its name free for later use:
```
bszwarc:
```suggestion
To uninstall/delete the `kiali` release completely and make its name free for further use, run:
```
Maybe we can have `delete completely` without `unistall` here to distinguish it from the previous one?

Issue #4928, opened 2019-07-17
Installation overwrite schemas for all kyma charts


a-thaler:
**Description**
Have JSON schemas for all kyma charts (incl. sub-charts) except of the ones mentioned here: https://github.com/kyma-project/kyma/issues/4764

**Reasons**
See epic

PR #4940, opened 2019-07-18
Json schemas

obadrawi:
**Description**

Adding JSON schemas to all kyma charts in the resources directory and defined in the kyma docs. related issue (#4928 )


igor-karpukhin:
Looks good, but I guess this schema for Velero is outdated @suleymanakbas91 @clebs

resources/velero/values.schema.json
```diff
@@ -0,0 +1,95 @@
+{
+    "$schema": "http://json-schema.org/draft-07/schema#",
+    "description": "Schema definition for velero helm chart values",
+    "type": "object",
+    "properties": {
+        "global": {
```
suleymanakbas91: This is outdated. Please update it before merging

obadrawi: done, please check again

dariadomagala:
Could you please also add a schema for console values? https://github.com/kyma-project/kyma/blob/master/resources/core/charts/console/values.yaml

obadrawi:
@dariadomagala  I am only creating schemas for the component charts and sub-charts specified in the kyma docs, the unspecified charts their schemas should be added by the owners themselves or should be added to the docs first

m00g3n:
Sorry for silly question but what is the purpose for having those schemas?

obadrawi:
@m00g3n  to provide validation on the values.yaml and also a description for the required parameters to be put in the values.yaml file

a-thaler:
See also linked ticket and related epic

m00g3n:
I see some potential problems:

- Value overriding is not handled
  
  e.g. `assetstore/values.yaml` overrides minio's properties
  
  ```yaml
  ...
  minio:
  image:
      tag: RELEASE.2019-06-04T01-15-58Z
  
  persistence:
      enabled: true
      size: 10Gi
  ...
  ```
  
  `assetstore/charts/minio/values.yaml`
  
  ```yaml
  ...
  image:
      repository: minio/minio
      tag: RELEASE.2019-04-09T01-22-30Z
      pullPolicy: IfNotPresent
  ...
  ```
  
- Not all `value.yaml` properties are defined in schemas
    - istio, image, service in `resources/assetstore/charts/asset-metadata-service/values.schema.json` (and other asetstore values.schema.yaml files)
    - replicaCount in `resources/assetstore/charts/asset-store-controller-manager/values.schema.json`
    - most of properties in `resources/assetstore/charts/minio/values.schema.yaml`

- Missing schemas e.g. istio, image, service

- No schema [reuse](https://json-schema.org/understanding-json-schema/structuring.html#reuse)

    e.g. in `resources/assetstore/charts/asset-upload-service/values.schema.json` and `resources/assetstore/charts/asset-metadata-service/values.schema.json`

    ```json
    "virtualservice": {
            "description": "Configuration of the virtualservice properties.",
            "type": "object",
            "properties": {
                "enabled": {
                    "description": "Enables the use of an external service. If you activate the virtualservice, it is available at https://{VIRTUALSERVICE_NAME}.{DOMAIN}/, for example at https://asset-upload-service.kyma.local/.",
                    "default": false,
                    "type": "boolean"
                },
                "annotations": {
                    "description": "Defines the service annotation.",
                    "default": {},
                    "type": "object"
                },
                "name": {
                    "description": "Defines the service name.",
                    "default": "asset-upload-service",
                    "type": "string"
                }
            }
        }
    ```

a-thaler:
@m00g3n 
Thanks for sharing your thoughts and potential problems you see.

##Goal##
Let me summarize the main driver for that story. We want to make the installation and the related configurable values (aka installer overrides) more convenient in usage. So in best case having a UI rendering input forms with description, minimal scope is to document it.
Here, it should focus on overrides relevant for the regular kyma cluster administrator, not confusing him by irrelevant technical values (like generic annotations on some kubernetes resources), still allowing any other value conform to values.yaml as generic input.
So the focus is on defining metadata for the relevant attributes per kyma module so that a user know how to configure the installation.

As we want to use a standard way of doing it, we decided to go with the potential way on how Helm 3 will do it. Unfortunately, it is still not fully specified and implemented yet in helm. I mainly found that PR https://github.com/helm/helm/pull/5350/files and validation was working in alpha1 binary but somehow I cannot find it in the sources anymore in current dev branch.

However, coming back to your points.

##syntax##
I think it is important to follow the helm way of doing it from the specification side of view. So if we support referencing schemas from sub-charts it should be in accordance to the helm way, which is not defined yet it seems

##Value overrides in parent charts##
Having an override in a parent chart mainly will repeat or reference the attribute schema of the sub-chart and potentially re-define the default value. Validation must be executed against schema in sub-chart definition always.
Referring to the asset-store parent chart for me it sounds like that you already extracted all the attributes relevant for customers into the parent chart, providing meaningful default values on asset-store functionality level. So maybe here the answer is to not provide schemas in the sub-chart but only for the parent chart and that's it.
That solution will be not valid for example for the "core" chart as it is only a bundling chart, here we should continue having a schema per sub-chart.

##Missing values##
As I said I think we should not provide metadata for all possible templating values as it will cause investment/maintenance effort with not a big value. Let's focus on the relevant ones, start simple and extend it later if meaningful. The mentioned approach of aggregating the tings in a parent chart and providing a schema only here sound good for me. For the end-user it will be anyway not relevant from which sub-chart the attribute comes from. Still he can figure it out if needed (by naming convention)

##schema re-use##
If you design your attributes around technical kubernetes resources, you probably should start re-using schemas, I agree. When trying to define attributes for a functional module like asset-store, I have my doubts that you can really re-use a schema..  


##Suggestion/Summary##
Let's check the schemas again and identify charts like "asset-store" or "monitoring" where a top-level schema might be more meaningful then many sub-chart schemas, solving the override and missing values problem automatically.
Go with that for now and see what we need to improve when starting with docu generation or other tooling.

m00g3n:
Many thx for detailed explanation Andreas.

- It is pretty clear now what is the goal and what properties should we expect to have defined in schema in first iteration.
- It looks like the value override is nothing more than schema referencing.
- It looks like we miss some best practices / guidance for schema referencing if we will not have them from the beginning we may end up with inconsistent implementation among components.  
- I also do not think that it's a good idea to reuse schemas across components. However we should reuse them in component scope e.g.

```
{
    "$schema": "http://json-schema.org/draft-07/schema#",
    "description": "Schema definition for asset-metadata-service helm subchart values",
    "type": "object",
    "properties": {
        "replicaCount": {
            "description": "Defines the number of service replicas. For more details, see the Kubernetes documentation.",
            "default": 1,
            "type": "number"
        },
        "virtualservice": {
            "$ref": "common.json#/virtualservice"
        }
    }
}
```

```
{
    "$schema": "http://json-schema.org/draft-07/schema#",
    "definitions": {
        "virtualservice": {
            "description": "Configuration of the virtualservice properties.",
            "type": "object",
            "properties": {
                "enabled": {
                    "description": "Enables the use of an external service. If you activate the virtualservice, it is available at https://{VIRTUALSERVICE_NAME}.{DOMAIN}/, for example at https://asset-upload-service.kyma.local/.",
                    "default": false,
                    "type": "boolean"
                },
                "annotations": {
                    "description": "Defines the service annotation.",
                    "default": {},
                    "type": "object"
                },
                "name": {
                    "description": "Defines the service name.",
                    "default": "asset-upload-service",
                    "type": "string"
                }
            }
        }
    }
}
```

obadrawi:
@m00g3n  going through the components I don't think it will be applicable to create the common .json file because there is not a common object between component charts and event if the element in the schema is the same between component charts the default value is different

m00g3n:
@obadrawi 
We could have the `common virtualservice` schema defined **without** the `default` property and use [combining schemas](https://json-schema.org/understanding-json-schema/reference/combining.html#combining-schemas) allOf.

Anyway as we agreed offline I will approve this PR and we will create a followup story to define common approach / best practices that we should use, follow in all components.

Disper:
Goats team will not finish the review today, but from our side you can merge it. We will apply any fixes related to our components ourselves.

PR #4942, opened 2019-07-18
ensure kubeless tests can be run in parallel with others

k15r:
**Description**

introduce retries to compensate for api-server hickups

**Related issue(s)**

#4300

### DIFF
--- a/installation/scripts/testing.sh
+++ b/installation/scripts/testing.sh
@@ -1,13 +1,13 @@
 #!/usr/bin/env bash
 ROOT_PATH=$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )
 
-CONCURRENCY=10
+CONCURRENCY=1
 
 POSITIONAL=()
 
 function validateConcurrency() {
   if [[ -z "$1" ]]; then
-    echo "Error: --concurrency requres a value"
+    echo "Error: --concurrency requires a value"
     exit 1
   fi
 
--- a/installation/scripts/testing.sh
+++ b/installation/scripts/testing.sh
@@ -1,7 +1,7 @@
 #!/usr/bin/env bash
 ROOT_PATH=$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )
 
-CONCURRENCY=1
+CONCURRENCY=10
 
 POSITIONAL=()
 
--- a/tests/kubeless/Dockerfile
+++ b/tests/kubeless/Dockerfile
@@ -1,5 +1,4 @@
-FROM alpine:3.8
-
+FROM golang:1.10-alpine as builder
 LABEL source="git@github.com:kyma-project/kyma.git"
 
 RUN apk --no-cache upgrade && apk --no-cache add curl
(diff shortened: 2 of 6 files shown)


tests/kubeless/kubeless-tests.go
```diff
@@ -4,12 +4,14 @@ import (
...
 	"io/ioutil"
-	"log"
 	"math/rand"
 	"net/http"
 	"os/exec"
 	"strings"
 	"time"
+
+	log "github.com/sirupsen/logrus"
+	"github.com/avast/retry-go"
```
antoineco: We could avoid that new dependency by using `k8s.io/apimachinery/pkg/util/wait` instead.

tests/kubeless/kubeless-tests.go
```diff
@@ -4,12 +4,14 @@ import (
...
 	"fmt"
 	"io/ioutil"
-	"log"
 	"math/rand"
 	"net/http"
 	"os/exec"
 	"strings"
 	"time"
+
+	log "github.com/sirupsen/logrus"
```
antoineco: Is switching to a new log library really justified considering the original scope?

tests/kubeless/kubeless-tests.go
```diff
@@ -230,35 +237,51 @@ func randomString(n int) string {
 	return string(b)
 }
 
-func cleanup(namespace, functionName string) {
-	log.Println("Cleaning up")
-	deleteFun(namespace, functionName)
-	deleteNamespace(namespace)
+func cleanup(namespace, functionName string) error{
+	log.Info("Cleaning up")
+	err:=deleteFun(namespace, functionName)
```
antoineco: `go fmt` is not too happy about that syntax.

tests/kubeless/kubeless-tests.go
k15r: true, i just use it as lots of other kyma components also make use of this library

tests/kubeless/kubeless-tests.go
k15r: during the tests i had to modify logging. otherwise i was not able to see the reason for the error.

tests/kubeless/kubeless-tests.go
k15r: fixed

tests/kubeless/kubeless-tests.go
antoineco: Probably because the Logger was [not initialized with an appropriate Writer](https://golang.org/pkg/log/#New) ([stderr by default](https://golang.org/src/log/log.go?s=9437:9465#L73)). No big deal for me, I just wanted to point that out.

k15r: nope, not because of that... it was just not giving enough information. But in order to not confuse people i removed those additional logs. And while i was at it, i quickly switched to logrus as it simplified implementing my needs

k15r:
> Looking good overall, providing that a concurrency of 10 doesn't make those tests flaky.
I asked @sjanota on how to proceed. Not sure if this should be enabled, yet. If we don't enable it i will revert the corresponding commit.

tests/kubeless/Dockerfile
```diff
@@ -11,11 +10,29 @@ RUN curl -Lo /usr/bin/kubectl https://storage.googleapis.com/kubernetes-release/
...
 # To automatically get the latest version:
 #RUN curl -Lo /tmp/kubeless.zip "$(curl -s https://api.github.com/repos/kubeless/kubeless/releases/latest | jq -r '.assets[]|select(.name=="kubeless_linux-amd64.zip").browser_download_url')" && unzip -q /tmp/kubeless.zip -d /tmp/ && mv /tmp/bundles/kubeless_linux-amd64/kubeless /usr/bin/ && rm -r /tmp/kubeless.zip /tmp/bundles && chmod +x /usr/bin/kubeless
 
+FROM alpine:3.8
+
+COPY --from=builder /usr/bin/kubectl /usr/bin/kubectl
+COPY --from=builder /usr/bin/kubeless /usr/bin/kubeless
+
+RUN chmod +x /usr/bin/kubeless
+RUN chmod +x /usr/bin/kubectl
```
antoineco: Both these directives should be removed, exec flags are already set explicitly in the builder.

installation/scripts/testing.sh
```diff
@@ -1,7 +1,7 @@
 #!/usr/bin/env bash
 ROOT_PATH=$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )
 
-CONCURRENCY=1
+CONCURRENCY=10
```
sjanota: Default CONCURRENCY should stay at `1`. CI plans that allows running tests concurrently will pass `--concurrency` flag to enable it.

k15r: fixed in latest commit

Issue #4754, opened 2019-07-03
Scalability of KNATIVE Publisher


akrausesap:
**Description**

Dear Colleagues,

for a custom event Gateway I was performing some scalability tests. During these tests I found out that with Kyma 1.2.2 standard setup the maximum event inbound rate is around 5-10 per second. After that the `event-bus-publish-knative-*-*` pods start crashing with OOM errors.

![2019-07-03_14-07-52](https://user-images.githubusercontent.com/40229965/60590394-27d15500-9d9c-11e9-87fc-72aa56c90f8f.png)

Should you require the entire test setup and further details please let me know.

Regards

Andreas

**Expected result**

Higher sustained throughput

**Actual result**

Throughput plateaus at 5-10 events per second

**Steps to reproduce**

Initiate loadtest for the http://event-bus-publish.kyma-system.svc.cluster.local:8080/v1/events endpoint.

**Troubleshooting**

N/A

abbi-gaurav:
Did some quick testing today and it seems to point out that the publish and nats-dispatcher might have a memory leak that could be contributing to the problem at hand.

My test scenario was pretty simple.

**1st run**
request per second = 10 with a concurrency of 10. It failed with OOM after around 30 minutes.

**2nd run**
request per second = 8 with a concurrency of 8. The memory build up keeps on happening. I stopped after around 25 minutes to see if the GC cleans up. a very little memory was released.

Here are the snapshots.

Publish app which points to a memory leak.
![Screenshot 2019-07-10 at 17 20 26](https://user-images.githubusercontent.com/4436741/60981957-959af500-a337-11e9-9906-163dce100100.png)

Nats dispatcher.
The memory footprint is small but there is a continuous build up.
![Screenshot 2019-07-10 at 17 21 22](https://user-images.githubusercontent.com/4436741/60982056-bb27fe80-a337-11e9-8532-98362a83dc66.png)

Event service
Memory footprint is really small. Whatever buildup happened during the run was released.
![Screenshot 2019-07-10 at 17 21 11](https://user-images.githubusercontent.com/4436741/60982158-e3176200-a337-11e9-94c2-d4f844c8fc9a.png)

abbi-gaurav:
I used [npm package loadtest](https://www.npmjs.com/package/loadtest) for simulating the load. It was the same originally used by @akrausesap (thanks for that).

montaro:
[...]

PR #4943, opened 2019-07-18
Scalability of knative publisher

montaro:
See also  #4754

### DIFF
--- a/components/event-bus/internal/knative/util/kn-eventing.go
+++ b/components/event-bus/internal/knative/util/kn-eventing.go
@@ -215,7 +215,10 @@ func (k *KnativeLib) SendMessage(channel *evapisv1alpha1.Channel, headers *map[s
 	}
 	defer func() {
 		_ = res.Body.Close()
-		tran, _ := k.httpClient.Transport.(*http.Transport)
+		tran, ok := k.httpClient.Transport.(*http.Transport)
+		if !ok {
+			return
+		}
 		tran.CloseIdleConnections()
 	}()
 
--- a/components/event-bus/internal/knative/util/kn-eventing.go
+++ b/components/event-bus/internal/knative/util/kn-eventing.go
@@ -8,6 +8,7 @@ import (
 	"log"
 	"net/http"
 	"net/http/httputil"
+	"sync"
 	"time"
 
 	evapisv1alpha1 "github.com/knative/eventing/pkg/apis/eventing/v1alpha1"
@@ -54,6 +55,8 @@ if err := k.CreateSubscription("my-sub", namespace, channelName, &uri); err != n
 return
 */
 
+var once sync.Once
+
 // KnativeAccessLib encapsulates the Knative access lib behaviours.
 type KnativeAccessLib interface {
 	GetChannel(name string, namespace string) (*evapisv1alpha1.Channel, error)
@@ -75,7 +78,8 @@ func NewKnativeLib() (KnativeAccessLib, error) {
 
 // KnativeLib represents the knative lib.
 type KnativeLib struct {
-	evClient eventingv1alpha1.EventingV1alpha1Interface
+	evClient   eventingv1alpha1.EventingV1alpha1Interface
+	httpClient http.Client
 }
 
 // Verify the struct KnativeLib implements KnativeLibIntf
@@ -96,6 +100,11 @@ func GetKnativeLib() (*KnativeLib, error) {
 	k := &KnativeLib{
 		evClient: evClient.EventingV1alpha1(),
 	}
+	once.Do(func() {
+		k.httpClient = http.Client{
+			Transport: initHTTPTransport(),
+		}
+	})
 	return k, nil
 }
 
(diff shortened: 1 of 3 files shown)


components/event-bus/internal/knative/util/kn-eventing.go
```diff
@@ -264,7 +278,11 @@ func resendMessage(httpClient *http.Client, channel *evapisv1alpha1.Channel, hea
 				log.Printf("ERROR: resendMessage(): could not resend HTTP request: %v", err)
 				return err
 			}
-			defer func() { _ = res.Body.Close() }()
+			defer func() {
+				_ = res.Body.Close()
+				tran, _ := httpClient.Transport.(*http.Transport)
```
sayanh: I think `tran.CloseIdleConnections()` needs to be in a separate `defer` right after `httpClient.Do(req)`.
If `httpClient.Do(req)` returns an error then we wouldn't be closing idle connections.

components/event-bus/internal/knative/util/kn-eventing.go
```diff
@@ -192,27 +201,30 @@ func (k *KnativeLib) UpdateSubscription(sub *evapisv1alpha1.Subscription) (*evap
...
+	res, err := k.httpClient.Do(req)
 	if err != nil {
 		log.Printf("ERROR: SendMessage(): could not send HTTP request: %v", err)
 		return err
 	}
 	defer func() {
 		_ = res.Body.Close()
+		tran, ok := k.httpClient.Transport.(*http.Transport)
+		if !ok {
+			return
```
sayanh: I think this is more readable:
```
if tran, ok := k.httpClient.Transport.(*http.Transport); ok {
  tran.CloseIdleConnections()
}
```

components/event-bus/internal/knative/util/kn-eventing.go
```diff
@@ -245,6 +257,8 @@ func resendMessage(httpClient *http.Client, channel *evapisv1alpha1.Channel, hea
 	}
 	defer func() {
 		_ = res.Body.Close()
+		tran, _ := httpClient.Transport.(*http.Transport)
```
sayanh: If typecasting fails then this will panic. Let's use the `ok` pattern here.

components/event-bus/internal/knative/util/kn-eventing.go
montaro: Honestly, the first one is more readable to me :-)

components/event-bus/internal/knative/util/kn-eventing.go
montaro: @sayanh Thanks for the catch, this was missed in the refactor, will fix it!

components/event-bus/internal/knative/util/kn-eventing.go
montaro: `defer()` should be called after `resendMessage()` exits, right?

components/event-bus/internal/knative/util/kn-eventing.go
montaro: I tried the other way but, I got a compilation error as `tran` would be unused var.

components/event-bus/internal/knative/util/kn-eventing.go
montaro: fixed

components/event-bus/internal/knative/util/kn-eventing.go
montaro: fixed as agreed

montaro:
```
root@my-shell-65ffc94dc4-4gvnb /# ab -n 5000 -c 100 -p data -T application/json http://10.0.23.71:8080/v2/events
This is ApacheBench, Version 2.3 <$Revision: 1807734 $>
Copyright 1996 Adam Twiss, Zeus Technology Ltd, http://www.zeustech.net/
Licensed to The Apache Software Foundation, http://www.apache.org/

Benchmarking 10.0.23.71 (be patient)
Completed 500 requests
Completed 1000 requests
Completed 1500 requests
Completed 2000 requests
Completed 2500 requests
Completed 3000 requests
Completed 3500 requests
Completed 4000 requests
Completed 4500 requests
Completed 5000 requests
Finished 5000 requests


Server Software:
Server Hostname:        10.0.23.71
Server Port:            8080

Document Path:          /v2/events
Document Length:        130 bytes

Concurrency Level:      100
Time taken for tests:   499.097 seconds
Complete requests:      5000
Failed requests:        0
Total transferred:      1195000 bytes
Total body sent:        1795000
HTML transferred:       650000 bytes
Requests per second:    10.02 [#/sec] (mean)
Time per request:       9981.948 [ms] (mean)
Time per request:       99.819 [ms] (mean, across all concurrent requests)
Transfer rate:          2.34 [Kbytes/sec] received
                        3.51 kb/s sent
                        5.85 kb/s total

Connection Times (ms)
              min  mean[+/-sd] median   max
Connect:        0    0   0.5      0       9
Processing:   376 9874 1061.7   9619   10865
Waiting:      376 9874 1061.7   9619   10865
Total:        379 9874 1061.4   9620   10865

Percentage of the requests served within a certain time (ms)
  50%   9620
  66%  10404
  75%  10580
  80%  10587
  90%  10595
  95%  10600
  98%  10630
  99%  10671
 100%  10865 (longest request)
```
image
image

resources/event-bus/values.yaml
```diff
@@ -6,8 +6,8 @@ global:
   containerRegistry:
     path: eu.gcr.io/kyma-project
   event_bus:
-    dir: develop/
-    version: "36059183"
+    dir: pr/
```
sayanh: The changes in values.yaml can be in separate PR if you wanna avoid merging an image with PR tag.

montaro: I've seen folks doing following this approach recently

PR #4958, opened 2019-07-21
Test unification

a-thaler:
**Description**

- Unified file naming
- Unified test resource naming
- Unified test resurce laelling


docs/kyma/03-02-testing.md
```diff
@@ -40,15 +40,19 @@ The following example presents TestDefinition with a container that calls the De
...
+    heritage: {{ .Release.Service }}
 spec:
   template:
     metadata:
       annotations:
         sidecar.istio.io/inject: "false"
     spec:
       containers:
-      - name: "test-{{ template "fullname" . }}-connection-dex"
+      - name: {{ .Chart.Name }}-tests
```
sjanota: I would go with just `test`. Or `main`. It has to be unique at pod scope so let's keep it short and simple.

docs/kyma/03-02-testing.md
```diff
@@ -40,15 +40,19 @@ The following example presents TestDefinition with a container that calls the De
 apiVersion: "testing.kyma-project.io/v1alpha1"
 kind: TestDefinition
 metadata:
-  name: "test-{{ template "fullname" . }}-connection-dex"
+  name: {{ .Chart.Name }}
+  labels:
+    app: {{ .Chart.Name }}-tests
+    release: {{ .Release.Name }}
+    heritage: {{ .Release.Service }}
```
sjanota: Maybe we can also clean up the labels? Helm [suggestions](https://helm.sh/docs/chart_best_practices/#labels-and-annotations) regarding labels are different.

docs/kyma/03-02-testing.md
a-thaler: done, used `tests`

docs/kyma/03-02-testing.md
a-thaler: I applied the suggestions. However I still added `app`as functionality might break without.

resources/application-connector/charts/connection-token-handler/templates/tests/test.yaml
```diff
@@ -2,10 +2,13 @@
 apiVersion: "testing.kyma-project.io/v1alpha1"
 kind: TestDefinition
 metadata:
-  name: {{ .Chart.Name }}-tests
+  name: application-{{ .Chart.Name }}
```
Szymongib: Shouldn't it be just the `name: {{ .Chart.Name }}`?

a-thaler: I thought it might be nice to prefix it with the main chart name like for the "core" charts. But application-connector-connection-xxx-xx" sounded strange.
But if you like I can change it

Szymongib: I do not have a particular preference, the thing is that it is just different than in the other charts under the `application-connector` like here https://github.com/kyma-project/kyma/pull/4958/files/7d32049154b5a8cc3de8ae306ef60e439c30a7f8#diff-ab13db9dc84bd2067c299d2e4e071abeR6 or here https://github.com/kyma-project/kyma/pull/4958/files/7d32049154b5a8cc3de8ae306ef60e439c30a7f8#diff-7b858abc7c31b5ba35e6505af3f45c1fR6 so I think we should stick to one convention.

a-thaler: In the references the sub-charts starting with "application-xx" so I thought to better not call it "application-application-xxx"

Either we use the plain sub chart name or we prefix it all with "application-connector"

Szymongib: Fair point.
Then I would suggest using a plain sub chart name.

a-thaler: fixed

PR #4956, opened 2019-07-19
Refactor eventing components: resources update

anishj0shi:
**Description**
As per #2502 certain components from the event-bus have to renamed and this PR proposes changes to comply that.

Changes proposed in this pull request:

-  Update event bus helm charts to comply with the event-bus new naming conventions 
-  update pre-upgrade knative-eventing script to delete old helm resources 
- Bump the image version for event-bus components
**Related issue(s)**

w.r.t #3024

PR #5044, opened 2019-07-26
Update event publish url references in tests: source update

anishj0shi:
As per #4956. event-publish-service urls have been modified to http://event-publish-service:8080/v1/events


tests/end-to-end/upgrade/pkg/tests/event-bus/event_bus.go
```diff
@@ -98,10 +98,6 @@ func (f *eventBusFlow) createResources() error {
 		f.createSubscription,
 		f.createSubscriber,
 		f.checkSubscriberStatus,
```
sayanh: There is an issue https://github.com/kyma-project/kyma/issues/5056 to add this back again.

Issue #5037, opened 2019-07-26
ClusterTestSuite resources are not removed


piotrmiskiewicz:
**Description**

ClusterTestSuites (and all Pods created from it) exists but should be removed, see:
```
kyma-system        oct-tp-testsuite-all-2019-07-26-03-11-assetstore-0                 0/2     Completed   0          65m
kyma-system        oct-tp-testsuite-all-2019-07-26-03-11-core-0                       0/2     Completed   0          56m
kyma-system        oct-tp-testsuite-all-2019-07-26-03-11-core-1                       0/2     Completed   0          53m
kyma-system        oct-tp-testsuite-all-2019-07-26-03-11-core-api-controller-0        0/1     Completed   0          135m
kyma-system        oct-tp-testsuite-all-2019-07-26-03-11-core-apiserver-proxy-0       0/2     Completed   0          131m
kyma-system        oct-tp-testsuite-all-2019-07-26-03-11-core-cluster-users-0         0/2     Completed   0          136m
kyma-system        oct-tp-testsuite-all-2019-07-26-03-11-core-console-0               0/3     Completed   0          64m
kyma-system        oct-tp-testsuite-all-2019-07-26-03-11-core-kubeless-0              0/2     Completed   0          133m
kyma-system        oct-tp-testsuite-all-2019-07-26-03-11-dex-connection-0             0/1     Completed   0          136m
kyma-system        oct-tp-testsuite-all-2019-07-26-03-11-dex-integration-0            0/2     Completed   0          132m
kyma-system        oct-tp-testsuite-all-2019-07-26-03-11-e2e-kubeless-0               0/2     Error       0          127m
kyma-system        oct-tp-testsuite-all-2019-07-26-03-11-e2e-kubeless-1               0/2     Error       0          118m
kyma-system        oct-tp-testsuite-all-2019-07-26-03-11-event-bus-0                  0/3     Error       0          50m
kyma-system        oct-tp-testsuite-all-2019-07-26-03-11-event-bus-1                  0/3     Error       0          46m
kyma-system        oct-tp-testsuite-all-2019-07-26-03-11-kiali-0                      0/1     Completed   0          132m
kyma-system        oct-tp-testsuite-all-2019-07-26-03-11-logging-0                    0/1     Completed   0          136m
kyma-system        oct-tp-testsuite-all-2019-07-26-03-11-monitoring-0                 0/2     Completed   0          129m
[...]

PR #5049, opened 2019-07-26
Add retries to kubectl delete cts command

piotrmiskiewicz:
**Description**

- Add retries for deleting ClusterTestSuite at the end of testing phase

**Related issue(s)**
Resolves: #5037

### DIFF
--- a/installation/scripts/testing-common.sh
+++ b/installation/scripts/testing-common.sh
@@ -152,6 +152,28 @@ function waitForTestPodsTermination() {
     return 1
 }
 
+function deleteCTS() {
+  local retry=0
+  local suiteName=$1
+
+  log "Deleting ClusterTestSuite ${suiteName}" nc bold
+  while [[ ${retry} -lt 20 ]]; do
+        msg=$(kubectl delete cts ${suiteName} 2>&1)
+        status=$?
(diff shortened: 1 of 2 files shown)


installation/scripts/testing-common.sh
```diff
@@ -152,6 +152,26 @@ function waitForTestPodsTermination() {
     return 1
 }
 
+function deleteCTS() {
+  local retry=0
+  local suiteName=$1
+
+  log "Deleting ClusterTestSuite ${suiteName}" nc bold
+  while [[ ${retry} -lt 20 ]]; do
+        kubectl delete cts ${suiteName}
```
Demonsthere: Doing this this way, we lose the info about what happened, what error occurred, please do it like this:
```bash
msg=$(kubectl foo 2>&1)
status=$?
```
Then check status, and log msg. In case of cts we might fail due to too long grace period (it is by default determined on the size of the resource). Might want to consider adding `--now=true --wait=true`

piotrmiskiewicz: there is no need to wait for the deletion and there is no need for "immediate shutdown". It is important to delete the CTS in a near future

PR #5089, opened 2019-07-30
Cbs integration tests

sjanota:
**Description**

*No support for BackendModules yet*
 

**Related issue(s)**


components/console-backend-service/integration/main_test.go
```diff
@@ -0,0 +1,158 @@
...
+	environment := &envtest.Environment{
+		KubeAPIServerFlags: append([]string{"--authorization-mode=RBAC"}, envtest.DefaultKubeAPIServerFlags...),
+	}
+	restConfig, err = environment.Start()
+	if err != nil {
+		panic(err)
+	}
+
+	_, err = envtest.InstallCRDs(restConfig, envtest.CRDInstallOptions{
+		Paths:              []string{"../../../resources/cluster-essentials/templates/crds"},
```
pkosiec: Also all BackendModule CRs should be applied as well to enable all CBS modules

components/console-backend-service/integration/main_test.go
```diff
@@ -0,0 +1,158 @@
...
+var (
+	restConfig    *rest.Config
+	gqlEndpoint   string
+	rbacClientset *rbacClient.RbacV1Client
+)
+
+func TestMain(m *testing.M) {
+	// setup and start kube-apiserver
+	var err error
+	environment := &envtest.Environment{
```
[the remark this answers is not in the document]
sjanota: As I understood it was about the environment getting up slowly. The increased timeout did the trick.

components/console-backend-service/integration/main_test.go
sjanota: 👍 I would like to add them gradually while migrating tests.

sjanota:
@pkosiec In cluster integration suite will be:
* test for authentication
* smoke test to check if chart configuration is valid
* any test that depends on logic from other Kyma controllers (if any)

sjanota:
this is blocked for now, as controller-runtime does not support k8s 1.5 yet.

sjanota:
It needs to be refined and rebase with the current master. I'm closing it for now.

Issue #5039, opened 2019-07-26
Add integration test for checking git feature in Helm Broker


mszostok:
**Description**

Add integration test for checking git feature in Helm Broker

**AC**

- HB Git feature is fully covered by integration tests

PR #5114, opened 2019-07-31
Add integration test for (Cluster)AddonConfiguration git feature

jasiu001:
Related to #5039


components/helm-broker/test/integration/integration_test.go
```diff
@@ -6,151 +6,305 @@ import (
...
+		})
+	}
+
+	for name, c := range map[string]struct {
+		kind      string
+		addonName string
+		redisID   string
+		testID    string
+	}{
+		"cluster": {
```
polskikiel: `cluster-http`

Issue #5113, opened 2019-07-31
Update documentation for Knative serving and overview file


rakesh-garimella:
**Description**
Update the documentation for knative-serving to make it experimental for the function controller. Also update the overview files to reflect all the modules.

- [x] Update Knative Serving and knative Build documentation
- [x] Updated overview file to have: kantive-serving, knative-eventing
- [x] Make knative-build disabled by default
- [x] Update the image for Knative-build tests.

PR #5116, opened 2019-07-31
Add missing modules to documentation

rakesh-garimella:
**Description**

* Update documentation to represent all the modules
* Image Bump for knative build acceptance test.

**Related issue(s)**

#5113


resources/knative-build/values.yaml
```diff
@@ -5,5 +5,5 @@ global:
   containerRegistry:
     path: eu.gcr.io/kyma-project
   test_knative_build_acceptance:
-    dir: pr/
-    version: PR-4606
+    dir: develop/
```
anishj0shi: Hi Rakesh,
could you please mention this image bump in the proposed changes for this PR as well and map it to the component change in the PR, so that we have reference to the actual change as well.

rakesh-garimella: done.

Issue #5165, opened 2019-08-02
 core-cluster-users


Disper:
**Description**
Test status: core-cluster-users is failing often in the CI.

**Expected result**

Test is stable.

**Actual result**

Test is not stable.

**Steps to reproduce**

**Troubleshooting**

Please see following prow jobs:
- https://status.build.kyma-project.io/view/gcs/kyma-prow-logs/pr-logs/pull/kyma-project_kyma/5117/pre-master-kyma-gke-integration/1156910324880773120 
- https://status.build.kyma-project.io/view/gcs/kyma-prow-logs/logs/post-master-kyma-gke-integration/1157151665262956545
- https://status.build.kyma-project.io/view/gcs/kyma-prow-logs/logs/post-master-kyma-gke-integration/1157151665262956545

PR #5178, opened 2019-08-02
Add retries to test to improve resilience

Tomasz-Smelcerz-SAP:
**Description**

Add retries to network operations to improve test resilience

- Add retries to network operations (kubectl, curl)

**Related issue(s)**
Implements #5165


tests/integration/cluster-users/sar-test.sh
```diff
@@ -11,30 +11,142 @@
 #  - VIEW_PASSWORD: password for the view user
 #  - NAMESPACE: namespace in which we perform the tests
```
piotrmsc: Those should be also taken from ENV

Issue #5053, opened 2019-07-26
Register default Runtime


PK85:
**Description**

Compass chart, will register a default runtime and create configuration that will be used by compass-runtime-agent.

**AC**
- Create configMap in the compass namespace, content:
  - URL to Director endpoint
  - tenant
  - runtimeId

PR #5218, opened 2019-08-05
Agent configuration job

PK85:
**Description**

- new job in compass chart
- job is responsible to register default kyma runtime in compass and create config map for runtime agent

**Related issue(s)**
#5053

### DIFF
--- a/resources/compass/templates/agent-configuration-job.yaml
+++ b/resources/compass/templates/agent-configuration-job.yaml
@@ -1,0 +1,121 @@
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: compass-agent-configuration
+  namespace: {{ .Release.Namespace }}
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: RoleBinding
+metadata:
+  name: compass-agent-configuration
+  namespace: {{ .Release.Namespace }}
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: Role
+  name: compass-agent-configuration
+subjects:
+  - kind: ServiceAccount
+    name: compass-agent-configuration
+    namespace: {{ .Release.Namespace }}
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  name: compass-agent-configuration
+  namespace: {{ .Release.Namespace }}
+rules:
+  - apiGroups: [""]
+    resources: [configmaps]
+    verbs: [create, get]
+---
+apiVersion: batch/v1
+kind: Job
+metadata:
+  name: compass-agent-configuration
+  namespace: {{ .Release.Namespace }}
+  annotations:
+    "helm.sh/hook": post-install,post-upgrade
+    "helm.sh/hook-weight": "1"
+    "helm.sh/hook-delete-policy": before-hook-creation
+spec:
+  backoffLimit: 0
+  template:
+    metadata:
+      name: compass-agent-configuration
+      annotations:
+        sidecar.istio.io/inject: "false"
+    spec:
+      {{ if .Values.global.isLocalEnv }}
+      hostAliases:
+        - ip: {{ .Values.global.minikubeIP }}
+          hostnames:
+            - "{{ .Values.global.gateway.host }}.{{ .Values.global.ingress.domainName }}"
+      {{ end }}
+      serviceAccountName: compass-agent-configuration
+      restartPolicy: Never
+      containers:
+        - name: compass-agent-configuration
+          image: "linkyard/kubectl:1.14.2"
+          command:
+            - bash
+            - -c
+            - |
+              MAX_RETRIES=60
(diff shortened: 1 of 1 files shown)


resources/compass/templates/agent-configuration-job.yaml
```diff
@@ -0,0 +1,104 @@
...
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  name: compass-agent-configuration
+  namespace: compass-system
+rules:
+  - apiGroups: [""]
+    resources: [configmaps]
+    verbs: [create, list, get, watch]
```
pkosiec: You don't need other permissions than creating configmaps, so I would remove other verbs

resources/compass/templates/agent-configuration-job.yaml
```diff
@@ -0,0 +1,104 @@
...
+                  set -o errexit
+              }
+              checkIfDirectorReady
+
+              export RUNTIME_ID=`curl "${DIRECTOR_URL}" \
+                  -H 'Content-Type: application/json' \
+                  -H 'tenant: {{ .Values.global.defaultTenant }}' \
+                  --data-binary '{
+                    "query":"mutation createRuntime {\n  result: createRuntime(\n    in: { name: \"kymaruntime\", description: \"Default Kyma runtime\" }\n  ) {\n    id\n  }\n}\n"
+                  }' | jq '.data .result .id'`
```
pkosiec: What if there will be a different status code than 200? The script should print error and exit

PK85: done, when jq checks a string, in case of failure the whole job will fail

resources/compass/templates/agent-configuration-job.yaml
```diff
@@ -0,0 +1,122 @@
...
+                        sleep 5
+                      fi
+                  done
+
+                  set -o errexit
+              }
+
+              kubectl get configmaps compass-agent-configuration -n compass-system > /dev/null 2>&1
+              if [[ $? -eq 1 ]]
+              then
```
pkosiec: Please invert the condition and don't do else:
```bash
   if [[ $? -eq 0 ]]; then
      echo "Compass agent already configured and paired."
       exit 0
   fi

   ...
```

The code will be much simpler

resources/compass/templates/agent-configuration-job.yaml
```diff
@@ -0,0 +1,122 @@
...
+                 set +o errexit
+                 echo "Compass agent configuration - in progress."
+                 checkIfDirectorReady
+
+                 RESPONSE_BODY=$(curl -k "${DIRECTOR_URL}" \
```
pkosiec: Please print response body, it will allow us to debug the code

resources/compass/templates/agent-configuration-job.yaml
```diff
@@ -0,0 +1,122 @@
...
+spec:
+  backoffLimit: 0
+  template:
+    metadata:
+      name: compass-agent-configuration
+      annotations:
+        sidecar.istio.io/inject: "false"
+    spec:
+      {{ if .Values.global.isLocalEnv }}
+      hostNetwork: true
```
pkosiec: Please see if the `hostNetwork` is really necessary - `nats-streaming` and `asset-store` migration jobs don't need them.

resources/compass/templates/agent-configuration-job.yaml
```diff
@@ -0,0 +1,122 @@
...
+      hostAliases:
+        - ip: {{ .Values.global.minikubeIP }}
+          hostnames:
+            - "{{ .Values.global.gateway.host }}.{{ .Values.global.ingress.domainName }}"
+      {{ end }}
+      shareProcessNamespace: true
```
pkosiec: Do you need that?

resources/compass/templates/agent-configuration-job.yaml
```diff
@@ -0,0 +1,122 @@
...
+      serviceAccountName: compass-agent-configuration
+      restartPolicy: Never
+      containers:
+        - name: compass-agent-configuration
+          image: "linkyard/kubectl:1.14.2"
```
pkosiec: Please try it out: `image: "{{ .Values.global.containerRegistry.path }}/{{ .Values.global.alpine_kubectl.dir }}{{ .Values.global.alpine_kubectl.image }}:{{ .Values.global.alpine_kubectl.version }}"`

resources/compass/templates/agent-configuration-job.yaml
PK85: not anymore. Was needed when job has istio sidecar.

resources/compass/templates/agent-configuration-job.yaml
PK85: Not needed.  Tests with current setup. 

Before I had istio sidecar and somehow that setting has solved the issue with connecting to API-Server.

resources/compass/templates/agent-configuration-job.yaml
PK85: environment: line 45: jq: command not found

PR #5236, opened 2019-08-06
Disable URL path escape in the gateway proxy

### LINKED ISSUE #5164: Odata apis are URL encoded by Kyma causing the API call to fail
**Description**
I am seeing problems when making the Odata API via the application gateway.

It seems the URLs of the format `${gateway-url}/Object('some-id')` are getting encoded into `https://<my.tenant>.com/Object%28%27some-id%27%29`. This causes the api call to fail with 404. 

I confirmed that this encoding is being done by the application gateway by making a direct call to the application from the same lambda which works fine.

**Expected Result**
Able to call Odata APIs from the lambda code.

**Actual result**

Call fails with possibly application gateway doing a URL encoding.

```
time="2019-08-01T14:00:59Z" level=info msg="Modified request url : 'https://<my-tenant>/sap/c4c/odata/v1/c4codataapi/CorporateAccountCollection%28%2700163E7841DC1ED9ACD99799567445B8%27%29?$select=CurrentDefaultAddressUUID,Name,CountryCodeText,HouseNumber,Street,StreetPostalCode,City', schema : 'https', path : '/sap/c4c/odata/v1/c4codataapi/CorporateAccountCollection('00163E7841DC1ED9ACD99799567445B8')'"
```

crabtree:
**Description**

- Disable URL path escape in the gateway proxy

**Related issue(s)**

Fixes #5164

### DIFF
--- a/components/application-gateway/internal/proxy/proxy_test.go
+++ b/components/application-gateway/internal/proxy/proxy_test.go
@@ -115,6 +115,51 @@ func TestProxy(t *testing.T) {
 		csrfStrategyMock.AssertExpectations(t)
 	})
 
+	t.Run("should proxy without escaping the URL path characters when target URL contains full path", func(t *testing.T) {
+		// given
+		ts := NewTestServer(func(req *http.Request) {
+			assert.Equal(t, "/somepath/Xyz('123')", req.URL.String())
+		})
+		defer ts.Close()
+
+		req, err := http.NewRequest(http.MethodGet, "?$search=XXX", nil)
+		require.NoError(t, err)
+
+		req.Host = "app-test-uuid-1.namespace.svc.cluster.local"
+
+		authStrategyMock := &authMock.Strategy{}
+		authStrategyMock.
+			On("AddAuthorization", mock.AnythingOfType("*http.Request"), mock.AnythingOfType("TransportSetter")).
+			Return(nil).
+			Once()
+
+		credentials := &authorization.Credentials{}
+		authStrategyFactoryMock := &authMock.StrategyFactory{}
+		authStrategyFactoryMock.On("Create", credentials).Return(authStrategyMock).Once()
+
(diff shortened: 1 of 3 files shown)


components/application-gateway/internal/proxy/proxy_test.go
```diff
@@ -26,6 +25,96 @@ func TestProxy(t *testing.T) {
...
+
+		credentials := &authorization.Credentials{}
+		authStrategyFactoryMock := &authMock.StrategyFactory{}
+		authStrategyFactoryMock.On("Create", credentials).Return(authStrategyMock).Once()
+
+		csrfFactoryMock, csrfStrategyMock := mockCSRFStrategy(authStrategyMock, calledOnce)
+
+		serviceDefServiceMock := &metadataMock.ServiceDefinitionService{}
+		serviceDefServiceMock.On("GetAPI", "uuid-1").Return(&metadatamodel.API{
+			TargetUrl:   ts.URL + "/somepath",
```
Szymongib: I think it would be good to also use the part that was being encoded in the target URL, which would be then `ts.URL + "/somepath/Xyz('123')`.
What do you think?

crabtree: Done

akrausesap:
looking at the fix, it somehow feels like it should be:

```
req.URL.RawPath =  joinPaths(target.RawPath, req.URL.RawPath)
req.URL.Path =  joinPaths(target.Path, req.URL.Path)
```
I suppose this avoids unintended side effects where path contains parameters that definitely need to be encoded

crabtree:
See your point @akrausesap but the target URL in the `makeProxy` func can have `RawPath` empty and in the same time Path is filled with the appropriate value. In this case there again will be the difference between the `Path` and `RawPath` and the encoding will be triggered.

Issue #4895, opened 2019-07-13
Tests missing for publish events using v2 API


sayanh:
**Description**

Following tests need to be added for publish-service using v2 API:
- upgrade tests(tests/end-to-end/upgrade/pkg/tests/event-bus/event_bus.go)
- event-bus e2etester(tests/event-bus/e2e-tester/e2e-tester.go)

**AC**

- Upgrade tests include v2 tests.
- Event bus integration tests `tests/event-bus/e2e-tester/e2e-tester.go` includes v2 tests.

sayanh:
Related example: https://github.com/kyma-incubator/examples/tree/master/cloud-events-e2e-scenario

PR #5289, opened 2019-08-12
Extend the EventBus test utils to allow creating named subscriber deployments and services

marcobebway:
**Description**

- Extend the EventBus test utils to allow creating named subscriber deployments and services.

**Related issue(s)**
- #4895

### DIFF
--- a/components/event-bus/test/util/subscriber_yaml.go
+++ b/components/event-bus/test/util/subscriber_yaml.go
@@ -7,28 +7,33 @@ import (
 )
 
 const (
-	// SubscriberName used in the tests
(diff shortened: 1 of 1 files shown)


antoineco:
Code OK, however the original issue was "missing tests...", but this doesn't add any test. Are we that in a hurry that we need to do that small refactoring in a separate PR? Can't we just ship all changes as one work unit not to lose the context?

marcobebway:
Agreed not to merge.

marcobebway:
Reopen to extend the EventBus test subscriber.

PR #5355, opened 2019-08-19
Prepare Kyma 1.4.1 patch release

nachtmaar:
**Description**

- Kyma 1.4.1 release
- Fix initialization of dynamic client #5340
- Use newer console image to fix fetching cmfs via apiserver
- Fix vendored golint 

**Related issue(s)**

https://github.com/kyma-project/kyma/pull/5340


resources/compass/values.yaml
```diff
@@ -4,23 +4,23 @@ global:
       path: eu.gcr.io/kyma-project/incubator
     connector:
       dir:
-      version: 1.4.0
+      version: 1.4.1
```
Szymongib: I am not sure if you can bump images of `compass` as you do not create release branch for it, so the images will not be built.

nachtmaar:
`pre-rel14-kyma-components-function-controller` is optional. I informed @rakesh-garimella about the problem

resources/compass/values.yaml
pkosiec: The PR is already created on compass repo, so all images have been built.

nachtmaar: @Szymongib Please take another look now :)

Issue #5168, opened 2019-08-02
Upgrade Event Bus dependencies to Kubernetes 1.15


Abd4llA:
**Description**

Upgrade Event Bus subscription controller to Kubernetes 1.15 . 

**Notes**

- Coordinated with other components *UPFRONT* regarding merging strategy. 

**AC**

- Subscription controller starts error free on a Kubernetes  1.15 cluster.
- Event bus publish starts error free on a Kubernetes 1.15 cluster.
- Event bus acceptance tests pass on a Kubernetes 1.15 cluster.
- Event bus upgrade tests are passing correctly.

marcobebway:
@montaro, why these [tests](https://github.com/kyma-project/kyma/blob/cc65bfa42a855aa4f70549a460a83a8e6681b86f/components/event-bus/internal/knative/subscription/controller/subscription/reconcile_test.go#L109) were commented out? Please reactivate them and make sure that they are working correctly.

marcobebway:
Moving back to todo, because these [tests](https://github.com/kyma-project/kyma/blob/cc65bfa42a855aa4f70549a460a83a8e6681b86f/components/event-bus/internal/knative/subscription/controller/subscription/reconcile_test.go#L109) were commented by mistake. The tests should be uncommented and verified to be working correctly.

derberg:
Removed from epic https://github.com/kyma-project/kyma/issues/5077 as it is not related

montaro:
Blocked by #5058

Abd4llA:
unstalling

anishj0shi:
event-bus deps are already on 1.15.3

PR #5356, opened 2019-08-19
Upgrade event bus dependencies to kubernetes 1.12.6

montaro:
See also #5168

### DIFF
--- a/components/event-bus/cmd/event-publish-service/handlers/publish.go
+++ b/components/event-bus/cmd/event-publish-service/handlers/publish.go
@@ -3,7 +3,7 @@ package handlers
 import (
 	"encoding/json"
 
-	"github.com/kyma-project/kyma/components/event-bus/api/publish/v1"
+	publishv1 "github.com/kyma-project/kyma/components/event-bus/api/publish/v1"
 	publishv2 "github.com/kyma-project/kyma/components/event-bus/api/publish/v2"
 
 	"log"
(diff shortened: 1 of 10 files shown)


components/event-bus/cmd/event-publish-service/handlers/publish.go
```diff
@@ -3,7 +3,7 @@ package handlers
 import (
 	"encoding/json"
 
-	"github.com/kyma-project/kyma/components/event-bus/api/publish/v1"
+	v1 "github.com/kyma-project/kyma/components/event-bus/api/publish/v1"
```
anishj0shi: i see there's a reference name "publishv2" can we be consistent here ?

components/event-bus/internal/knative/util/kn-eventing_test.go
```diff
@@ -163,7 +167,8 @@ func Test_SendMessage(t *testing.T) {
 	assert.Nil(t, err)
 	u, err := url.Parse(srv.URL)
 	assert.Nil(t, err)
-	ch.Status.SetAddress(u.Hostname() + ":" + u.Port())
+	url := &apis.URL{Scheme: "http", Host: fmt.Sprintf("%s:%s", u.Hostname(), u.Port())}
+	ch.Status.SetAddress(url)
```
anishj0shi: for readability, can we create "url" as a value and pass it's reference in SetAddress()? something like SetAddress(&url)

montaro: Thanks for the comment.
IMHO, it's readable and I'd not also change the signature of the `SetAddress()`

resources/event-bus/charts/event-publish-service/templates/deployment.yaml
```diff
@@ -1,4 +1,4 @@
-apiVersion: extensions/v1beta1
```
sayanh: May be, we can move these changes to another PR(will be created anyway) which will have the bumped image of event-bus.

PR #5448, opened 2019-08-28
Revert "Run xip-patch job as non-root user"

mwieczorek:
Reverts kyma-project/kyma#5390

There was an issue with clusters created with xip domain, looks like that is not tested before merge...


kubadz:
logs from xip-patch with securityContext:
```
unable to write 'random state'
writing new private key to '/app/key.pem'
/app/key.pem: Permission denied
```

PR #5551, opened 2019-09-04
Add query and subscriptions for namespaces 

dariadomagala:
**Description**

- Add query and subscriptions for namespaces 

**Related issue(s)**


components/console-backend-service/internal/domain/k8s/namespace_resolver.go
```diff
@@ -75,6 +60,20 @@ func (r *namespaceResolver) NamespacesQuery(ctx context.Context, applicationName
 		glog.Error(errors.Wrapf(err, "while listing %s", pretty.Namespaces))
 		return nil, gqlerror.New(err, pretty.Namespaces)
 	}
+
+	var namespaces []types.NamespaceWithAdditionalData
+	for _, ns := range rawNamespaces {
+		isSystem := isSystemNamespace(*ns, r.systemNamespaces)
+		passedSystemNamespaceCheck := !isSystem || (withSystemNamespaces != nil && *withSystemNamespaces && isSystem)
```
sjanota: Can be a little more clear IMO:
```suggestion
		passedSystemNamespaceCheck := (withSystemNamespaces != nil && *withSystemNamespaces) || !isSystem
```

components/console-backend-service/internal/domain/k8s/resourcequotastatus_converter_test.go
```diff
@@ -17,18 +17,18 @@ func TestResourceQuotaStatusConverter_ToGQL(t *testing.T) {
 	}{
 		"success": {
 			InputMap: map[string]map[v1.ResourceName][]string{
-				"rq-a": {
```
sjanota: Sure you wanted to change that?

components/console-backend-service/internal/domain/k8s/resourcequotastatus_service.go
```diff
@@ -109,7 +109,7 @@ func (svc *resourceQuotaStatusService) checkResourcesRequests(namespace string,
 		}
 	}
 
-	// For each StatefulSet that has a number of replicas lower than expected, check if any ResourceQuota blocks the progress of the StatefulSet.
```
sjanota: Same here

components/console-backend-service/internal/domain/k8s/selfsubjectrules_resolver_test.go
```diff
@@ -73,7 +73,7 @@ func TestSelfSubjectRulesQuery(t *testing.T) {
 func fixExampleSSRRGQLResponse() []gqlschema.ResourceRule {
 	return []gqlschema.ResourceRule{
 		gqlschema.ResourceRule{
-			Verbs:     []string{"a", "b"},
```
sjanota: And here

components/console-backend-service/internal/domain/k8s/namespace_service.go
```diff
@@ -16,19 +17,25 @@ import (
 )
 
 type namespaceService struct {
-	informer cache.SharedIndexInformer
-	client   corev1.CoreV1Interface
+	namespacesInformer cache.SharedIndexInformer
```
sjanota: I like the previous version `informer` more. You're not going to have more informers here.

dariadomagala: we had two informers in one of our versions, and this name stayed here after we removed it ;) will change it 👍

components/console-backend-service/internal/domain/k8s/k8s.go
```diff
@@ -46,17 +46,20 @@ func New(restConfig *rest.Config, informerResyncPeriod time.Duration, applicatio
...
 	informerFactory := informers.NewSharedInformerFactory(clientset, informerResyncPeriod)
 
-	namespaceSvc, err := newNamespaceService(informerFactory.Core().V1().Namespaces().Informer(), client)
+	podService := newPodService(informerFactory.Core().V1().Pods().Informer(), client)
+	namespaceSvc, err := newNamespaceService(informerFactory.Core().V1().Namespaces().Informer(), podService, client)
 	if err != nil {
 		return nil, errors.Wrap(err, "while creating namespace service")
 	}
+	namespaceResolver := newNamespaceResolver(namespaceSvc, applicationRetriever, systemNamespaces, podService)
+	namespaceResolver.namespaceConverter = *newNamespaceConverter(systemNamespaces)
```
sjanota: It's already set [here](https://github.com/kyma-project/kyma/pull/5551/files#diff-e6a7d56d82d12a25a102bac0d6b8fea7R42). You probably don't need to set it once again.
```suggestion
```

components/console-backend-service/internal/domain/k8s/namespace_converter_test.go
```diff
@@ -60,41 +65,62 @@ func TestNamespaceConverter_ToGQLs(t *testing.T) {
...
+		assert.Equal(t, expectedName, result.Name)
+		assert.Equal(t, expectedLabels, result.Labels)
+		assert.True(t, result.IsSystemNamespace)
+	})
+
 	t.Run("Success", func(t *testing.T) {
-		converter := namespaceConverter{}
 		expectedName := "exampleName"
 		expectedLabels := gqlschema.Labels{"test": "label"}
+		converter := namespaceConverter{}
```
sjanota: Add here different system namespace to check if it differentiates them properly

components/console-backend-service/internal/domain/k8s/namespace_resolver_test.go
```diff
@@ -4,45 +4,76 @@ import (
...
-	"github.com/kyma-project/kyma/components/console-backend-service/internal/gqlerror"
 	"github.com/kyma-project/kyma/components/console-backend-service/internal/gqlschema"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
 func TestNamespaceResolver_NamespacesQuery(t *testing.T) {
 	t.Run("Success", func(t *testing.T) {
```
sjanota: Do I see correctly that there is no test for a subscription? Please add it. Remember to check if all is cleaned up when the context is closed. Otherwise, we would have routines leakage.

components/console-backend-service/internal/domain/k8s/namespace_service_test.go
```diff
@@ -5,6 +5,7 @@ import (
 	"time"
```
sjanota: You missing tests for Subscribe/Unsubscribe

components/console-backend-service/internal/gqlschema/schema.graphql
```diff
@@ -920,7 +928,7 @@ type Query {
     connectorService(application: String!): ConnectorService! @HasAccess(attributes: {resource: "applications", verb: "create", apiGroup: "applicationconnector.kyma-project.io", apiVersion: "v1alpha1"})
 
     # Depends on 'application'
-    namespaces(application: String): [Namespace!]! @HasAccess(attributes: {resource: "namespaces", verb: "list", apiGroup: "", apiVersion: "v1"})
+    namespaces(withSystemNamespaces: Boolean, withInactiveStatus: Boolean): [Namespace!]! @HasAccess(attributes: {resource: "namespaces", verb: "list", apiGroup: "", apiVersion: "v1"})
```
sjanota: So, you cannot query by application, but you will still get applications in the namespace in the result, right?

dariadomagala: Right :)

components/console-backend-service/internal/domain/k8s/namespace_resolver_test.go
```diff
@@ -330,48 +310,79 @@ func TestNamespaceResolver_DeleteNamespace(t *testing.T) {
...
+
+		appRetriever := new(appAutomock.ApplicationRetriever)
+		resolver := k8s.NewNamespaceResolver(svc, appRetriever, []string{}, podSvc)
+
+		channel, err := resolver.NamespaceEventSubscription(ctx, nil)
+		require.NoError(t, err)
+
+		<-channel
+		svc.AssertCalled(t, "Unsubscribe", mock.Anything)
+		podSvc.AssertCalled(t, "Unsubscribe", mock.Anything)
```
sjanota: check if the event channel is closed. If you can find a way to check if the pods event channel is closed that would be awesome too. I can't think of an easy way, though. Maybe you can get it somehow from the mock?

sjanota: Also, the subscription runs in a different goroutine, and I worry that the code here may be subject to a race condition. Run `go test -count 100` to check if this test is stable.

tests/console-backend-service/internal/domain/k8s/namespace_test.go
```diff
@@ -25,32 +23,46 @@ func TestNamespace(t *testing.T) {
...
+	suite.thenUpdateEventIsSent(t, subscription)
 
 	t.Log("Deleting namespace...")
 	deleteRsp, err := suite.whenNamespaceIsDeleted()
 	suite.thenThereIsNoError(t, err)
 	suite.thenThereIsNoGqlError(t, deleteRsp.GqlErrors)
 	suite.thenDeleteNamespaceResponseIsAsExpected(t, deleteRsp)
 	suite.thenNamespaceIsRemovedFromK8sEventually(t)
+	//namespace changes its status to 'Terminating' first - delete event is sent after few seconds
+	suite.thenUpdateEventIsSent(t, subscription)
```
sjanota: Can we check if it works as expected for pods? Most simplistic scenario, create a pod which will stop just after startup. If it's too complicated leave it out. We have it covered by unit tests for now.

tests/console-backend-service/internal/domain/k8s/namespace_test.go
```diff
@@ -82,20 +94,19 @@ func givenNewTestNamespaceSuite(t *testing.T) testNamespaceSuite {
 	return suite
 }
 
+//subscribe
+func (s testNamespaceSuite) whenNamespacesAreSubscribedTo() *graphql.Subscription {
+	subscription := s.gqlClient.Subscribe(s.fixNamespacesSubscription())
```
sjanota: Thinking out loud. It's pretty dumb to send update events triggered by pods when the client is not subscribed to pods, but there's nothing we can do now.

tests/console-backend-service/internal/domain/k8s/namespace_test.go
```diff
@@ -145,61 +180,105 @@ func (s testNamespaceSuite) thenDeleteNamespaceResponseIsAsExpected(t *testing.T
...
+	return namespaceEvent.NamespaceEvent, err
+}
+
+func checkNamespaceEvent(expected NamespaceEventObj, subscription *graphql.Subscription) error {
+	for {
+		event, err := readNamespaceEvent(subscription)
+		if err != nil {
+			return err
+		}
+		if expected.Type == event.Type && expected.Namespace.Name == event.Namespace.Name {
```
sjanota: Let's make this check more exhaustive. Why not `require.Equal`?

components/console-backend-service/internal/domain/k8s/namespace_resolver_test.go
sjanota: passed

Issue #5474, opened 2019-08-29
Test coverage metrics


clebs:
**Description**
Coverage from unit and end to end tests needs to be generated and stored so that we can see the quality of our code and see its evolution.

**Acceptance**
- Test coverage is periodically measured by CI/CD.
- Coverage data is stored in the quality metrics storage with an appropriate format.
- Coverage data is queryable.

**Related issues**
#5473

clebs:
Coverage data will be stored and queriable after #5473.

franpog859:
Hey! Printing test coverage introduced in [this PR](https://github.com/kyma-project/kyma/pull/5568) lefts the `cover.out` file not removed if the tests failed. I suggest to delete them in this case or add a line to the `.gitignore` file.

To reproduce it, go to for example [this file](https://github.com/kyma-project/kyma/blob/master/components/compass-runtime-agent/internal/kyma/service_test.go) and add a line
```go
require.True(t, false)
```
somewhere in the test and run [`before-commit.sh` script](https://github.com/kyma-project/kyma/blob/master/components/compass-runtime-agent/before-commit.sh)

lilitgh:
@franpog859, thanks for reporting this. I created a bug for this issue (https://github.com/kyma-project/kyma/issues/5715) and I am closing this one again.

PR #5568, opened 2019-09-05
Add total test coverage calculation where possible

clebs:
**Description**

- Now tests show total coverage at the end for golang

**Related issue(s)**
#5474

### DIFF
--- a/components/application-operator/before-commit.sh
+++ b/components/application-operator/before-commit.sh
@@ -57,12 +57,15 @@ fi
 # GO TEST
 ##
 echo "? go test"
-go test ./...
+go test -short -coverprofile=cover.out ./...
 # Check if tests passed
 if [ $? != 0 ]; then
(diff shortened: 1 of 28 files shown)


components/application-operator/before-commit.sh
```diff
@@ -57,12 +57,15 @@ fi
 # GO TEST
 ##
 echo "? go test"
-go test ./...
+go test -short -coverprofile=cover.out ./...
```
Szymongib: Running test coverage on all directories also counts the generated code. It is not covered with tests, which therefore decreases overall test coverage of the component significantly. This is the case for several components.
I do not think this is desired behavior.

clebs: Yes, we definitely do not want to count those packages, thanks for pointing that out.
For now I just enabled the coverage to have at least a number.

Teams can further improve on their respective components as they see fit by either adding unit tests or removing packages that shall be excluded.

Issue #5467, opened 2019-08-29
Enable by default validation and conversion service for asyncapi in hCMS


derberg:
**Description**

- release asyncapi/parser and use the release version in the server
- update parser in the service to use latest and have nice errors in the status of the CR
- update converter in asyncapi org to use latest RC2 and release and then use the converter in Kyma
- update CBS, the EventActivation query to properly parse it and extract info needed for Lambda view. Don't break API.
- modify helm to make sure the asyncapi service is on by default
- manually test registration of application:
  - with broken spec 1.0 and make sure:
    - in UI event tab is not rendered
    - in Lambda view event triggers are not available
  - with not broken spec 1.0:
    - that UI is rendered
    - that Lambda event triggers are listed
- manualy test that with upgrade between kyma 1.5 and 1.6 we will also convert old asyncapi specs
- describe everything in details in release notes, especially the behavior on upgrade, so it is clear for the users that all asyncapi specs will be converted to 2.0 during upgrade
- update docs for app connector and headless cms

**Reasons**

RC2 released and in 2 weeks official 2.0 will be available and we want to support it the same time of announcement.

kazydek:
Documentation is covered in #https://github.com/kyma-project/kyma/pull/5638

PR #5594, opened 2019-09-10
Migrate to AsyncAPI 2.0.0

michal-hudy:
**Description**

- Migrated to AsyncAPI 2.0.0

**Related issue(s)**

#5467


components/cms-services/deploy/asyncapi/Dockerfile
```diff
@@ -1,9 +1,7 @@
-FROM alpine:3.10
+FROM gcr.io/distroless/static:latest
```
aerfio: `:latest`?

michal-hudy: Yes, there are no semver tags for this image :(. We are also using `latest` in controllers ;)

Issue #5605, opened 2019-09-11
Add OWNERS files


lilitgh:
All teams all requested to have OWNERS file that will replace CODEOWNERS when [Tide](https://github.com/kubernetes/test-infra/tree/master/prow/cmd/tide) is fully integrated. It applies to all repositories in kyma-project and kyma-incubator organisation. For more information please consult [owners.md](https://github.com/kubernetes/community/blob/master/contributors/guide/owners.md#owners-spec) file.

PR #5612, opened 2019-09-11
Adding OWNERS file for event-bus and serverless components

anishj0shi:
w.r.t #5605 

Adding Owner alias for event-bus and serverless components and OWNERS file for respective components

### DIFF
--- a/OWNERS_ALIASES
+++ b/OWNERS_ALIASES
@@ -31,7 +31,7 @@ aliases:
     - dariadomagala
     - akucharska
     - sjanota
-  skydiving-tunas
+  skydiving-tunas-approvers:
     - @Abd4llA
     - @joek
     - @abbi-gaurav
--- a/OWNERS_ALIASES
+++ b/OWNERS_ALIASES
@@ -31,3 +31,31 @@ aliases:
     - dariadomagala
     - akucharska
     - sjanota
+  skydiving-tunas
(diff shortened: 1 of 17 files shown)


OWNERS_ALIASES
```diff
@@ -31,3 +31,31 @@ aliases:
...
+    - @lilitgh
+    - @k15r
+    - @nachtmaar
+    - @anishj0shi
+    - @montaro
+    - @marcobebway
+    - @radufa
+    - @sayanh
+    - @antoineco
+  skydiving-tunas
```
k15r: twice twice?

OWNERS_ALIASES
```diff
@@ -31,3 +31,31 @@ aliases:
     - dariadomagala
     - akucharska
     - sjanota
+  skydiving-tunas
```
k15r:
```suggestion
  skydiving-tunas-approvers:
```

OWNERS_ALIASES
k15r: i see the other teams differentiate between reviewers and approvers. Shall we do the same even if they would be the same initially?
```suggestion
  skydiving-tunas-reviewers:
```


if we decide to do so all the OWNERS files need to be updated

anishj0shi:
@k15r @antoineco Please review it again.

Issue #5150, opened 2019-08-01
Migrate to tekton for knative-function-controller


rakesh-garimella:
**Description**
We have knative-function-controller which uses knative-build. Knative-build has now been deprecated, we need to migrate  to tekton

derberg:
pretty fresh blog post about it https://medium.com/google-cloud/migrating-from-knative-build-to-tekton-pipelines-68fc6de14373

PR #5560, opened 2019-09-05
Upgrade Knative Serving to v0.8.1

rakesh-garimella:
**Description**

Upgrade knative serving v0.8.1
- Copied serving and build yamls into config/crd directory as the yamls added field `versions` which is only supported in k8s 1.15. However kubebuilder as of now supports k8s v .1.14
- Updated knative serving tests to 0.8.1 also upgraded the Gopkg.toml.
- Modified `knative-ingress-gateway` by copying the spec from `kyma-gateway` to make https work

**Related issue(s)**

#5497

PR #5652, opened 2019-09-17
Migrate function-controller build backend to Tekton

antoineco:
**Description**

- Migrate build backend from Knative Build to Tekton 

**Related issue(s)**

#5150


components/function-controller/Gopkg.toml
```diff
@@ -1,72 +1,50 @@
...
 
 [[override]]
   name = "k8s.io/apimachinery"
-  version="kubernetes-1.13.4"
+  version = "kubernetes-1.13.4"
 
 [[override]]
   name = "k8s.io/api"
-  version="kubernetes-1.13.4"
+  version = "kubernetes-1.13.4"
```
anishj0shi: should we follow the kubernetes version from knative 0.8?

components/function-controller/Gopkg.toml
```diff
@@ -1,72 +1,50 @@
...
-  version="v0.1.9"
+  version = "v0.1.9"
 
+[[constraint]]
+  name = "github.com/tektoncd/pipeline"
+  version = "v0.6.0"
+
+[[constraint]]
+  name = "github.com/knative/serving"
+  version = "v0.6.1"
```
anishj0shi: how about knative serving 0.8?

components/function-controller/config/rbac/rbac_role_binding.yaml
```diff
@@ -9,4 +9,5 @@ roleRef:
   name: manager-role
 subjects:
 - kind: ServiceAccount
-  name: ""
+  name: default
+  namespace: system
```
anishj0shi: where is this namespace being created? via tekton?

components/function-controller/pkg/controller/function/function_controller.go
```diff
@@ -106,385 +98,436 @@ var (
...
+func (r *ReconcileFunction) Reconcile(req reconcile.Request) (reconcile.Result, error) {
 	// Get Function instance
-	fn := &serverlessv1alpha1.Function{}
-	err := r.getFunctionInstance(request, fn)
-	if err != nil {
-		if errors.IsNotFound(err) {
-			return reconcile.Result{}, nil
+	fn, err := r.getFunctionInstance(req)
+	switch {
+	case errors.IsNotFound(err):
```
anishj0shi: doesn't isNotFound() also mean that the function is probably not even created and then you flag it for creation?

components/function-controller/Gopkg.toml
antoineco: That's the topic of another PR (currently in review)

components/function-controller/config/rbac/rbac_role_binding.yaml
antoineco: This file is generated and meant to be applied by kustomize, which replaces these entries, by the actual namespace.

components/function-controller/pkg/controller/function/function_controller.go
antoineco: The first thing we do in Reconcile() is checking whether the function exists, we would have returned early if the Function didn't exist.

components/function-controller/Gopkg.toml
antoineco: That's the topic of another PR: #5560 (currently in review)

components/function-controller/pkg/controller/function/function_controller.go
```diff
@@ -106,385 +98,436 @@ var (
...
 	}
 
-	// Unique Build name base on function sha
-	shortSha := ""
-	if len(functionSha) > 10 {
-		shortSha = functionSha[0:10]
-	} else if len(functionSha) > 0 && len(functionSha) < 10 {
-		shortSha = functionSha
+	shortSha := fnSha
+	if len(fnSha) > 10 {
```
k15r: not that it matters, but this check will never fail. sha256 always require 64characters in hex representation

antoineco: Nice catch

components/function-controller/pkg/controller/function/function_controller.go
```diff
@@ -373,167 +371,154 @@ func generateFunctionHash(fnCm *corev1.ConfigMap) (string, error) {
...
-		args := foundBuild.Spec.Template.Arguments
-		for _, arg := range args {
-			if arg.Name == "IMAGE" && arg.Value == imageName {
-				return true
-			}
+		if err := r.Create(ctx, desiredTr); err != nil {
+			return nil, err
+		}
+
+		// TODO(antoineco): it would be enough to compute the status ONCE, at the end of the sync
```
k15r: we should change the status resource to allow multiple conditions. then we could have functions in status building and simultaneously serving. If we do so we would probably keep the current updates along the way

antoineco: Controllers (in the Kubernetes core at least) usually set status by observation at the _end_ of the reconciliation. Things are created/updated/deleted in the main logic, and then `setStatus()` (example name) observes the current state of the world to compute the entire status, ideally from informers cache for speed.

components/function-controller/pkg/controller/function/function_controller.go
antoineco: See https://github.com/kyma-project/kyma/pull/5664/commits/7abd9284ecc4c9a2e81ec8f34116b15c3a863e80

components/function-controller/pkg/controller/function/function_controller.go
antoineco: Moving the discussion to #5673

components/function-controller/Gopkg.toml
```diff
@@ -1,72 +1,50 @@
...
-    "sigs.k8s.io/kustomize",
     "sigs.k8s.io/controller-tools/cmd/controller-gen", # for crd/rbac generation
-    "sigs.k8s.io/controller-runtime/pkg/client/config",
-    "sigs.k8s.io/controller-runtime/pkg/controller",
-    "sigs.k8s.io/controller-runtime/pkg/handler",
-    "sigs.k8s.io/controller-runtime/pkg/manager",
-    "sigs.k8s.io/controller-runtime/pkg/runtime/signals",
-    "sigs.k8s.io/controller-runtime/pkg/source",
-    "sigs.k8s.io/testing_frameworks/integration", # for integration testing
-	  "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1beta1",
```
antoineco: All these packages had no reason to be added as requirements, it's `dep`'s role to figure that out. Only packages that are not referenced in the code should be there (third-party dependencies).

components/function-controller/pkg/controller/function/function_controller.go
```diff
@@ -211,7 +207,8 @@ func (r *ReconcileFunction) Reconcile(req reconcile.Request) (reconcile.Result,
 	buildName := fmt.Sprintf("%s-%s", fn.Name, fnSha[:buildNameSuffixLen])
 	log.Info("Build info", "namespace", fn.Namespace, "name", fn.Name, "buildName", buildName, "imageName", imgName)
 
-	err = r.buildFunctionImage(rnInfo, fn, imgName, buildName)
+	// Run Function build (Tekton TaskRun)
+	fnTr, err := r.buildFunctionImage(rnInfo, fn, imgName, buildName)
```
antoineco: This function now returns the `TaskRun` it has found or created, so we can pass it later on to `setFunctionCondition()` and compute the function status. This way we save ourselves extra (and expensive) calls to `client.Get()`.

components/function-controller/pkg/controller/function/function_controller.go
```diff
@@ -611,56 +599,41 @@ func (r *ReconcileFunction) serveFunction(rnInfo *runtimeUtil.RuntimeInfo, found
...
-		return
-	}
-
-	// if build show error, set function status to error too
-	for _, condition := range foundBuild.Status.Conditions {
-		if condition.Type == duckv1alpha1.ConditionSucceeded && condition.Status == corev1.ConditionFalse {
-			err := r.updateFunctionStatus(fn, serverlessv1alpha1.FunctionConditionError)
-			if err != nil {
-				log.Error(err, "Error while trying to update the function Status", "namespace", fn.Namespace, "name", fn.Name)
+func (r *ReconcileFunction) setFunctionCondition(fn *serverlessv1alpha1.Function, tr *tektonv1alpha1.TaskRun,
```
antoineco: Notice how `setFunctionCondition()` now accepts the objects it needs to compute the Function status from as parameter, this way we can avoid calling `Get()` again for no reason. See previous comment.

components/function-controller/pkg/controller/function/function_controller.go
```diff
@@ -367,168 +369,154 @@ func generateFunctionHash(fnCm *corev1.ConfigMap) (string, error) {
...
 	}
 
-	if _, err = r.updateFunctionBuildTemplate(currentBt, desiredBt); err != nil {
+	if _, err = r.updateFunctionBuildTask(currentTask, desiredTask); err != nil {
 		if err := r.updateFunctionStatus(fn, serverlessv1alpha1.FunctionConditionError); err != nil {
 			log.Error(err, "Error setting Function status", "namespace", fn.Namespace, "name", fn.Name)
 		}
 
-		log.Error(err, "Error during sync of the Function BuildTemplate", "namespace", desiredBt.Namespace, "name", desiredBt.Name)
+		log.Error(err, "Error during sync of the Function build Task", "namespace", desiredTask.Namespace, "name", desiredTask.Name)
```
k15r: Why keep this errormessage but remove a similar one in 391?

antoineco: Nice catch, that error moved at line 190, this one is a leftover.
Removed.

antoineco:
@k15r agree, I'm on it.

components/function-controller/README.md
```diff
@@ -1,14 +1,14 @@
...
+The Knative Function Controller is a Kubernetes controller that enables Kyma to manage Function resources. It uses Tekton Pipelines and Knative Serving under the hood.
 
 ## Prerequisites
 
 The Function Controller requires the following components to be installed:
 
-- Knative Build (v0.6.0)
-- Knative Serving (v0.8.1)
-- Istio (v1.0.7)
+- [Tekton Pipelines](https://github.com/tektoncd/pipeline/releases) (v0.7.0)
```
nachtmaar: 👍 for adding links

components/function-controller/pkg/utils/equality.go
```diff
@@ -0,0 +1,147 @@
...
+	"reflect"
+
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/conversion"
+
+	servingv1alpha1 "knative.dev/serving/pkg/apis/serving/v1alpha1"
+)
+
+// Semantic can do semantic deep equality checks for API objects.
+// Fields which are not relevant for Model-T are intentionally omitted.
```
k15r: lets assume i dont know what Model-T is...
btw: what is Model-T

components/function-controller/pkg/utils/equality.go
```diff
@@ -0,0 +1,147 @@
...
+var Semantic = conversion.EqualitiesOrDie(
+	ksvcEqual,
+)
+
+// ksvcEqual asserts the equality of two Knative Service objects.
+func ksvcEqual(s1, s2 *servingv1alpha1.Service) bool {
+	if s1 == s2 {
```
k15r: check for `nil` first, `==` is not defined on `nil`

components/function-controller/pkg/utils/equality.go
```diff
@@ -0,0 +1,147 @@
...
+			return false
+		}
+	}
+
+	return true
+}
+
+// podSpecEqual asserts the equality of two PodSpec objects.
+func podSpecEqual(ps1, ps2 *corev1.PodSpec) bool {
+	if ps1 == ps2 {
```
k15r: same `nil` handling here please

components/function-controller/pkg/utils/equality.go
antoineco: I'm guilty on that one 😅

components/function-controller/pkg/utils/equality.go
```diff
@@ -0,0 +1,147 @@
...
+)
+
+// Semantic can do semantic deep equality checks for API objects.
+// Fields which are not relevant for Model-T are intentionally omitted.
+var Semantic = conversion.EqualitiesOrDie(
+	ksvcEqual,
+)
+
+// ksvcEqual asserts the equality of two Knative Service objects.
+func ksvcEqual(s1, s2 *servingv1alpha1.Service) bool {
```
k15r: i see no handling of the RouteSpec that you can put into a ksvc. is this on purpose?

antoineco: It's on purpose, because AFAIK we don't fiddle with it. I don't mind adding it for completion if you think we should care.

k15r: lets put it in when we need it

Issue #5227, opened 2019-08-06
Compass Runtime Agent should use Application name as a CR name


Szymongib:
**Description**

For now, Compass Runtime Agent uses the Application ID as a name of the Custom Resource. As now the name is guaranteed to be valid Kubernetes name so we can start using it as a name of the CR.

Acceptance:
- Application Name used as a name of CR
- Tests adjusted accordingly

janmedrek:
We also need to remove `app-` prefix in services deployed by Application Operator, it broke contract with some of the connectors and also may influence maximum length of application name.

Issue #5309, opened 2019-08-13
Apps created by Compass Runtime Agent should be labeled


Szymongib:
**Description**

Currently, Compass Runtime Agent operates on all Applications inside the cluster. This causes that only Applications registered in Director will exist in a cluster.

The Applications created by Runtime Agent should be labeled and the component should operate only on those Applications.

Consider using annotations or spec flag.

PR #5697, opened 2019-09-19
Use Application name as a CR name in Compass Runtime Agent

franpog859:
**Description**

- Use Application name to name the CR instead of Application ID
- Validate Runtime Agent manageable Applications using ClientIDs from Compass Metadata

**Related issue(s)**

See the issue #5227


franpog859:
Labeling Applications will be implemented in [PR 5805](https://github.com/kyma-project/kyma/pull/5805). This change will resolve [issue 5309](https://github.com/kyma-project/kyma/issues/5309)

components/application-connectivity-validator/internal/validationproxy/handler.go
```diff
@@ -113,22 +153,36 @@ func hasValidSubject(subjects []string, appName, group, tenant string) bool {
...
+		return len(subjectField) == 1 && subjectField[0] == expectedValue
 	}
-}
 
-func validateSubjectField(subjectField []string, expectedValue string) bool {
-	return len(subjectField) == 1 && subjectField[0] == expectedValue
+	switch {
+	case group == "" || tenant == "":
+		return validateCommonNameWithAppName
+
```
Szymongib: As discussed we should adjust this logic.

components/application-connectivity-validator/internal/validationproxy/handler_test.go
```diff
@@ -30,6 +36,36 @@ type event struct {
...
+			Kind:       "Application",
+			APIVersion: "applicationconnector.kyma-project.io/v1alpha1",
+		},
+		ObjectMeta: metav1.ObjectMeta{
+			Name: applicationMetaName,
+		},
+		Spec: v1alpha12.ApplicationSpec{
+			Description:     "Description",
+			Services:        []v1alpha12.Service{},
+			CompassMetadata: &v1alpha12.CompassMetadata{Authentication: v1alpha12.Authentication{ClientIds: []string{applicationName}}},
```
Szymongib: We should not use `applicationName` as a `ClientIds` as it caused our test to pass despite the logic error.

components/compass-runtime-agent/internal/kyma/service.go
```diff
@@ -32,9 +32,9 @@ const (
 )
 
 type Result struct {
-	ApplicationID string
-	Operation     Operation
-	Error         apperrors.AppError
+	ApplicationName string
```
Szymongib: I think it would be good to include both `ApplicationName` and `ApplicationId` here.

components/application-connectivity-validator/internal/cache/id_cache.go
```diff
@@ -0,0 +1,38 @@
+package cache
```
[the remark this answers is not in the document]
franpog859: I think casting to `[]string` is quite low level to do it in the handler. This wrapper makes it more clear in my opinion. Are you sure we want to get rid of this abstraction?

components/application-connectivity-validator/cmd/applicationconnectivityvalidator/applicationconnectivityvalidator.go
```diff
@@ -54,3 +70,20 @@ func main() {
...
 	wg.Wait()
 }
+
+func newApplicationGetter() (applications.Getter, apperrors.AppError) {
+	cfg, err := config.GetConfig()
+	if err != nil {
+		return nil, apperrors.Internal("failed to get k8s config: %s", err)
+	}
+
+	applicationEnvironmentClientset, err := versioned.NewForConfig(cfg)
```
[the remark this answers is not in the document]
franpog859: Yeah, you're right ^^

components/application-connectivity-validator/internal/applications/getter.go
```diff
@@ -0,0 +1,26 @@
...
+
+func NewGetter(applicationInterface v1alpha12.ApplicationInterface) Getter {
+	return &applicationGetter{
+		applicationInterface: applicationInterface,
+	}
+}
+
+func (ai *applicationGetter) Get(applicationName string) (*v1alpha1.Application, error) {
+	return ai.applicationInterface.Get(applicationName, metav1.GetOptions{})
+}
```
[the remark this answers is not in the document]
franpog859: Removed!

components/application-connectivity-validator/internal/validationproxy/handler.go
franpog859: Done!

components/application-connectivity-validator/internal/validationproxy/handler_test.go
franpog859: Done!

components/compass-runtime-agent/internal/kyma/service.go
franpog859: I added ApplicationName and changed ApplicationID to ApplicationIDs

components/application-connectivity-validator/internal/validationproxy/handler.go
```diff
@@ -82,6 +101,34 @@ func (ph *proxyHandler) ProxyAppConnectorRequests(w http.ResponseWriter, r *http
...
+	if !found {
+		var err apperrors.AppError
+		applicationClientIDs, found, err = ph.getApplicationClientIDs(applicationName)
+		if err != nil {
+			return []string{}, err
+		}
+
+		if found {
+			ph.cache.SetClientIDs(applicationName, applicationClientIDs)
+		}
```
Szymongib: Correct me if I am wrong, but from what I understand if `CompassMetadata` is `nil` we do not save anything in cache. As a result, for standalone Applications, we do the API Server call every time which is not desired behavior. We should either cache the empty array for this app or skip lookup all together.

components/application-connectivity-validator/internal/cache/id_cache.go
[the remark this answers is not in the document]
Szymongib: But does it make sense to have a wrapper struct just for this one operation?
Unless we want to extract whole caching logic with reading Application to this cache I do not see a point in having this wrapper.

components/compass-runtime-agent/internal/kyma/service.go
```diff
@@ -32,9 +32,10 @@ const (
...
+	ApplicationIDs  []string
```
Szymongib: I was rather thinking of `ApplicationID` which always is (and will be) a single value.

components/compass-runtime-agent/internal/kyma/service.go
```diff
@@ -412,8 +413,9 @@ func getEventApiType(eventApiSpec *model.EventAPISpec) docstopic.ApiType {
 
 func newResult(application v1alpha1.Application, operation Operation, appError apperrors.AppError) Result {
 	return Result{
-		ApplicationID: application.Name,
-		Operation:     operation,
-		Error:         appError,
+		ApplicationName: application.Name,
+		ApplicationIDs:  application.Spec.CompassMetadata.Authentication.ClientIds,
```
Szymongib: We should get the `ApplicationID` from the data we get from the Director. In the future `Spec.CompassMetadata.Authentication.ClientIds` will contain something different than the `ApplicationID`.
This will require a little refactoring as we will need to pass the ID to `newResult` function.

components/application-connectivity-validator/internal/validationproxy/handler.go
franpog859: Done!

components/compass-runtime-agent/internal/kyma/service.go
franpog859: Sure!

components/compass-runtime-agent/internal/kyma/service.go
franpog859: Done!

PR #5709, opened 2019-09-20
Correct a Conditional statement for event-bus

anishj0shi:
Initially in the function GetChannelByLabels() labels were passed as pointers and its null check was done as if it was a pointer with no reference, After changing the method signature to value based, this null check was not updated.

with this PR we check the map with 0 length instead of nil pointer.


anishj0shi:
> This condition was fine how it was, curious to know why you want to change it?

A zero value map is not always a nil pointer, that's why

Issue #5524, opened 2019-09-02
Use upstream Service Catalog 0.3.0 (CRD only) in Kyma


PK85:
**Description**

Use upstream Service Catalog 0.3.0 (CRD only) in Kyma.

**Reasons**

Finally no ETCD and API-Server in Kyma for Service Catalog component

**AC**

- Kyma has updated and cleanuped chart.
- update service-catalog clients in all components that users SC (appBroker, SBU, CBS)
- adjust kyma-project documentation (if we mentioned there etcd/api-server things)

adamwalach:
Components:
- Application Broker - https://github.com/kyma-project/kyma/pull/5687
- Service Binding Usage Controller - https://github.com/kyma-project/kyma/pull/5693
- Console Backend Service
   - Removed Kubeless (requires older version of k8s than we can support) - is not really neded anymore. 
   I encountered this issue: https://github.com/kyma-project/test-infra/issues/1502
   -

PR #5724, opened 2019-09-23
Update dependencies for Console Backend Service

adamwalach:
**Description**

Upgraded:
  - kubernetes - 1.15,
  - service-catalog - crds

**Related issue(s)**
https://github.com/kyma-project/kyma/issues/5524


components/console-backend-service/hack/update-mocks.sh
```diff
@@ -0,0 +1,7 @@
+#!/usr/bin/env bash
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
+mockery -case underscore -name DiscoveryInterface -dir ./vendor/k8s.io/client-go/discovery/ -output ./internal/domain/k8s/automock -outpkg automock
```
sjanota: Any chance we could make it a `go generate` directive instead of additional script?

adamwalach: I could add 
```
//go:generate mockery -name=DiscoveryInterface -case=underscore -output=automock -outpkg=automock -dir ../../../vendor/k8s.io/client-go/discovery/
```
to resource_service.go
But I do not like the relative path with ../../../ cause it will fail when this file moves in the directory structure.

sjanota: How about adding a simple wrapper interface? 
```
type DiscoveryInterface interface {
  discovery.DiscoveryInterface
}
```
and mock this one? Then IDE will update paths whenever you move the file.

adamwalach: Right, fixed

Issue #5726, opened 2019-09-23
Get Rid of GetKnativeChannelName() function in event-bus/pkg/util


anishj0shi:
GetKnativeChannelName() function was initially used to derive knative channel names based on event source, type and version. this function usage has been completely removed and use "generateName" functionality from Kubernetes and fetch channels based on labelSelectors, hence we need to clean this up and update all the unit tests with current approach.

PR #5731, opened 2019-09-23
Get Rid of GetKnativeChannelName() function in event-bus/pkg/util

anishj0shi:
w.r.t #5726 

- Get rid of dead code event-bus/pkg/util/util.go and fix tests relying on dead code

#### Proposed Changes for test fix
- Create a prefix => channel Name map for getting channel name via a prefix
- makeKnChannel function enhanced to create a Knative Channel stub for unit testing

GetKnativeChannelName() function was initially used to derive knative channel names based on event source, type and version. this function usage has been completely removed and use "generateName" functionality from Kubernetes and fetch channels based on labelSelectors, hence we need to clean this up and update all the unit tests with current approach.


components/event-bus/internal/knative/subscription/controller/subscription/reconcile_test.go
```diff
@@ -426,8 +431,8 @@ func makeKnChannel(namespace string, name string, labels map[string]string) *mes
 	}
 }
 
-func makeKnChannelName(kySub *eventingv1alpha1.Subscription) string {
-	return eventBusUtil.GetKnativeChannelName(&kySub.SourceID, &kySub.EventType, defaultMaxChannelLength)
+func makeKnChannelNamePrefix(kySub *eventingv1alpha1.Subscription) string {
```
antoineco: I'm a bit puzzled about the func name. It says it makes something with a prefix but returns another something called EventType. One of those are off.

components/event-bus/internal/knative/subscription/controller/subscription/reconcile_test.go
```diff
@@ -366,8 +368,8 @@ func (k *MockKnativeLib) GetChannelByLabels(namespace string, labels map[string]
 	return k.GetChannel(channelName, namespace)
 }
 
-func (k *MockKnativeLib) CreateChannel(name string, namespace string, labels map[string]string, timeout time.Duration) (*messagingV1Alpha1.Channel, error) {
-	channel := makeKnChannel(namespace, name, labels)
+func (k *MockKnativeLib) CreateChannel(prefix, namespace string, labels map[string]string, timeout time.Duration) (*messagingV1Alpha1.Channel, error) {
```
antoineco: /begin note for the future (NOT this PR in particular)
All those mocks seem too far away from the actual implementation to be relied upon, and we would be better off using the Fake clientset against the real implementation instead.
/end

components/event-bus/internal/knative/subscription/controller/subscription/reconcile_test.go
```diff
@@ -130,7 +131,7 @@ var testCases = []controllertesting.TestCase{
 				if _, ok := knSubscriptions[makeKnSubscriptionName(makeEventsActivatedSubscription())]; !ok {
 					t.Errorf("Knative subscription was NOT created")
 				}
-				if channel, ok := knChannels[makeKnChannelName(makeEventsActivatedSubscription())]; ok {
+				if channel, ok := knChannels[knChannelNames[makeKnChannelNamePrefix(makeEventsActivatedSubscription())]]; ok {
```
antoineco: An intermediary variable could make those calls slightly more readable I believe.

components/event-bus/internal/knative/subscription/controller/subscription/reconcile_test.go
anishj0shi: so initially the function name was *makeKnChannelName* this function was returning a derived name for channel. I've renamed it to **makeKnChannelNamePrefix** so it doesn't means make something with prefix. it just means make a channel name prefix. and we create prefix via Event-type parameter of kyma subscription.

components/event-bus/internal/knative/subscription/controller/subscription/reconcile_test.go
anishj0shi: i think, we use knative's unit testing f/w over here. for most of the reconciliation ops. however i agree to use fake client for knativelib operations as well. could you please create a ticket on this, @antoineco ?

components/event-bus/internal/knative/subscription/controller/subscription/reconcile_test.go
anishj0shi: on it

components/event-bus/internal/knative/subscription/controller/subscription/reconcile_test.go
antoineco: OK I get it now. It's an accessor with an obscure :)
I would argue that it introduces unnecessary obscurity but I won't insist.

anishj0shi: yea, i agree, but everywhere the accessors in this test start with *make* i don't think refactoring all of them is a good idea for this PR, whose intention is to get rid of dead code and fix tests based on that.

components/event-bus/internal/knative/subscription/controller/subscription/reconcile_test.go
anishj0shi: done

sayanh:
Since we are not limiting the length of the Knative channel name. What if prefix is really long(say 53 chars)? Then the additional strings will be added as prefixes and suffixes, which may prevent creation of k8s svcs. I have seen really long event names.

components/event-bus/internal/knative/util/kn-eventing_test.go
```diff
@@ -330,4 +330,14 @@ func Test_MakeChannelWithPrefix(t *testing.T) {
 
 	// makeChannel should not add double hyphens if already present
 	assert.False(t, strings.HasSuffix(a.GenerateName, "--"))
+
+	//Test prefix length is trucated if the event-type is too big
```
sayanh: > truncated

PR #5766, opened 2019-09-25
Remove powershell installation scripts

crabtree:
**Description**

This PR removes PowerShell scripts that perform the kyma installation on Windows. Currently, the supported way for kyma installation on Windows is via kyma cli, so the scripts are no longer needed.

- Remove PowerShell installation scripts

**Related issue(s)**


derberg:
@crabtree Hey man, just to make sure, you remove those scripts as we do not need them anymore as desired installation way for windows user is to do it with Kyma CLI? Please extend the description of PR to make it clear as there is no related issue

Issue #5315, opened 2019-08-14
Upgrade istio to 1.3


piotrmsc:
**Description**

In istio 1.3 pilot agent will have `/quitquitquit` handler released which will allow us to move on with k8s jobs having sidecars on azure. The problem is known from #4719. 

Before upgrading make sure it is indeed released (anything can happen ;-) ) So either look for commit in changelog or deploy on a minikube pure charts quickly, deploy app with sidecar, exec into istio-proxy and make a call to localhost:$PORT/quitquitquit

**Reasons**

Unblock failing tests jobs on Azure

**Attachments**

Demonsthere:
The upgrade is temporarily halted due to a change in SDS handling in Istio. A bug report has been created: https://github.com/istio/istio/issues/17409

piotrmsc:
1.3.4 will contain fix for that, we should update to it as soon as released

piotrmsc:
Fix with sds should improve stablity of ingressgateway itself which give from time to time errora : https://github.com/kyma-project/kyma/issues/6005

Demonsthere:
New PR: https://github.com/kyma-project/kyma/pull/6200

PR #5775, opened 2019-09-26
Move gateway hosts using mtls to a separate gateways

kubadz:
**Description**

- move gateway hosts with mtls to a separate gateways

**Related issue(s)**

Related to #5315


kubadz:
As the whole issue of upgrading Istio is currently blocked, further work on this PR will also be suspended. It needs investigation on the reason of failing tests. Tests are failing on components using gateway configured with mTLS, for which we moved the configuration to separate Gateway resources.

kubadz:
Done in: #6214

PR #5801, opened 2019-09-27
Introduce dex config for compass UI

y-kkamil:
**Description**

- Add dex as an identity provider in Compass UI

**Related issue(s)**
See https://github.com/kyma-project/console/issues/1268


y-kkamil:
due to failing tests it is introduced in: https://github.com/kyma-project/kyma/pull/5822

PR #5803, opened 2019-09-27
Add ability to execute only one test in testing.sh

michal-hudy:
**Description**

- Added selective testing

**Related issue(s)**


derberg:
Merging without waiting for CI because we are 100% sure this change is not blocking anyone, but rather unblocking other PRs

PR #5810, opened 2019-09-27
Fix test flag reading in testing.sh

michal-hudy:
**Description**

- Fixed flags reading in testing.sh

**Related issue(s)**


derberg:
Related to https://github.com/kyma-project/kyma/pull/5803

Merging without waiting for CI because we are 100% sure this change is not blocking anyone, but rather unblocking other PRs

Issue #5869, opened 2019-10-03
Service Catalog Upgrade fails if executed second time


mszostok:
**Description**

Currently, the Service Catalog upgrade action can be executed only once.  When it is executed the second time then such err is returned: 

```
service-catalog:
  Helm upgrade error: rpc error: code = Unknown desc = persistentvolumeclaims "service-catalog-catalog-migration-storage" already exists
```

The problem is also that in first iteration the Service Catalog resources cannot be restored because of temporary CRD unavailability in main api server.

mszostok:
As a first step we will and waiter for ready CRDs definition and also retries on restore operation.

If such workaround will fail then we will try to figure out the way for handling the multi upgrade execution

mszostok:
This issue was solved by updating the Service Catalog to v0.3.0-beta.1 where discussed fixes were merged, check this PR for more info: https://github.com/kyma-project/kyma/pull/5879

PR #5879, opened 2019-10-04
Update the Service Catalog image to v0.3.0-beta.1

mszostok:
**Description**

- Add missing `svcat: "true"` labels to Service Catalog CRDs
- Add CRD waiter for the migration job and the webhook server.
   It may take some time before Catalog CRDs are registered in the main API Server. For this reason, the CRD waiter is executed at the initial state of each Service Catalog binary because Service Catalog clients/informers can be created only when CRDs are available.
    
    Additionally, we added retries when restoring Service Catalog resources during the migration process.

    See the discussion in this thread for more info: 
    https://kubernetes.slack.com/archives/C232SF3TK/p1570107013007500

- Add iteration gap for readiness/liveness probe executions which are checking if CRDs for ServiceCatalog exist. Previously, the probes were executed every 10 secs, which generates a heavy load on the main API Server. 

- Previously, we listed all K8s CRDs. Now we fetch only CRDs which belong to Service Catalog using proper label selector. As a result, the response uses fewer resources.

**Related issue(s)**

- https://github.com/kyma-project/kyma/issues/5869


resources/service-catalog/charts/catalog/values.yaml
```diff
@@ -1,6 +1,6 @@
 # Default values for Service Catalog
 # service-catalog image to use
-image: quay.io/kubernetes-service-catalog/service-catalog:v0.3.0-beta.0
+image: quay.io/kubernetes-service-catalog/service-catalog:v0.3.0-beta.0-2-g943b96d
```
mszostok: will be updated after Service Catalog release

PR #5886, opened 2019-10-04
Use kubectl delete instead of scale in Event Bus pre-upgrade

antoineco:
**Description**

Follows the usual pattern, clearer and less error prone.

**Related issue(s)**

#5853


antoineco:
Done in #6072

PR #5890, opened 2019-10-07
Improve stability of Application Operator Tests

Szymongib:
**Description**

- Event Service tests - Wait for Istio Sidecar to be up before starting tests
- Application Gateway Tests - Use resilient client

**Related issue(s)**


tests/integration/event-service/test/apitests/gateway_events_test.go
```diff
@@ -42,9 +42,6 @@ const (
 
 func TestGatewayEvents(t *testing.T) {
 
-	config, err := testkit.ReadConfig()
-	require.NoError(t, err)
```
Abd4llA: Can you give more context why would it be ok for ReadConfig to return an error ?

Szymongib: It is not. Reading config was moved to the `TestMain` function as it needs to be run only once https://github.com/kyma-project/kyma/pull/5890/files#diff-5d812308d56ccc84812557a61b547728R25

Issue #5938, opened 2019-10-09
Enable knative-serving test


aerfio:
**Description**

As we agreed in https://github.com/kyma-project/community/issues/369 we are turning off flaky tests in [this](https://github.com/kyma-project/kyma/pull//5937) PR, and `knative-serving` is one of them. This needs to be re-enabled.

nachtmaar:
Works on minikube on yesterday's master, might only occur in stability-checker. Taking a closer look now.

aerfio:
**Update after quality SIG meeting:**

During the meeting it was discussed that disabling whole test suite is too much most of the time, as only few tests inside are really unstable. It was proposed that a person on duty on CI-Force should have an easy way to disable one particular test by regexp or some other mechanism.

nachtmaar:
TL;DR

```
 textPayload: "	knative_serving_test.go:81: Cannot create test Service: services.serving.knative.dev "test-service" already exists
```
Possible caused by force deleting the knative service: https://github.com/kyma-project/kyma/blob/d2afb5838f900765287100a1b533527e2b4fb686/tests/knative-serving/knative_serving_test.go#L179

The two pictures show that the failure message always comes then the test failed:
<img width="799" alt="Screenshot 2019-10-10 at 18 05 22" src="https://user-images.githubusercontent.com/8013145/66587064-009fd600-eb8a-11e9-88f8-6c09a15d7aa6.png">
<img width="876" alt="Screenshot 2019-10-10 at 18 15 34" src="https://user-images.githubusercontent.com/8013145/66587065-01d10300-eb8a-11e9-805a-810eb225fd2b.png">

nachtmaar:
Implemented ...

PR #5984, opened 2019-10-11
fix knative-serving test

k15r:
**Description**

knative serving tests failed probably due to a force delete of the `ksvc` created during the test. The ksvc will now be removed without specifying immediate deletetion and it fails the test if the deletion fails.

**Related issue(s)**

#5938

### DIFF
--- a/tests/knative-serving/knative_serving_test.go
+++ b/tests/knative-serving/knative_serving_test.go
@@ -81,7 +81,7 @@ func TestKnativeServingAcceptance(t *testing.T) {
 		t.Fatalf("Cannot create test Service: %v", err)
 	}
 
-	defer deleteService(serviceClient, svc.Name)
+	defer deleteService(t, serviceClient, svc.Name)
 	var testServiceURL string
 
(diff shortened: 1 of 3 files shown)


tests/knative-serving/knative_serving_test.go
```diff
@@ -173,9 +173,9 @@ func loadKubeConfigOrDie(t *testing.T) *rest.Config {
 	return kubeConfig
 }
 
-func deleteService(servingClient servingtyped.ServiceInterface, name string) {
-	var deleteImmediately int64
-	_ = servingClient.Delete(name, &meta.DeleteOptions{
-		GracePeriodSeconds: &deleteImmediately,
-	})
+func deleteService(t *testing.T, servingClient servingtyped.ServiceInterface, name string) {
+	err := servingClient.Delete(name, &meta.DeleteOptions{})
```
antoineco: Don't forget to call `t.Helper()` before that line, so the error is logged at the correct location is the error trace.

PR #6018, opened 2019-10-15
Adding watch Verb in event-publish-service ClusterRole

anishj0shi:
evennt publish service is continuously overwhelmed with logs as the knative channel informer is unable to retrieve knative channels due to missing permission. with this PR we introduce the missing verb required for the channel informer to retrieve those Knative Channels

### DIFF
--- a/resources/event-bus/charts/event-publish-service/templates/service-account.yaml
+++ b/resources/event-bus/charts/event-publish-service/templates/service-account.yaml
@@ -17,6 +17,7 @@ rules:
     verbs:
       - get
       - list
+      - watch
   - apiGroups:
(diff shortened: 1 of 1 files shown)


resources/event-bus/charts/event-publish-service/templates/service-account.yaml
```diff
@@ -17,13 +17,15 @@ rules:
     verbs:
       - get
       - list
+      - watch
```
antoineco: This group should be removed entirely.

antoineco: This group should be removed entirely.

Issue #5639, opened 2019-09-13
Refactor acceptance tests


polskikiel:
AC:
- remove application-test
- create and use a helper, which skips test(s) on demand, hint: you can use testing.T.Name() method
- improve reports by displaying proper custom resources in case of test failure (for example ApplicationMapping, ClusterServiceClasses, ClusterServiceBroker, ServiceInstance etc.)

piotrmiskiewicz:
Proposed AC:
 - remove application-test
 - create and use a helper, which skips test(s) on demand, hint: you can use testing.T.Name() method
 - when a test is failing, improve report by displaying proper custom resources (for example ApplicationMapping, ClusterServiceClasses, ClusterServiceBroker, ServiceInstance etc.)

PR #6019, opened 2019-10-15
Refactor Service Catalog acceptance tests

ksputo:
**Description**

- Removed disabled and reduntant application-test
- Renamed "acceptance" directory to "service-catalog"
- Refactored report so it allows to choose which resources should be dumped in case of particular test failure
- Added missing resources to the report

**Related issue(s)**

#5639


tests/service-catalog/test/servicecatalog_test.go
```diff
@@ -158,19 +162,26 @@ func TestHelmBrokerClusterAddonsConfiguration(t *testing.T) {
...
+	// when
+	err = dynamicClient.Create(context.TODO(), addonsConfig)
+	require.NoError(t, err)
+
 	// then
 	repeat.AssertFuncAtMost(t, func() error {
 		_, err := scClient.ServicecatalogV1beta1().ClusterServiceClasses().Get(addonId, metav1.GetOptions{})
 		return err
-	}, time.Second*20)
+	}, time.Second*30)
```
polskikiel: FYI: We bumped the timeouts because of the fails on this step on ci-force

tests/service-catalog/test/servicecatalog_test.go
```diff
@@ -158,19 +162,26 @@ func TestHelmBrokerClusterAddonsConfiguration(t *testing.T) {
...
 
-	// when
-	err = dynamicClient.Create(context.TODO(), addonsConfig)
-	require.NoError(t, err)
 	defer func() {
+		if t.Failed() {
+			namespaceReport := report.NewReport(t,
+				k8sConfig,
+				report.WithHB())
+			namespaceReport.PrintJsonReport("default")
```
polskikiel: use empty string rather than "default", you can use `v1.NamespaceAll` const from k8s

tests/service-catalog/test/pkg/report/report.go
```diff
@@ -0,0 +1,258 @@
...
+		r.test.Logf("ServiceBindingUsage list unreachable: %v \n", err)
+	}
+	r.printObject(printer, sbus, "ServiceBindingUsage")
+}
+
+func (r Report) hbResourceDump(printer printers.ResourcePrinter, clientset *hbClient.Clientset, ns string) {
+	if clientset == nil {
+		return
+	}
+
```
polskikiel: not needed empty line, in above func we don't have one, let's be consistent

tests/service-catalog/test/pkg/report/report.go
```diff
@@ -0,0 +1,258 @@
...
+func NewReport(t *testing.T, config *rest.Config, options ...Option) *Report {
+	return &Report{
+		test:      t,
+		logWriter: TestLogWriter{testing: t},
+		cfg:       config,
+		opt:       options,
+	}
+}
+
+type Option func(r *Report, ns string)
```
polskikiel: We can call that type `ReportOption`, `LoggingOption` or `PrinterOption`. We can discuss it and decide

tests/service-catalog/test/pkg/report/report.go
```diff
@@ -0,0 +1,258 @@
...
+}
+
+func (t TestLogWriter) Write(p []byte) (n int, err error) {
+	t.testing.Log(string(p))
+
+	return len(p), nil
+}
+
+func (r *Report) setPrinter(pnr printers.ResourcePrinter) {
+	r.printer = pnr
```
polskikiel: can't we just set a printer in the NewReport function?

tests/service-catalog/test/common.go
```diff
@@ -1,4 +1,4 @@
-package servicecatalog
+package test
```
polskikiel: IMO it's better to move that file to `test/pkg/common/common.go` path

resources/service-catalog/templates/tests/rbac.yaml
```diff
@@ -25,6 +25,10 @@ rules:
     resources: ["clusterserviceclasses"]
     verbs: ["list", "delete", "get"]
     nonResourceURLs: []
+  - apiGroups: ["servicecatalog.k8s.io"]
+    resources: ["clusterservicebrokers"]
+    verbs: ["get", "list"]
+    nonResourceURLs: []
```
polskikiel: you can remove that line as it's not needed

polskikiel: you can even remove all of the other occurrences of that in this file :D

Issue #5829, opened 2019-09-30
Use generic prow jobs for Huskies' components


clebs:
**Description**
Huskies have no components that need to be adapted to the new generic Prow job. Instead the following need to be adapted:
- Monitoring integration test 
- Logging integration test
- ~~Backup/restore end-to-end test~~
- Alpine-net tool

For information on how to do the migration please see: #5481

PR #6037, opened 2019-10-16
Use generic makefiles for Huskies' components

suleymanakbas91:
**Description**

- Use generic makefiles

**Related issue(s)**

See https://github.com/kyma-project/kyma/issues/5829

### DIFF
--- a/scripts/docker.mk
+++ b/scripts/docker.mk
@@ -1,50 +1,6 @@
 # Default configuration
 IMG_NAME := $(DOCKER_PUSH_REPOSITORY)$(DOCKER_PUSH_DIRECTORY)/$(APP_NAME)
 TAG := $(DOCKER_TAG)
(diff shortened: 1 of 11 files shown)


scripts/docker.mk
```diff
@@ -0,0 +1,14 @@
+# Default configuration
+IMG_NAME := $(DOCKER_PUSH_REPOSITORY)$(DOCKER_PUSH_DIRECTORY)/$(APP_NAME)
+TAG := $(DOCKER_TAG)
+
+.PHONY: release
+
+release: build-image push-image
```
sjanota: Let's have two colons here, so one can extend it on component side.
```suggestion
release:: build-image push-image
```

sjanota:
Would it be possible to use docker.mk in go-dep.mk?

PR #6072, opened 2019-10-21
Fix upgrade hook for Knative-eventing chart

sayanh:
**Description**

- Fix to not scale down the subscription controller before deleting knative-subscriptions. This is to delete the `finalizers` in knative subscriptions and enable GC by k8s.

**Related issue(s)**

### DIFF
--- a/components/application-gateway/README.md
+++ b/components/application-gateway/README.md
@@ -90,3 +90,4 @@ The Console Backend Service has the following obligatory requirements:
 ### Contribution
 
 To learn how you can contribute to this project, see the [Contributing](/CONTRIBUTING.md) document.
+
--- a/resources/application-connector/values.yaml
+++ b/resources/application-connector/values.yaml
@@ -39,7 +39,7 @@ global:
     version: e029fcf4
   event_service:
     dir: develop/
-    version: "36059183"
+    version: e029fcf4
   event_service_integration_tests:
     dir: pr/
     version: PR-5890
--- a/resources/application-connector/values.yaml
+++ b/resources/application-connector/values.yaml
@@ -17,8 +17,8 @@ global:
   containerRegistry:
     path: eu.gcr.io/kyma-project
   application_operator:
-    dir: pr/
-    version: PR-5697
+    dir: develop/
+    version: c9f7a5f5
   application_operator_tests:
     dir: develop/
     version: 2569744b
--- a/resources/application-connector/values.yaml
+++ b/resources/application-connector/values.yaml
@@ -47,8 +47,8 @@ global:
     dir: pr/
     version: PR-5697
   application_gateway:
-    dir: develop/
-    version: e029fcf4
+    dir: pr/
+    version: PR-6072
   application_gateway_tests:
     dir: pr/
     version: PR-5890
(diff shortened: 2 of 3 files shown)


resources/knative-eventing/charts/knative-eventing/templates/pre-upgrade-job.yaml
```diff
@@ -57,6 +56,7 @@ data:
       subscriptions.eventing.knative.dev \
       clusterchannelprovisioners.eventing.knative.dev \
       channels.eventing.knative.dev --ignore-not-found
+    kubectl scale -n kyma-system deploy/event-bus-subscription-controller --replicas=0 || true
```
k15r: i know this is not part of this pr and also not its purpose but:
why would this fail:
- no event-bus-subscription-controller is available => we dont need to scale
- for whatever reason scaling fails (missing permissions ... ) => we definitely want to fail
- other internal issues => we want to fail

anishj0shi: @k15r => from 1.6 => master upgrade scenario we need the presence of our kyma subscription controller so that once knative subscriptions are deleted during thr pre upgrade process, we immediately reconcile them, once the reconciliation is don we scale the kyma subscription to 0. @sayanh please correct me If I'm wrong or add any more details.

antoineco: I still believe that thing is shaky. Please use `delete` instead of `scale` we use it in all other places.
See #5886

sayanh: @anishj0shi That's correct. We all believe `|| true` is not a good practice. Hence, changed `scale` to `delete`.

sayanh: True that. I changed it.

sayanh:
@nachtmaar I added a comment.

resources/application-connector/values.yaml
```diff
@@ -39,16 +39,16 @@ global:
     version: e029fcf4
```
antoineco: How did this file end up in this PR?

components/application-gateway/README.md
```diff
@@ -90,3 +90,4 @@ The Console Backend Service has the following obligatory requirements:
 ### Contribution
 
 To learn how you can contribute to this project, see the [Contributing](/CONTRIBUTING.md) document.
```
antoineco: How did this file end up in this PR?

resources/application-connector/values.yaml
anishj0shi: we need to force the image update in application operator upgrade test.

components/application-gateway/README.md
Szymongib: It is done to force rebuilding the component to unlock the upgrade job.

resources/application-connector/values.yaml
Szymongib: It is done to fix Application Operator upgrade tests, as it is not passing.

Issue #5627, opened 2019-09-12
Loki crashes repeatedly under load


davyodom:
When running some load tests, Loki crashes due to OOM errors repeatedly.  It is not clear in any documentation what is the expected load that the default configuration of Loki can handle, and what approach should be taken to scale Loki.  This was running against Kyma 1.4.

derberg:
@hisarbalik can you please take a look on https://github.com/kyma-project/kyma/issues/5700 as it seams like related/duplicated

a-thaler:
We configured a proper data retention policy and increased memeory.

PR #6099, opened 2019-10-24
Increase size of storage claim for prometheus https://github.com/kyma…

### LINKED ISSUE #6094: Prometheus runs out of disk space
**Description**

It seems that the 10GB for the PVC is not sufficient for a long running cluster. Prometheus (monitoring) crashes because the PV created by the prometheus-monitoring-db-prometheus-monitoring-0 PVC is not large enough. 

Suggestion is to increase the size and change the retention size (in combination with time) to ensure we don't go over the limit

brendandobbs:
Fixes #6094


a-thaler:
Will also fix #5627

PR #6191, opened 2019-10-31
Expose only necessary Hydra endpoints

colunira:
**Description**

- Virtual service doesn't expose admin endpoints except `/oauth2/introspect`
- Virtual service exposes public endpoints

**Related issue(s)**


resources/ory/templates/hydra-virtualservice.yaml
```diff
@@ -1,46 +1,43 @@
...
-    - {{ .Values.global.istio.gateway.name }}.{{ .Values.global.istio.gateway.namespace }}
-  hosts:
-    - oauth2-admin.{{ .Values.global.ingress.domainName }}
-  http:
+    {{- end -}}
+    {{- if .Values.hydra.service.admin.enabled -}}
     - match:
         - uri:
-            regex: /.*
+            prefix: "/oauth2/introspect"
```
Tomasz-Smelcerz-SAP: I think this should be `exact` match

PR #6200, opened 2019-11-04
Upgrade istio to 1.3.4

Demonsthere:
**Description**

- Upgrade istio to 1.3.4
- Add missing service port names
- Initial gateway split

**Related issue(s)**

- Implements #5315


tests/end-to-end/upgrade/main.go
```diff
@@ -195,7 +195,12 @@ func getDomainNameFromCluster(k8sCli *k8sClientSet.Clientset) (string, error) {
 		return "", err
 	}
 
-	value, _ := overrides.FindOverrideStringValue(coreOverridesMap, "global.ingress.domainName")
+	value, found := overrides.FindOverrideStringValue(coreOverridesMap, "global.ingress.domainName")
+	logrus.Infof(fmt.Sprintf("using domainName: %v", value))
```
piotrmiskiewicz: maybe something like:
```
logrus.Infof("using domainName: %v", value)
```

tests/end-to-end/upgrade/main.go
```diff
@@ -205,6 +210,7 @@ func getDexConfigFromCluster(k8sCli *k8sClientSet.Clientset, userSecret, dexName
 		secret, err := k8sCli.CoreV1().Secrets(dexNamespace).Get(userSecret, metav1.GetOptions{})
 		if err != nil {
 			logrus.Infof("while getting dex secret: %v", err)
+			time.Sleep(5 * time.Second)
```
piotrmiskiewicz: I'd like to remove this sleep  or put a comment why in this place you really need 5 seconds. On the other hand the waiter util has 1000 milliseconds sleep so maybe it is enough

resources/istio-init/templates/kyma-job-cleaner.yaml
```diff
@@ -25,10 +25,10 @@ spec:
           - /bin/bash
           - -c
           - |
-            kubectl delete -n istio-system job istio-init-crd-10 --ignore-not-found=true
-            kubectl delete -n istio-system job istio-init-crd-11 --ignore-not-found=true
-            kubectl delete -n istio-system job istio-init-crd-certmanager-10 --ignore-not-found=true
-            kubectl delete -n istio-system job istio-init-crd-certmanager-11 --ignore-not-found=true
+            kubectl delete -n istio-system job istio-init-crd-10-{{ .Values.global.tag | printf "%v" | trunc 32 }} --ignore-not-found=true
+            kubectl delete -n istio-system job istio-init-crd-11-{{ .Values.global.tag | printf "%v" | trunc 32 }} --ignore-not-found=true
```
kubadz: First - that job lacks istio-init-crd-12.
Second - isn't this job useless? Jobs already have a hook `helm.sh/hook-delete-policy: "before-hook-creation"` which will delete the appropriate jobs if such already exist.

There is also one other thing to consider: now in case of Istio version upgrade old jobs will never be removed. Maybe we should instead run this job in post-upgrade step?

resources/istio/values.yaml
```diff
@@ -125,8 +139,12 @@ pilot:
     requests:
       cpu: 100m
       memory: 128Mi
+  rollingMaxSurge: 1
+  rollingMaxUnavailable: 0
   env:
     PILOT_HTTP10: 1
+    PILOT_SKIP_VALIDATE_TRUST_DOMAIN: "true"
+  enableProtocolSniffing: false
```
kubadz: that value is not used anywhere in the chars. Probably should be `enableProtocolSniffingForOutbound` and `enableProtocolSniffingForInbound` instead

PR #6214, opened 2019-11-05
Split gateways

colunira:
**Description**

- Split kyma-gateway into kyma-gateway, kyma-gateway-application-connector and kyma-gateway-compass 
- Update virtual services with proper gateway

**Related issue(s)**


resources/core/charts/gateway/templates/compass-gateway.yaml
```diff
@@ -0,0 +1,18 @@
+apiVersion: networking.istio.io/v1alpha3
```
Szymongib: Can we get rid of this Gateway? I feel like we are complicating things a whole bunch instead of simplifying them. I guess the desired scenario would be to have Compass Gateways always in Compass chart and install them no matter if we are installing Compass with or without Core.
Can we achieve it?

resources/compass/charts/gateway/templates/gateway.yaml
```diff
@@ -34,4 +25,25 @@ spec:
         httpsRedirect: true # automatic 301 redirect from http to https
       hosts:
         - "*.{{.Values.global.ingress.domainName}}"
+
+---
+
+apiVersion: networking.istio.io/v1alpha3
```
Szymongib: To make things simpler those Gateways should be installed always so we do not need `.Values.gateway.enabled` which is switched depending if we install Compass with or without Core.

jakkab: modified as agreed.

resources/core/charts/gateway/templates/compass-gateway.yaml
jakkab: removed

jakkab:
@kfurgol @Szymongib would you mind taking another look? I've addressed your comment.

Issue #1623, opened 2018-11-10
UI API memory consumption constantly growing


michal-hudy:
**Description**

UI API memory consumption constantly growing in time. UI API has multiple in-memory caches, so maybe it related to this - informers and cache for content (BigCache). It is also possible that there is a memory leak. It should be investigated.

Memory consumption from ~10H:
![image](https://user-images.githubusercontent.com/26445970/48304726-805a9600-e51e-11e8-9c0b-64854d6d8d50.png)

**Expected result**

Memory should be released if not needed.

Szymongib:
Should be resolved by #1910

PR #6227, opened 2019-11-06
Increasing limit because of prometheus container memory demand.

dekiel:
**Description**
After deploying stackdriver collector and defining additional grafana charts, prometheus container was suffering OOMKiler. We need incraese max container limit in limitrange to 6Gi as prometheus presented demand for at least 5Gi.

Increasing max memory in kyma-default limitrange to 6Gi.

**Related issue(s)**
See also test infra issue #1623

### DIFF
--- a/resources/cluster-essentials/templates/limit-range.yaml
+++ b/resources/cluster-essentials/templates/limit-range.yaml
@@ -5,7 +5,7 @@ metadata:
 spec:
   limits:
   - max:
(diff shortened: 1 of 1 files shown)


derberg:
@dekiel do I get it right, that the need is only for some pipelines of ours? if so, I would say we should not increate limits for entire Kyma but rather template the value and enable for overriding in our pipelines only

Issue #6226, opened 2019-11-06
Tracing does not work from publish-service to subscriber


sayanh:
**Description**

Trace spans are not available from publish service to subscriber.

**Expected result**

Complete trace spans should be available from event-service to subscriber.

**Steps to reproduce**

Change istio trace sampling to 100% and create an event. Check in jaeger for trace spans.

Kyma versions: 1.6+

k15r:
Will be included in kyma patch 1.7.1

PR #6230, opened 2019-11-06
Fix tracing

sayanh:
**Description**

- Enabled tracing for natss dispatcher. Using sample rate 0.1 for now but that can be configured as per need.

**Related issue(s)**

See #6226


resources/event-bus/charts/event-publish-service/templates/_helpers.tpl
```diff
@@ -20,6 +20,7 @@ app.kubernetes.io/name: {{ template "event-publish-service.name" . }}
 app.kubernetes.io/instance: {{ .Release.Name }}
 app.kubernetes.io/managed-by: {{ .Release.Service }}
 app.kubernetes.io/part-of: kyma
+app: {{ template "event-publish-service.name" . }}
```
antoineco: We had a PR some time ago to explicitly avoid those labels. Why are they back?

resources/knative-provisioner-natss/charts/knative-provisioner-natss/templates/pre-upgrade-job.yaml
```diff
@@ -0,0 +1,105 @@
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: {{ .Release.Name }}-pre-upgrade
+  annotations:
+    "helm.sh/hook": "pre-upgrade"
```
antoineco: That's JSON notation. YAML has no quotes.

resources/event-bus/charts/event-publish-service/templates/_helpers.tpl
sayanh: These are not needed hence I removed it.

resources/knative-provisioner-natss/charts/knative-provisioner-natss/templates/pre-upgrade-job.yaml
sayanh: True that hence removed except the weights where we wanna force them to be strings.

sayanh:
@antoineco 
> Could you please add one sentence to the issue description explaining why using a sample rate value of 0.1 fixes the missing traces? The README says 0.1 is not enough, so I'm a bit confused.

We are enabling traces as you see in the config map and using 0.1 as the sampling rate. Just using "sample rate value of 0.1" doesn't fix the issue. This is good for a start. Meaning, if a trace is sampled, then the complete trace is there otherwise the complete trace is missing. Documentation talks about what needs to be done if one needs more trace sampling.

docs/tracing/10-01-traces-troubleshooting.md
```diff
@@ -5,11 +5,20 @@ type: Troubleshooting
 
 ## Jaeger shows only a few traces
 
-The current Istio Pilot settings define the trace sampling rate at `1.0`, where `100` is the maximum value. This means that only 1 out of 100 requests is sent to Jaeger for trace recording. To change this system behavior, run:
+1. The current Istio Pilot settings define the trace sampling rate at `1.0`, where `100` is the maximum value. This means that only 1 out of 100 requests is sent to Jaeger for trace recording. To change this system behavior, run:
```
bszwarc:
```suggestion
Jaeger showing just a few traces may be caused by the following settings:
1. Istio Pilot sets the trace sampling rate at `1.0`, where `100` is the maximum value. This means that only 1 out of 100 requests is sent to Jaeger for trace recording. To change this system behavior, run:
```
Let's add a short introductory sentence to the points.

docs/tracing/10-01-traces-troubleshooting.md
```diff
@@ -5,11 +5,20 @@ type: Troubleshooting
...
 
 ```bash
 kubectl -n istio-system edit deploy istio-pilot
 ```
 Set the **traceSampling** parameter to a desired value, such as `60`.
 
 >**NOTE**: Using a very high value may affect Istio's performance and stability.
+
+2. The current Knative trace sampling rate is also at `0.1` where `1` is the maximum value. To have complete trace recordings, run:
```
bszwarc:
```suggestion
2. The current Knative trace sampling rate is at `0.1` where `1` is the maximum value. To have complete trace recordings, run:
```

resources/knative-eventing/values.yaml
```diff
@@ -1,2 +1,8 @@
 global:
   isLocalEnv: false
+
+knative-eventing:
+  tracing:
+    enable: true
```
k15r: is there a way to get this value from kymas istio configuration? It should be the same value across all kyma

resources/knative-provisioner-natss/charts/knative-provisioner-natss/templates/natss.yaml
```diff
@@ -162,7 +163,9 @@ spec:
         - port: 80
           protocol: TCP
           targetPort: 8080
+          name: http
     selector:
+        app: natss-ch-dispatcher
```
k15r: Do we really need the app in the selector? The selector should also apply to this pod without relying on the app-label.
If i understand it correctly we put app here just for proper trace naming.

resources/knative-eventing/values.yaml
```diff
@@ -1,2 +1,8 @@
...
+    collectorUrl: http://zipkin.kyma-system.svc.cluster.local:9411/api/v2/spans
```
sayanh: Add a comment on why?

resources/knative-eventing/values.yaml
sayanh: Definitely, there is a way by using "overrides". But that will bloat the PR scope IMO and we have to holistically solve it for all the components e.g. publish-service, console-backend-service. How about solving it in a different issue?

resources/knative-eventing/values.yaml
sayanh: Comment is added.

resources/knative-provisioner-natss/charts/knative-provisioner-natss/templates/natss.yaml
sayanh: No, we don't need selector. Hence removed it as well as the pre-upgrade hook.

PR #6236, opened 2019-11-07
Knative Source HTTP Adapter for Broker/Trigger

nachtmaar:
**Description**

- Implement an HTTP Adapter to receive CloudEvents on a HTTP endpoint
- Endpoint is CloudEvents 1.0 compatible, declines CE events lower than 1.0
- Implement Readiness Probe on `/healthz`

Credits also go to @k15r 

**Related issue(s)**

Due to an open issue in knative/eventing#2052, we need to patch vendor :/


components/event-sources/Makefile
```diff
@@ -16,6 +16,8 @@ APP_NAME = event-sources
 APP_PATH = components/event-sources
 BUILDPACK = eu.gcr.io/kyma-project/test-infra/buildpack-golang:go1.12
 SCRIPTS_DIR = ../../common/makefiles
+
+patched_trace = vendor/go.opencensus.io/trace/trace.go
```
antoineco: I feel like this variable belongs closer to where it's actually used, instead of mixed with the variables used by the shared Makefile.

components/event-sources/Makefile
```diff
@@ -53,13 +56,16 @@ vendor: Gopkg.lock
 	@echo '+ Pulling vendored dependencies'
 	@dep ensure -v --vendor-only
 
+$(patched_trace): hack/set-span-id.patch vendor
+	@echo '+ Patching opencensus dependency'
+	@git apply ${REPO_ROOT_DIR}/components/event-sources/hack/set-span-id.patch
```
antoineco: You could avoid that complicated path-fu if you just trimmed "components/event-sources/" from the paths inside your patch file:
```diff
--- a/vendor/go.opencensus.io/trace/trace.go
+++ b/vendor/go.opencensus.io/trace/trace.go
```

components/event-sources/Makefile
```diff
@@ -53,13 +56,16 @@ vendor: Gopkg.lock
 	@echo '+ Pulling vendored dependencies'
 	@dep ensure -v --vendor-only
 
+$(patched_trace): hack/set-span-id.patch vendor
```
antoineco: Your patch is _modifying_ an existing file if I'm not mistaken, which means your target must be PHONY, otherwise `make` will report "xyz is up-to-date."

components/event-sources/adapter/http/adapter.go
```diff
@@ -0,0 +1,196 @@
+package http
+
+import (
+	"context"
```
antoineco: Grouping imports would be nice, at the very least separate stdlib from third-parties.

components/event-sources/adapter/http/adapter.go
```diff
@@ -0,0 +1,196 @@
...
+	if !ok {
+		panic(fmt.Sprintf("cannot create adapter, expecting a *envconfig, but got a %T", processed))
+	}
+
+	return &httpAdapter{
+		adapterContext: ctx,
+		ceClient:       ceClient,
+		statsReporter:  reporter,
+		accessor:       accessor,
+		logger:         logging.FromContext(ctx).Desugar(),
```
antoineco: Any particular reason for de-sugarring the logger? The sugarred logger provides an API that's slightly more use-friendly.

components/event-sources/adapter/http/adapter.go
```diff
@@ -0,0 +1,196 @@
...
+	}
+}
+
+// Start is the entrypoint for the adapter and is called by sharedmain coming from pkg/adapter
+func (h *httpAdapter) Start(_ <-chan struct{}) error {
+	logger := h.logger
```
antoineco: (nit) That assignment could be removed since you make a single call to the logger.

components/event-sources/adapter/http/adapter.go
```diff
@@ -0,0 +1,196 @@
...
+
+	log.Printf("listening on :%d%s\n", h.accessor.GetPort(), endpointCE)
+
+	// note about graceful shutdown:
+	// TLDR; StartReceiver unblocks as soon as a stop signal is received
+	// `StartReceiver` waits internally until `ctx.Done()` does not block anymore
+	// the context `h.adapterContext` returns a channel (when calling `ctx.Done()`)
+	// which is closed as soon as a stop signal is received, see https://github.com/knative/pkg/blob/master/signals/signal.go#L37
+	if err := c.StartReceiver(h.adapterContext, h.serveHTTP); err != nil {
+		return fmt.Errorf("failed to start receiver: %v", err)
```
antoineco: The error  message is misleading, an error could also occur after the receiver has been successfully started if I'm not mistaken.

components/event-sources/adapter/http/adapter.go
```diff
@@ -0,0 +1,196 @@
...
+func (h *httpAdapter) Start(_ <-chan struct{}) error {
+	logger := h.logger
+
+	t, err := cloudevents.NewHTTPTransport(
+		cloudevents.WithPort(h.accessor.GetPort()),
+		cloudevents.WithPath(endpointCE),
+		cloudevents.WithMiddleware(WithReadinessMiddleware),
+	)
+	if err != nil {
+		return fmt.Errorf("failed to create transport, %v", err)
```
antoineco: (nit) You could spare all that error formatting by standardizing on `errors.Wrap(err, "some context")`, which is already vendored and seems widely used in Knative as well.

components/event-sources/adapter/http/adapter.go
```diff
@@ -0,0 +1,196 @@
...
+	GetPort() int
+}
+
+const resourceGroup = "http." + sources.GroupName
+
+const (
+	// endpoint for cloudevents
+	endpointCE = "/"
+	// endpoint for readiness check
+	readinessReadiness = "/healthz"
```
antoineco: Did you mean "readinessEndpoint"?

components/event-sources/adapter/http/adapter.go
```diff
@@ -0,0 +1,196 @@
...
+		EventType:     event.Type(),
+		Name:          "http_adapter",
+		ResourceGroup: resourceGroup,
+	}
+
+	logger.Debug("sending event", zap.Any("sink", h.accessor.GetSinkURI()))
+	rctx, revt, err := h.ceClient.Send(ctx, event)
+	if err != nil {
+		h.logger.Error("failed to send cloudevent to sink", zap.Error(err), zap.Any("sink", h.accessor.GetSinkURI()))
+		resp.Error(http.StatusBadGateway, "")
```
antoineco: Will it always be a 502 here? Doesn't the error code depend on the response from the sink?

components/event-sources/adapter/http/adapter.go
```diff
@@ -0,0 +1,196 @@
...
+		resp.RespondWith(http.StatusInternalServerError, revt)
+		return nil
+	}
+
+	// report a sent event
+	if err := h.statsReporter.ReportEventCount(reportArgs, rtctx.StatusCode); err != nil {
+		h.logger.Warn("cannot report event count", zap.Error(err))
+	}
+
+	if rtctx.StatusCode/100 == 2 {
```
antoineco: wat

components/event-sources/cmd/http-adapter/main.go
```diff
@@ -0,0 +1,8 @@
+package main
+
+import "knative.dev/eventing/pkg/adapter"
+import "github.com/kyma-project/kyma/components/event-sources/adapter/http"
```
antoineco: Good candidate for an `import()` block since you have more than 1.

components/event-sources/cmd/http-adapter/main.go
```diff
@@ -0,0 +1,8 @@
...
+
+func main() {
+	adapter.Main("application-source", http.NewEnvConfig, http.NewAdapter)
```
antoineco: "application" -> "http"

components/event-sources/Makefile
antoineco: `patch -p0 < hack/set-span-id.patch` would be preferable here. Your file is not checked out due to the ignored vendor directory, so you `git apply` a file unknown to Git. It may work, but it feels a bit clunky to me.

components/event-sources/adapter/http/README.md
```diff
@@ -0,0 +1,70 @@
...
+```
+
+The adapter will send events to the `SINK_URI`. When running the adapter locally, you can simple use `netcat` as a sink:
+```bash
+printf "HTTP/1.1 200 OK\r\n\r\n" | nc -vl 55555
+```
+
+Send an event to the adapter:
+
+```bash
```
antoineco: (nit) `console` provides a better highlighting for console commands (anything starting with `#`, `$` followed by a command and then some output)

components/event-sources/adapter/http/adapter.go
```diff
@@ -0,0 +1,196 @@
...
+	"log"
+	"net/http"
+)
+
+type envConfig struct {
+	adapter.EnvConfig
+	EventSource string `envconfig:"EVENT_SOURCE" required:"true"`
+
+	// PORT as required by knative serving runtime contract
+	Port int `envconfig:"PORT" required:"true" default:"8080"`
```
antoineco: Do these get populated without calling `envconfig.Process()`? If yes, how does this black magic work?

components/event-sources/adapter/http/adapter.go
nachtmaar: We can also use the sugared logger if you prefer. I know the desugared logger's performance is a bit better, but that is not the reason why I choose it. I am used to the desugared one from knative code.

components/event-sources/adapter/http/adapter.go
nachtmaar: Good catch! I meant `endpointReadiness`

components/event-sources/adapter/http/adapter.go
nachtmaar: Ideally there would be a helper in `http`. `http.is2XX(statusCode)` which would improve readability. Let me check if there is something like this.

components/event-sources/cmd/http-adapter/main.go
nachtmaar: Good catch. Thanks

components/event-sources/adapter/http/adapter.go
nachtmaar: True.

components/event-sources/adapter/http/adapter.go
nachtmaar: Did not find a nice helper. Refactored it to a function to improve readability.

components/event-sources/adapter/http/adapter.go
nachtmaar: Take from: https://github.com/cloudevents/sdk-go/blob/master/pkg/cloudevents/client/client.go#L82

> // Send transmits the provided event on a preconfigured Transport.
> // Send returns a response event if there is a response or an error if there
> // was an an issue validating the outbound event or the transport returns an
> // error.

Because we already validate the event beforehand, there should only be transport errors left for this case.

Http status checks are performed after `err` is handled. 

Does that awnser your question?

components/event-sources/Makefile
nachtmaar: Sure, why not.

components/event-sources/adapter/http/README.md
nachtmaar: I separated the commands into two bash blocks since the curl output is never shown I did not use `console`. I also removed the leading `$`. Please check if that is ok for you :)

components/event-sources/Makefile
nachtmaar: Very good suggestion 👍

components/event-sources/adapter/http/adapter.go
antoineco: I'm fine with both 👍

components/event-sources/adapter/http/adapter.go
antoineco: Yup, thanks! 👍

components/event-sources/Makefile
```diff
@@ -53,13 +55,15 @@ vendor: Gopkg.lock
...
+	@echo '+ Patching opencensus dependency'
+	@patch -p0 < hack/set-span-id.patch
 
 .PHONY: test
-
-testpkgs := $(shell go list ./... | grep -vE '/apis/|/client/|/cmd/|/testing')
+testpkgs := $(shell go list ./... | grep -vE '/apis/|/client/|/testing')
```
antoineco: There is no test in `cmd`, and `cmd` tests are not very relevant considering we just call `shared.Main` there, so this should remain probably remain ignored.

nachtmaar: True, not anymore. We were calling sharedmain in one of our early tests. 
Will revert the file.

nachtmaar: Done

components/event-sources/adapter/http/README.md
```diff
@@ -0,0 +1,74 @@
+# HTTP Adapter Source
+
+The http adapter is a http server which receives [CloudEvents](https://github.com/cloudevents/spec) in version 1.0 and proxies them to a preconfigured sink.
+The adapter is written with the [cloudevents sdk-go](https://github.com/cloudevents/sdk-go) and is fully compatible with CloudEvents specification.
+
+## Specification
+
+The http adapter is CloudEvents 1.0 compatible. It accepts binary as well as structured content mode. Only batched mode is not implemented.
+Read more about the content types and the http protocol binding [here](https://github.com/cloudevents/spec/blob/master/http-protocol-binding.md#13-content-modes).
```
bszwarc:
```suggestion
See [this document](https://github.com/cloudevents/spec/blob/master/http-protocol-binding.md#13-content-modes) for details about the content types and the HTTP protocol binding.
```
Not much I can say here - to me `this document` sounds better than `here` at the end of the sentence.

components/event-sources/adapter/http/README.md
```diff
@@ -0,0 +1,74 @@
...
+
+## Run
```
bszwarc:
```suggestion
## Usage
```
Again, taken from the template.

components/event-sources/adapter/http/README.md
```diff
@@ -0,0 +1,74 @@
...
+
+The adapter can be run locally.
```
bszwarc:
```suggestion
You can run the adapter locally.
```
If possible, it's better to use the active voice.

components/event-sources/adapter/http/README.md
```diff
@@ -0,0 +1,74 @@
...
+
+It requires a few variables to be set:
```
bszwarc:
```suggestion
1. Start with exporting the variables. Run: 
```
I added the steps, to introduce the logical order.

components/event-sources/adapter/http/README.md
```diff
@@ -0,0 +1,74 @@
...
+
+```bash
+export SINK_URI=http://localhost:55555
+export NAMESPACE=foo
+export K_METRICS_CONFIG='{"config-observability":"{\"metrics.backend\":\"prometheus\"}"}'
+export K_LOGGING_CONFIG='{"zap-logger-config": "{\"level\":\"info\",\"development\":\"true\",\"outputPaths\":[\"stdout\"],\"errorOutputPaths\":[\"stderr\"],\"encoding\":\"console\",\"encoderConfig\":{\"timeKey\":\"ts\",\"levelKey\":\"level\",\"nameKey\":\"logger\",\"callerKey\":\"caller\",\"messageKey\":\"msg\",\"stack traceKey\":\"stacktrace\",\"lineEnding\":\"\",\"levelEncoder\":\"\",\"timeEncoder\":\"iso8601\",\"durationEncoder\":\"\",\"callerEncoder\":\"\"}}"}'
+export EVENT_SOURCE="varkes"
+```
+
+Run it:
```
bszwarc:
```suggestion
2. Run the adapter:
```

components/event-sources/adapter/http/README.md
```diff
@@ -0,0 +1,74 @@
...
+```bash
+go run cmd/http-adapter/main.go
+```
+
+The adapter will send events to the `SINK_URI`. When running the adapter locally, you can simple use `netcat` as a sink:
```
bszwarc:
```suggestion
As a result, the adapter will send events to the `SINK_URI`. When running the adapter locally, you can simply use `netcat` as a sink:
```

components/event-sources/adapter/http/README.md
```diff
@@ -0,0 +1,74 @@
...
+    -H "ce-source: will be replaced" \
+    -H "ce-type: foo" \
+    -H "ce-eventtypeversion: 0.3" \
+    -H "ce-id: A234-1234-1234" \
+    -H "content-type: text/xml" \
+    -d '<much wow=\"xml\"/>' \
+    http://localhost:8080
+```
+
+## Testing
```
bszwarc:
```suggestion
## Development

### Testing
```
I added the `Development` section - if there are any additional things apart from the testing we can add them here

components/event-sources/adapter/http/README.md
```diff
@@ -0,0 +1,74 @@
...
+
+Unit and integration tests can be run using the following command:
```
bszwarc:
```suggestion
Run unit and integration tests:
```
Active voice + shorter version :)

components/event-sources/adapter/http/README.md
```diff
@@ -0,0 +1,74 @@
...
+```bash
+go run cmd/http-adapter/main.go
+```
+
+The adapter will send events to the `SINK_URI`. When running the adapter locally, you can simple use `netcat` as a sink:
+```bash
+printf "HTTP/1.1 200 OK\r\n\r\n" | nc -vl 55555
+```
+
+Send an event to the adapter:
```
bszwarc:
```suggestion
### Send events to the adapter

You can send events using the structured or binary content mode. 
```

components/event-sources/adapter/http/README.md
```diff
@@ -0,0 +1,74 @@
...
+
+### Structured Content Mode
```
bszwarc:
```suggestion

To use the structured content mode, run:  
```
I would use an introductory line instead of a heading, to reduce the number of headings a bit and also introduce the code snippet.

components/event-sources/adapter/http/README.md
```diff
@@ -0,0 +1,74 @@
...
+    "source": "will be replaced",
+    "type": "foo",
+    "eventtypeversion": "0.3",
+    "id" : "A234-1234-1234",
+    "datacontenttype" : "text/xml",
+    "data" : "<much wow=\"xml\"/>"
+}' -H "Content-Type: application/cloudevents+json" -X POST http://localhost:8080
+```
+
+### Binary Content Mode
```
bszwarc:
```suggestion
To use the binary content mode, run: 
```

components/event-sources/adapter/http/README.md
```diff
@@ -0,0 +1,74 @@
...
+The http adapter is a http server which receives [CloudEvents](https://github.com/cloudevents/spec) in version 1.0 and proxies them to a preconfigured sink.
+The adapter is written with the [cloudevents sdk-go](https://github.com/cloudevents/sdk-go) and is fully compatible with CloudEvents specification.
+
+## Specification
+
+The http adapter is CloudEvents 1.0 compatible. It accepts binary as well as structured content mode. Only batched mode is not implemented.
+Read more about the content types and the http protocol binding [here](https://github.com/cloudevents/spec/blob/master/http-protocol-binding.md#13-content-modes).
+
+## Run
+
```
bszwarc:
```suggestion
### Run the adapter locally
```
I would have two subsections to the `Usage` part:
- `Run locally`
- `Send events to adapter`

components/event-sources/adapter/http/README.md
```diff
@@ -0,0 +1,78 @@
...
+The HTTP adapter is an HTTP server that receives [CloudEvents](https://github.com/cloudevents/spec) in version 1.0 and proxies them to a preconfigured sink.
+The adapter is written using [Go SDK for CloudEvents](https://github.com/cloudevents/sdk-go) and is fully compatible with the CloudEvents 1.0 specification.
+
+
+It accepts binary and structured content modes. The batched mode is not implemented.
+See [this document](https://github.com/cloudevents/spec/blob/master/http-protocol-binding.md#13-content-modes) for details about the content types and the HTTP protocol binding.
+
+## Usage
+### Run the adapter locally
+You can run the adapter locally.
```
bszwarc:
```suggestion
To use the adapter locally:
```
We can actually change this part a bit into an introductory sentence.

components/event-sources/adapter/http/adapter.go
```diff
@@ -0,0 +1,203 @@
...
+
+	"github.com/kyma-project/kyma/components/event-sources/apis/sources"
+)
+
+type envConfig struct {
+	adapter.EnvConfig
+	EventSource string `envconfig:"EVENT_SOURCE" required:"true"`
+
+	// PORT as required by knative serving runtime contract
+	Port int `envconfig:"PORT" required:"true" default:"8080"`
```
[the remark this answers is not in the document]
antoineco: It is set by the controller, and then Knative exports that variable. The "hardcoded" value just provides a sand default to run locally.

components/event-sources/adapter/http/adapter.go
```diff
@@ -0,0 +1,203 @@
...
+
+// is2XXStatusCode checks whether status code is a 2XX status code
+func is2XXStatusCode(statusCode int) bool {
+	return statusCode >= http.StatusOK && statusCode < http.StatusMultipleChoices
+}
+
+// isSupportedCloudEvent determines if an incoming cloud event is accepted
+func isSupportedCloudEvent(event cloudevents.Event) bool {
+	eventVersion := event.SpecVersion()
+	return eventVersion != cloudevents.VersionV01 && eventVersion != cloudevents.VersionV02 && eventVersion != cloudevents.VersionV03
```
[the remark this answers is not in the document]
nachtmaar: We could do so. With the given code we will also support newer cloudevents versions as soon as support is build into the sdk (of course we still need to update the sdk dependency).
We don't need to update the code to e.g. `return eventVersion == cloudevents.VersionV1 || eventVersion == cloudevents.VersionV12` as we would with your code.
I would prefer going with the blacklisting approach. Could I convince you?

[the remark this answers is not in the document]
montaro: You could! Thanks :D

Issue #6113, opened 2019-10-25
Improve makefiles in all Application Connectivity components 


akgalwas:
**Description**

This is a follow up for #5481. 

We need to make the following improvements:
- Use latest build packs (buildpack-golang:go1.11 is used) 
- Improve Compass Runtime Agent makefile (turning go modules on)

**Reasons**

The currently used build pack is not up to date.

PR #6244, opened 2019-11-08
Use newest buildpack in Makefiles and newest images in Dockerfiles of Application Connectivity

franpog859:
**Description**

- Use newest buildpack in Makefiles of Application Connectivity
- Use newest images in Dockerfiles of Application Connectivity
- Add `licenses` to .gitignores 
- Adjust old development scripts to match new Makefile approach

**Related issue(s)**

See the issue #6113


components/application-registry/scripts/can-i-commit.sh
```diff
@@ -24,15 +24,15 @@ if [ $? != 0 ]
 fi
 
 ##
-# MAKE BUILD
+# MAKE VERIFY
```
Szymongib: Do we still need this file? I guess we can do basically the same just by running `make verify`.

components/connector-service/scripts/can-i-commit.sh
```diff
@@ -24,15 +24,15 @@ if [ $? != 0 ]
 fi
 
 ##
-# MAKE BUILD
+# MAKE VERIFY
```
Szymongib: Same as in `application-registry`, not sure if we still need it.

tests/application-connector-tests/Dockerfile
```diff
@@ -1,13 +1,14 @@
-FROM golang:1.10-alpine as builder
+FROM golang:1.13-alpine as builder
 
 ENV SRC_DIR=/go/src/github.com/kyma-project/kyma/tests/application-connector-tests
+ENV CGO_ENABLED=0
```
Szymongib: I think we can do it the same as in other components, setting this var only for building binary like this: `RUN CGO_ENABLED=0 go test...`

components/application-registry/scripts/can-i-commit.sh
franpog859: Done!

components/connector-service/scripts/can-i-commit.sh
franpog859: Done!

tests/application-connector-tests/Dockerfile
franpog859: Done!

Issue #6036, opened 2019-10-16
Reduce the number of requests to API server


michal-hudy:
**Description**

Based on metrics provided by (https://github.com/kyma-project/kyma/issues/6035) number of requests to API server should be reduced to an absolute minimum.

AC:
 - Identified components/tests are generating fewer requests (calculate the improvements)
 - Based on fixes create a document that will describe best practices for Kyma developers

**Reasons**

As we already have a list of components that are making too many requests to API server we need to improve them. Thanks to that Kyma will be a lighter solution and probably performance will be also improved.

derberg:
@michal-hudy please specify exactly what and in which component things have to be improved and provide data that proves how we know it is a big problem. At the moment it is impossible to decide who can execute

adamwalach:
This issue is killing our long-leaving-cluster - weekly. The api-server is hammered down by the traffic. Basically we can't run any kubectl queries.
We increased number of nodes from 3 to 6 and it improved situation a bit. 
These are charts describing api load on a 6 node weekly cluster:
<img width="1362" alt="Screen Shot 2019-11-13 at 12 22 32" src="https://user-images.githubusercontent.com/8530211/68760034-3ca6da80-0611-11ea-840b-b7b4ed721c67.png">


<img width="1342" alt="Screen Shot 2019-11-13 at 12 22 45" src="https://user-images.githubusercontent.com/8530211/68760063-4d575080-0611-11ea-936a-0fb1c8882196.png">

adamwalach:
By suggestion from GCP support we changed type of weekly and nightly clusters to [regional](https://cloud.google.com/kubernetes-engine/docs/concepts/regional-clusters). Since we did the change the clusters are more stable, and we could not observe any issue with api-server not being available.

kasiakepka:
Looking at the AC I believe we can't assume that this issue can be closed.
Goats already implemented [dashboards](https://github.com/kyma-project/kyma/pull/6079) that help us to measure potential improvements.
@piotrmsc did you notice any improvements in the number of requests to API?

If we want to create "document that will describe best practices for Kyma developers" I think we should push this issue further. Can POs take care of that?

kasiakepka:
On SIG Quality meeting it was agreed to close the issue and create separate one focused on Mixer (the biggest contributor when it comes to the number of requests to the APiserver).
Potential Mixer improvements are documented here: https://github.com/kyma-project/kyma/issues/6348

PR #6247, opened 2019-11-08
Disable the pod-not-running rule

majakurcius:
**Description**

- Disable the `pod-not-running` Prometheus rule until the problem with the unstable k8s API on infra clusters is fixed, as currently it results in an excessive number of alerts being sent. 

**Related issue(s)**
See also #6036

### DIFF
--- a/resources/monitoring/charts/alert-rules/templates/kyma-rules.yaml
+++ b/resources/monitoring/charts/alert-rules/templates/kyma-rules.yaml
@@ -11,16 +11,16 @@ metadata:
     app: kyma.rules
 spec:
   groups:
-  #- name: pod-not-running-rule
-  #  rules:
-    #- alert: SystemPodNotRunning
-    #  expr: sum(kube_pod_container_status_running { namespace=~"kyma-.*|kube-.*|istio-.*|natss", pod!~"(test.*)|((dummy|sample)-.*)|(.*(docs|backup|test)-.*)|((oct-tp-testsuite-all)-.*)|(.*-(tests|dummy))" } == 0 )by (pod,namespace) * on(pod, namespace) (kube_pod_status_phase{phase="Succeeded"} != 1)
-    #  for: 60s
-    #  labels:
-    #    severity: critical
-    #  annotations:
-    #    description: "{{`{{ $labels.namespace }}`}}/{{`{{ $labels.pod }}`}} is not running"
-    #    summary: "{{`{{ $labels.pod }}`}} is not running"
+  - name: pod-not-running-rule
+    rules:
+    - alert: SystemPodNotRunning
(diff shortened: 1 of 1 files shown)


resources/monitoring/charts/alert-rules/templates/kyma-rules.yaml
```diff
@@ -11,16 +11,16 @@ metadata:
...
-    rules:
-    - alert: SystemPodNotRunning
-      expr: sum(kube_pod_container_status_running { namespace=~"kyma-.*|kube-.*|istio-.*|natss", pod!~"(test.*)|((dummy|sample)-.*)|(.*(docs|backup|test)-.*)|((oct-tp-testsuite-all)-.*)|(.*-(tests|dummy))" } == 0 )by (pod,namespace) * on(pod, namespace) (kube_pod_status_phase{phase="Succeeded"} != 1)
-      for: 60s
-      labels:
-        severity: critical
-      annotations:
-        description: "{{`{{ $labels.namespace }}`}}/{{`{{ $labels.pod }}`}} is not running"
-        summary: "{{`{{ $labels.pod }}`}} is not running"
+  #- name: pod-not-running-rule
```
a-thaler: Checking https://github.com/kyma-project/kyma/blob/master/resources/monitoring/charts/alertmanager/templates/alertmanager.config.yaml you can just change the severity label and no notification is happening anymore.

a-thaler: With that alert is still active and you can see on the cluster if it is firing or not, but it is not notifying anymore.

majakurcius: Thanks for the suggestion, sounds good! I've applied the suggested change.

resources/monitoring/charts/alert-rules/templates/kyma-rules.yaml
```diff
@@ -17,7 +17,7 @@ spec:
       expr: sum(kube_pod_container_status_running { namespace=~"kyma-.*|kube-.*|istio-.*|natss", pod!~"(test.*)|((dummy|sample)-.*)|(.*(docs|backup|test)-.*)|((oct-tp-testsuite-all)-.*)|(.*-(tests|dummy))" } == 0 )by (pod,namespace) * on(pod, namespace) (kube_pod_status_phase{phase="Succeeded"} != 1)
       for: 60s
       labels:
-        severity: critical
+      #  severity: critical
```
a-thaler: Seems to be invalid, see build logs: spec.groups.rules.labels in body must be of type object: "null"

Either change the value to something like "warning" or comment out labels section fully

majakurcius: Thanks for the tip!

Issue #825, opened 2018-09-20
Remove logic that detect if SCI is enabled in UI API Acceptance Tests


michal-hudy:
**Description**

Currently UI API Acceptance Tests skip tests execution when SCI is enabled. We provided such logic, because stability checker works on cluster with SCI enabled. When dynamic password generation will be implemented (#197) and stability checker will work on cluster without SCI then such logic will be useless and we should remove it.

To remove:
 - Injecting DEX configuration to tests: https://github.com/kyma-project/kyma/blob/master/resources/core/templates/tests/test-ui-api-acceptance.yaml
 - Logic responsible for reading DEX configuration and detecting if SCI is enabled: https://github.com/kyma-project/kyma/blob/master/tests/console-backend-service/internal/dex/dex.go

michal-hudy:
/remind @derberg to check this issue in 4 weeks

michal-hudy:
It should be also removed from UI Acceptance tests

pkosiec:
Still valid.

a-thaler:
I finally removed all findings, was more work than expected :)

PR #6255, opened 2019-11-08
removed non-static-user check and bumped test image versions

a-thaler:
**Description**

Follow up of https://github.com/kyma-project/kyma/pull/6254 and https://github.com/kyma-project/console/pull/1380

- removed logic from testing.sh to skip non-static-user tests
- removed label from static-users-tests
- removed injection of dex config for console-backend-service test
- removed injection of dex config for console-test

**Related issue(s)**
#6254 #825 https://github.com/kyma-project/console/pull/1380

### DIFF
--- a/installation/scripts/testing.sh
+++ b/installation/scripts/testing.sh
@@ -91,20 +91,6 @@ if [[ -n "${TEST_NAME}" && -n "${TEST_NAMESPACE}" ]]; then
       - name: ${TEST_NAME}
         namespace: ${TEST_NAMESPACE}
 "
-else
-  ${kc} get cm dex-config -n kyma-system -ojsonpath="{.data}" | grep --silent "#__STATIC_PASSWORDS__"
-  if [[ $? -eq 1 ]]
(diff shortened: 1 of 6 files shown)


mszostok:
One thing, I can see that u removed the 
- `resources/core/charts/apiserver-proxy/templates/tests/test.yaml`
- `resources/core/charts/cluster-users/templates/tests/test.yaml`
but you didnt change the images. That means that those tests already supports that?

a-thaler:
@piotrmsc about your concern that the nightly test runs will not work anymore.
I checked the nightly and the tests having the labels attached like core-apiserver-proxy  gets executed already (and succeeds) as the label condition is not met. So all will be good.

a-thaler:
@mszostok will check the mentioned scripts in test-infra repo and create a new PR.
about the two test cases you mentioned. I removed the labels, that's all. So they will be executed always. Before the testing.sh was checking the labels and skipped execution in specific scenario which we do not have.

PR #6175, opened 2019-10-30
Add Knative event sources component

antoineco:
**Description**

Changes proposed in this pull request:

- Add a controller for the reconciliation of custom Knative event sources (`HTTPSource` only for the time being)

**Related issue(s)**

#6033 (Preparation work)
#6150

PR #6272, opened 2019-11-13
Integrate HTTP adapter with Reconciler

antoineco:
**Description**

Integrate HTTP adapter with Reconciler. Round 2 of #6236

* Wait for Channel readiness before synchronizing adapter Service
* Add Ports and EnvVars to adapter Service
* Pass metrics and logging configs to adapter Service
* Configure ReadinessProbe of adapter Service
* Inject adapter image via envconfig instead of os.Getenv()
* Refactor some of the reconciliation flow

**Related issue(s)**

#6236
#6175


components/event-sources/reconciler/httpsource/adapter_config.go
```diff
@@ -0,0 +1,78 @@
...
+	"knative.dev/pkg/metrics"
+)
+
+const component = "httpsource"
+
+// httpAdapterEnvConfig contains properties used to configure the HTTP receive
+// adapter. These are automatically populated by envconfig.
+type httpAdapterEnvConfig struct {
+	// Container image
+	Image string `required:"true"`
```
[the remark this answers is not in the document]
antoineco: Done

components/event-sources/config/500-sources-controller.yaml
```diff
@@ -48,4 +48,4 @@ spec:
           - name: METRICS_DOMAIN
             value: kyma-project.io/event-sources
           - name: HTTP_ADAPTER_IMAGE
-            value: github.com/kyma-project/kyma/components/event-sources/cmd/httpsource-adapter
+            value: github.com/kyma-project/kyma/components/event-sources/cmd/http-adapter
```
k15r: to make it clearer that these parts will be replaces by ko we should use strict references
means: `ko://github.com/kyma-project/kyma/components/event-sources/cmd/http-adapter`.

antoineco: Didn't know that trick, thanks! (and done)

components/event-sources/reconciler/objects/service.go
```diff
@@ -54,66 +54,84 @@ type ServiceOption func(*servingv1alpha1.Service)
...
-		}
-		(*containers)[0].Image = img
+// WithContainerPort sets the container port of a Service.
+func WithContainerPort(port int32) ServiceOption {
+	return func(s *servingv1alpha1.Service) {
+		ports := &firstServiceContainer(s).Ports
+
+		*ports = append(*ports, corev1.ContainerPort{
+			// empty name defaults to http/1.1 protocol
+			ContainerPort: port,
```
k15r: please add a name for the port. Unnamed port cause weird issues with istio

antoineco: It's illegal in a Kn Service, unless you name it exactly "h2c" to force http2.

k15r: oki doki

antoineco: No worries, Kn _does_ set a name on the Pod itself. This constructor only sets the port on the Kn Service, which gets converted to a deployment with all the missing stuff injected.

PR #6294, opened 2019-11-18
Bump Oathkeeper image to support alternative token location

kubadz:
**Description**

- update oathkeeper config in ory chart values due to the recent changes in access rules configuration (https://github.com/ory/oathkeeper/pull/258)
- change oathkeeper image to the recent Oathkeeper version

**Related issue(s)**

See also https://github.com/ory/oathkeeper/issues/257

### DIFF
--- a/resources/ory/values.yaml
+++ b/resources/ory/values.yaml
@@ -85,34 +85,39 @@ oathkeeper:
           enabled: true
         anonymous:
           enabled: true
-          subject: anonymous
+          config:
+            subject: anonymous
         cookie_session:
           enabled: false
-          # REQUIRED IF ENABLED - The session store to forward request method/path/headers to for validation
-          check_session_url: https://session-store-host
-          # Optionally set a list of cookie names to look for in incoming requests.
-          # If unset, all requests are forwarded.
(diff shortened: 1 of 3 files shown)


resources/ory/values.yaml
```diff
@@ -123,21 +128,32 @@ oathkeeper:
...
+            # jwks_url: https://fetch-keys/from/this/location.json
+            # jwks_url: file:///from/this/absolute/location.json
+            # jwks_url: file://../from/this/relative/location.json
+            # Sets the time-to-live of the ID token. Defaults to one minute. Valid time units are: s (second), m (minute), h (hour).
+            ttl: 60s
         header:
           enabled: true
+          config:
+            headers:
+              bar: foo
```
piotrmsc: Please use default headers more meaningful like `x-server: oathkeeper`

resources/ory/values.yaml
```diff
@@ -123,21 +128,32 @@ oathkeeper:
...
         cookie:
           enabled: true
+          config:
+            cookies:
+              foo: bar
```
piotrmsc: Same here, `foo:bar` does not seem good for production ;-)

Issue #6299, opened 2019-11-18
Publish-service pod OOMs out


sayanh:
**Description**

It is observed that memory consumption for publish-service pod increases overtime with traffic of 1 event per second.

Following are the concerns:
- 32Mi is too low a memory limit for publish service.
- Although memory limit is increased to 64Mi the publish-service OOMs out as the memory usage gets built with time. Probably a memory leak, need to figure.

phamilton:
We had to increase the limits to "128Mi" during recent performance tests we ran to avoid OOM.  It appeared to be stable at that level.

PR #6318, opened 2019-11-20
Bump event-bus image

sayanh:
**Description**

- Bump event-bus image
- Bump memory limits for event-bus-publish

**Related issue(s)**

See #6299

### DIFF
--- a/resources/event-bus/values.yaml
+++ b/resources/event-bus/values.yaml
@@ -6,8 +6,8 @@ global:
   containerRegistry:
     path: eu.gcr.io/kyma-project
   event_bus:
-    dir: develop/
-    version: e026d9e0
+    dir: pr/
+    version: PR-6311
     publisherImage: event-publish-service
     subscriptionControllerImage: subscription-controller
   event_bus_tests:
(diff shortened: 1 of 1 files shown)


sayanh:
After 5+ hours of publishing events, here's what we see: 

Screenshot 2019-11-20 at 18 48 50
Screenshot 2019-11-20 at 18 48 41

resources/event-bus/values.yaml
```diff
@@ -34,7 +34,7 @@ global:
         cpu: 20m
         memory: 32Mi
       limits:
-        memory: 32Mi
+        memory: 128Mi
```
antoineco: Still needed?

sayanh:
@nachtmaar @marcobebway @antoineco With decent traffic of events(refer to the screenshot above), the current usage of publish-service is `23.8Mi` and the memory usage is showing a slight upward tendency. Keep in mind that, this is 5+ hours of data. Hence, I feel, `32Mi` is way too less a limit for this container considering the current usage.

This is definitely an improvement from before where the pod was OOM-ing in 10 minutes.

Now, the primary rationale behind 128Mi is preparing for the worst. Of course, it could have been 64Mi or less. As this will be a pre-mature optimization, a larger number won't hurt IMO.

sayanh:
@nachtmaar The other limits are for subscriptioncontroller as rest are unused. Subscriptioncontroller is unimpacted with this exercise.

Issue #6106, opened 2019-10-24
Compass Runtime Agent tests should authenticate via Dex


Szymongib:
**Description**

Currently, Compass Runtime Agent Tests uses OAuth to authenticate to Director.
The better way to handle authentication can be done via Dex, as this approach simulates the user interaction instead of Application/Runtime/IntegrationSystem.

Acceptance:
- Compass Runtime Agent Tests authenticates to Director via Dex tokens

PR #6333, opened 2019-11-21
Fetch Dex token after waiting for API Server in Runtime Agent tests

franpog859:
**Description**

- Fetch Dex token after waiting for API Server in Runtime Agent tests. Not doing so could cause bugs with not ready sidecars and flaky tests failures

**Related issue(s)**

See the issue #6106

### DIFF
--- a/tests/compass-runtime-agent/test/runtimeagent/suite.go
+++ b/tests/compass-runtime-agent/test/runtimeagent/suite.go
@@ -58,7 +58,8 @@ type TestSuite struct {
 
 	mockServiceServer *mock.AppMockServer
 
-	Config testkit.TestConfig
+	Config        testkit.TestConfig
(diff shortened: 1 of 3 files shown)


tests/compass-runtime-agent/test/runtimeagent/suite.go
```diff
@@ -58,7 +58,8 @@ type TestSuite struct {
 
 	mockServiceServer *mock.AppMockServer
 
-	Config testkit.TestConfig
+	Config        testkit.TestConfig
+	CoreClientset *kubernetes.Clientset
```
Szymongib: I think this line is not needed :)

franpog859: Yeah, you're right! Deleted 😎

PR #6336, opened 2019-11-21
Bump Provisioner images

Maladie:
**Description**

- Test images for Provisioner added

**Related issue(s)**


resources/compass/charts/provisioner/values.yaml
```diff
@@ -10,7 +10,7 @@ database:
   schemaFilePath: "/assets/database/provisioner.sql"
 
 tests:
-  enabled: false
+  enabled: true
```
Szymongib:
```suggestion
  enabled: false
```
It is set to true by the overrides in the pipeline.

Issue #6282, opened 2019-11-15
Readiness probe fails on new Knative Service


Abd4llA:
**Description**

When creating a new Knative Service in Kyma, the service revision never becomes ready because readiness probe fails.

More precisely, the port 8022 of the `queue-proxy` container (injected by Knative) seems unreachable.

**Expected result**

A revision becomes ready.

**Actual result**

A revision never gets ready because the readiness probe fails.

**Steps to reproduce**

TBD
**Troubleshooting**

TBD

antoineco:
Fixed by #6345. Closing.

PR #6296, opened 2019-11-18
Add Helm Chart for the event-sources controller manager

antoineco:
**Description**

Changes proposed in this pull request:

- Add Helm Chart for the event-sources controller manager

This PR also includes the necessary changes to the Makefile so that Prow can build and push images:

<details>
 <summary>:memo: "make push-image" output</summary>

```console
$ export DOCKER_PUSH_REPOSITORY=my.gcr.io
$ export DOCKER_PUSH_DIRECTORY=/develop
$ export DOCKER_TAG=0fag13lf
$ make push-image
+ Building controller-manager
+ Building container image event-sources-controller-manager
docker image build -f cmd/controller-manager/Dockerfile -t event-sources-controller-manager .
Sending build context to Docker daemon  57.86MB
Step 1/2 : FROM scratch
 --->
Step 2/2 : COPY controller-manager ./
 ---> Using cache
 ---> 21568c53b9cc
Successfully built 21568c53b9cc
Successfully tagged event-sources-controller-manager:latest
+ Pushing container image event-sources-controller-manager
docker image tag event-sources-controller-manager my.gcr.io/develop/event-sources-controller-manager:0fag13lf
docker image push my.gcr.io/develop/event-sources-controller-manager:0fag13lf
+ Building http-adapter
+ Building container image event-sources-http-adapter
docker image build -f cmd/http-adapter/Dockerfile -t event-sources-http-adapter .
Sending build context to Docker daemon  106.4MB
Step 1/2 : FROM scratch
 --->
Step 2/2 : COPY http-adapter ./
 ---> Using cache
 ---> 8cce66d3974f
Successfully built 8cce66d3974f
Successfully tagged event-sources-http-adapter:latest
+ Pushing container image event-sources-http-adapter
docker image tag event-sources-http-adapter my.gcr.io/develop/event-sources-http-adapter:0fag13lf
docker image push my.gcr.io/develop/event-sources-http-adapter:0fag13lf
```

</details>

**Related issue(s)**

#6288
kyma-project/test-infra#1745

PR #6345, opened 2019-11-21
Set Prometheus port for event-sources' HTTP adapter metrics

antoineco:
**Description**

- Set Prometheus port to 9092 for adapter metrics. The default port 9090 is reserved ([Godoc](https://godoc.org/knative.dev/serving/pkg/apis/networking#pkg-constants)).

**Related issue(s)**

#6282


antoineco:
The prow job requires #6296

PR #6350, opened 2019-11-22
Fix service-catalog test dangling resources

polskikiel:
**Description**

- Fix service-catalog test dangling resources on long-lasting clusters

**Related issue(s)**


tests/service-catalog/test/binding_usage_test.go
```diff
@@ -409,7 +454,13 @@ func (ts *TestSuite) createAndWaitBindingUsage(bindingName, sbuName, envPrefix s
 		}
 	}
 
-	_, err := usageCli.Create(sbu)
+	err := wait.Poll(time.Second, time.Minute, func() (bool, error) {
```
ksputo: you can use const here:
```suggestion
	err := wait.Poll(time.Second, timeoutPerStep, func() (bool, error) {
```

PR #6371, opened 2019-11-22
Set application-name label on HTTP source Channel

antoineco:
**Description**

- Set `application-name` label on HTTP source Channel. Required by application broker.
- Add Channel reconciliation logic to ensure the label remains up-to-date

**Related issue(s)**

marcobebway/kyma#2028

### DIFF
--- a/components/event-sources/reconciler/httpsource/reconcile.go
+++ b/components/event-sources/reconciler/httpsource/reconcile.go
@@ -83,6 +83,8 @@ const (
 
 const adapterHealthEndpoint = "/healthz"
 
+const applicationNameLabelKey = "application-name"
+
 // Reconcile compares the actual state of a HTTPSource object referenced by key
 // with its desired state, and attempts to converge the two.
 func (r *Reconciler) Reconcile(ctx context.Context, key string) error {
@@ -220,6 +222,7 @@ func (r *Reconciler) makeKnService(src *sourcesv1alpha1.HTTPSource,
 func (r *Reconciler) makeChannel(src *sourcesv1alpha1.HTTPSource) *messagingv1alpha1.Channel {
 	return objects.NewChannel(src.Namespace, src.Name,
 		objects.WithChannelControllerRef(src.ToOwner()),
+		objects.WithChannelLabel(applicationNameLabelKey, src.Name),
 	)
 }
 
--- a/components/event-sources/reconciler/httpsource/reconcile.go
+++ b/components/event-sources/reconciler/httpsource/reconcile.go
@@ -126,7 +126,21 @@ func httpSourceByKey(key string, l sourceslistersv1alpha1.HTTPSourceLister) (*so
 
 // reconcileSink reconciles the state of the HTTP adapter's sink (Channel).
 func (r *Reconciler) reconcileSink(src *sourcesv1alpha1.HTTPSource) (*messagingv1alpha1.Channel, error) {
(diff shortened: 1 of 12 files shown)


components/event-sources/reconciler/httpsource/reconcile.go
```diff
@@ -124,7 +126,21 @@ func httpSourceByKey(key string, l sourceslistersv1alpha1.HTTPSourceLister) (*so
 
 // reconcileSink reconciles the state of the HTTP adapter's sink (Channel).
 func (r *Reconciler) reconcileSink(src *sourcesv1alpha1.HTTPSource) (*messagingv1alpha1.Channel, error) {
```
antoineco: The logic of this function is now a little bit more than just "create if not exists" because we now also ensure the Channel's labels are synchronized.

components/event-sources/reconciler/httpsource/reconcile_test.go
```diff
@@ -105,9 +113,7 @@ func TestReconcile(t *testing.T) {
 				NewHTTPSource(tNs, tName),
 			},
 			WantCreates: []runtime.Object{
-				NewChannel(tNs, tName,
```
antoineco: I moved all the redundant calls to `NewXYZ()` to helper funcs. It should make the tests slightly shorter and more readable.

antoineco:
The prow job requires #6296

components/event-sources/reconciler/objects/channel.go
```diff
@@ -47,3 +56,32 @@ func WithChannelControllerRef(or *metav1.OwnerReference) ChannelOption {
...
+// ApplyExistingChannelAttributes copies some important attributes from a given
+// source Channel to a destination Channel.
+func ApplyExistingChannelAttributes(src, dst *messagingv1alpha1.Channel) {
+	// resourceVersion must be returned to the API server
+	// unmodified for optimistic concurrency, as per Kubernetes API
+	// conventions
+	dst.ResourceVersion = src.ResourceVersion
+
+	// immutable Knative annotations must be preserved
+	for _, ann := range knativeMessagingAnnotations {
```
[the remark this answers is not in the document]
antoineco: Nope, there is no label to preserve. All of them are either set by our own Channel builder, or injected by web hooks.

components/event-sources/reconciler/httpsource/reconcile.go
```diff
@@ -242,6 +260,25 @@ func (r *Reconciler) syncKnService(src *sourcesv1alpha1.HTTPSource,
...
+func (r *Reconciler) syncChannel(src *sourcesv1alpha1.HTTPSource,
+	currentCh, desiredCh *messagingv1alpha1.Channel) (*messagingv1alpha1.Channel, error) {
+
+	if objects.Semantic.DeepEqual(currentCh, desiredCh) {
+		return currentCh, nil
+	}
+
+	ch, err := r.messagingClient.Channels(currentCh.Namespace).Update(desiredCh)
+	if err != nil {
+		r.eventWarn(src, failedUpdateReason, "Update failed for Channel %q", ch.Name)
```
marcobebway: At this point, `ch` could be nil.

Spotted the same for line **255**, `ksvc` could be nil as well.

antoineco: Nice catch, fixed

PR #6385, opened 2019-11-25
Bump director images

KarolJaksik:
**Description**

- Update director images


components/compass-runtime-agent/internal/kyma/applications/converter_test.go
```diff
@@ -22,8 +22,8 @@ func TestConverter(t *testing.T) {
 			ID:          "App1",
 			Name:        "Appname1",
 			Description: "Description",
-			Labels: map[string][]string{
-				"key": {"value1", "value2"},
+			Labels: map[string]interface{}{
+				"key": []string{"value1", "value2"},
```
Szymongib: I think it would be good to add another label of type `string` to hit the second `case` statement in the switch.

Issue #6288, opened 2019-11-15
Create a helm chart for the new event-source controller


k15r:
**Description**

The new event-source controller needs to be installed using the kyma installer. This means we require a helm chart for it

**Reasons**

No helm chart, no event-source

**Acceptance Criteria**
* install event source
* install test definitions

PR: #6296
Other: kyma-project/test-infra#1745

antoineco:
We also need to think about the release process. Our adapters (well, adapter, singular) need to be built as well, and their image should be references by the controller.

PR #6400, opened 2019-11-26
Set images versions of event-sources component

antoineco:
**Description**

- Set images versions in Helm chart values ([corresponding Prow jobs](https://status.build.kyma-project.io/?type=postsubmit&job=post-kyma-components-event-sources))

**Related issue(s)**

issue #6288
pr #6296

### DIFF
--- a/resources/event-sources/values.yaml
+++ b/resources/event-sources/values.yaml
@@ -9,4 +9,4 @@ global:
   httpAdapter:
     image: event-sources-http-adapter
     dir: ''
-    version: 0b31784e
+    version: dc48c27c
--- a/resources/event-sources/values.yaml
+++ b/resources/event-sources/values.yaml
@@ -4,9 +4,9 @@ global:
   controllerManager:
     name: controller-manager
     image: event-sources-controller-manager
-    dir: develop/
-    version: #TODO
+    dir: ''
+    version: dc48c27c
   httpAdapter:
     image: event-sources-http-adapter
-    dir: develop/
-    version: #TODO
+    dir: ''
+    version: 0b31784e


resources/event-sources/values.yaml
```diff
@@ -4,9 +4,9 @@ global:
   controllerManager:
     name: controller-manager
     image: event-sources-controller-manager
-    dir: develop/
-    version: #TODO
+    dir: ''
```
sayanh: I think we need to be consistent with the other images which have `develop/` in their path.

antoineco: The Prow job sends an empty dir and builds images like that. If I add that dir here, the image name becomes invalid. I think this changed with the shared Makefile.

sayanh: True

resources/event-sources/values.yaml
```diff
@@ -4,9 +4,9 @@ global:
...
+    version: dc48c27c
   httpAdapter:
     image: event-sources-http-adapter
-    dir: develop/
-    version: #TODO
+    dir: ''
+    version: 0b31784e
```
k15r: there is also this version: [dc48c27c](https://console.cloud.google.com/gcr/images/kyma-project/EU/event-sources-http-adapter@sha256:44b3283569284d9b31584ad30da839a118ebcee7b07de7531aac4e20f9c40471/details?tab=info)` for the event-sources-http-adapter
I think - even if no changes to this particular code have been introduced - we should use this as it is the latest version

antoineco:
```suggestion
    version: dc48c27c
```

antoineco: Done

Issue #4719, opened 2019-07-02
Non-deterministic failing test [Azure]


Demonsthere:
**Description**

In some cases the octopus tests fails with an error log like this: 

```
Test summary
Test status: test-api-controller-acceptance - Succeeded
Test status: test-dex-integration - Running
Test status: application-operator-tests - NotYetScheduled
Test status: application-connector-tests - NotYetScheduled
```
This means that the test that is running is blocking the queue -> the pod containing the test is never killed because some process inside did not complete. 
This can happen if the test process is not killed properly, or if the test uses a istio sidecar, and the sidecar process is not killed after the test. 

To handle this we have a kill-switch mechanism for the istio process, which seems to be failing in some cases (race condition?). This seems to be happening more often on Azure than GKE and needs to be investigated. 

**Expected result**
Istio-proxy process is killed after the test is done

**Actual result**
Istio-proxy persists causing a never killed pod and never ending test

**Steps to reproduce**

**Troubleshooting**

Demonsthere:
Interesting find: running `kubectl get pod -n kyma-system -o yaml | grep shareProcessNamespace | wc -l` on GKE produces a positive number, and on AKS it is 0. The flag `shareProcessNamespace` is missing in pod definitions on AKS.

Demonsthere:
This issue requires a non signal based method of killing isito-proxy after the main container is done. This is being developed [here](https://github.com/istio/istio/issues/15041) and [here](https://github.com/istio/istio/pull/15320). We are currently blocked by this.

piotrmsc:
Latest istio release 1.2.3 does not have /quitquitquit handler in pilot-agent yet.

Tomasz-Smelcerz-SAP:
Tests that are not capable of using /quitquitquit endpoint at the moment:

Because there's no `curl` in the image:
- resources/application-connector/charts/connector-service/templates/tests/test.yaml
- resources/assetstore/templates/tests/test.yaml
- resources/core/charts/console-backend-service/templates/tests/test.yaml
- resources/core/charts/kubeless/templates/tests/test.yaml
- resources/dex/templates/tests/test-integration.yaml
- resources/service-catalog/templates/tests/test.yaml

Because of `manager` container (will not work in Azure):
- resources/core/charts/console/templates/test/test.yaml
- resources/core/templates/tests/test-external-solution.yaml
- resources/event-bus/templates/tests/test.yaml
[...]

Issue #6076, opened 2019-10-21
Use multi-stage builds and from scratch in dockerfiles in all service-management components


PK85:
**Description**

Use multi-stage builds and from scratch in `Dockerfile`'s in all service-management components.
All `debugging` tools should be available on-demand as a sidecars, `Debug sidecar`.

**Reasons**

Eliminate binary vulnerabilities.

**Components**
- ApplicationBroker https://github.com/kyma-project/kyma/pull/6316
- HelmBroker https://github.com/kyma-project/kyma/pull/6316
- ServiceBindingUsageController https://github.com/kyma-project/kyma/pull/6316
- PodPresetController 
- test/service-catalog https://github.com/kyma-project/kyma/pull/6316
- octopus https://github.com/kyma-incubator/octopus/pull/38
- service-catalog-tester https://github.com/kyma-incubator/service-catalog-tester/pull/13

mszostok:
Here is example how to do it:https://github.tools.sap/gophers-team/kyma-environment-service-broker/blob/ac4d02ab378b2b5daa7ddea707052cd0d9e78082/Dockerfile

Tomasz-Smelcerz-SAP:
Please ensure that service catalog tests image contains curl as well as some shell (e.g. busybox).
This is necessary to invoke curl *after* test finishes, regardless of test outcome: success, failure, crash. See #4719 for details.

polskikiel:
**Update**
I've changed the base image to scratch in our components, but 2 images are still built from the `alpine:3.10`:
- `service-catalog-test`- requires a shell and curl to terminate istio sidecar after the test. We don't bother about binary vulnerabilities in the tests.
- `helm-controller` - requires a shell (executes `eval(ssh_agent -s)`), curl was exchanged to the health-check-proxy solution, so the binary vulnerability is removed.

PR #6404, opened 2019-11-26
Add curl to service catalog tests Dockerfile

colunira:
**Description**
We can resolve the problem described in #4719 with newly added Istio endpoint `/quitquitquit`. We need to add curl to our Dockerfile to be able to call it.

- Install curl in Dockerfile

**Related issue(s)**
#4719

### DIFF
--- a/tests/service-catalog/Dockerfile
+++ b/tests/service-catalog/Dockerfile
@@ -12,6 +12,8 @@ RUN CGO_ENABLED=0 go test -tags=acceptance -c ./test -o /servicecatalog.test
 
 FROM alpine:3.10
 
+RUN apk add --no-cache curl
+
 COPY --from=builder env-tester.bin /go/bin/env-tester.bin
 COPY --from=builder ./servicecatalog.test .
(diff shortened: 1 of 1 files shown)


polskikiel:
Hello, that solution won't work after we change our base image to scratch in this story: https://github.com/kyma-project/kyma/issues/6076

Tomasz-Smelcerz-SAP:
> Hello, that solution won't work after we change our base image to scratch in this story: #6076

@polskikiel What is your plan, then? Currently, this test uses "pkill" approach, which also won't work after the change the image to "scratch".
"pkill" is not acceptable, though, because it doesn't work in AKS and probably in other managed K8s environments.
"curl" approach works everywhere. Actually, it doesn't have to be `curl`, any CLI tool capable of doing HTTP POST will do.

polskikiel:
We'll continue this PR in this story: #6076

Tomasz-Smelcerz-SAP:
I am closing this one because of #6076

PR #6432, opened 2019-11-27
Add curl to event-bus-tests Docker image

Tomasz-Smelcerz-SAP:
**Description**
Add curl binary to event-bus-tests Docker image.
This is necessary in order to run the tests in Azure environment - See #4719 
In addition, the shell is also required, so that we can call curl after the test process is finished.
Example invocation that ensures proper termination:
```
sleep 20; ./test.sh; exit_code=$?; curl -XPOST http://127.0.0.1:15020/quitquitquit; sleep 4; exit $exit_code;
```

- Add curl binary to event-bus-tests Docker image

**Related issue(s)**
See also #4719

### DIFF
--- a/tests/event-bus/e2e-tester/Dockerfile
+++ b/tests/event-bus/e2e-tester/Dockerfile
@@ -9,7 +9,7 @@ COPY e2e-tester.go .
 RUN ls ./
 RUN CGO_ENABLED=0 GOOS=linux go build -v -a -installsuffix cgo -o e2e-tester .
 
-FROM scratch
+FROM curlimages/curl:7.66.0
 LABEL source=git@github.com:kyma-project/kyma.git
 
 ARG version


tests/event-bus/e2e-tester/Dockerfile
```diff
@@ -9,7 +9,7 @@ COPY e2e-tester.go .
 RUN ls ./
 RUN CGO_ENABLED=0 GOOS=linux go build -v -a -installsuffix cgo -o e2e-tester .
 
-FROM scratch
+FROM curlimages/curl:7.66.0
```
Tomasz-Smelcerz-SAP: This image contains `curl` and shell, which fits our requirement.

antoineco: Was there any reason not to depend on `alpine` directly and adding a simple `apk add --no-cache curl`?

Tomasz-Smelcerz-SAP: @antoineco This is an official curl image and it is based on alpine anyway. Using this image has the advantage of pinning to exact curl version. In alpine, you get the version that is supported by maintainers.

Issue #6142, opened 2019-10-28
go-jsonhash: Old library with no LICENSE file


jose-cortina:
**Description**

This library is included in the helm-broker and cli, and was last updated in June 2016 (https://github.com/komkom/go-jsonhash)

The library is reported by [FOSSA](https://app.fossa.com/projects/git%2Bgithub.com%2Fkyma-project%2Fkyma/refs/branch/master/78f6ad19112c28e3cd7489341398c14d81f7acb3/issues/licensing?filter=unlicensed_dependency&revisionScanId=1351913) as it has no LICENSE file.

Evaluate the use of the library and find a suitable replacement to remove this dependency.

https://github.com/search?q=org%3Akyma-project+go-jsonhash&type=Code

**Expected result**

There are no libraries without a license.

**Actual result**

Three libraries reported with no license.

jasiu001:
The `https://github.com/komkom/go-jsonhash` library is used in one place in `internal/broker/provision.go` file (plus test files), one time, one method. 
Method hashes request parameters using standard go libraries, the library is not complicated, it could be replaced by our implementation. After that, the library can be removed.

PK85:
@polskikiel as we discussed. In case customer has upgraded a cluster, make sure service instances created before the upgrade, works as before.

Take into account that can resend SI creation with the same id.

PR #6453, opened 2019-11-28
Remove outdated library in Application Broker

ksputo:
**Description**
The library was used to generate a hash from Provisioning Parameters provided in Provision Request. We used it to identify conflicts described in OSB API specification - trying to provision Service Instance with ID that already exists, but with different parameters should result in HTTP 409 Conflict status code in response. However, Application Broker doesn't use Provisioning Parameters at all and to avoid having dead code we are not going to implement it at the moment. Instead we can just remove the logic for Provisioning Parameters completely.

**Possible follow-up:**
The idea is to add Provisioning Parameters to existing Instance and Instance Operation models, which is better than hashing it in case of possible future enhancements like adding UPDATE method to Application Broker etc. A proper validation logic should be implemented too, so the system knows how to interpret provided parameters. Additionally, if Parameters provided in request could contain some sensitive data, the implementation should be extended with some kind of encryption OR the Provisioning Parameters should just contain reference to corresponding k8s secret.

- Removed outdated library from Application Broker
- Added rejecting requests with Provisioning Parameters

**Related issue(s)**

#6142


components/application-broker/internal/broker/provision.go
```diff
@@ -74,14 +75,19 @@ func (svc *ProvisionService) Provision(ctx context.Context, osbCtx osbContext, r
...
+	}
 
 	switch state, err := svc.instanceStateGetter.IsProvisioned(iID); true {
 	case err != nil:
 		return nil, &osb.HTTPStatusCodeError{StatusCode: http.StatusBadRequest, ErrorMessage: strPtr(fmt.Sprintf("while checking if instance is already provisioned: %v", err))}
 	case state:
-		if err := svc.compareProvisioningParameters(iID, paramHash); err != nil {
-			return nil, &osb.HTTPStatusCodeError{StatusCode: http.StatusConflict, ErrorMessage: strPtr(fmt.Sprintf("while comparing provisioning parameters %v: %v", req.Parameters, err))}
+		if err := svc.compareProvisioningParameters(iID, provisioningParams); err != nil {
+			if strings.Contains(err.Error(), "conflict") { // TODO: implement custom errors for database access problems
```
piotrmiskiewicz: Do not analyze the error message here! You can introduce a dedicated error or just return (bool, error) instead of (error). Such value could indicate if the parameters are equal or not (and consider change of method name).

components/application-broker/internal/storage/populator/instances.go
```diff
@@ -69,7 +69,11 @@ func (p *Instances) Do(ctx context.Context) error {
 	for _, si := range serviceInstances {
 		if si.Spec.ServiceClassRef != nil {
 			if _, ex := abClassNames[si.Spec.ServiceClassRef.Name]; ex {
-				if err := p.inserter.Insert(p.converter.MapServiceInstance(si)); err != nil {
+				convertedInstance, err := p.converter.MapServiceInstance(si)
+				if err != nil {
+					return errors.Wrap(err, "while converting SC service instance to internal service instance model")
```
piotrmiskiewicz: The message is misleading. It should be just something like: "while converting service instance to internal service". There is nothing about service class

components/application-broker/internal/broker/provision.go
```diff
@@ -74,25 +74,45 @@ func (svc *ProvisionService) Provision(ctx context.Context, osbCtx osbContext, r
...
+	provisioningParams := &internal.ProvisioningParameters{
+		Parameters: req.Parameters,
+	}
 
 	switch state, err := svc.instanceStateGetter.IsProvisioned(iID); true {
 	case err != nil:
 		return nil, &osb.HTTPStatusCodeError{StatusCode: http.StatusBadRequest, ErrorMessage: strPtr(fmt.Sprintf("while checking if instance is already provisioned: %v", err))}
 	case state:
-		if err := svc.compareProvisioningParameters(iID, paramHash); err != nil {
+		conflictOccured, err := svc.hasInstanceDifferentProvisiongParameters(iID, provisioningParams)
```
mszostok: We should not check if it is `not found` error, cause its problem of our inconsistent state that we know that sth is provisioned, but we do not have that instance in our storage. 
In case of the OSB API: 
> 400 Bad Request	MUST be returned if the request is malformed or missing mandatory data

so we cannot use the 400 error. We should simply return in such case the Internal Error

```go
conflictOccured, err := svc.hasInstanceDifferentProvisiongParameters(iID, provisioningParams)
if err != nil { // error handling ALWAYS FIRST
  return nil, &osb.HTTPStatusCodeError{StatusCode: http.StatusInternalServerError, ErrorMessage: strPtr(fmt.Sprintf("while comparing provisioning parameters %v: %v", req.Parameters, err))}
}
if conflictOccured {
  return nil, &osb.HTTPStatusCodeError{StatusCode: http.StatusConflict, ErrorMessage: strPtr(fmt.Sprintf("while comparing provisioning parameters %v: %v", req.Parameters, err))}
}
```

components/application-broker/internal/broker/provision.go
```diff
@@ -271,21 +291,24 @@ func (svc *ProvisionService) ensureEaUpdate(appID, ns string, si *v1beta1.Servic
 	return nil
 }
 
-func (svc *ProvisionService) compareProvisioningParameters(iID internal.InstanceID, newHash string) error {
+func (svc *ProvisionService) hasInstanceDifferentProvisiongParameters(iID internal.InstanceID, paramInput *internal.ProvisioningParameters) (bool, error) {
```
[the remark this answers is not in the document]
ksputo: I agree, we should avoid dead code - modified the pull request description and removed Provisioning Parameters completely.

components/application-broker/internal/broker/provision.go
ksputo: I agree, to avoid having dead code I just removed the whole logic and added proper information and possible follow-up in PR description

PR #6466, opened 2019-11-29
Create health-check tool

polskikiel:
**Description**

- Created a health-check tool

**Related issue(s)**


a-thaler:
Please consider moving it into the components folder. As it is not really producing an active component but rather a base image it is not strictly falling into that category. However, I prefer to have all in the component directory which is part of the actual helm charts and with that part of the actual runtime of kyma, so also that base image. I'm in favor of removing the fuzzy "tools" folder and trying to reduce the amount of entries.

tools/health-check-proxy/README.md
```diff
@@ -0,0 +1,69 @@
+# Health check proxy
+
+## Overview
+
+The Health check proxy is designed to check liveness or readiness endpoints in the applications with mTLS enabled which are built from the scratch image.
+
+The Kubernetes API server cannot reach Pods with mTLS enabled via httpGet, because of that we cannot make a direct call to the liveness or readiness probe for status.
+
+In this case, the best way to check the status of our application is to
```
a-thaler: Please complete the sentence

a-thaler:
Hust a thought: somehow it is not ideal that it is a base image. Think of tomorrow, when somehow wants to add another utility, he either adds it here (and the name will not fit anymore) or build another image based on this image? Will it not be better to just have a library that you can include in your build process. Then you can pick what you want instead of building layers dependent on each other.

polskikiel:
According to the @a-thaler comment, I moved the codebase into the `components` folder. I will also provide documentation on how to build health-check binary manually, without the need to use this app as the base image.

components/health-check-proxy/cmd/health-check/main.go
```diff
@@ -0,0 +1,43 @@
...
+)
+
+func main() {
+	var (
+		path       string
+		host       string
+		statusPort int
+		retries    int
+	)
+	flag.StringVar(&path, "path", "", "liveness or readiness endpoint")
```
mszostok: just accept only one flag e.g. called URL. It's simple and generic. When you split that with different flag like host, port, path and then building your final url is more complicated and end-users do not have a full control e.g. `http` protocol is right now hardcoded

mszostok:
@a-thaler We will even make you happier 😀 **we will not add this component at all 🚀** 

**"No code is the best way to write secure and reliable applications. Write nothing; deploy nowhere."**

We discovered such an option:
https://istio.io/docs/ops/configuration/mesh/app-health-check/#use-annotations-on-pod

Issue #6472, opened 2019-11-29
Add rafter domain to console-backend-service


magicmatatjahu:
**Description**
- Add `rafter` domain to console-backend-service
- Write unit and integration test for new domain
- Remove old `asset-store` and `cms` domains

PR #6473, opened 2019-11-30
Remove asset-store and cms domains in console-backend-service

magicmatatjahu:
**Description**

- Remove `asset-store` and `cms` domain in console-backend-service and adjust in `rafter` domain resource names

**Related issue(s)**
https://github.com/kyma-project/kyma/issues/6472


components/console-backend-service/README.md
```diff
@@ -23,7 +23,7 @@ Use the following tools to set up the project:
 To run the application without building the binary, run this command:
 
 ```bash
-APP_KUBECONFIG_PATH=/Users/$USER/.kube/config APP_VERBOSE=true APP_ASSET_STORE_ADDRESS=https://minio.{kymaDomain} APP_ASSET_STORE_VERIFY_SSL=false APP_APPLICATION_GATEWAY_INTEGRATION_NAMESPACE=kyma-integration APP_APPLICATION_CONNECTOR_URL=http://dummy.url APP_OIDC_ISSUER_URL=https://dex.{kymaDomain} APP_OIDC_CLIENT_ID=kyma-client go run main.go
+APP_KUBECONFIG_PATH=/Users/$USER/.kube/config APP_VERBOSE=true APP_RAFTER_ADDRESS=https://storage.{kymaDomain} APP_RAFTER_VERIFY_SSL=false APP_APPLICATION_GATEWAY_INTEGRATION_NAMESPACE=kyma-integration APP_APPLICATION_CONNECTOR_URL=http://dummy.url APP_OIDC_ISSUER_URL=https://dex.{kymaDomain} APP_OIDC_CLIENT_ID=kyma-client go run main.go
```
michal-hudy: Hm... CBS is using internal cluster communication, so I think we should use the internal address, especially we want to secure the storage endpoint with OAuth.

The internal endpoint is without SSL :)

michal-hudy: My bad, it is for testing outside of the cluster

magicmatatjahu: Great :)

components/console-backend-service/internal/domain/rafter/assetgroup_resolver.go
```diff
@@ -28,7 +28,7 @@ func newAssetGroupResolver(assetGroupService assetGroupSvc, assetGroupConverter
 	}
 }
 
-func (r *assetGroupResolver) AssetGroupAssetsField(ctx context.Context, obj *gqlschema.AssetGroup, types []string) ([]gqlschema.RafterAsset, error) {
+func (r *assetGroupResolver) AssetGroupAssetsField(ctx context.Context, obj *gqlschema.AssetGroup, types []string) ([]gqlschema.Asset, error) {
```
sjanota: So, this resolver operates on `Asset`s. Why the domain is called `rafter` then? Wouldn't `assets` be more clear and meaningful?

components/console-backend-service/internal/gqlschema/schema.graphql
```diff
@@ -26,7 +26,7 @@ input ResourceAttributes {
     isChildResolver: Boolean! = false
 }
 
-# Rafter & Asset Store & CMS
+# Rafter
```
sjanota: There is no mention about rafter in the schema. this comment is just exposing the implementation and should be removed.

components/console-backend-service/internal/domain/rafter/assetgroup_resolver.go
magicmatatjahu: It's called `rafter` because in this domain we operate on `rafter` resources. For me it makes no difference.

components/console-backend-service/internal/gqlschema/schema.graphql
magicmatatjahu: This comment point that below types reference to Rafter resources, but as I wrote in another comment, for me it makes no difference.

components/console-backend-service/internal/domain/rafter/assetgroup_resolver.go
magicmatatjahu: Also as I know, we called always domains as apiGroups, not as mention for operated resources.

components/console-backend-service/internal/gqlschema/schema.graphql
sjanota: I agree that domain can/should be named after k8s `apiGroup` as you mentioned above. However, I don't think this information should leak to the API. It's not a blocker, though.

PR #6474, opened 2019-11-30
Add integration tests for rafter domain in console-backend-service

magicmatatjahu:
**Description**

- Add integration tests for `rafter` domain in console-backend-service
- Remove tests for `cms` domain
- Update tests for `service-catalog` domain

**Related issue(s)**
https://github.com/kyma-project/kyma/issues/6472


michal-hudy:
For me, it looks OK, but I would like to know what was the reason for disabling these tests because I think we can improve them in this PR.

@kwiatekus @akucharska ?

Issue #6330, opened 2019-11-21
Kyma running on Azure provisioned via Gardener.


piotrmsc:
**Description**

Currently, we have problems on clusters provisioned via Gardener (Azure for example) with istio.
What we have observed is that istio is installed as well kyma however at least 2 things are failing : 
a) istio rbac
b) after you deploy new app (with sidecar) mTLS is not working

We need to debug what's happening on that clusters, it's 100% reproducible. Gardener adds PSPs.

**Reasons**

We need to have kyma offering up and running on clusters provisioned via Gardener.

**Attachments**

piotrmsc:
Useful links for debugging https://istio.io/docs/ops/common-problems/security-issues/

piotrmsc:
The problems was with ip addresation on Gardener, we had 10.0.0.1/8 but services on Gardener have 100.x.x.x so envoy was not intercepting requests.

PR #6499, opened 2019-12-03
Change IPRanges in istio to *

colunira:
**Description**

Due to Gardener addressing we need to change the includeIPRanges field to allow pods to communicate with each other.
Also, during testing the changes it turned out that e2e-kubeless tests require more resources than it's specified as default in the kyma-system namespace.

- Change includeIPRanges value to *
- Add resources limits and requests to e2e-kubeless-tests TestDefinition

**Related issue(s)**
#6330


a-thaler:
I tested the setting on gardener and can confirm that it solved our problems, e2e scenario was working well. Thanks for the investigation

Issue #6239, opened 2019-11-07
Integrate Rafter and remove Asset Store and Headless CMS


derberg:
**Description**

- `Change installation`: 
  - remove old `assetstore` and `cms` charts and replace with rafter chart
  - figure out where to place CRDs
- `console-backend-service`
  - add new or modify existing query that instead of docstopics reads assetgroups
- `application-connector`
  - contribute to application connector so it doesn't create docstopic but assetgroups
  - probably integration with upload service will also require update
  - compas runtime agent contribution
- `helm-broker`
  - contribute to the broker to make sure it creates assetgroups and not docstopics. 
  - talk to owners to make a release once change is in the master
- `console`
  - service catalog views and docs ui should fetch documentation and specifications from a new query
- `website`
  - searching for new `rafter` apiGroup in `content-loader-tool` and change to assetgroups
- modify existing or create missing backup tests and upgrade test 
- create very simple integration test that just creates an assetgroup, to make sure integration of rafter works fine
- have migration jobs of existing resources to the new ones
- Release notes with clear description of what we are doing
- `MinIO Gateway` - we use asset-store-overrides to switch to MinIO Gateway mode (4 tutorials) so we must adjust Secret and ConfigMap to work with rafter-overrides.
- `Custom metrics & dashboards` - check and adjust Asset Store and Headless CMS custom metrics & Garafana dashboards.

Stretch:
- instead of using default bucket for AssetGroups consider introduction of mutation webhook that basing on a label we agree on, will mutate created AssetGroup by adding a custom bucket we created for Kyma use case - topic for discussion

**Reasons**

Asset Store and Headless CMS got integrated into single standalone project called Rafter. Now Kyma need to migrate to use Rafter

**Human resources**
- [x] `installation` - @magicmatatjahu
- [x] `console-backend-service` - @magicmatatjahu
- [x] `application-registry` - @m00g3n
- [x] `compass-runtime-agent` - @m00g3n
- [x] `helm-broker` - @pPrecel
- [x] `console` - @magicmatatjahu 
- [x] `website` - @magicmatatjahu 
- [x] `integration-test` - @aerfio 
- [x] `backup-test` - @michal-hudy 
- [x] `upgrade-test` - @magicmatatjahu 
- [x] `migration-job` - @magicmatatjahu 
- [x] `MinIO Gateway (overrides)` - ___
- [x] `Custom metrics & dashboards` - @pPrecel
- [x] `docs` - @mmitoraj & @kazydek
[...]

PR #6517, opened 2019-12-04
Add to CODEOWNERS mention about new rafter chart in resources folder

magicmatatjahu:
**Description**

- Add to CODEOWNERS mention about new rafter chart in `resources` folder

**Related issue(s)**
https://github.com/kyma-project/kyma/issues/6239


derberg:
There is no voting needed for this one and our open governance rules do not apply as rafter is basically new chart for asset store and hcms so maintainers for all of them are actually the same, soon asset store and hcms will be removed

Issue #6298, opened 2019-11-18
Revisit API specification description for the Application Registry


majakurcius:
**Description**

Revisit API specification description in the Application Registry and the Application Connector. Make sure the fields are up to date and well-documented and that they're not missing description. 

**Reasons**

We got feedback that some of the parameters in the API specs are missing descriptions or their description is confusing.

PR #6553, opened 2019-12-09
Update App Registry API spec description

majakurcius:
**Description**

- Update the Application Registry API specification description in `docs` to match the one in the `components\...` directory.

**Related issue(s)**
See #6298

### DIFF
--- a/docs/application-connector/assets/metadataapi.yaml
+++ b/docs/application-connector/assets/metadataapi.yaml
@@ -2,24 +2,23 @@ openapi: '3.0.0'
 info:
   version: '1.0.0'
   title: 'Kyma Application Registry API'
-  description: 'API for registering services in Kyma' 
-  license:
-    name: Apache 2.0
-    url: https://www.apache.org/licenses/LICENSE-2.0
+tags:
+  - name: 'Application Registry'
+    description: 'API for registering services in Kyma.'
 paths:
   /v1/health:
     get:
       summary: 'Returns health of a service'
       operationId: 'getHealth'
       tags:
-      - 'health'
+        - 'health'
       responses:
         '200':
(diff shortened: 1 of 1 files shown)


docs/application-connector/assets/metadataapi.yaml
```diff
@@ -376,6 +388,7 @@ components:
         url:
           type: 'string'
           format: 'uri'
+          description: 'URL to fetch OAuth access token'
```
kazydek:
```suggestion
          description: 'URL to fetch an OAuth access token'
```

docs/application-connector/assets/metadataapi.yaml
```diff
@@ -272,17 +280,19 @@ components:
     ApiUpdate:
       type: 'object'
       properties:
-        targetUrl:
+        targetUrl: 'URL that the API exposes and that the proxy calls'
```
kazydek:
```suggestion
        targetUrl: 'URL that the API exposes and the proxy calls'
```

docs/application-connector/assets/metadataapi.yaml
```diff
@@ -250,15 +256,17 @@ components:
         targetUrl:
           type: 'string'
           format: 'uri'
+          description: 'URL that the API exposes and that the proxy calls'
```
kazydek:
```suggestion
          description: 'URL that the API exposes and the proxy calls'
```

docs/application-connector/assets/metadataapi.yaml
majakurcius: Since articles etc. are usually omitted in code, and this file is a duplication of an already merged API spec description, let's leave it as it is for consistency's sake:)

PR #6555, opened 2019-12-09
create new images built using the new makefiles

k15r:
**Description**

- all images related to event-bus and function-controller will be recreated

**Related issue(s)**

### DIFF
--- a/components/event-bus/Makefile
+++ b/components/event-bus/Makefile
@@ -64,7 +64,7 @@ clean:
 
 .PHONY: test
 
-vendor: Gopkg.lock Gopkg.toml
+vendor: Gopkg.lock
 	@echo '+ Pulling vendored dependencies'
 	@dep ensure -v --vendor-only
 
--- a/components/event-bus/Makefile
+++ b/components/event-bus/Makefile
@@ -31,18 +31,14 @@ verify:: all
 PKG := github.com/kyma-project/kyma/components/event-bus
 # produce goals for each cmd program ("bin1 bin2 ...")
 #cmds := $(foreach cmd,$(shell go list -f '{{ if eq .Name "main" }}{{ .ImportPath }}{{end}}' ./...),$(notdir $(cmd)))
-cmds := $(foreach cmd,$(shell docker run $(DOCKER_INTERACTIVE) \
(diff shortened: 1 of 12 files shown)


components/event-bus/Makefile
```diff
@@ -63,14 +63,15 @@ clean:
...
-vendor: Gopkg.lock Gopkg.toml
+vendor: Gopkg.lock
 	@echo '+ Pulling vendored dependencies'
 	@dep ensure -v --vendor-only
 
 
 GOTEST_FLAGS := -cover
 test-local: vendor
 	@echo "+ Performing unit tests"
+	$(eval testpkgs := $(shell go list -test -f '{{ if .TestGoFiles }}{{ .ForTest }}{{end}}' ./... ))
```
antoineco: Is the `eval` call really needed? Can't we just have the variable defined like any other variable outside of the function since it technically doesn't clash with anything?

components/event-bus/Makefile
```diff
@@ -29,15 +30,14 @@ verify:: all
 # base import path
 PKG := github.com/kyma-project/kyma/components/event-bus
 # produce goals for each cmd program ("bin1 bin2 ...")
-cmds := $(foreach cmd,$(shell go list -f '{{ if eq .Name "main" }}{{ .ImportPath }}{{end}}' ./...),$(notdir $(cmd)))
+#cmds := $(foreach cmd,$(shell go list -f '{{ if eq .Name "main" }}{{ .ImportPath }}{{end}}' ./...),$(notdir $(cmd)))
```
antoineco: Is the leftover comment on purpose?

components/event-bus/Makefile
antoineco: Is the `eval` call really needed? Can't we just have the variable defined like any other variable outside of the function since it technically doesn't clash with anything?

components/event-bus/Makefile
k15r: left in just in case we figure out a way to have this evaluated either: 
- using docker and not slow down execution of the makefile
- using go on the machine that executes the Makefile

components/event-bus/Makefile
k15r: this target is evaluated in the docker container. We need this as we need go and currently we dont have go on the machine that executes the Makefile

PR #6556, opened 2019-12-09
bump images for event-bus and kubeless

k15r:
**Description**

- bump all event-bus and kubeless related images

**Related issue(s)**


resources/application-connector/values.yaml
```diff
@@ -38,11 +38,11 @@ global:
     dir:
     version: fbf1000b
   event_service:
-    dir: develop/
-    version: e029fcf4
+    dir:
```
montaro: do you think we still need this key? should we remove it?

resources/nats-streaming/values.yaml
```diff
@@ -16,8 +16,8 @@ global:
   containerRegistry:
     path: eu.gcr.io/kyma-project
   event_bus:
-    dir: develop/
-    version: 5e4264e9
+    dir: ""
```
montaro: In the other occurrences, you did not use the `""`, is it safe to use both styles?

resources/application-connector/values.yaml
k15r: it is still referenced in the templates, so we cannot just remove it from the values

resources/nats-streaming/values.yaml
k15r: correct me if im wrong: "" == empty string , without the "" it evaluates to null... the result is the same

resources/application-connector/values.yaml
montaro: Yes, I understand, would it be cleaner if we removed it from the templates as well?

resources/nats-streaming/values.yaml
montaro: Yes, sure, but for the sake of consistency, should we make all of them with the same value?
My concern was only if the template engine would panic if he found the empty string, but fine forget about it 👍

antoineco: The templating language does the right thing in both case: generate an emtpy string.

antoineco: The templating language does the right thing in both cases: generate an emtpy string.

Issue #4799, opened 2019-07-05
Application Connector should support  adding additional headers and query parameters to OAuth token calls


akgalwas:
**Description**

Additional headers and query parameters are supported for target API calls and specification fetching. Additional headers and query parameters should be supported for OAuth calls as well.

[Here](https://github.com/kyma-project/kyma/issues/6445) is a follow up for the Compass Runtime Agent part of this issue

**Acceptance Criteria**

- OAuth calls support additional headers and query parameters if provided
- Additional headers and query parameters for OAuth calls are documented in Kyma Docs

jcawley5:
Have a use case this is blocking.  Any updates on the progress of the request?

janmedrek:
Hello @jcawley5, we have included this ticket in the scope the next release - you can expect it to be present in 1.9. :)

PR #6565, opened 2019-12-09
Update documentation on additional headers and query parameters in Application Connector

majakurcius:
**Description**

- Since additional headers and query parameters are supported for target API calls and specification fetching, and now also for OAuth calls, we need to update our documentation.

**Related issue(s)**
See also #4799

### DIFF
--- a/docs/application-connector/03-06-api-registration.md
+++ b/docs/application-connector/03-06-api-registration.md
@@ -17,9 +17,9 @@ For the OpenAPI format, both methods are supported.
 You can register OData APIs only with `SpecificationUrl`.
 
 You can specify additional headers and query parameters with requests to:
-- make a call to the target URL
-- fetch the API specification
-- authenticate with a custom OAuth method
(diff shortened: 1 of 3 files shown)


docs/application-connector/03-06-api-registration.md
```diff
@@ -15,3 +15,11 @@ You can pass the API specification in two ways:
 
 For the OpenAPI format, both methods are supported.
 You can register OData APIs only with `SpecificationUrl`.
+
+You can specify additional headers and query parameters with requests to:
+- make a call to the target URL
+- fetch the API specification
+- fetch an OAuth access token
```
franpog859: I think I wouldn't mention the OAuth call this way. The calls to the target URL and for the API specification are explicit. On the other the OAuth call is implicit, user doesn't even have to remember that such call is performed. It's done under the hood. I'd put here something like `- authenticate with custom OAuth method`

PR #6468, opened 2019-11-29
Integrate Rafter in Kyma

magicmatatjahu:
**Description**

Changes proposed in this pull request:

Integrate Rafter in Kyma:
  - enable installation of Rafter
  - disable installation on of `asset-store` and `cms`
  - update image of `cbs` - create `rafter` domain in `console-backend-service`
  - enable integration tests of `rafter` domain in `console-backend-service-tests`
  - update `helm-broker` image - integration with rafter
  - update relevant rbacs
  - change old resources from AssetStore/HCMS to new one in Rafter - create migration job
  - update image of `content-ui`, `service-catalog-ui` with new `rafter` queries in `cbs`
  - update image of `ui-tests` - update tests after introduce new queries
  - write and enable rafter integration tests
  - write tests for Rafter in e2e-upgrade tests (with checking behaviour of migration job)
  - write tests for Rafter in e2e-backup tests and removing old from AssetStore and HCMS
  - migrate compass-runtime to Rafter

**Related issue(s)**
https://github.com/kyma-project/kyma/issues/6239

PR #6570, opened 2019-12-09
Add destination rule to rafter-upload-service

m00g3n:
**Description**

- Add destination rule to rafter-upload-service

**Related issue(s)**

### DIFF
--- a/resources/rafter/charts/upload-service/templates/destination-rule.yaml
+++ b/resources/rafter/charts/upload-service/templates/destination-rule.yaml
@@ -1,0 +1,9 @@
+apiVersion: "networking.istio.io/v1alpha3"
+kind: "DestinationRule"
+metadata:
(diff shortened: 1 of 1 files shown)


resources/rafter/charts/upload-service/templates/destination-rule.yaml
```diff
@@ -0,0 +1,9 @@
+apiVersion: "networking.istio.io/v1alpha3"
```
magicmatatjahu: Please add this template to this https://github.com/kyma-project/kyma/pull/6468 PR, or if you don't have a permission to do this, please write in comment and I will do it for you.

m00g3n: This patch is for `rafter` umbrella chart that is already in master. 
We do not need to wait on #6468 to be merged.

PR #6622, opened 2019-12-13
Add kyma environment broker chart

polskikiel:
**Description**

- Added kyma environment broker chart

**Related issue(s)**

### DIFF
--- a/resources/compass/charts/kyma-environment-broker/values.yaml
+++ b/resources/compass/charts/kyma-environment-broker/values.yaml
@@ -1,9 +1,9 @@
-replicaCount: 1
-
-image:
-  repository: mszostok/kyma-env-broker
-  tag: 0.1.0
-  pullPolicy: IfNotPresent
+deployment:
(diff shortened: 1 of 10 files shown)


resources/compass/charts/kyma-environment-broker/values.yaml
```diff
@@ -0,0 +1,53 @@
...
+
+affinity: {}
+
+provisioner:
+  URL: "http://compass-provisioner.compass-system.svc.cluster.local:3000/graphql"
+  gardener:
+    # name of the secret with kubeconfig to the gardener cluster
+    secretName: "gardener"
+    # name of the gardener project
+    projectName: "gopher"
```
aszecowka: `gopher` project name seems to be not the production configuration

polskikiel: We plan to make this field configurable in the future when the whole scenario with our broker will work.

Issue #6523, opened 2019-12-05
Integrate compass-runtime-agent with rafter


m00g3n:
**Description**

**Reasons**

In 1.9 rafter is replacing asset-store, compass-runtime-agent needs to be adapted to this change.

**Attachments**

PR #6627, opened 2019-12-16
Migrate Compass Runtime to Rafter

michal-hudy:
**Description**

- Migrated from Asset Store to Rafter

**Related issue(s)**

#6523


components/compass-runtime-agent/internal/kyma/apiresources/rafter/clusterassetgrouprepository_test.go
```diff
@@ -42,39 +42,39 @@ func TestCreateDocsTopic(t *testing.T) {
...
-	t.Run("Should update DocsTopic", func(t *testing.T) {
+func TestUpdateClusterAssetGroup(t *testing.T) {
+	t.Run("Should update ClusterAssetGroup", func(t *testing.T) {
 		// given
 		resourceInterfaceMock := &mocks.ResourceInterface{}
-		repository := NewDocsTopicRepository(resourceInterfaceMock)
+		repository := NewAssetGroupRepository(resourceInterfaceMock)
 
-		dt := createK8sDocsTopic()
+		dt := createK8sClusterAssetGroup()
```
franpog859: It's cool to sometimes find some `re`s for the old Remote Environments. It feels like finding easter eggs ^^ Anyway, I think it should be renamed to `ag` for the Asset Group instead of the Docs Topis.
There are four of them. I don't want to spam you with similar comments so just find the `dt :=` here

tests/compass-runtime-agent/test/testkit/assertions/k8s_resource_checker.go
```diff
@@ -329,7 +318,7 @@ func (c *K8sResourceChecker) assertResourcesDoNotExist(t *testing.T, resourceNam
 	assert.True(t, k8serrors.IsNotFound(err))
 
 	//assert Docs Topics have been removed
```
franpog859: `//assert Asset Groups have been removed`

PR #6422, opened 2019-11-26
application-connector: Add support for Knative Event Mesh

antoineco:
**Description**

Changes proposed in this pull request:

Enable support for the new Knative based event mesh (alpha).

- Provision HTTPSource object upon binding (satisfied by event-sources controller).
- Rename event-service VirtualService from `<app>` to `<app>-validator` to avoid clashing the Knative Serving's VirtualService created for the HTTPSource object above.
- Add reverse proxy for the event mesh (HTTP adapter) to the application-connectivity-validator.

Quick demo:

```console
$ kubectl get applications,applicationmappings

NAME                                                     AGE
application.applicationconnector.kyma-project.io/dummy   3m59s

NAME                                                            AGE
applicationmapping.applicationconnector.kyma-project.io/dummy   2m47s
```
```console
$ kubectl -n kyma-integration get httpsource,ksvc

NAME                                       READY   REASON
httpsource.sources.kyma-project.io/dummy   True

NAME                                URL                                               LATESTCREATED   LATESTREADY   READY   REASON
service.serving.knative.dev/dummy   http://dummy.kyma-integration.svc.cluster.local   dummy-wlwsr     dummy-wlwsr   True
```

```console
$ kubectl -n kyma-integration get pod dummy-connectivity-validator-123xyz-123xyz -o jsonpath='{ .spec.containers[0].args }'

[/applicationconnectivityvalidator --proxyPort=8081 ... --eventMeshPathPrefix=/dummy/events --eventMeshHost=dummy.kyma-integration ...]
```

```console
$ kubectl -n kyma-integration get virtualservices.networking.istio.io

NAME                     GATEWAYS                                                                          HOSTS
connector-service        [kyma-system/kyma-gateway]                                                        [connector-service.35.242.147.25.xip.io]
connector-service-mtls   [kyma-system/kyma-gateway-application-connector]                                  [gateway.35.242.147.25.xip.io]
dummy                    [knative-serving/cluster-local-gateway knative-serving/knative-ingress-gateway]   [dummy.kyma-integration dummy.kyma-integration.svc dummy.kyma-integration.svc.cluster.local 0c0563e5ba4f5b17aa30914b52acbff7.probe.invalid]
dummy-validator          [kyma-gateway-application-connector.kyma-system.svc.cluster.local]                [gateway.35.242.147.25.xip.io]
dummy-mesh               [mesh]                                                                            [dummy.kyma-integration.svc.cluster.local 107e7d872485708d4218540d5a5a2380.probe.invalid]
[...]

PR #6634, opened 2019-12-16
Bump application-connector images

antoineco:
**Related issue(s)**

#6422

### DIFF
--- a/resources/application-connector/values.yaml
+++ b/resources/application-connector/values.yaml
@@ -18,10 +18,10 @@ global:
     path: eu.gcr.io/kyma-project
   application_operator:
     dir:
-    version: PR-6422
+    version: 83cc9c71
   application_operator_tests:
     dir:
-    version: PR-6422
+    version: 83cc9c71
   connector_service:
     dir:
     version: 4e3576cd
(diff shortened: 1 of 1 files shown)


resources/application-connector/values.yaml
```diff
@@ -87,5 +87,7 @@ tests:
     skipSslVerify: true
     image:
       dir:
+      # Prow job 'pre-kyma-tests-application-connector-tests" doesn't get
```
nachtmaar: Does a `/test pre-kyma-tests-application-connector-tests` help in order to create the image?

nachtmaar: To trigger the prow job manually from within a PR.

antoineco: I did that, but it will only create another (identical) PR image.

nachtmaar: True, normally only the post submit jobs create a real sha.

antoineco: It doesn't for this image, hence my comment.

nachtmaar: `post-kyma-tests-application-connector-tests` should do we trick, but you are right, it didn't run. At least I don't see it on `https://status.build.kyma-project.io/?type=postsubmit&job=post-kyma-tests-application-connector-tests`

PR #6684, opened 2019-12-18
External solution test for the new Knative Eventing Mesh

anishj0shi:
Update *End to End External Solution Test* with the new eventing mesh alpha
Image bump #6748

PR #6748, opened 2019-12-30
Bump image for external solution test

anishj0shi:
Bump image of external solution test with the new Eventing Mesh alpha
see also: https://github.com/kyma-project/kyma/pull/6684


resources/core/templates/tests/rbac-external-solution.yaml
```diff
@@ -55,6 +55,15 @@ rules:
   - apiGroups: ["gateway.kyma-project.io"]
     resources: ["apis"]
     verbs: ["create", "delete"]
+  - apiGroups: ["messaging.knative.dev"]
```
nachtmaar: Why are the create permissions required for channels and brokers? Also delete?

anishj0shi: we would still support the old test, there  subscriptions are created from the test which are now part of messaging api group, To unify all the messaging group. i gave default access to the test service account

nachtmaar: As discussed: we reduced the RBAC permissions and removed write access there possible

resources/core/templates/tests/test-external-solution.yaml
```diff
@@ -41,7 +41,7 @@ spec:
             - |
               echo 'TESTING start'
               sleep 10
-              /e2e e2e \
+              /e2e e2e-event-mesh \
```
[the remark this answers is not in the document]
nachtmaar: We discussed a bit more and thought that having the new event-mesh-alpha test in it's own testdefinition is better, because the shell script will be easier to write and you can distinguish between both tests in case of a failure.

PR #6780, opened 2020-01-03
Host value split in half in API Rule Header

Wawrzyn321:
**Description**

- Improve api rule details header
- Add customizable caption to `CopiableText`.
- Add customizable column width to `PageHeader`.

**Related issues**
[Console](https://github.com/kyma-project/console/pull/1509)


resources/core/charts/console/values.yaml
```diff
@@ -15,7 +15,7 @@ console:
 core_ui:
   image:
     dir:
```
akucharska: /pr

Issue #6320, opened 2019-11-20
POC: Make Service Catalog awaiting time reasonable


PK85:
**Description**

We need to assure that user can create ServiceInstance, ServiceBinding and ServiceBindingUsage at the same time and ServiceCatalog will give proper time to accomplish all the operations, for example in 24hours.

_Assumption: ServiceBroker is already registered and ServiceClass names/ids are known and static._

**Reasons**

This enhancement will open the way for automation. For example single helm chart which activates Services and make use of them in Lambda.

**AC**
- make sure that creation of SB pointing to not existing SI will work (I think reconcile on failed SB will do a job)
- make sure that creation of  SBU which points to not existing SB_name and App/lambda_name will work (I think reconcile on failed SBU will do a job)
- find how to configure time to be ready per SI,SB,SBU and pickup/document reasonable values like hours (6,12,24hours)
- create an integration test (could be with shorter timeouts in ServiceCatalog)
- documentation with summary and estimation about all implementations that needs to be done in the next release 1.10

mszostok:
**Hints:**

Consider pinging the ServiceBinding when the ServiceInstance is done.
In SBU think about having informers for the ServiceBinding.

jasiu001:
**ServiceCatalog - ServiceBinding**
- If SB is in the `failed` state each informer synchronization will put this resource to the queue and reconciliation for this resource is processed. Delay of time duration for informer synchronization is set in chart values [file](https://github.com/kubernetes-sigs/service-catalog/blob/master/charts/catalog/values.yaml#L63) (default values is 5 min). The resource is removed constantly from a queue when achieving ready/true status. 

Below test was made for: 1 controller worker, 0 `maxRetries` value for the controller and setting synchronization time to one minute, you can see reconciliation for SB is made each minute:
```
I1217 08:20:44.845061       1 controller_binding.go:171] ServiceBinding "default/test-binding" v6317: beginning to process resourceVersion: 6317
I1217 08:21:43.943260       1 controller_binding.go:171] ServiceBinding "default/test-binding" v6317: beginning to process resourceVersion: 6317
I1217 08:22:43.974794       1 controller_binding.go:171] ServiceBinding "default/test-binding" v6317: beginning to process resourceVersion: 6317
I1217 08:23:43.347174       1 controller_binding.go:171] ServiceBinding "default/test-binding" v7305: beginning to process resourceVersion: 7305
```

(The longest period for the test was made for 1h)

[...]

PR #6803, opened 2020-01-08
add a trigger for false status ServiceBindingUsage if related objects…

jasiu001:
Related to #6320 
All details you can find in this [comment](https://github.com/kyma-project/kyma/issues/6320#issuecomment-566558959)


components/service-binding-usage-controller/internal/controller/controller.go
```diff
@@ -127,13 +136,18 @@ func NewServiceBindingUsage(
...
+		guard:                    sbuGuard,
 		maxRetires:               defaultMaxRetries,
 		log:                      log.WithField("service", "controller:service-binding-usage"),
 		queue:                    workqueue.NewNamedRateLimitingQueue(workqueue.DefaultControllerRateLimiter(), "ServiceBindingUsage"),
 		prefixGetter:             &envPrefixGetter{},
 		metric:                   cbm,
 	}
 
+	bindingInformer.Informer().AddEventHandler(cache.ResourceEventHandlerFuncs{
+		UpdateFunc: c.triggerServiceBindingUsageReconciliation,
```
piotrmiskiewicz: what about addFunc?

components/service-binding-usage-controller/internal/controller/controller.go
```diff
@@ -195,6 +209,43 @@ func (c *ServiceBindingUsageController) onUpdateOrRelistServiceBindingUsage(old,
...
+	if !ok {
+		c.log.Errorf("Error handling ServiceBinding update: cannot covert obj [%+v] of type %T to *ServiceBindingUsage", cur, cur)
+		return
+	}
+	if !isServiceBindingReady(sb) {
+		c.log.Infof("ServiceBinding %s/%s is not ready. ServiceBindingUsage trigger action delayed", sb.Namespace, sb.Name)
+		return
+	}
+
+	sbus, err := c.bindingUsageLister.ServiceBindingUsages(sb.Namespace).List(labels.NewSelector())
```
piotrmiskiewicz: you can use labels.Everything()

components/service-binding-usage-controller/internal/controller/guard_test.go
```diff
@@ -0,0 +1,114 @@
+package controller_test
+
```
piotrmiskiewicz: you are testing here if a message was logged or not. It is not the goal of the Guard behaviour.  I would check if the SBU has increased reprocessRequest

components/service-binding-usage-controller/internal/controller/controller.go
jasiu001: `addFunc` is not necessary, in the first iteration (add action) of controller binding, SB is initiating with empty status value (https://github.com/kubernetes-sigs/service-catalog/blob/master/pkg/controller/controller_binding.go#L181-L188)

piotrmiskiewicz: Imagine scenario:
1. Create SBU but SB is not existing.
2. The controller is being restarted.
3. SB is created and get ready (status ready)
4. The controller started (after few seconds)
4.1 The controller is getting Add event for ServiceBinding

After the start of the controller, Binding is not updated, so the controller won't be triggered

jasiu001: Then IMO both events are triggered, add and update.
I tested and it works, what I did:
- create SBU without SB, after 15 attempts SBU failed
- scale down SBU controller
- create SB, wait it will be ready/true
- scale-up SBU controller
SBU status immediately changes to ready/true

piotrmiskiewicz: It seems the SBU controller got Add event and processed just after scaling the controller up.

Issue #6713, opened 2019-12-20
Compass migrator-job template not handling case of managedGCP PostgreSQL


JackCheng01:
**Description**

Compass migrator-job template not handling case of managedGCP PostgreSQL

**Reasons**

The migrator job for compass will fail due to not able to connect to database.
As both director and provisioner deployments are utilising cloudsql-proxy from google. 
We might align with them and add  cloudsql-proxy container in our job definition

**Attachments**
We might add below container and volumes definition to [migrator job template](https://github.com/kyma-project/kyma/blob/967e65ed5098d46d05bf2d585e3d1a5b67879f4e/resources/compass/templates/migrator-job.yaml#L21)
```yaml
        {{if eq .Values.global.database.embedded.enabled false}}
        - name: cloudsql-proxy
          image: gcr.io/cloudsql-docker/gce-proxy:1.11
          command:
          - /bin/sh
          args:
          - -c
          - '/cloud_sql_proxy -instances={{ .Values.global.database.managedGCP.instanceConnectionName }}=tcp:5432 -credential_file=/secrets/cloudsql-instance-credentials/credentials.json & sleep 30; exit'
          volumeMounts:
            - name: cloudsql-instance-credentials
              mountPath: /secrets/cloudsql-instance-credentials
              readOnly: true
          {{- with .Values.deployment.securityContext }}
          securityContext:
{{ toYaml . | indent 12 }}
          {{- end }}
          {{end}}

...

      {{if eq .Values.global.database.embedded.enabled false}}
      volumes:
        - name: cloudsql-instance-credentials
          secret:
            secretName: cloudsql-instance-credentials
      {{end}}
```

pkosiec:
Hi @life0215, is this task already resolved by your PR? Can we close it?

life0215:
Hi, @pkosiec 
Yes, closing.

Issue #6711, opened 2019-12-20
compass/director chart deployment template missing volume mounts in case of managed GCP database


JackCheng01:
**Description**

Deployment template for director component missing volume mounts.
Chart won't work when managedGCP database is enabled.

**Expected result**

deployment finish successfully

**Actual result**

deployment not ready with pod unable to read unmounted resource

**Steps to reproduce**

install Compass with GCP managed Postgres database, set the database.embedded.enabled value to false inside the ./chart/compass/values.yaml file, and fill these values
see [Configure Managed GCP PostgreSQL](https://github.com/kyma-project/kyma/blob/967e65ed5098d46d05bf2d585e3d1a5b67879f4e/resources/compass/configure-managed-gcp-postgresql.md)

**Troubleshooting**

Add volumemounts for container cloudsql-proxy should fix this issue.
https://github.com/kyma-project/kyma/blob/967e65ed5098d46d05bf2d585e3d1a5b67879f4e/resources/compass/charts/director/templates/deployment.yaml#L93

PR #6343, opened 2019-11-21
Quitquitquit

Tomasz-Smelcerz-SAP:
**Description**
New Istio version 1.3.4 brings a standard way to terminate istio-pilot container with a dedicated HTTP endpoint: `/quitquitquit`.
We have to terminate this container in our tests, because when the test container finishes, the istio-pilot container still runs and the test infrastructure observes this as a never-ending run of the test scenario.
At the moment we're using brute-force approach: pkill of the istio-pilot process.
But this solution requires to use the flag `shareProcessNamespace: true`, which doesn't work in the Azure environment.

In order to run our tests in Azure, we have to switch to the `quitquitquit` approach.
Edit: It seems this flag is now working in AKS clusters, but we can encounter similar problems in other managed K8s environments, so the switch to `quitquitquit` is still important.

Changes proposed in this pull request:

- Switch to explicit termination of test pods using /quitquitquit istio-pilot endpoint.

*See also*
Implements: #4719

PR #6806, opened 2020-01-08
Fix Helm templates for Compass managed GCP PostgreSQL

life0215:
**Description**

- Add cloudsql-proxy container to compass-migration job
- Add missing volumemounts to compass-director deploy template

**Related issue(s)**
https://github.com/kyma-project/kyma/issues/6711
compass/director chart deployment template missing volume mounts in case of managed GCP database
https://github.com/kyma-project/kyma/issues/6713
Compass migrator-job template not handling case of managedGCP PostgreSQL

### DIFF
--- a/resources/compass/templates/migrator-job.yaml
+++ b/resources/compass/templates/migrator-job.yaml
@@ -17,25 +17,19 @@ spec:
                 release: {{ .Release.Name }}
         spec:
             restartPolicy: Never
+            shareProcessNamespace: true
             containers:
                 {{if eq .Values.global.database.embedded.enabled false}}
                 - name: cloudsql-proxy
-                  image: gcr.io/cloudsql-docker/gce-proxy:1.11
-                  command: ["/bin/sh", "-c"]
+                  image: gcr.io/cloudsql-docker/gce-proxy:1.13
+                  command:
+                  - /bin/sh
                   args:
-                    - |
-                      /cloud_sql_proxy \
-                      -instances={{ .Values.global.database.managedGCP.instanceConnectionName }}=tcp:5432 \
-                      -credential_file=/secrets/cloudsql-instance-credentials/credentials.json & CHILD_PID=$!
-                      (while true; do if [[ -f "/tmp/pod/migration-terminated" ]]; then kill $CHILD_PID; fi; sleep 1; done) &
-                      wait $CHILD_PID
-                      if [[ -f "/tmp/pod/migration-terminated" ]]; then exit 0; fi
+                  - -c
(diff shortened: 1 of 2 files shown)


resources/compass/templates/migrator-job.yaml
```diff
@@ -18,6 +18,19 @@ spec:
...
             restartPolicy: Never
             containers:
+                {{if eq .Values.global.database.embedded.enabled false}}
+                - name: cloudsql-proxy
+                  image: gcr.io/cloudsql-docker/gce-proxy:1.11
+                  command:
+                  - /bin/sh
+                  args:
+                  - -c
+                  - '/cloud_sql_proxy -instances={{ .Values.global.database.managedGCP.instanceConnectionName }}=tcp:5432 -credential_file=/secrets/cloudsql-instance-credentials/credentials.json & sleep 30; exit'
```
pkosiec: Why do we need to sleep for 30 seconds in parallel?

life0215: This way, this container will exit running status after 30 seconds,  won't block the job mark itself completed.

pkosiec: What if the schema migration takes longer than 30 seconds? The proxy will exit and then the migrator will fail - do I understand correctly?

If so, the solution doesn't seem to be right. Can we do it the same way as killing Istio sidecar? Please see my other comment.

resources/compass/templates/migrator-job.yaml
```diff
@@ -60,3 +73,9 @@ spec:
                   args:
                     - "-c"
                     - "./run.sh; exit_code=$?; echo '# KILLING PILOT-AGENT #'; curl -XPOST http://127.0.0.1:15020/quitquitquit; sleep 5; exit $exit_code;"
```
pkosiec: If `cloud_sql_proxy` doesn't have a dedicated endpoint, please use the `pkill -INT` command like we did before `quitquitquit` endpoint for Istio sidecar: https://github.com/kyma-project/kyma/pull/6343/files#diff-44b7cd4a24bf84a8d012823ad10723a6L19

resources/compass/templates/migrator-job.yaml
life0215: Thanks for the suggestion, has updated this PR with an workaround.

resources/compass/templates/migrator-job.yaml
life0215: Has updated the solution again using pkill, gcr.io/cloudsql-docker/gce-proxy:1.11 is not working with pkill -INT well, using gcr.io/cloudsql-docker/gce-proxy:1.13

Issue #6611, opened 2019-12-13
Persistence profile for ory components


piotrmsc:
**Description**

Currently ory components like hydra do not have any persistence. If you create an Oauth2 client hydra is storing it locally on container disk. If application restarts information about clients is lost.

We need to have an easily configurable profile (helm override) that will setup ory components with persistence enabled.

Write backup and restore tests for api gateway.
Run acceptance tests in parallel with persistence enabled,

**Reasons**

In memory or basically local storage is good for playing around but not for the production use case 

**Attachments**

PR #6812, opened 2020-01-08
Add persistance to ory components

Demonsthere:
**Description**

- Add database subchart to ory
- Enable persistance in hydra
- Create job for connection string creation

**Related issue(s)**

- Implements #6611


Demonsthere:
During development https://github.com/ory/hydra-maester/issues/43 was discovered and reported

installation/resources/production/ory-persistance-profile.yaml.tpl
```diff
@@ -0,0 +1,19 @@
+apiVersion: v1
```
piotrmsc: Production profile should be a single file in installation resources where all components will have its production profiles

installation/resources/production/ory-persistance-profile.yaml.tpl
```diff
@@ -0,0 +1,19 @@
...
+  namespace: kyma-installer
+  labels:
+    installer: overrides
+    component: ory
+    kyma-project.io/installation: ""
+data:
+  postgresql.enabled: "true"
+  hydra.hydra.autoMigrate: "true"
+  global.ory.hydra.persitance.enabled: "true"
+  # global.ory.hydra.persitance.user: "someUser"
```
piotrmsc: it looks the same as commented values in `values.yaml` chart. We have separate PR with docs that should describe and override for external DB.

PR #6865, opened 2020-01-14
Add EnvoyFilter for rewriting connector token to headers

crabtree:
**Description**

- Add EnvoyFilter for rewriting connector token to headers for ory hydrator

**Related issue(s)**

See also https://github.com/kyma-incubator/compass/pull/644 https://github.com/kyma-incubator/compass/pull/645


a-thaler:
That filters are getting active on a plain kyma installation (without compass) in my understanding. Please try to keep the things where they belong to. If there is a way to bundle the filter with compass itself please do so.

Issue #6795, opened 2020-01-07
Event bus deletion and upgrade investigation


Abd4llA:
**Description**
Investigate how to delete Event Bus chart, objects and trigger upgrade script which will be implemented in #6763

**Note**
- Check if we can delete Event Bus related object during deleting Event Bus chart (any hook for deletion in the chart?)

**AC**
1. Investigation is documented properly.

antoineco:
### Expectation

At the end of the Kyma upgrade, the user is left with an event mesh that has feature parity with the previous Kyma version, without the need for any manual configuration and without message loss.

### Logical steps

1. Replicate existing eventing components onto the new event mesh, for each Kyma application.
    For the duration of the update, each application can technically support both event meshes, old and new.
2. Deploy compatibility layer (_to be defined_)
    Messages sent to legacy APIs get routed to the new event mesh.
3. Drain legacy message channels
    In-flight messages must be delivered before the old event mesh gets shut down.
4. Purge old event mesh

### Technical considerations

* **The update logic for _step 1_ needs to be hooked into an existing component, but the event mesh encompasses multiple components.**
    Proposed action: consolidate all custom update steps into update hooks at the level of the `knative-eventing` component, which all other components depend on.
* **The Kyma operator does not currently uninstall components which are no longer required.**
    Proposed action: implement custom logic at the end of the operator's reconciliation to purge the `event-mesh` chart, since this will not be supported anymore moving forward.

antoineco:
### Expectation

At the end of the Kyma upgrade, the user is left with an event mesh that has feature parity with the previous Kyma version, without the need for any manual configuration and without message loss.

### Logical steps

1. **Replicate existing eventing components onto the new event mesh, for each Kyma application.**
    For the duration of the update, each application can technically support both event meshes, old and new.
2. **Deploy compatibility layer** (_to be defined_)
    Messages sent to legacy APIs get routed to the new event mesh.
3. **Drain legacy message channels**
    In-flight messages sent to legacy APIs must be delivered before the old event mesh gets shut down.
4. **Purge old event mesh**

### Technical considerations

* **The update logic for _step 1_ needs to be hooked into an existing component, but the event mesh encompasses multiple components.**
    Proposed action: consolidate all custom update steps into update hooks at the level of the `knative-eventing` component, which all other components depend on.
* **The Kyma operator does not currently uninstall components which are no longer required.**
[...]

PR #6873, opened 2020-01-15
Proposal: migration to Knative Event Mesh

antoineco:
⚠️ This PR is used as a mean to review a proposal, it is **not meant to be merged**.

---

📃  [Rendered markdown](https://github.com/antoineco/kyma/blob/proposal-eventmesh-upgrade/proposal-eventmesh-upgrade.md) (always on latest commit)

**Description**

Proposal for the migration to the Knative Event Mesh.

**Related issue(s)**

#6795

### DIFF
--- a/proposal-eventmesh-upgrade.md
+++ b/proposal-eventmesh-upgrade.md
@@ -3,32 +3,91 @@
 ## Contents
 
 1. [Expectation](#expectation)
-2. [Logical steps](#logical-steps)
-3. [Technical considerations](#technical-considerations)
+2. [Steps](#steps)
+   * [First Kyma update](#first-kyma-update)
+     1. [Set up new event mesh for existing Applications](#1-set-up-new-event-mesh-for-existing-applications)
+     2. [Deploy compatibility layer](#2-deploy-compatibility-layer)
+     3. [Drain legacy message channels](#3-drain-legacy-message-channels)
+     4. [Purge event-bus component](#4-purge-event-bus-component)
+   * [Second Kyma update](#second-kyma-update)
 
 ## Expectation
 
 At the end of the Kyma upgrade, the user is left with an event mesh that has feature parity with the previous Kyma
 version, without the need for any manual configuration and without message loss.
 
-## Logical steps
+## Steps
 
-1. **Replicate existing eventing components onto the new event mesh, for each Kyma application.**
-    For the duration of the update, each application can technically support both event meshes, old and new.
-2. **Deploy compatibility layer** (_to be defined_)
-    Messages sent to legacy APIs get routed to the new event mesh.
-3. **Drain legacy message channels**
-    In-flight messages sent to legacy APIs must be delivered before the old event mesh gets shut down.
-4. **Purge old event mesh**
-    Includes both control (chart) and data (API objects) planes.
+The upgrade happens in two steps to avoid creating a major disruption and to keep the logical path a little clearer. In
+the first step, we ensure existing Kyma applications have a backing event mesh in place and delete the `event-bus`
+component alone. In a second step, we remove leftovers from all other charts and rely on the natural deletion (via Helm)
+that will happen upon the next Kyma update.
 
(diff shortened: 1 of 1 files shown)


antoineco:
Open question: once the upgrade has run, how do we prevent the Application broker from provisioning new EventActivations?

antoineco:
I am not really sold on the idea of doing this migration all in one step - and by that I mean update and delete, not the 2 phases the above proposal - mainly for 2 reasons:

1. Removing the `event-bus` component alone is not sufficient. The eventing features of Kyma are spread across multiple components, therefore we would have to remove features across the board _in one atomic unit of work_ (a single commit), otherwise components such as the Application Broker will start failing immediately after dependencies go missing.
2. We are limited by the current design of the Kyma operator. As mentioned in the proposal, components are processed iteratively, so the ordering if components will bite us. For example, after deleting CRDs from the `cluster-essentials` component, we won't be able to purge eventing objects in a later step, and I'm sure there are other similar edge cases I couldn't think of. A solution could be to hook all the upgrade logic in a completely unrelated component just to let it kick early in the installation process, but does it really make any sense? 

I'd rather revisit this proposal and do the migration in 2 steps:
1. Just update and migrate applications to use the new Knative event mesh. Users can still technically _use_ the old eventing, but they are warned that this thing goes away for good in the next release.
2. Clean up eventing leftovers from the rest of Kyma at our pace, one component at a time, and make it easy on code owners.

proposal-eventmesh-upgrade.md
```diff
@@ -0,0 +1,93 @@
...
+
+**Technical**
+
+_To be defined (may be part of an existing component, or a separate component)_
+
+#### 3. Drain legacy message channels
+
+**Logical**
+
+In-flight messages sent to legacy APIs must be delivered before the old event mesh gets shut down.
```
k15r: Do we really need to actively monitor metrics? The update is done during a kyma upgrade which imho implies that there are no messages sent to the cluster

proposal-eventmesh-upgrade.md
```diff
@@ -0,0 +1,93 @@
...
+#### 4. Purge event-bus component
+
+**Logical**
+
+The `event-bus` component gets uninstalled from the cluster, including the control plane (chart) and data plane (API
+objects).
+
+**Technical**
+
+The Kyma operator does not currently uninstall components which are no longer required, so we need to implement custom
```
k15r: wouldnt it make sense to just remove all objects specified in the chart in this step, but keep the chart.
removing the then empty chart could be done in update phase two from literally any kyma chart. This would remove the necessity modify the installer

k15r: on the other hand: we might want to remove other components in the future as well. so adding some removal functionality to the operator could be beneficial

proposal-eventmesh-upgrade.md
```diff
@@ -0,0 +1,93 @@
...
+
+_target: 1.12_
+
+Remove leftovers from other Kyma components:
+
+* cluster-essentials:
+  * Subscription CRD
+  * EventActivation CRD
+
+_Others?_
```
k15r: in case we dont remove the event-bus chart in phase 1:
* remove the event-bus-chart

proposal-eventmesh-upgrade.md
antoineco: You can't assume there won't be any senders generating events during the update, because you don't have control over how Kyma is being used. You _can_ decide to ignore those events though. Maybe that position is OK with us, maybe not (we need a clear statement).

proposal-eventmesh-upgrade.md
antoineco: Sure, we could also hook that logic at the beginning of the update process, but I'm not sure whether hacking stuff into other components just for the sake of avoiding touching the installer really makes sense. I'd favor a reasonably decent design considering our process is already super hacky. You call.

Issue #6786, opened 2020-01-07
Knative Eventing mesh dashboard - Pods 


Abd4llA:
**Description**
We need a dashboard that shows the status of all Knative Eventing mesh related pods.

1. Metrics to show are
    1. status and restarts
    1. CPU/Memory/Network IO
1. The pods to observe are:
	* Validator
	* HttpSource (Controller & Adapter)
	* Channel pods (http source & broker )
		* Kafka: 3 pods (controller, producer, and consumer)
	* Filter pod per trigger
	* Knative eventing and serving pods (controllers, webhook, activator..etc)

**Note**: Checkout results of #6656

**AC**
- Kyma includes the mentioned dashboard
- In the case of missing metrics, proper issues are created for them

PR #6880, opened 2020-01-15
Eventing mesh dashboard pods

montaro:
Description

    Added label kyma-project.io/event-mesh: true to facilitate the event-mesh dashboard

Related issue(s)

See #6786

### DIFF
--- a/components/application-operator/charts/application/templates/validator.yaml
+++ b/components/application-operator/charts/application/templates/validator.yaml
@@ -22,6 +22,7 @@ spec:
         sidecar.istio.io/inject: "true"
       labels:
         app: {{ .Release.Name }}-connectivity-validator
+        kyma-project.io/dashboard: event-mesh
         release: {{ .Release.Name }}
     spec:
       containers:
--- a/components/event-sources/config/500-sources-controller.yaml
+++ b/components/event-sources/config/500-sources-controller.yaml
@@ -28,6 +28,7 @@ spec:
         sidecar.istio.io/inject: 'false'
       labels:
         app: event-sources-controller-manager
+        kyma-project.io/dashboard: event-mesh
     spec:
       serviceAccountName: event-sources-controller-manager
       containers:
--- a/components/event-sources/reconciler/httpsource/adapter_config.go
+++ b/components/event-sources/reconciler/httpsource/adapter_config.go
@@ -23,7 +23,15 @@ import (
 	"knative.dev/pkg/metrics"
 )
 
-const component = "httpsource"
+const (
+	component = "httpsource"
+
+	// Ports 9090-9091 are reserved for the Serving queue proxy (knative.dev/serving/pkg/apis/networking)
+	adapterMetricsPort = 9092
+
+	dashboardLabelKey   = "kyma-project.io/dashboard"
+	dashboardLabelValue = "event-mesh"
+)
 
 // httpAdapterEnvConfig contains properties used to configure the HTTP adapter.
 // These are automatically populated by envconfig.
@@ -35,9 +43,6 @@ type httpAdapterEnvConfig struct {
 	// CloudEvents receiver port
 	Port int32 `default:"8080"`
 }
-
-// Ports 9090-9091 are reserved for the Serving queue proxy (knative.dev/serving/pkg/apis/networking)
-const adapterMetricsPort = 9092
 
 // updateAdapterMetricsConfig serializes the metrics config from a ConfigMap to
 // JSON and updates the existing config stored in the Reconciler.
--- a/components/event-sources/reconciler/httpsource/reconcile.go
+++ b/components/event-sources/reconciler/httpsource/reconcile.go
@@ -233,6 +233,7 @@ func (r *Reconciler) makeKnService(src *sourcesv1alpha1.HTTPSource,
 		object.WithProbe(adapterHealthEndpoint),
 		object.WithControllerRef(src.ToOwner()),
 		object.WithLabel(routeconfig.VisibilityLabelKey, routeconfig.VisibilityClusterLocal),
+		object.WithPodLabel(dashboardLabelKey, dashboardLabelValue),
 	)
 }
 
--- a/components/event-sources/reconciler/httpsource/reconcile_test.go
+++ b/components/event-sources/reconciler/httpsource/reconcile_test.go
@@ -435,6 +435,9 @@ func newService() *servingv1alpha1.Service {
 						Annotations: map[string]string{
 							autoscaling.MinScaleAnnotationKey: "1",
 						},
+						Labels: map[string]string{
+							dashboardLabelKey: dashboardLabelValue,
+						},
 					},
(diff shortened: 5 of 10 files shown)


components/event-sources/reconciler/httpsource/reconcile.go
```diff
@@ -233,6 +233,7 @@ func (r *Reconciler) makeKnService(src *sourcesv1alpha1.HTTPSource,
 		object.WithProbe(adapterHealthEndpoint),
 		object.WithControllerRef(src.ToOwner()),
 		object.WithLabel(routeconfig.VisibilityLabelKey, routeconfig.VisibilityClusterLocal),
+		object.WithSpecTemplateLabel(eventMeshLabel, "true"),
```
antoineco: An empty struct would be enough and signal that we only care about the label being _present_ (it would make no sense to set `event-mesh: "false"` right?)

components/event-sources/reconciler/object/equality_test.go
```diff
@@ -182,6 +182,14 @@ func TestKsvcEqual(t *testing.T) {
 			},
 			false,
 		},
+		"not equal when template label differ": {
```
antoineco: label **s**

components/event-sources/reconciler/object/service.go
```diff
@@ -77,6 +77,23 @@ func WithMinScale(replicas int) ObjectOption {
 	}
 }
 
+// WithSpecTemplateLabel sets the value of an API object's metadata label
+func WithSpecTemplateLabel(key, val string) ObjectOption {
```
antoineco: I think `WithPodLabel` would be clear enough.

components/event-sources/reconciler/object/service.go
```diff
@@ -77,6 +77,23 @@ func WithMinScale(replicas int) ObjectOption {
...
+
+		tpl := &s.Spec.ConfigurationSpec.Template
+		if *tpl == nil {
+			*tpl = &servingv1alpha1.RevisionTemplateSpec{}
+		}
+		obj := (*tpl).ObjectMeta
+		if obj.Labels == nil {
+			obj.Labels = make(map[string]string)
+		}
+		obj.Labels[key] = val
```
antoineco: Won't work because `obj` is a _copy_ of ObjectMeta, not the original ObjectMeta. https://play.golang.org/p/D4vazyZiZcU
You have to work with pointers, or just use `(*tpl).ObjectMeta` directly.

components/event-sources/test/fixtures/ksvc.json
```diff
@@ -30,6 +30,9 @@
     "spec": {
         "template": {
             "metadata": {
+                "labels": {
+                    "kyma-project.io/event-mesh": "true"
```
antoineco: Same comment about the empty string instead of that misleading "true" value (we only care that the label _exists_).

components/event-sources/test/fixtures/ksvc.json
```diff
@@ -30,6 +30,9 @@
     "spec": {
         "template": {
             "metadata": {
+                "labels": {
```
antoineco: (nit) No big deal but if you `kubectl get` that object again, it will move `labels` _after_ `annotations` due to the alphabetical order. We should respect that.

components/event-sources/reconciler/httpsource/adapter_config.go
```diff
@@ -23,7 +23,14 @@ import (
 	"knative.dev/pkg/metrics"
 )
 
-const component = "httpsource"
+const (
+	component = "httpsource"
+
+	// Ports 9090-9091 are reserved for the Serving queue proxy (knative.dev/serving/pkg/apis/networking)
+	adapterMetricsPort = 9092
```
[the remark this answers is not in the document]
montaro: I found my self adding a new const, so I thought of grouping the, is it a bad idea? specially the file is pretty small. What do you think?

components/event-sources/reconciler/object/service.go
```diff
@@ -77,20 +77,19 @@ func WithMinScale(replicas int) ObjectOption {
 	}
 }
 
-// WithSpecTemplateLabel sets the value of an API object's metadata label
-func WithSpecTemplateLabel(key, val string) ObjectOption {
+// WithPodLabel sets the value of an API object's metadata label
```
antoineco: The comment doesn't really match anymore. How about something along the lines of "sets a label on a Service's template".

components/event-sources/reconciler/httpsource/adapter_config.go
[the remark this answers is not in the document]
antoineco: Sure, no big deal

antoineco:
@montaro tests are failing with
```diff
- 		Labels: nil,
+ 		Labels: map[string]string{"kyma-project.io/event-mesh": ""},
```
You can add the missing label to the test Service below [these lines](https://github.com/kyma-project/kyma/blob/release-1.9/components/event-sources/reconciler/httpsource/reconcile_test.go#L435-L437).

components/application-operator/charts/application/templates/validator.yaml
```diff
@@ -22,6 +22,7 @@ spec:
         sidecar.istio.io/inject: "true"
       labels:
         app: {{ .Release.Name }}-connectivity-validator
+        kyma-project.io/dashboard: "event-mesh"
```
antoineco: Quotes are unnecessary here

components/event-sources/config/500-sources-controller.yaml
```diff
@@ -28,6 +28,7 @@ spec:
         sidecar.istio.io/inject: 'false'
       labels:
         app: event-sources-controller-manager
+        kyma-project.io/dashboard: "event-mesh"
```
antoineco: Quotes are unnecessary here

components/event-sources/reconciler/httpsource/reconcile_test.go
```diff
@@ -435,6 +435,9 @@ func newService() *servingv1alpha1.Service {
 						Annotations: map[string]string{
 							autoscaling.MinScaleAnnotationKey: "1",
 						},
+						Labels: map[string]string{
+								"kyma-project.io/dashboard": "event-mesh",
```
antoineco: Is the indentation off?

components/event-sources/reconciler/httpsource/reconcile_test.go
```diff
@@ -435,6 +435,9 @@ func newService() *servingv1alpha1.Service {
 						Annotations: map[string]string{
 							autoscaling.MinScaleAnnotationKey: "1",
 						},
+						Labels: map[string]string{
+							"kyma-project.io/dashboard": "event-mesh",
```
antoineco: (nit) You could use your constants here. The idea is the have an object that is identical to what the "real" _make_ function generates. Not check the value of constants. It's a minor, and the code is fine regardless. Sorry for not spotting earlier 🙏

PR #6893, opened 2020-01-16
Wait for channel during broker test

k15r:
**Description**

- minimal refactor to enable/disable tests easier
- update dependencies to kyma 1.9
- a channel is required during app broker test. Application creation cannot work without installing the required components anymore (SkipInstallation in Application-CR)
- wait for channel to be ready 
- split into two test-binaries (commandline arguments for tests do not really work with go 1.13)

**Related issue(s)**


tests/end-to-end/backup/pkg/common/common.go
```diff
@@ -99,53 +96,60 @@ func TestBackupAndRestoreCluster(t *testing.T) {
...
 				err := myBackupClient.CreateNamespace(e2eTest.namespace)
-				So(err, ShouldBeNil)
+				convey.So(err, convey.ShouldBeNil)
 				t.Logf("[CreateResources: %s] Starting execution", e2eTest.name)
 				e2eTest.backupTest.CreateResources(e2eTest.namespace)
-				t.Logf("[CreateResources: %s] End with success", e2eTest.name)
-			}
-			for _, e2eTest := range e2eTests {
-				t.Logf("[TestResources: %s] Starting execution", e2eTest.name)
+				logrus.Infof("Testing resources in namespace: %s", e2eTest.namespace)
```
suleymanakbas91: Printing the test name would be better than printing the namespace.

tests/end-to-end/backup/test-cmd/before-backup/Dockerfile
```diff
@@ -0,0 +1,17 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+FROM curlimages/curl:7.66.0
+COPY before-backup.test /
+ENTRYPOINT ["/before-backup.test"]
```
suleymanakbas91: Please add label to be consistent with the others and also copy the license folder as done in the previous dockerfile.  Please do the same for the other dockerfile as well.

tests/end-to-end/backup/pkg/common/common.go
k15r: oh, i missed that

tests/end-to-end/backup/test-cmd/before-backup/Dockerfile
k15r: sure

tests/end-to-end/backup/pkg/tests/servicecatalog/app_broker.go
```diff
@@ -211,6 +230,58 @@ func (f *appBrokerFlow) createApplication() error {
...
+	}
+
+	_, err := f.messagingInterface.Channels(integrationNamespace).Create(channel)
+	if errors.IsAlreadyExists(err) {
+		return nil
+	}
+	return err
+}
+
+func (f *appBrokerFlow) waitForChannel() error {
```
anishj0shi: why are we waiting for messagingchannel but not the httpsource?

tests/end-to-end/backup/pkg/tests/servicecatalog/app_broker.go
```diff
@@ -211,6 +230,58 @@ func (f *appBrokerFlow) createApplication() error {
 	return err
 }
 
+// createChannel creates a Knative channel that would normally be installed by the application controller upon helm-chart installation. As we skip the installation part,
+// we need to ensure that the channel exists as it is required by the knative broker.
+func (f *appBrokerFlow) createChannel() error {
```
anishj0shi: for an app broker does it makes sense to create channel? shouldn't this be done by the httpsource, right?

tests/end-to-end/backup/pkg/tests/servicecatalog/app_broker.go
k15r: In this case we do not send an event. so the only thing we are waiting for should be the channel.

PR #6900, opened 2020-01-17
Increase time delay for the ksvc

anishj0shi:
External solution test fails as the underlying ksvc takes time to be ready. Hence Increasing number of attempts and delay till we figure out the core reason behind the delay.


anishj0shi:
> bump the image

It's bumped in this PR: https://github.com/kyma-project/kyma/pull/6901

PR #6903, opened 2020-01-17
Bump compass chart to the latest version

aszecowka:
**Description**

- ...
- ...
- ...

**Related issue(s)**


resources/compass/values.yaml
```diff
@@ -13,10 +13,13 @@ global:
       path: eu.gcr.io/kyma-project/incubator
     connector:
       dir:
-      version: "ed26afa1"
+      version: "PR-634"
```
crabtree: Please put here the final image tag

resources/compass/charts/cockpit/values.yaml
```diff
@@ -3,7 +3,8 @@ images:
     path: eu.gcr.io/kyma-project
   ui:
     dir:
-    version: ab79f919
+    version: 56a4db12
```
dariadomagala: `ab79f919` is a newer commit than `56a4db12`, why did you change it back?

aszecowka: thanks for spotting that, this was a version used in https://github.com/kyma-incubator/compass/blob/master/chart/compass/charts/cockpit/values.yaml
I revert that change.

resources/compass/charts/connectivity-adapter/templates/deployment.yaml
```diff
@@ -0,0 +1,31 @@
+apiVersion: apps/v1beta2
```
Szymongib:
```suggestion
apiVersion: apps/v1
```
`apps/v1beta2` is deprecated and is not supported from Kubernetes 1.16 up.

aszecowka: thanks @Szymongib, I will solve this in another PR

ebensom:
Hi,
This change breaks deployments of director and migrator-job with Cloud SQL DB. May I ask if the cloud_sql_proxy code deletions were intentional in the corresponding helm charts?

PR #6932, opened 2020-01-21
Remove add runtime to default scenario step in E2E tests

crabtree:
**Description**

- Remove add runtime to default scenario step in E2E tests

**Related issue(s)**

See also https://github.com/kyma-incubator/compass/issues/657


resources/compass-runtime-agent/values.yaml
```diff
@@ -9,8 +9,8 @@ global:
       dir:
       version: "PR-6835"
     compassExternalSolutionTests:
-      dir: "develop/"
-      version: "79acee1e"
+      dir: "pr/"
```
tgorgol: Please, bump external-solution-integration values in other charts that use that component (different scenarios) to make sure that the changes didn't affect them.

Issue #6646, opened 2019-12-16
Production profile for service mesh


piotrmsc:
**Description**

Cluster (default) values for service mesh are not enough/ not suitable for production use cases. 
We have limited sidecar resources (requests and limits for CPU and memory) to provide minimal set working on a reasonably small cluster (n1-standard4 on GCP). HPA are enabled but should be reconsidered for critical control plane components.

We need to offer production profile - set of overrides that will deploy istio capable of performing on level around 800 req/s (with whole kyma installed). Istio which will not crash under load but on the other hand will not consume up to 2vcpu like it is in one of the builtin istio profiles.

We should also think of persistence for istio policy enforcement, currently, it's in-memory.

Another thing is can we limit the configuration that envoy holds? https://banzaicloud.com/blog/istio-sidecar/
Maybe it's possible to hold minimum config instead of all envoys in single envoy instance?

**Reasons**

The default cluster profile is actually enough for playground purposes but not for production use case. 
With increased load it starts to act weirdly. Enovy with current resource limits can take up to only 300 req/s

**Attachments**

PR #6946, opened 2020-01-22
Istio & Ory performance tests

strekm:
**Description**

- Istio performance tests
- Ory performance tests

**Related issue(s)**
see https://github.com/kyma-project/kyma/issues/6646 and


tests/perf/components/istio/istio.js
```diff
@@ -0,0 +1,84 @@
...
+            "status was NOT 501": (r) => r.status !== 501,
+            "status was NOT 503": (r) => r.status !== 503,
+            "transaction time < 150 ms": (r) => r.timings.duration < 150,
+            "transaction time in (150;400) ms": (r) => (r.timings.duration > 150 && r.timings.duration < 400),
+            "transaction time > 400 ms": (r) => r.timings.duration > 400,
+        }, {
+            "test": manyToOwnName,
+        });
+    });
+}
```
suleymanakbas91: Please put a new line to the end of all the files :)

Tomasz-Smelcerz-SAP:
I've applied suggestions from shellcheck and added more verbose logging.

tests/perf/prerequisites/istio/setup.sh
```diff
@@ -0,0 +1,57 @@
...
+export WORKLOAD_SIZE=${WORKLOAD_SIZE:-10}
+export VUS=${VUS:-64}
+
+common_resources=(
+    namespace.yaml
+)
+
+workload_resources=(
+    app.yaml
+    api.yaml
```
Tomasz-Smelcerz-SAP: `api.yaml` does not exist, there's `apirule.yaml` file.

tests/perf/prerequisites/istio/setup.sh
```diff
@@ -0,0 +1,57 @@
...
+    export WORKER=$((i + 1))
+    for resource in "${workload_resources[@]}"; do
+        echo "setting up resource: ${resource} for worker ${WORKER}"
+        envsubst <"${WORKING_DIR}/$resource" | kubectl -n "${NAMESPACE}" apply -f -
+
+        LIMIT=15
+        COUNTER=0
+        SUCCESS="false"
+
+        while [ ${COUNTER} -lt ${LIMIT} ]; do
```
Tomasz-Smelcerz-SAP: We are waiting for every pod before we start the next one. Is that on purpose? What if we start all pods at once, then wait for their readiness? A cluster can start pods in parallel, this could reduce setup time significantly

tests/perf/prerequisites/istio/setup.sh
```diff
@@ -0,0 +1,57 @@
...
+    api.yaml
+)
+
+for resource in "${common_resources[@]}"; do
+    envsubst <"${WORKING_DIR}/$resource" | kubectl -n "${NAMESPACE}" apply -f -
+done
+
+for (( i = 0; i < WORKLOAD_SIZE; i++ )); do
+    export WORKER=$((i + 1))
+    for resource in "${workload_resources[@]}"; do
```
Tomasz-Smelcerz-SAP: Only one of the files actually starts k8s pods: `app.yaml`. The other, `apirule.yaml` only applies apiRules. But we're waiting for pods in both cases. I'd suggest to extract the entire loop into a function. This function could have two params: `resourceName:string` and `waitForPods: bool`. Then we can invoke this function twice: `doIt("app.yaml", "true")` and doIt("apirule.yaml", "false")`. Perhaps you can find a better solution.

tests/perf/prerequisites/istio/setup.sh
```diff
@@ -0,0 +1,57 @@
+#!/usr/bin/env bash
```
Tomasz-Smelcerz-SAP: See my last commit for changes I made - shellcheck suggestions.

PR #6967, opened 2020-01-23
Update Director image with new API Spec name validation

crabtree:
**Description**

- Update Director image with new API Spec name validation

**Related issue(s)**

See also https://github.com/kyma-incubator/compass/issues/652


dbadura:
Images updated in
https://github.com/kyma-project/kyma/pull/6962

PR #6972, opened 2020-01-24
Make the Ports configurable based on cloud provider

rakesh-garimella:
**Description**
Make the core DNS service port configurable based on cloud provider as the metrics are exposed on different ports on Azure and on GKE.

**Related issue(s)**


a-thaler:
Solved in a different way already

Issue #6787, opened 2020-01-07
Knative Eventing mesh dashboard  - delivery


Abd4llA:
**Description**
We need a dashboard that shows the eventing delivery status of all Knative Eventing mesh related components. Initially, we want to monitor HTTP Status codes (2XX, 4XX, 5XX..etc) for following pods
* Validator
* HttpSource (event_count)
* HttpSource Channel pod (producer pods for Kafka)
* Broker Channel pod
* Filter pods
* Subscriber (lambda)

**AC**
- Kyma includes the mentioned dashboard changes
- In case of missing metrics, proper issues are created for them

montaro:
@Abd4llA Should we add the Broker-Ingress metrics as well?
https://github.com/knative/eventing/blob/master/docs/metrics.md#broker

PR #6988, opened 2020-01-24
Eventing mesh dashboard delivery [component]

montaro:
This PR adds the logic of creating a new istio policy when the service provisioning happens in order to make the namespace's  knative broker (ingress and filter) scrapeable for Prometheus.

See also #6787


components/application-broker/internal/broker/provision.go
```diff
@@ -234,6 +246,12 @@ func (svc *ProvisionService) do(iID internal.InstanceID, opID internal.Operation
 		return
 	}
 
+	// Create istio policy
+	if err := svc.createIstioPolicy(ns); err != nil {
+		//TODO
+		return
```
montaro: TODO

components/application-broker/internal/broker/provision.go
```diff
@@ -394,6 +412,44 @@ func (svc *ProvisionService) channelForApp(applicationName internal.ApplicationN
...
+	svc.log.Infof("Creating Policy %s in namespace: %s", policyName, string(ns))
+
+	brokerTargetSelector := &istioauthenticationalpha1.TargetSelector{
+		Name: brokerTargetSelectorName,
+	}
+	filterTargetSelector := &istioauthenticationalpha1.TargetSelector{
+		Name: filterTargetSelectorName,
+	}
+	mtls := &istioauthenticationalpha1.MutualTls{
+		//PERMISSIVE
```
sayanh: Would you make 1 a constant out of it? We could even name it as PERMISSIVE_MODE. This would get rid of the comment IMO.

components/application-broker/internal/broker/provision.go
```diff
@@ -394,6 +412,44 @@ func (svc *ProvisionService) channelForApp(applicationName internal.ApplicationN
...
+		Mode: 1,
+	}
+	peerAuthenticationMethod := &istioauthenticationalpha1.PeerAuthenticationMethod{
+		Params: &istioauthenticationalpha1.PeerAuthenticationMethod_Mtls{
+			Mtls: mtls,
+		},
+	}
+
+	_, error := svc.istioClient.AuthenticationV1alpha1().Policies(string(ns)).Create(&istiov1alpha1.Policy{
```
sayanh: Would it make sense to scoop out a method to get a new Policy? That will make it more readable IMO.

components/application-broker/internal/broker/provision.go
```diff
@@ -410,3 +466,5 @@ func getSvcByID(services []internal.Service, id internal.ApplicationServiceID) (
 func strPtr(str string) *string {
 	return &str
 }
+
+//
```
sayanh: Probably a fat-finger

components/application-broker/internal/broker/provision_test.go
```diff
@@ -710,6 +716,7 @@ func TestDoProvision(t *testing.T) {
 				mockServiceInstanceGetter,
 				fake.NewSimpleClientset().ApplicationconnectorV1alpha1(),
 				knative.NewClient(knCli, k8sCli),
+				istioversionedclientfake.NewSimpleClientset(),
```
sayanh: How do we validate the policy which got created?

montaro: Still not validated. I suggest creating a ticket for proper testing. What do you think?
CC: @Abd4llA ?

components/application-broker/internal/broker/provision.go
montaro: Well, I would say it is only 6 lines to be separated, plus this function would have a single client.

components/application-broker/internal/broker/provision_test.go
montaro: Now its tested.

components/application-broker/internal/broker/deprovision_test.go
```diff
@@ -234,7 +234,7 @@ func TestDoDeprovision(t *testing.T) {
 				fixDeprovisionSucceeded(),
 			).Return(nil).Once()
 
-			knCli, k8sCli := bt.NewFakeClients(tc.initialObjs...)
+			knCli, k8sCli, _ := bt.NewFakeClients(tc.initialObjs...)
```
sayanh: I know in deprovisioning process policy removal is not needed. It is based on the decision that we will keep the broker after the deprovisioning process along with the policy. I think a comment on the background with a Github issue link will be helpful so that in future we know why `istioclientset` is ignored here.

components/application-broker/internal/broker/provision.go
```diff
@@ -11,6 +11,9 @@ import (
 	"github.com/pkg/errors"
 	osb "github.com/pmorie/go-open-service-broker-client/v2"
 	"github.com/sirupsen/logrus"
+	istioauthenticationalpha1 "istio.io/api/authentication/v1alpha1"
+	istiov1alpha1 "istio.io/client-go/pkg/apis/authentication/v1alpha1"
+	istioversionedclient "istio.io/client-go/pkg/clientset/versioned"
```
sayanh: (nit): IMO it makes sense to separate the imports from k8s.io and club it with other third-party imports.

components/application-broker/internal/broker/provision.go
```diff
@@ -229,7 +245,19 @@ func (svc *ProvisionService) do(iID internal.InstanceID, opID internal.Operation
...
-		opDesc = fmt.Sprintf("provisioning failed while enabling default Knative Broker for namespace: %s on error: %s", ns, err)
+		opDesc = fmt.Sprintf("provisioning failed while enabling default Knative Broker for namespace: %s"+
+			" on error: %s", ns, err)
+		svc.updateStates(iID, opID, instanceState, opState, opDesc)
+		return
+	}
+
+	// Create istio policy
+	if err := svc.createIstioPolicy(ns); err != nil {
+		svc.log.Printf("Error creating istio policy: %v", err)
```
sayanh: Isn't this `svc.log.Printf` redundant as `opDesc` is actually meant for that?

components/application-broker/internal/broker/provision.go
```diff
@@ -394,6 +422,53 @@ func (svc *ProvisionService) channelForApp(applicationName internal.ApplicationN
 	return svc.knClient.GetChannelByLabels(integrationNamespace, labels)
 }
 
+// Create a new policy
+func (svc *ProvisionService) createIstioPolicy(ns internal.Namespace) error {
+	policyName := string(ns) + policyNameSuffix
```
sayanh: If we can make a namespace variable that is of string type then we can avoid casting twice.
1: line 427
2: line 448

components/application-broker/internal/broker/provision_test.go
```diff
@@ -9,6 +9,7 @@ import (
 	"github.com/pkg/errors"
 	osb "github.com/pmorie/go-open-service-broker-client/v2"
 	"github.com/stretchr/testify/assert"
+	istioversionedclientfake "istio.io/client-go/pkg/clientset/versioned/fake"
```
sayanh: (nit): `istioversionedclientfake` can be `fakeistioversionedclient` or something else to sound good.

components/application-broker/internal/broker/provision_test.go
```diff
@@ -655,6 +665,9 @@ func TestDoProvision(t *testing.T) {
 				bt.NewAppNamespace(string(appNs), false),
 				bt.NewAppSubscription(string(appNs), string(appName)),
 			},
+			expectCreates: []runtime.Object{
+				bt.NewIstioPolicy(string(appNs), string(appNs)+policyNameSuffix),
```
sayanh: Same as before. I don't see a reason to cast the variables multiple times.

components/application-broker/internal/broker/testing/listers.go
```diff
@@ -0,0 +1,57 @@
...
+
+type Listers struct {
+	sorter testing.ObjectSorter
+}
+
+func NewScheme() *runtime.Scheme {
+	scheme := runtime.NewScheme()
+
+	for _, addTo := range clientSetSchemes {
+		addTo(scheme)
```
ksputo: you should handle the error from underlying method `func (sb *SchemeBuilder) AddToScheme(s *Scheme) error` here.
```suggestion
		if err := addTo(scheme); err != nil {
		return nil, errors.Wrapf(err, "while adding to scheme")
		}
```

components/application-broker/internal/broker/testing/client_assertion.go
```diff
@@ -5,22 +5,23 @@ import (
 
 	"github.com/google/go-cmp/cmp"
 
+	istioversionedclientfake "istio.io/client-go/pkg/clientset/versioned/fake"
```
sayanh: (nit): I see you used fakeistioclientset([here](components/application-broker/internal/broker/testing/listers.go)) which is great IMO. Lets be consistent.

components/application-broker/internal/broker/testing/listers.go
```diff
@@ -0,0 +1,57 @@
+package testing
+
+import (
+	fakeistioclientset "istio.io/client-go/pkg/clientset/versioned/fake"
```
sayanh: (nit): Lets cluster the imports thridparty and k8s.io

components/application-broker/internal/broker/testing/api_object.go
```diff
@@ -97,6 +107,34 @@ func NewDefaultBroker(ns string) *eventingv1alpha1.Broker {
...
+	peerAuthenticationMethod := &istioauthenticationalpha1.PeerAuthenticationMethod{
+		Params: &istioauthenticationalpha1.PeerAuthenticationMethod_Mtls{
+			Mtls: mtls,
+		},
+	}
+
+	return &istiov1alpha1.Policy{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      policyName,
+			Namespace: ns,
```
sayanh: Let's assign some coherent labels.

components/application-broker/internal/broker/provision.go
```diff
@@ -394,6 +422,53 @@ func (svc *ProvisionService) channelForApp(applicationName internal.ApplicationN
...
+			Targets: []*istioauthenticationalpha1.TargetSelector{brokerTargetSelector, filterTargetSelector},
+			Peers:   []*istioauthenticationalpha1.PeerAuthenticationMethod{peerAuthenticationMethod},
+		},
+	}
+
+	_, error := svc.istioClient.AuthenticationV1alpha1().Policies(string(ns)).Create(policy)
+	if error != nil {
+		if apiErrors.IsAlreadyExists(error) {
+			if _, err := svc.istioClient.AuthenticationV1alpha1().Policies(string(ns)).Update(policy); err != nil {
+				svc.log.Printf("Updating Policies %s in namespace: %s failed with error:\n %s", policyName, ns,
```
sayanh: How about using `svc.log.Error`

components/application-broker/internal/broker/provision_test.go
sayanh: It turns out `NewIstioPolicy` is only used for testing. I think a better name(e.g. `createTestPolicy`) should be used to clear things out. What do you think?

components/application-broker/cmd/broker/main.go
```diff
@@ -136,7 +142,7 @@ func SetupServerAndRunControllers(cfg *config.Config, log *logrus.Entry, stopCh
 	srv := broker.New(sFact.Application(), sFact.Instance(), sFact.InstanceOperation(), accessChecker,
 		mClient.ApplicationconnectorV1alpha1(), siFacade,
 		mInformersGroup.ApplicationMappings().Lister(), brokerService,
-		&mClient, knClient, log, livenessCheckStatus)
+		&mClient, knClient, istioClient, log, livenessCheckStatus)
```
sayanh: In order to be consistent with mClient, usage of a pointer for istioversioned.Interface makes sense for me.

montaro: @sayanh But the already used `knClient` for example is not a pointer. 🤔

components/application-broker/internal/broker/provision_test.go
montaro: @sayanh I'd say it's already in the testing package, and the second thing, I'm following the convention of naming helpers in the same file.

components/application-broker/internal/broker/testing/api_object.go
montaro: @sayanh Would you please suggest examples?

components/application-broker/internal/broker/provision.go
```diff
@@ -394,6 +422,53 @@ func (svc *ProvisionService) channelForApp(applicationName internal.ApplicationN
 	return svc.knClient.GetChannelByLabels(integrationNamespace, labels)
 }
 
+// Create a new policy
+func (svc *ProvisionService) createIstioPolicy(ns internal.Namespace) error {
```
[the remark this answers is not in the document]
montaro: I wanted to follow the convention of using the defined type

components/application-broker/internal/broker/testing/listers.go
```diff
@@ -0,0 +1,64 @@
...
+type Listers struct {
+	sorter testing.ObjectSorter
+}
+
+func NewScheme() *runtime.Scheme {
+	scheme := runtime.NewScheme()
+
+	for _, addTo := range clientSetSchemes {
+		if err := addTo(scheme); err != nil {
+			_ = errors.Wrapf(err, "error while adding to scheme")
```
ksputo: now the NewScheme() method would fail silently if error occurs and return nil, which shouldn't happen. As this code is part of testing package, you could use panic here. This way you won't have to modify the whole method to return error. See:
https://github.com/kyma-project/kyma/blob/c24dec578c54fa35a8a9278d19e48b2621688536/components/event-bus/internal/knative/subscription/controller/testing/listers.go#L59-L68

components/application-broker/internal/broker/testing/api_object.go
sayanh: How about `eventing.knative.dev/broker: default`? Rationale is the policy is for the `default` broker

components/application-broker/internal/broker/provision.go
```diff
@@ -394,6 +425,53 @@ func (svc *ProvisionService) channelForApp(applicationName internal.ApplicationN
...
+			Targets: []*istioauthenticationalpha1.TargetSelector{brokerTargetSelector, filterTargetSelector},
+			Peers:   []*istioauthenticationalpha1.PeerAuthenticationMethod{peerAuthenticationMethod},
+		},
+	}
+
+	_, createError := svc.istioClient.AuthenticationV1alpha1().Policies(string(ns)).Create(policy)
+	if createError != nil {
+		if apiErrors.IsAlreadyExists(createError) {
+			if _, updateError := svc.istioClient.AuthenticationV1alpha1().Policies(string(ns)).Update(policy); updateError != nil {
+				svc.log.Printf("Updating Policies %s in namespace: %s failed with createError:\n %s", policyName, ns,
```
ksputo: logging an error here is redundant, because if it occurs you first log it and then return it, and then log the returned error again in line 259. Instead you can wrap it with a message informing about the context of the error occurance:
```suggestion
				return errors.Wrapf(err, "while creating policies %s in namespace: %s failed with error:\n %s", policyName, ns)
```

components/application-broker/internal/broker/provision.go
montaro: @sayanh But the `opDesc` is not logged anywhere, no?
BTW, I changed it to `svc.log.Errorf("Error creating istio policy: %v", err)`

sayanh: Sure, the log doesn't give any value as the error is already logged in `updateStates`

components/application-broker/internal/broker/provision.go
```diff
@@ -394,6 +425,56 @@ func (svc *ProvisionService) channelForApp(applicationName internal.ApplicationN
...
+			Namespace: string(ns),
+			Labels:    policyLabels,
+		},
+		Spec: istioauthenticationalpha1.Policy{
+			Targets: []*istioauthenticationalpha1.TargetSelector{brokerTargetSelector, filterTargetSelector},
+			Peers:   []*istioauthenticationalpha1.PeerAuthenticationMethod{peerAuthenticationMethod},
+		},
+	}
+
+	_, createError := svc.istioClient.AuthenticationV1alpha1().Policies(string(ns)).Create(policy)
```
ksputo: you can change `createError` and `updateError` to `err` in this block of code. It will work just fine as scoping will do the job

components/application-broker/internal/broker/provision.go
```diff
@@ -394,6 +425,56 @@ func (svc *ProvisionService) channelForApp(applicationName internal.ApplicationN
...
+	if createError != nil {
+		if apiErrors.IsAlreadyExists(createError) {
+			if _, updateError := svc.istioClient.AuthenticationV1alpha1().Policies(string(ns)).Update(policy); updateError != nil {
+				return errors.Wrapf(updateError, "while updating istio policy with name: %q in namespace: %q", policyName, ns)
+			}
+			return nil
+		}
+		return errors.Wrapf(createError, "while creating istio policy with name: %q in namespace: %q", policyName, ns)
+		return createError
```
ksputo: a little mistake here, you are already returning the error
```suggestion
```

PR #7060, opened 2020-01-30
Add namespace arg in create limit range query

piotrmsc:
**Description**

- Fixed RBAC check when creating limit range which is namespace scoped.

Tests : I'm working on UI tests for flow using namespace admin user

**Related issue(s)**


piotrmsc:
After discussing with @dariadomagala UI test will be skipped due to the fact they are planned to be deleted

Issue #5582, opened 2019-09-09
Expose Knative Function using API


k15r:
Currently Functions created by the function controller are getting exposed using the default knative mechanisms. We need to enable exposing using api-objects in order to fully integrate into kyma.

michal-hudy:
To achieve that few things need to be done:
 - Internal functions need to point to the mesh (#6867)
 - Function controller should create only internal functions (#5580)
 - API rule should point to function placeholder service that will be handled by Istio
 - CBS should filter out revision and private function services

Thanks to that we don't need to modify API controller and API UI will be compatible with functions.

michal-hudy:
Generally, it works, but I noticed one issue:
![image.png](https://images.zenhubusercontent.com/5b85435e6092773c0b8551d0/ee5e2972-3ed7-49dd-8213-e8ecc7a5893b)

Such a label shouldn't be visible on `function` CR. Yes, if someone explicitly set this value it should override the default that we are configuring on `kservice`, but by default, we should set any label on `Function` CR, especially label that is not possible to remove.

michal-hudy:
@magicmatatjahu Generally, it works, but I noticed one issue.

Label `serving.knative.dev/visibility: cluster-local` should be set by Function Controller on `kservice`, it shouldn't be visible on `Function` CR. It is something that can confuse the user because there is a label, so he can think that as he will remove that label, the lambda will be public. Another problem is that it is not possible to remove that label from `function` CR.

So, label `serving.knative.dev/visibility: cluster-local` should be set only on `kservice` not `function` CR.

michal-hudy:
@kazydek we should also remove the sentence that says the label is set on `function` CR from https://kyma-project.io/docs/master/components/serverless-v2/#details-exposing-lambdas

Issue #5580, opened 2019-09-09
Update knative-function-controller to make function internal


k15r:
to make functions internal we need to add the following label on the ksvc object that is created by the function-controller
```serving.knative.dev/visibility: cluster-local```

PR #7070, opened 2020-01-31
Make function as internal cluster scoped by default

magicmatatjahu:
**Description**

- Make function as default internal cluster scoped:
  - add to CRD optional field `visibility`, which defines visibility of function (internal cluster scoped or exposed - at the moment only those two options)
  - apply default value (internal cluster-scoped function), when `visibility` is not set

**Related issue(s)**
https://github.com/kyma-project/kyma/issues/5580
https://github.com/kyma-project/kyma/issues/5582


components/function-controller/pkg/apis/serverless/v1alpha1/function_types.go
```diff
@@ -23,29 +23,43 @@ import (
...
-// TemplateKind defines the type of BuildTemplate used by the build.
+// FunctionVisibility defines the visibility of function.
+// +kubebuilder:validation:Enum=cluster-local
+type FunctionVisibility string
+
+const (
+	// FunctionVisibilityClusterLocal is used to denote that the Function
+	// should be only be exposed locally to the cluster.
+	// This is the default value for FunctionVisibility.
+	FunctionVisibilityClusterLocal FunctionVisibility = "cluster-local"
```
m00g3n: I was wondering if we have plans to have other function visibilities than `cluster-local`?

components/function-controller/config/crd/bases/serverless.kyma-project.io_functions.yaml
```diff
@@ -18,6 +16,11 @@ spec:
     description: Runtime is the programming language used for a function e.g. nodejs8
     name: Runtime
     type: string
+  - JSONPath: .spec.visibility
+    description: Visibility defines function's visibility through cluster, available
```
m00g3n: ATM visibility may be set only to `cluster-local`. It would be good to have it in description.

> Visibility defines function's visibility through cluster (current version supports only in-cluster visibility)

components/function-controller/pkg/apis/serverless/v1alpha1/function_types.go
magicmatatjahu: Is is something like a `way` for improvements in future, but at the moment here will be only `cluster-local` value.

components/function-controller/config/crd/bases/serverless.kyma-project.io_functions.yaml
magicmatatjahu: I removed it from CRDs :)

components/function-controller/pkg/apis/serverless/v1alpha1/function_types.go
magicmatatjahu: I also removed `visibility` field from crd, but logic stays.

Issue #7054, opened 2020-01-30
Deprecate helm options overrides for istio


piotrmsc:
**Description**

Istio deprecated helm charts and its overrides aka helm options https://istio.io/docs/reference/config/installation-options/. Kyma installer support installation overrides and we should deprecate existing old overrides. We must provide new way of customizing istio installation using installer overrides but without relying on istio helm options.
Documentation should be updated as well, describing how to customize istio with examples.

We should either allow to use new istio installation options https://istio.io/docs/reference/config/istio.operator.v1alpha12.pb/ to be set as kyma overrides or allow overriding whole control plane file as an override.

**Reasons**

Charts and helm options are deprecated and around 1.6 will be gone.

**Attachments**

PR #7129, opened 2020-02-06
Istio control plane overrides

Tomasz-Smelcerz-SAP:
**Description**

Allow the Istio Control-Plane definition as an override for the Kyma-Installer.
There is a new override for Istio chart: `kyma_istio_control_plane`.
The value of this override should be a proper YAML definition of `install.istio.io/v1alpha2:IstioControlPlane` CR.
The settings defined in the override are merged with default settings from Istio chart.

- Introduce `kyma_istio_control_plane` override for Istio installation

**Related issue(s)**
See also #7054


resources/istio/templates/configmap-cluster.yaml
```diff
@@ -1,8 +1,15 @@
...
 metadata:
   namespace: {{ .Release.Namespace }}
-  name: kyma-cluster-config
+  name: kyma-istio-control-plane-config
 data:
   config.yaml: |-
 {{.Files.Get "files/kyma_cluster_istio_control_plane.yaml" | printf "%s" | indent 4}}
+{{ if .Values.kyma_istio_control_plane }}
+  #This additional user-provided definition is merged with the default one
+  overrides.yaml: |-
```
Tomasz-Smelcerz-SAP: Note for reviewer: Actual size of the configmap is currently about 40KiB. This is way below the etcd object limit of 1MiB.

installation/resources/installer-config-local.yaml.tpl
```diff
@@ -146,4 +146,366 @@ metadata:
     component: core
     kyma-project.io/installation: ""
 data:
-  kubeless.tests.enabled: "false"
\ No newline at end of file
+  kubeless.tests.enabled: "false"
+---
+apiVersion: v1
```
Tomasz-Smelcerz-SAP: Note for reviewer: This is for test purposes only, will be removed before merge

resources/istio/files/job-script.sh
```diff
@@ -45,8 +45,14 @@ if [ ! -z "$overrides" ]; then
...
-printf "istioctl manifest apply -f /etc/istio/config.yaml ${overrides_transformed}\n"
-istioctl manifest apply -f /etc/istio/config.yaml ${overrides_transformed}
+istio_control_plane_overrides=""
+
+if [ -f "/etc/istio/overrides.yaml" ]; then
+  istio_control_plane_overrides="-f /etc/istio/overrides.yaml"
+fi
+
+printf "istioctl manifest apply -f /etc/istio/config.yaml ${istio_control_plane_overrides} ${overrides_transformed}\n"
+istioctl manifest apply -f /etc/istio/config.yaml ${istio_control_plane_overrides} ${overrides_transformed}
```
piotrmsc: To consider : if istio control plane is provided perhaps it  'old' overrides aka `overrides_transformed` should be ignored. So if user provides old helm options they are taken into account but if CP is provided then it takes precedence.  Why? there's a chance that merge of values will not properly be applied.

resources/istio/templates/configmap-cluster.yaml
```diff
@@ -1,8 +1,15 @@
+{{ if not .Values.global.isLocalEnv }}
```
piotrmsc: nit: not needed TBH

installation/resources/installer-config-local.yaml.tpl
```diff
@@ -146,4 +146,5 @@ metadata:
     component: core
     kyma-project.io/installation: ""
 data:
-  kubeless.tests.enabled: "false"
\ No newline at end of file
+  kubeless.tests.enabled: "false"
+---
```
Demonsthere: Looks like a leftover change :)

Issue #7043, opened 2020-01-30
Application Gateway - deployment in customer's namespaces


janmedrek:
**Description**

In order to preserve resources, we would like to deploy Application Gateway only when it is required. There is no point in maintaining a deployment in central namespace while no calls are made.

The goal of this task is to have a solution that deploys Application Gateway only when there is a case that requires it in a given namespace. Moreso - with the new proxy approach it should be deployed once per namespace, not per Application.

**Acceptance Criteria**
- Application Gateway is deployed within customer's namespaces
- Application Gateway is only deployed in namespaces containing active Service Binding Usages

**Subtasks**
- [x] ServiceInstance based controller - @Maladie 
- [x] Gateway deployment - @Maladie 
- [x] Add feature toggle - @Maladie 
- [x] Application Operator tests refactoring - @Maladie 
- [x] Documentation updated - @majakurcius 

Descoped: 
- [ ] Application Operator refactoring cleanup - @Maladie 
- [ ] Migration guide created

PR #7149, opened 2020-02-07
Add Application Mapping Controller for single Gateway per namespace

Maladie:
**Description**

- Added Application Mapping controller in Application Operator
- Added fake Gateway Deployer

**Related issue(s)**

#7043


components/application-operator/pkg/application-mapping-controller/reconciler.go
```diff
@@ -33,17 +32,21 @@ type ApplicationMappingManagerClient interface {
...
 	return &appMappingReconciler{
 		appConnClient:   appConnClient,
 		gatewayDeployer: gatewayDeployer,
+		logger:          logger,
 	}
 }
 
 func (r *appMappingReconciler) Reconcile(request reconcile.Request) (reconcile.Result, error) {
-	if request.Namespace == kymaIntegrationNamespace || request.Namespace == kymaSystemNamespace {
+	log := r.logger.WithField("namespace", request.Namespace)
```
Szymongib: I would suggest using something like that `r.logger.WithField("application-mapping", request.NamespacedName)`

PR #7154, opened 2020-02-07
Bump compass chart

ebensom:
**Description**
Bump compass chart

- Sync to latest compass chart from kyma-incubator/compass

**Related issue(s)**
N/A


resources/compass/charts/gateway/values.yaml
```diff
@@ -6,4 +6,5 @@ deployment:
     allowPrivilegeEscalation: false
 
 gateway:
-  enabled: false
+  enabled: true
```
pkosiec: We should have it disabled here, as we assume that user installs `core` of Kyma, and, in the result, also Kyma Istio gateway. Two Istio Gateways for the same host makes the tests fail.

resources/compass/charts/kyma-environment-broker/templates/deployment.yaml
```diff
@@ -44,10 +44,6 @@ spec:
               value: "{{ .Values.enablePlans }}"
             - name: APP_PROVISIONING_URL
               value: "{{ .Values.provisioner.URL }}"
-            - name: APP_PROVISIONING_SECRET_NAME
```
pkosiec: Are the values not needed anymore? Or they should be defined also on `compass` repo? @polskikiel @mszostok @piotrmiskiewicz @jasiu001

resources/compass/charts/kyma-environment-broker/values.yaml
```diff
@@ -36,14 +37,14 @@ provisioner:
   URL: "http://compass-provisioner.compass-system.svc.cluster.local:3000/graphql"
   gardener:
     # name of the secret with kubeconfig to the gardener cluster
-    secretName: "gardener"
+    secretName: "gopher-canary-gardener"
```
pkosiec: should we really have `gopher` in values? 😛  @polskikiel @mszostok @piotrmiskiewicz @jasiu001

resources/compass/values.yaml
```diff
@@ -143,18 +134,24 @@ global:
...
+
   istio:
     gateway:
-      name: "kyma-gateway"
-      namespace: "kyma-system"
+      name: compass-istio-gateway
```
pkosiec: By default we should have `kyma-gateway here`

resources/compass/values.yaml
```diff
@@ -143,18 +134,24 @@ global:
...
+      namespace: compass-system
 
   database:
+    manageSecrets: true
```
pkosiec: 🤔 Should we really have `manageSecrets` turned on by default? Let's say someone takes kyma from the repo and tries to install it along with compass. All secrets will be missing from default installation, right? I think we should do this with overrides for our internal pipeline.

The same issue which I didn't think of previously is on `compass ` repo.

resources/compass/values.yaml
```diff
@@ -143,18 +134,24 @@ global:
...
       host: adapter-gateway-mtls
-  
+
   rewriteFilters:
     workloadLabel: oathkeeper
     tokenDataHeader: "Connector-Token"
     certificateDataHeader: "Certificate-Data"
 
+  ingress:
+    domainName: kyma.local
```
pkosiec: I wonder if we should keep it, as these values are not specified for all Kyma components. We have override for local installation for our `kyma.local` domain. By removing them from here we enforce users to specify their own configuration for the domain.

resources/compass/charts/provisioner/values.yaml
```diff
@@ -9,7 +9,14 @@ deployment:
 database:
   schemaFilePath: "/assets/database/provisioner.sql"
 
-hydraURL: "http://ory-hydra-public.kyma-system:4444/oauth2/token"
+gardener:
+  project: "" # Gardener project connected to SA
+  kubeconfigPath: "/gardener/kubeconfig/kubeconfig"
+  kubeconfig: "" # Base64 encoded Gardener SA key
+  secretName: "gardener-credentials"
+  manageSecrets: true
```
pkosiec: As in the https://github.com/kyma-project/kyma/pull/7154, I'm thinking about the `manageSecrets: true` value. I know, previously the approach has been approved, but should we really have it turned on by default? It prevents installing Compass for anyone without providing secrets manually before installation. Could you please toggle it in all places, and create override for our internal pipeline?

resources/compass/charts/kyma-environment-broker/templates/deployment.yaml
ebensom: FYI This were removed on purpose by @Szymongib and @polskikiel by https://github.com/kyma-incubator/compass/commit/a69b051b9fef360d63f6e44c6f7ecb8f7a29b5aa#diff-3e1525df3ec2118e9e767b61d0d9aefa

resources/compass/values.yaml
ebensom: I tried to keep the default bahavior here, so `manageSecrets: true` actually means let the helm chart "manage" the DB secrets by default.

resources/compass/values.yaml
ebensom: Will remove the whole `ingress: ` block

resources/compass/values.yaml
ebensom: Will revert these values back.

resources/compass/charts/gateway/values.yaml
ebensom: Changing it back.

resources/compass/charts/provisioner/values.yaml
ebensom: I think the installation would succeed without providing the gardener kubeconfig , but the generated secret would have empty data, making the provisioning though gardener functionality non-functioning. 
Anyways I'm okay with changing the default behavior to `false`, but then we need to guard all the references to the secrets with `{{- if }} ` to avoid K8S throwing errors due to nonexistent Secret resources.

resources/compass/values.yaml
ebensom: Also, for the default kyma installation with compass comes with the embedded DB enabled, so the managed Cloud SQL secret will not play a role here.

resources/compass/charts/provisioner/values.yaml
ebensom: I think by default the KEB and provisioner subcharts are guarded by the global.provisioning.enabled value, which is false by default. If someone enables it, the default behavior now is to supply credentials for gardener.

resources/compass/values.yaml
pkosiec: Ah, okay, I get it. Makes sense 👍

resources/compass/charts/provisioner/values.yaml
pkosiec: I mean I think the secrets should be created by default.

I've thought incorrectly that `manageSecrets: true` means that the secrets are not created - but they are. Which is totally fine. Sorry for making confusion.

ebensom:
retest roulette, spin again!

ebensom:
@pkosiec Can I get an approve? Should something still be addressed?

resources/compass/charts/cockpit/OWNERS
```diff
@@ -3,5 +3,8 @@ reviewers:
 
 approvers:
   - kwiatekus
+  - dariadomagala
   - y-kkamil
-  - sjanota
```
kwiatekus: Please remove y-kkamil

ebensom: Ok, but please let me merge this one and quickly open a new PR for that, I would like to avoid another round of retest roulette.

PR #7165, opened 2020-02-10
Restrict monitoring for kyma namespaces

hisarbalik:


hisarbalik:
> I see that prometheus operator also has a [namespace selector](https://github.com/kyma-project/kyma/blob/a61917837d22fa6a40912e7e6d5a9d15c9bbb1e5/resources/monitoring/values.yaml#L1044). Aren't we supposed to set that as well?

This configuration parameter not used in current version 0.34, for this one we have to upgrade to version 0.35

Issue #5531, opened 2019-09-03
Provide an in-cluster container registry for hosting function images


marcobebway:
**Description**

Provide an in-cluster container registry for hosting function images as per the function controller [proposal](https://github.com/kyma-project/community/pull/314).

nachtmaar:
Blocked for now until @Abd4llA gives green light that the feature will make it into Kyma. Some more internal discussions are required.

derberg:
From my experience, and how I see other companies, I think we should integrate in-cluster container registry and leave it up to the administrator to decide if he wants to use it and install it or if he prefers to use the existing registry outside the cluster. Why?
- We have to consider users that want to use Kyma internally only where their whole infrastructure, including docker registry, is hidden behind a firewall and it is not so common to use public registries. I assume, not basing on any data, that many enterprises are very restrictive here
- This in-cluster registry would be available only internally inside the cluster and used only for Knative. That would be a secure solution, and also much faster than pushing to the external registry

Best would be to have it as a component, that if you install, is out of the box configured with Knative.

tricky42:
I second the request for a local, in-cluster docker registry. Since we started Kyma we always emphasised the principal of **Batteries-Included**. From me this is a perfect example. 

We need something, which is out-of-the box available and can be used without any additional configuration step necessary. At the same time we need to provide configuration examples, how external docker registries can be used in a secure way and include in this example how the local registry would not have to be installed in this case.

PR #7175, opened 2020-02-11
Add docker-registry subchart for function-controller

aerfio:
**Description**

- add subchart now to have easier review process later
- this is PURE, UNCHANGED registry subchart, just freshly copied from helm repo, will be tweaked later

**Related issue(s)**

See also #5531


resources/function-controller/charts/docker-registry/OWNERS
```diff
@@ -0,0 +1,6 @@
+approvers:
```
magicmatatjahu: Please remove that file

PR #7179, opened 2020-02-11
Generate JWKS for ory oathkeeper

piotrmsc:
**Description**
Rotate oathkeeper keys for id_token mutator instead of using a single one, hardcoded.

- Added Job creating keys if not provided by user
- Added cron job responsible for rotating keys

**Related issue(s)**


resources/ory/values.yaml
```diff
@@ -80,26 +80,9 @@ hydra:
     globalAccess: true
 # Overrides for the Oathkeeper chart
 oathkeeper:
+  job:
```
Demonsthere: Maybe change the name to something less generic? `certJob`?

resources/ory/charts/oathkeeper/templates/cronjob.yaml
```diff
@@ -0,0 +1,79 @@
...
+                --from-file=/etc/secrets/jwks.json \
+                -n kyma-system -o yaml --dry-run | kubectl apply -f -
+            volumeMounts:
+            - name: {{ include "oathkeeper.name" . }}-keys-volume
+              mountPath: /etc/secrets
+              readOnly: true
+          volumes:
+          - name: {{ include "oathkeeper.name" . }}-keys-volume
+            emptyDir: {}
+          restartPolicy: OnFailure
```
Demonsthere: You shouldn't really restart jobs, better to create new jobs when it fails. `restartPolicy: Never`

resources/ory/charts/oathkeeper/templates/cronjob.yaml
```diff
@@ -0,0 +1,79 @@
...
+  successfulJobsHistoryLimit: 1
+  jobTemplate:
+    spec:
+      template:
+        metadata:
+          annotations:
+            sidecar.istio.io/inject: "false"
+        spec:
+          containers:
+          - name: keys-generator
```
Demonsthere: This could be an init container, if we need it to go first

resources/ory/charts/oathkeeper/templates/cronjob.yaml
piotrmsc: This is cron job, if it fails to do the job we have to restart it, otherwise it will not rotate keys at all

resources/ory/charts/oathkeeper/templates/cronjob.yaml
```diff
@@ -0,0 +1,80 @@
...
+  kind: Role
+  name: ory-oathkeeper-keys-job-role
+  apiGroup: rbac.authorization.k8s.io
+---
+apiVersion: batch/v1beta1
+kind: CronJob
+metadata:
+  name: oathkeeper-jwks-rotator
+spec:
+  schedule: "0 0 * */1 *"
```
Demonsthere: Default option for a cronjob is `concurrencyPolicy: Allow`. We may want to have it set to `Forbid`

PR #7197, opened 2020-02-12
Log cluster metrics in Runtime Agent

franpog859:
**Description**

- Log cluster metrics in Runtime Agent so FluentBit can pass them further

**Related issue(s)**

_none (!)_


components/compass-runtime-agent/cmd/main.go
```diff
@@ -114,10 +113,40 @@ func main() {
...
+	metricsLogger, err := newMetricsLogger(options.MetricsLoggingTimeInterval)
+	if err != nil {
+		log.Error(errors.Wrap(err, "Unable to create metrics logger"))
 	}
+
+	runManagerAndLoggerConcurrently(mgr, metricsLogger)
+	os.Exit(1)
+}
+
+func runManagerAndLoggerConcurrently(manager manager.Manager, logger metrics.Logger) {
```
koala7659: I suppose it is possible that logger maybe passed as nil here - after newMetricsLogger function failed (for example when creation of Resource fetcher has failed).  In such a case logger.Log(quitChannel) (line 146) will crash

franpog859: Good catch! Thanks, I'll fix it ;>

Issue #7042, opened 2020-01-30
Application Gateway - refactoring of proxy approach


janmedrek:
**Description**

The current state of Application Gateway is not ideal - it is deployed in a system namespace, scaling is problematic and resource usage (Kubernetes services, pod's memory, and CPU) can be improved. With a few changes, we can achieve a state in which the workflow is simple, easy to understand for both users and developers and not depending on the resources contained in a single namespace.

The idea is to create a generic solution that would not strongly rely on Application CR. Information such as auth method or credentials to use, headers, query parameters or target URL can be passed along the call, ideally in path (so that internal Kyma URL that is called defines the way the call is proxied). This can be implemented in many ways, the proposed solution is to use path segments for each piece of information, or by pointing Application Gateway to corresponding binding / configmap.

With this approach, a single Gateway can handle many applications and services, there is an additional security measure that needs to be taken into consideration - we need a way to confirm that a request originating from given lambda or service has the privilege to call target URL.

**Acceptance Criteria**
- Application Gateway no longer uses Access Services
- Auth method, credentials and other required information are not tied to the Application entry, those should be passed with the call (multiple possible approaches, should be discussed and aligned with other components)
- Host verification implemented (check whether specific origin has the access to the target)
- Documentation, tutorials and links updated

**Subtasks**
- [x] Proxy refactoring - @Szymongib 
- [x] Gateway tests refactoring - @Szymongib 
- [x] Use the feature flag
- [x] Documentation updated

PR #7224, opened 2020-02-17
Application Operator deploys Gateway once per namespace

Maladie:
**Description**

- Application Mapping controller changed to Service Instance Controller
- Deploying gateways once per namespace implemented.

**Related issue(s)**

#7042


components/application-operator/cmd/manager/manager.go
```diff
@@ -55,14 +61,21 @@ func main() {
...
+
+	helmClient, err := kymahelm.NewClient(options.tillerUrl, options.helmTLSKeyFile, options.helmTLSCertificateFile, options.tillerTLSSkipVerify, options.installationTimeout)
 	if err != nil {
 		log.Fatal(err)
 	}
 
 	log.Printf("Preparing Release Manager.")
 
-	releaseManager, err := newReleaseManager(options, cfg)
+	releaseManager, err := newReleaseManager(options, cfg, helmClient)
```
Szymongib: The name `ReleaseManager` is now confusing because Gateway is also deployed with Helm so it also has releases. I would suggest renaming it to `ApplicationReleaseManager`.

components/application-operator/cmd/manager/manager.go
```diff
@@ -74,16 +87,31 @@ func main() {
...
+
+	gatewayManager, err := newGatewayManager(options, cfg, helmClient)
+
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	log.Printf("Upgrading gateways")
+
+	err = gatewayManager.UpgradeGateways()
```
Szymongib: We now call `releaseManager.UpgradeReleases()` and  `gatewayManager.UpgradeGateways()`. I think that `UpgradeReleases` method should be renamed to `UpgradeApplicationReleases` becase Gateways will also be releases.

components/application-operator/pkg/kymahelm/gateway/gatewaymanager.go
```diff
@@ -0,0 +1,199 @@
...
+
+	if len(namespaces) == 0 {
+		return nil
+	}
+
+	err = g.updateGateways(namespaces)
+
+	if err != nil {
+		return errors.Errorf("Error updating Gateway: %s", err.Error())
+	}
```
Szymongib: If we for some reason cannot update the Gateway release, we end up in the crash loop and the whole Application Operator is not working. I do not think it is the desired behavior and we should just log and error and proceed as we do in case of Applications.

components/application-operator/pkg/kymahelm/gateway/gatewaymanager.go
```diff
@@ -0,0 +1,199 @@
...
+	namespacesList, err := g.namespaces.List(metav1.ListOptions{})
+
+	if err != nil {
+		return nil, errors.Errorf("Error listing namespaces: %s", err.Error())
+	}
+
+	var namespaces []string
+
+	for _, namespace := range namespacesList.Items {
+		instances := g.serviceCatalogueClient.ServiceInstances(namespace.Name)
```
Szymongib: I wonder if we can just pass `""` as a namespace to `ServiceInstances("")`. In core Kubernetes client we can then operate on resources from all namespaces. That would simplify this code.

resources/application-connector/values.yaml
```diff
@@ -49,7 +49,7 @@ global:
     version: 83cc9c71
   application_gateway:
     dir:
-    version: 97ee9c85
+    version: c608dbf3
```
Szymongib: Why do we use the older version?

PR #7238, opened 2020-02-17
Fix gardener tls rotation

piotrmsc:
**Description**

- Provide dex cert for iam-kubeconfig, apiserver-proxy, cbs if not running on Gardener
- Request a separate certificate for apiserver-proxy if running on Gardener
- Allow TLS cert rotation for kyma running on Gardener
- Unify application-connector Gateway to use kyma-gateway-certs  TLS secret
- Provide additional secret with cacert for application connector, adjust helper jobs that creates cacert
- Adjust compass charts (values) to work with changed secrets
**Related issue(s)**


resources/core/charts/gateway/templates/application-connector-certs.yaml
```diff
@@ -1,8 +1,10 @@
 ---
 apiVersion: v1
 data:
```
piotrmsc: Note to self : we need to check if we can reuse kyma secret or issue new cert from Gardener directly on gateway.$DOMAIN

resources/core/charts/apiserver-proxy/templates/deployment.yaml
```diff
@@ -53,7 +53,9 @@ spec:
         - --v=10
         - --oidc-issuer=https://dex.{{ .Values.global.ingress.domainName }}
         - --oidc-clientID=kyma-client
+        {{ if not .Values.global.environment.gardener }}
```
Demonsthere: Suggestion: If we are doing multiple ifs that all depend on the same condition (isGardener here), it may be better to have 2 depoyment files (deployment and deployment-gardener) and pick which one we use.

resources/core/charts/apiserver-proxy/templates/generate-certs-job.yaml
```diff
@@ -29,9 +29,37 @@ spec:
         - -c
```
Demonsthere: This script is quite long, maybe it would be better to move it as a file and include it as a file into the yaml? We can always do `| tpl` to keep the helm variables

resources/core/charts/console-backend-service/templates/deployment.yaml
```diff
@@ -80,8 +80,10 @@ spec:
             value: "https://dex.{{ .Values.global.ingress.domainName }}"
```
Demonsthere: Suggestion: If we are doing multiple ifs that all depend on the same condition (isGardener here), it may be better to have 2 depoyment files (deployment and deployment-gardener) and pick which one we use.

resources/core/charts/iam-kubeconfig-service/templates/deployment.yaml
```diff
@@ -32,7 +32,11 @@ spec:
         command:
```
Demonsthere: Suggestion: If we are doing multiple ifs that all depend on the same condition (isGardener here), it may be better to have 2 depoyment files (deployment and deployment-gardener) and pick which one we use.

components/xip-patch/xip-patch.sh
```diff
@@ -56,12 +58,23 @@ generateCerts() {
...
 
 createOverridesConfigMap() {
-    if [ -z "$(kubectl get configmap -n kyma-installer net-global-overrides --ignore-not-found)" ]; then
+    if [[ "$GARDENER_ENVIRONMENT" == true ]]; then
         kubectl create configmap net-global-overrides \
-            --from-literal global.ingress.domainName="$INGRESS_DOMAIN" \
-            --from-literal global.ingress.tlsCrt="$INGRESS_TLS_CERT" \
-            --from-literal global.ingress.tlsKey="$INGRESS_TLS_KEY" \
-            -n kyma-installer
+        --from-literal global.ingress.domainName="$INGRESS_DOMAIN" \
```
Tomasz-Smelcerz-SAP: We can shorten this a bit by extracting common parameters to a variable and then execute like this: `kubectl create configmap --from-literal someKey=someValue $COMMON_PARAMS`

resources/core/charts/apiserver-proxy/files/generate-certs.sh
```diff
@@ -0,0 +1,56 @@
...
+  END_TIME=$((SECONDS+600)) #600 seconds = 10 minutes
+  while [ ${SECONDS} -lt ${END_TIME} ];do
+    STATUS="$(kubectl get -n {{ .Release.Namespace }} certificate.cert.gardener.cloud {{ template "name" . }}-tls-cert -o jsonpath='{.status.state}')"
+    if [ "${STATUS}" = "Ready" ]; then
+      break
+    fi
+    echo "Waiting for Certicate generation, status is ${STATUS}"
+    sleep 10
+  done
+
```
piotrmsc: We need to check properly if it's ready. After 10 mins it can be still in error state, then it try process but should exit instead

Demonsthere: A check here is needed to see if the cert was created after the waiting loop. Currently, it can continue without a propper cert

resources/core/charts/apiserver-proxy/files/generate-certs.sh
```diff
@@ -0,0 +1,59 @@
...
+    if [ "${STATUS}" = "Ready" ]; then
+      break
+    fi
+    echo "Waiting for Certicate generation, status is ${STATUS}"
+    sleep 10
+  done
+  if [ "${STATUS}" != "Ready" ]; then
+    echo "Certificate is still not ready, status is ${STATUS}. Exiting.."
+    exit 1
+  fi
```
Demonsthere:
```suggestion
  DONE=false
  SECONDS=0
  END_TIME=$((SECONDS+600)) #600 seconds = 10 minutes
  while [ ${SECONDS} -lt ${END_TIME} ];do
    STATUS="$(kubectl get -n {{ .Release.Namespace }} certificate.cert.gardener.cloud {{ template "name" . }}-tls-cert -o jsonpath='{.status.state}')"
    if [ "${STATUS}" = "Ready" ]; then
      DONE=true
      break
    fi
    echo "Waiting for Certicate generation, status is ${STATUS}"
    sleep 10
  done
  if [ ! ${DONE} ]; then
    echo "Certificate is still not ready, status is ${STATUS}. Exiting.."
    exit 1
  fi
```

PR #7246, opened 2020-02-18
Extend Gateway to work in namespace only mode

Szymongib:
**Description**

Application Gateway will be deployed once per namespace instead of once per every Application. As a result, it needs to work in a different way - not relying on Application CR.

Namespaced Gateway will read data from the Secret provided in the path in the format:
```
/secret/{SECRET_NAME}/api/{API_NAME}
```

- Feature flag for Gateway in Namespaced mode introduced
- Reading proxy configuration from secrets implemented

**Related issue(s)**
#7042


components/application-gateway/pkg/proxy/provider/secrets_provider.go
```diff
@@ -0,0 +1,115 @@
+package provider
```
akgalwas: I'm wondering whether the package names are descriptive enough. Maybe something like : /pkg/proxyconfig/provider would be better.

Issue #7249, opened 2020-02-18
Override returns an error for istio-kyma-patch Pod 


bszwarc:
**Description**
After adding an override for Istio settings, the installation fails,  `istio-kyma-patch` pod returns an error.

Tested locally on Minikube.
I applied the following override to change the value for trace sampling. The override was added at runtime with the Kyma installation update triggered afterwards.

```
apiVersion: v1
kind: ConfigMap
metadata:
  name: istio-overrides
  namespace: kyma-installer
  labels:
    installer: overrides
    component: istio
    kyma-project.io/installation: ""
data:
  kyma_istio_control_plane: |-
    apiVersion: install.istio.io/v1alpha2
    kind: IstioControlPlane
    spec:
      trafficManagement:
        components:
          pilot:
            enabled: true
            k8s:
              env:
              - name: PILOT_TRACE_SAMPLING
                value: "60"
```

**Expected result**

Override is applied, Kyma updated, value is changed. 

**Actual result**

istio-kyma-patch pod returns an error:

```
istio-kyma-patch-56gwq                   0/1     Error     0          2m15s
istio-kyma-patch-sdjpj                   0/1     Error     0          114s
```
```
mTLS must be "STRICT"
```

**Steps to reproduce**

1. Create an override + apply it
2. Trigger installation update

PR #7258, opened 2020-02-19
Bump istio-installer image

Demonsthere:
**Description**

- Add YQ merging step
- Bump installer image

**Related issue(s)**

- Fix to #7249

### DIFF
--- a/resources/istio/files/job-script.sh
+++ b/resources/istio/files/job-script.sh
@@ -2,6 +2,11 @@
 if [ -f "/etc/istio/overrides.yaml" ]; then
   #New way: just merge default IstioControlPlane definition with a user-provided one.
   yq merge -a -x /etc/istio/config.yaml /etc/istio/overrides.yaml > /etc/combo.yaml
+  kubectl create cm "${CONFIGMAP_NAME}" -n "${NAMESPACE}" \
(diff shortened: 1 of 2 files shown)


resources/istio/files/job-script.sh
```diff
@@ -1,8 +1,14 @@
-
+#!/bin/bash -e
 if [ -f "/etc/istio/overrides.yaml" ]; then
   #New way: just merge default IstioControlPlane definition with a user-provided one.
-  printf "istioctl manifest apply -f /etc/istio/config.yaml -f /etc/istio/overrides.yaml\n"
-  istioctl manifest apply -f /etc/istio/config.yaml -f /etc/istio/overrides.yaml
+  yq merge -a -x /etc/istio/config.yaml /etc/istio/overrides.yaml > /etc/combo.yaml
+  kubectl create cm "${CONFIGMAP_NAME}" -n "${NAMESPACE}" \
```
piotrmsc: Won't this break during the upgrade in the future? `kubectl create` returns error like `configmap kyma-istio-control-plane-config` already exists. If we want to keep it up to date we should do sth like `kubectl create cm ..... --dry-run -o yaml | kubectl apply -f -`

Issue #7235, opened 2020-02-17
event-bus chart pre-upgrade job for migrating a user namespace to knative eventing mesh


Abd4llA:
**Description**
As per kyma-project/community#434 proposal, Kyma 1.11 needs to have a job that syncs user namespace with Knative Eventing mesh by doing the following :

- Recreates `ServiceInstance` objects.
- Creates `Trigger` objects for each existing Kyma Subscription object.

The job will be a `pre-upgrade` hook in `event-bus` chart.

**AC**
- `event-bus` chart has a `pre-upgrade` hook which syncs the user namespace to Knative Eventing mesh.

marcobebway:
### The following migration issues are fixed:

1. The migration job was running more than once. The fix for this is to add `restartPolicy: Never` to the job spec.

  ```bash
  $ kubectl get po -

  NAMESPACE     NAME                                                              READY   STATUS             RESTARTS   AGE
  kyma-system   event-bus-mesh-namespaces-migration-mesh-namespaces-migrat6zx44   0/1     Pending             0          0s
  kyma-system   event-bus-mesh-namespaces-migration-mesh-namespaces-migrat6zx44   0/1     Pending             0          0s
  kyma-system   event-bus-mesh-namespaces-migration-mesh-namespaces-migrat6zx44   0/1     ContainerCreating   0          0s
  kyma-system   event-bus-mesh-namespaces-migration-mesh-namespaces-migrat6zx44   1/1     Running             0          3s
  kyma-system   event-bus-mesh-namespaces-migration-mesh-namespaces-migrat6zx44   0/1     Completed           0          12s
  kyma-system   event-bus-mesh-namespaces-migration-mesh-namespaces-migrat6zx44   0/1     Terminating         0          20s
  kyma-system   event-bus-mesh-namespaces-migration-mesh-namespaces-migrat6zx44   0/1     Terminating         0          20s
  ```

2. The script was deleting service instances which are not related to the Application-Broker and not related to eventing. The fix for this is to filter out the unrelated service instances before doing the migration.

  ```bash
  $ kubectl get serviceinstances.servicecatalog.k8s.io -A -w

  NAMESPACE                      NAME               CLASS                           PLAN     STATUS  AGE
  applicationbrokerupgradetest   api-service-id     ServiceClass/api-service-id     default  Ready   14m << (untouched - not eventing related)
  applicationbrokerupgradetest   events-service-id  ServiceClass/events-service-id  default  Ready   46s
[...]

PR #7264, opened 2020-02-19
event-bus chart pre-upgrade job for migrating a user namespace to knative eventing mesh

montaro:
See also #7235

### DIFF
--- a/components/event-bus/cmd/mesh-namespaces-migration/Dockerfile
+++ b/components/event-bus/cmd/mesh-namespaces-migration/Dockerfile
@@ -1,0 +1,4 @@
+FROM curlimages/curl:7.68.0
+COPY mesh-namespaces-migration /
+COPY licenses ./licenses
+ENTRYPOINT ["/mesh-namespaces-migration"]
--- a/components/event-bus/cmd/mesh-namespaces-migration/subscription.go
+++ b/components/event-bus/cmd/mesh-namespaces-migration/subscription.go
@@ -1,0 +1,233 @@
+package main
+
+import (
+	"fmt"
+	"log"
+	"time"
+
+	"github.com/google/uuid"
+	"github.com/pkg/errors"
+
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/util/wait"
+
+	kneventingv1alpha1 "knative.dev/eventing/pkg/apis/eventing/v1alpha1"
+	kneventingclientset "knative.dev/eventing/pkg/client/clientset/versioned"
+	"knative.dev/pkg/apis"
+	knv1alpha1 "knative.dev/pkg/apis/v1alpha1"
+
+	kymaeventingv1alpha1 "github.com/kyma-project/kyma/components/event-bus/apis/eventing/v1alpha1"
+	kymaeventingclientset "github.com/kyma-project/kyma/components/event-bus/client/generated/clientset/internalclientset"
+)
+
+// CloudEvents attributes
+const (
+	sourceAttr           = "source"
+	eventTypeAttr        = "type"
+	eventTypeVersionAttr = "eventtypeversion"
+)
+
+const (
+	defaultBrokerName = "default"
+	brokerLabel       = "eventing.knative.dev/broker"
+)
+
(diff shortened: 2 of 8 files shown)


components/event-bus/cmd/mesh-namespaces-migration/Dockerfile
```diff
@@ -0,0 +1,4 @@
+FROM curlimages/curl:7.66.0
+COPY mesh-namespaces-migration /
+#COPY licenses ./licenses
```
montaro: will fix

components/event-bus/cmd/mesh-namespaces-migration/Dockerfile
```diff
@@ -0,0 +1,4 @@
+FROM curlimages/curl:7.66.0
```
anishj0shi: can we update the curl image, as per this: CxServer\SP\SAP\OpenSource

antoineco: @anishj0shi could you please indicate the full image name ?

components/event-bus/cmd/mesh-namespaces-migration/subscription.go
```diff
@@ -0,0 +1,236 @@
...
+	}
+
+	return m, nil
+}
+
+// populateSubscription populates the local list of Kyma Subscriptions.
+func (m *subscriptionMigrator) populateSubscriptions(namespaces []string) error {
+	var kymaSubscriptions []kymaeventingv1alpha1.Subscription
+
+	for _, ns := range namespaces {
```
anishj0shi: rather then iterating through namespaces and finding kyma-subscriptions, does it makes sense to use SharedInformerFactory.Listers() this might give you cluster wide resources, right?

components/event-bus/cmd/mesh-namespaces-migration/subscription.go
```diff
@@ -0,0 +1,236 @@
...
+	m.subscriptions = kymaSubscriptions
+
+	return nil
+}
+
+// populateTriggers populates the local triggersByNamespace map.
+func (m *subscriptionMigrator) populateTriggers(namespaces []string) error {
+	triggersByNamespace := make(triggersByNamespaceMap)
+
+	for _, ns := range namespaces {
```
anishj0shi: I'd really recommend using SharedInformerFactory.Lister() to get knative resources, to avoid nested loops.

components/event-bus/cmd/mesh-namespaces-migration/Dockerfile
anishj0shi: my bad, i thought the link was updated.. https://nvd.nist.gov/vuln/detail/CVE-2019-15601

antoineco: That's the CVE, what about the image name?

components/event-bus/cmd/mesh-namespaces-migration/subscription.go
antoineco: An informer does the same: it calls List() and puts the results in a local cache. The difference is that it keeps watching for changes afterwards, which we don't need. I think an Informer is not what you want to use here.

components/event-bus/cmd/mesh-namespaces-migration/Dockerfile
antoineco: I meant: what image should we use?

components/event-bus/cmd/mesh-namespaces-migration/subscription.go
antoineco: The example you mention is in the context of a long running program. An Informer is not what you would want to use here. See previous comment.

components/event-bus/cmd/mesh-namespaces-migration/Dockerfile
anishj0shi: image version: 7.68

antoineco: [done](https://github.com/kyma-project/kyma/compare/9f3745fbd22aa842cbb6307d4b7f138983bab958..5d7bc22aa931c741d78725979e7a7575c63413d2)

components/event-bus/cmd/mesh-namespaces-migration/subscription.go
montaro: @anishj0shi  Thanks for the feedback
@antoineco WHat do you think?

PR #7301, opened 2020-02-21
Namespaces resources migration from event bus to event-mesh pre-upgrade job

montaro:
Migrate the namespaces resources from event-bus to event-mesh

**Related issue(s)**
See also #7235


resources/compass-runtime-agent/values.yaml
```diff
@@ -10,7 +10,7 @@ global:
       version: "cfcfa726"
     compassExternalSolutionTests:
       dir:
-      version: 46b1e3dc
+      version: d4716e58
```
sayanh: Let's add the PR image(PR-7500) here to test out the upgrade job.

Issue #7036, opened 2020-01-30
Reduce prometheus memory footprint by reduced labelling and defensive querying


a-thaler:
**Description**
For prometheus the label [cardinality is key](https://www.robustperception.io/cardinality-is-key). As in kyma the default monitoring is just scraping any endpoint blindly and is labelling all metrics by default, the whole setup is not manageable (very exhaustive) from a memory perspective.

To successfully manage the memory footprint, endpoints must be scraped more conscious and queries should be optimized and cutted if bad behaving. 

Actions:
- Only metrics are getting scraped by prometheus which are in use by a dashboard/record/alertRule/kiali/jaeger
- Only labels are getting applied to metrics which are needed, especially check the default labelling introduced by the prometheus-operator
- Assure that complex queries are stopped at some point in time as they are too complex
- Check existing queries if there are complex ones hitting the new border and improve them.

**Reasons**
Reduce the memory footprint to make kyma more lightweight, have a more predictable memory consumption over time

**Attachments**

hisarbalik:
Storage usage before metric re-labeling, screenshoot below show time series being scraped before re-labeling, currently promethues uses over 100K time series
![Screen Shot 2020-02-21 at 10 40 02](https://user-images.githubusercontent.com/6337703/75022558-9469a300-5496-11ea-9f92-68836e70a1f1.png)

Storage usage after metric re-labeling, screenshoot below show time series being scraped after re-labeling, currently promethues uses around 50K time series
![Screen Shot 2020-02-21 at 10 40 42](https://user-images.githubusercontent.com/6337703/75022587-a21f2880-5496-11ea-8ff5-d156912c0c1b.png)

Amount of time series will be dramatically reduced after reducing label collected and attached scraped metric, current result shows only metric re-labeling

hisarbalik:
A dedicated kyma cluster with new configuration deployed on https://grafana.monitoring-reduced.cluster.extend.cx.cloud.sap/?orgId=1 and available for reviewing changes.

Please check all dashboards and alert-rules related to your component and ensure they are still working as expected. In case you miss some metrics please let us know with the component/servicce-monitor name so we can add missing metrics to the new configuration

PR #7309, opened 2020-02-24
Reduce prometheus memory footprint by reduced labelling and defensive querying

hisarbalik:
#7036


resources/event-bus/charts/event-publish-service/templates/service-monitor.yaml
```diff
@@ -16,5 +16,9 @@ spec:
 {{- if .Values.monitoring.scrapeInterval }}
     interval: {{ .Values.monitoring.scrapeInterval }}
 {{- end }}
+    metricRelabelings:
+    - sourceLabels: [ __name__ ]
+      regex: ^(go_goroutines|go_memstats_alloc_bytes|go_memstats_heap_alloc_bytes|go_memstats_heap_inuse_bytes|go_memstats_heap_sys_bytes|go_memstats_stack_inuse_bytes|process_cpu_seconds_total|process_max_fds|process_open_fds|process_resident_memory_bytes|process_start_time_seconds|process_virtual_memory_bytes)$
```
sayanh: `kyma-event-bus` dashboard does not have any data. It uses `total_published_messages` metrics. Do we need to include this here?

jasiu001:
Some metrics for HelmBroker are missing, GC duration shouldn't be empty. 
Below is a comparison with the `weekly` cluster:

Screen Shot 2020-02-27 at 13 40 57
Screen Shot 2020-02-27 at 13 41 12

The same situation is for BindingUsageComponent. (as an above comparison with `weekly`)

Screen Shot 2020-02-27 at 14 16 57
Screen Shot 2020-02-27 at 14 18 42

hisarbalik:
> What `my-kyma.yaml` is used for?

File removed, was mistakenly added

hisarbalik:
> Currently, metrics for Application Registry and Connector Service do not work here and on the master branch. There is already [an issue for it](https://github.com/kyma-project/kyma/issues/7478)
> I'll approve this PR from the Application Connector side not to block it. Potential oversights should be fixed as a follow-up

Be aware, after metric re-labeling some labels may be not more available with old names like instead of **pod_name** will be **pod** etc.

sjanota:
There is no memory, network and CPU information for Kyma / WebAPPs.
Screen Shot 2020-03-05 at 15 01 31
They are available on weekly.

hisarbalik:
> There is no memory, network and CPU information for Kyma / WebAPPs.
> Screen Shot 2020-03-05 at 15 01 31
> They are available on weekly.

dashboard updated, please check again

PR #7312, opened 2020-02-24
Apply default serving.knative.dev/visibility label on Knative Service, not on Function CR

magicmatatjahu:
**Description**

- Apply default `serving.knative.dev/visibility` label on Knative Service, not on Function CR:
  - apply suggestion described here: https://github.com/kyma-project/kyma/issues/5582#issuecomment-589958427

**Related issue(s)**
https://github.com/kyma-project/kyma/issues/5582

### DIFF
--- a/components/function-controller/pkg/controllers/function/function_controller.go
+++ b/components/function-controller/pkg/controllers/function/function_controller.go
@@ -23,6 +23,8 @@ import (
 	"reflect"
 	"strings"
 
+	"knative.dev/serving/pkg/reconciler/route/config"
+
 	tektonv1alpha1 "github.com/tektoncd/pipeline/pkg/apis/pipeline/v1alpha1"
 
 	"crypto/sha256"
@@ -55,10 +57,6 @@ var knativeServingAnnotations = []string{
 	servingapis.GroupName + knapis.CreatorAnnotationSuffix,
 	servingapis.GroupName + knapis.UpdaterAnnotationSuffix,
 }
-
-const (
-	kNativeServingVisibilityLabel = "serving.knative.dev/visibility"
-)
 
 // Add creates a new Function Controller and adds it to the Manager with default RBAC. The Manager will set fields on the Controller
 // and Start it when the Manager is Started.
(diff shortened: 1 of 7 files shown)


components/function-controller/pkg/controllers/function/function_controller.go
```diff
@@ -56,6 +56,10 @@ var knativeServingAnnotations = []string{
 	servingapis.GroupName + knapis.UpdaterAnnotationSuffix,
 }
 
+const (
+	kNativeServingVisibilityLabel = "serving.knative.dev/visibility"
```
aerfio: please use vars from knative serving
image

components/function-controller/pkg/webhook/validation_mutation_test.go
```diff
@@ -40,49 +40,14 @@ func TestMutation(t *testing.T) {
...
 	// ensure defaults are set
-	g.Expect(functionWithoutVisibilityLabel.ObjectMeta).To(gstruct.MatchFields(gstruct.IgnoreExtras, gstruct.Fields{
+	g.Expect(function.ObjectMeta).To(gstruct.MatchFields(gstruct.IgnoreExtras, gstruct.Fields{
 		"Name":      gomega.BeEquivalentTo("example-123"),
 		"Namespace": gomega.BeEquivalentTo("ns"),
-		"Labels": gomega.BeEquivalentTo(map[string]string{
-			"serving.knative.dev/visibility": string(serverlessv1alpha1.FunctionVisibilityClusterLocal),
-		}),
 	}))
-	g.Expect(functionWithoutVisibilityLabel.Spec).To(gstruct.MatchFields(gstruct.IgnoreExtras, gstruct.Fields{
```
aerfio: you've deleted quite a lot of tests - could you tell why? if they no longer suit you needs then rewrite that part

magicmatatjahu: Please read variables `functionWithoutVisibilityLabel` and `functionWithVisibilityLabel` :P Those only were nedeed when we had logic for labels in webhooks, not in reconcile.

components/function-controller/pkg/controllers/function/function_controller.go
magicmatatjahu: ok

PR #7317, opened 2020-02-24
Disable istio sidecar injection for fluentbit

zralt:
Installing fluentbit on a PSP enabled k8s cluster doesn't work because of istio (capabilities missing which are needed for istio).
Options are:

Modify the PSP to add NET_ADMIN (and NET_RAW ?) capabilities when you have Istio installed. The problem with this is the more complex configuration, without any added benefits.
Turn off Istio via an override with kyma-installer. This is not an option as the kyma-installer doesn't support annotations with "/" in them.
Just turn off istio for fluentbit, as there's no sane reason to configure it for this service.

### DIFF
--- a/installation/resources/installer-cr-cluster-compass.yaml.tpl
+++ b/installation/resources/installer-cr-cluster-compass.yaml.tpl
@@ -32,3 +32,5 @@ spec:
       namespace: "compass-system"
     - name: "monitoring"
       namespace: "kyma-system"
+    - name: "logging"
+      namespace: "kyma-system"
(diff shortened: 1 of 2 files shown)


zralt:
Installing fluentbit on a PSP enabled k8s cluster doesn't work because of istio (capabilities missing which are needed for istio). 
Options are:
1. Modify the PSP to add NET_ADMIN (and NET_RAW ?) capabilities when you have Istio installed. The problem with this is the more complex configuration, without any added benefits.
2. Turn off Istio via an override with kyma-installer. This is not an option as the kyma-installer doesn't support annotations with "/" in them.
3. Just turn off istio for fluentbit, as there's no sane reason to configure it for this service.

zralt:
@a-thaler updated the description

PR #7328, opened 2020-02-25
Remove SAR checks from apiserver-proxy

### LINKED ISSUE #7057: Authorization check for apiserver proxy takes a lot of time
**Description**

As seen on the image, authN check is super fast in comparison to authZ check. When you use kyma kubeconfig it takes a lot of time to get the result from kubectl command. 
Screenshot 2020-01-30 at 14 47 16
Authorization checks do not have caching, they directly make SAR checks in k8s. 
Do we really need it? IMO we can make a request to apiserver directly after authN. 

**Expected result**

Authorization check are done much faster. 

**Actual result**

**Steps to reproduce**
Fetch kyma kubeconfig and fetch namespaces/deployments any CR

**Troubleshooting**

Demonsthere:
**Description**

- Remove SAR checks from proxy operations
- Remove SAR related grafana dashboards 

**Related issue(s)**

- Fixes #7057

### DIFF
--- a/components/apiserver-proxy/cmd/proxy/main.go
+++ b/components/apiserver-proxy/cmd/proxy/main.go
@@ -178,19 +178,12 @@ func main() {
 		}
 	}
 
-	sarClient := kubeClient.AuthorizationV1beta1().SubjectAccessReviews()
-	authorizer, err := authz.NewAuthorizer(sarClient)
-
-	if err != nil {
-		glog.Fatalf("Failed to create authorizer: %v", err)
-	}
-
 	metrics, err := monitoring.NewProxyMetrics()
 	if err != nil {
 		glog.Fatalf("Failed to create metrics: %v", err)
 	}
 
-	authProxy := proxy.New(cfg.auth, authorizer, oidcAuthenticator, metrics)
+	authProxy := proxy.New(cfg.auth, nil, oidcAuthenticator, metrics)
 
 	if err != nil {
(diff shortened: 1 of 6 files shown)


components/apiserver-proxy/cmd/proxy/main.go
```diff
@@ -178,19 +178,12 @@ func main() {
...
-		glog.Fatalf("Failed to create authorizer: %v", err)
-	}
-
 	metrics, err := monitoring.NewProxyMetrics()
 	if err != nil {
 		glog.Fatalf("Failed to create metrics: %v", err)
 	}
 
-	authProxy := proxy.New(cfg.auth, authorizer, oidcAuthenticator, metrics)
+	authProxy := proxy.New(cfg.auth, nil, oidcAuthenticator, metrics)
```
colunira: I don't think this is the best solution to pass nil here. We should remove this field from the function definition.

resources/core/charts/cluster-users/values.yaml
```diff
@@ -21,6 +21,7 @@ clusterRoles:
       - "servicecatalog.k8s.io"
       - "settings.k8s.io"
       - "metrics.k8s.io"
+      - "coordination.k8s.io "
```
colunira: Why this is added?

Demonsthere: This is a "by-the-way" fix since k8s 1.16 a new group is required to perform `kubectl describe node`

Issue #7201, opened 2020-02-13
Application broker creates one service class for application, and plans per api-package


PK85:
**Description**
Application broker creates one service class for application, and plans per api-package.

**AC**
- make sure Application CR ID is a service class ID. Label that service class that UI will now is a new way how we represent applications in the Service Catalog.
- make sure Application#Service IDs are service plans IDs
- update existing documentation to reflect new design

**Related ticket**
- UI: https://github.com/kyma-project/console/issues/1544

mszostok:
**Alternative plan:**

- Align contract with X-team about API Packages.
- Change Appilcation broker to be a cluster-wide and get rid of ApplicationMappings
- Remove controller for Application CR and replace that with graphql client to Director
- Change mapping and hardcoded things around single API and Events to support multi-entries
- Make sure that proxy implemented: https://github.com/kyma-incubator/compass/issues/752

mszostok:
**New APP CR**

```
apiVersion: applicationconnector.kyma-project.io/v1alpha1
kind: Application
metadata:
  # {{ new requirement 
  #required, MUST only contain lowercase characters, numbers and hyphens (no spaces)
  # MUST be unique across all services. MUST be a non-empty string. !!!!!!!!
  # }}
  name: ec-prod
spec:
  accessLabel: access-label-1
  description: "EC description"

  # {{ move from service
  tags:
    - occ
    - promotions
  labels:
    connected-app: "ec-prod"
  # required, mapped to metadata.longDescription, used by catalog UI
  longDescription: "Long description for Promotions APIs"
  # mandatory, maps to metadata.providerDisplayName, provider name in catalog UI
  providerDisplayName: "HakunaMatata"
  # }}

  # {{ new fields
  displayName: "name"
  id: "foo-bar-123" # unique
  # }}

  services:
    # mandatory, unique ID
    - id: "ac031e8c-9aa4-4cb7-8999-0d358726ffaa"               # required, uuid, immutable
      # required, MUST only contain lowercase characters, numbers and hyphens (no spaces)
      # MUST be unique across all services. MUST be a non-empty string.
      name: "promotions-ac031e8c"
      # required, mapped to metadata.displayName, used by catalog UI
      displayName: "Promotions"
      # required, mapped to description, used by catalog UI
      description: "Short desc for Promotions APIs"

      # {{ new fields
      schemas:
        ServiceInstance: "JSON"
        ServiceBinding: "JSON"
      # }}

[...]

PR #7330, opened 2020-02-25
Add support to api packages in Application Broker and Application CRD

mszostok:
**Description**

Add support to api packages in Application Broker:
- Support for multiple GATEWAY and TARGET url
    - Pattern: `upperCase(<APP_NAME>)_GATEWAY_URL`, `upperCase(<APP_NAME>)_TARGET_URL`
- Remove restriction “provision only once”
- Implemented on feature flag `APP_API_PACKAGES_SUPPORT`
- Remove the OwnerReference from the EventActivation and replace that with removing that on deprovision action. (feature changed in both approaches with api-packages and without them)
- Fix updating the Istio Policy on provision action

**Related issue(s)**

- https://github.com/kyma-project/kyma/issues/7201


mszostok:
here is a task in which the old impl should be removed: https://github.com/kyma-project/kyma/issues/7415

the `TODO` statement in source code will be updated after review

components/application-broker/internal/syncer/validator.go
```diff
@@ -8,11 +8,74 @@ import (
...
+				}
+
+			case eventEntryType:
+				EventEntryCnt++
+			}
+		}
+
+		if APIEntryCnt == 0 && EventEntryCnt == 0 {
+			messages = append(messages, fmt.Sprintf("Service with id %q is invalid. Requires at least one API or Event entry", svc.ID))
+		}
```
jasiu001: First, number of Entries is check (`len(svc.Entries) == 0`) then the concrete amount of type is checked, it means we can expect other not supported types so maybe it is good to log (`default case`) other types which can appear in Entries

components/application-broker/internal/syncer/validator.go
```diff
@@ -8,11 +8,74 @@ import (
...
+			return true
+		}
+	}
+	return false
+}
+
+// Deprecated
 type appCRValidator struct{}
 
 const (
```
jasiu001: move const up

components/application-broker/internal/broker/error.go
```diff
@@ -1,8 +1,17 @@
 package broker
 
+import "github.com/pkg/errors"
+
+type notFoundError struct{}
```
jasiu001: Why this error is repated? Check `/internal/storage/driver/memory/error.go`, IMO it should be in one place.

components/application-broker/internal/model.go
```diff
@@ -10,44 +11,83 @@ type ApplicationName string
...
 	Labels map[string]string
+}
 
-	//TODO(entry-simplification): this is an accepted simplification until
-	// explicit support of many APIEntry and EventEntry
-	APIEntry      *APIEntry
-	EventProvider bool
+func (s *Service) IsBindable() bool {
+	for _, e := range s.Entries {
+		if e.Type == "API" {
```
jasiu001: Why don't you compare it with some const, e.g. `apiEntryType` from validator file? 
By the way, the same values (API, Events) are in both places `mapper.go` and `validator.go`.

mszostok: I've added proper const in `model.go` file and used them across all packages

components/application-broker/internal/broker/error.go
mszostok: I forget to remove it :)

Issue #7240, opened 2020-02-18
Add Istio authorization policy for app-connector 


Abd4llA:
**Description**

Following #7065 PoC, we need to create a specific service account and add Authorization Policy objects for:

1. app-connectivity-validator
1. HTTPSource receive adapter
1. Channel dispatcher

Consider adding the validator istio policy to the [application connector](https://github.com/kyma-project/kyma/tree/master/resources/application-connector/) chart and the HttpSource istio policy to the [event-source](https://github.com/kyma-project/kyma/tree/master/resources/event-sources) chart.

**Notes**

1. Changes should apply only when some flag is enabled for "strict eventing mode" that can be passed as an override.
1. Make sure to allow GET for metrics and probes.

**AC**

1. Only authorized apps can publish events to Knative Eventing mesh.
1. There are specific SA for pods mentioned above.

sayanh:
Since istio upgrade [PR](https://github.com/kyma-project/kyma/pull/7493) is merged, this needs to be tested  again.

sayanh:
Blocked as the verification on upgraded Istio is pending. This is because cluster creation fails in the rebased PR(as well as master) while installing `core` chart. Error is:
```
$ k logs -n kyma-system apiserver-proxy-certs-job-x7jzf -f
Error from server (NotFound): services "apiserver-proxy-ssl" not found
Error from server (NotFound): services "apiserver-proxy-ssl" not found
Error from server (NotFound): services "apiserver-proxy-ssl" not found
Error from server (NotFound): services "apiserver-proxy-ssl" not found
Error from server (NotFound): services "apiserver-proxy-ssl" not found
Error from server (NotFound): services "apiserver-proxy-ssl" not found
---> Could not retrive the IP address. Verify if service apiserver-proxy-ssl exists in the namespace kyma-system
---> Command executed: kubectl get service -n kyma-system apiserver-proxy-ssl -o jsonpath='{.status.loadBalancer.ingress[0].ip}'
```

sayanh:
Auth policy for dispatcher is not implemented in this issue as there is an overlap with auth policies for brokers, which is taken care of in this [#7241](https://github.com/kyma-project/kyma/issues/7241).

IMO dispatcher auth policy is a cross-cutting topic that goes beyond app-connector hence it needs to be dealt with separately.

Issue #7241, opened 2020-02-18
Add Istio Authorization policy for the broker pods


Abd4llA:
**Description**

Following #7065 PoC, Application Broker needs to create Authorization Policy objects for:

1. Broker ingress
1. Broker filter

Consider either querying the channel config to determine the SA or using a generic SA for Kyma channels.

**AC**
1. Only authorized apps can publish events to Knative Eventing mesh.

anishj0shi:
lets evaluate OpenPolicy by knative Security Task Force

PR #7349, opened 2020-02-26
Added auth policy for app-validator, app-event-service and source adapter

sayanh:
**Description**

- Created AuthorizationPolicy for:
  - {{application}}-connectivity-validator
  - {{application}}-event-service
  - {{application}}-http-source
- Creation of AuthorizationPolicy should be controlled by an override 
- Narrowed down PERMISSIVE policy for HTTP source adapter just to metrics
 
**Related issue(s)**

See #7240 
Chart bump PR: https://github.com/kyma-project/kyma/pull/7356


components/application-connectivity-validator/internal/validationproxy/handler.go
```diff
@@ -295,7 +295,8 @@ func createReverseProxy(destinationHost string, reqOpts ...requestOption) *httpu
 		Director: func(request *http.Request) {
 			request.URL.Scheme = "http"
 			request.URL.Host = destinationHost
-
+			request.Host = ""
```
antoineco: Are you certain we can safely alter the Host header for _every_ request?

components/application-operator/README.md
```diff
@@ -31,6 +31,7 @@ Application Operator (AO) performs different operations as a result of the follo
  - **eventServiceImage** is the Event Service image version to use in the Application chart.
  - **eventServiceTestsImage** is the Event Service Tests image version to use in the Application chart.
  - **applicationConnectivityValidatorImage** is the Application Connectivity Validator image version to use in the Application chart.
+ - **strictMode** is a toggle to enable/disable Istio authorization policy for validator and HTTP source adaptor. The default value is `disabled`.
```
antoineco: We should align on the way we do this. Ping @montaro, this concerns you since you used an env var.

components/application-connectivity-validator/internal/validationproxy/handler.go
k15r: the host header will be updated to the correct value automatically, but this only works if it is not specified before

components/application-connectivity-validator/internal/validationproxy/handler.go
```diff
@@ -295,7 +295,8 @@ func createReverseProxy(destinationHost string, reqOpts ...requestOption) *httpu
...
+			delete(request.Header, "X-Forwarded-Client-Cert")
```
k15r: 👍

components/application-connectivity-validator/internal/validationproxy/handler.go
```diff
@@ -82,7 +82,7 @@ func NewProxyHandler(
 		appRegistryHost:          appRegistryHost,
 
 		eventsProxy:      createReverseProxy(eventServiceHost),
-		eventMeshProxy:   createReverseProxy(eventMeshHost, withRewriteBaseURL("/"), withEnforceURLHost),
+		eventMeshProxy:   createReverseProxy(eventMeshHost, withRewriteBaseURL("/")),
```
k15r: imho it would be more obvious to use this request option everywhere instead of removing it from the mesh proxy and secretly sneak it in to the reverseproxy as a default

components/application-operator/charts/application/templates/authorization-policy.yaml
```diff
@@ -0,0 +1,121 @@
...
+        - operation:
+            methods:
+              - GET
+            paths:
+              - /{{ .Release.Name }}/v1/events/subscribed
+              - /v1/health
+    - from:
+        - source:
+            principals:
+              - cluster.local/ns/kyma-system/sa/core-console-backend-service
```
k15r: the 'core' comes from a different chart. would it make sense to parameterize it here as well instead of hardcoding?

components/application-operator/charts/application/templates/policy.yaml
```diff
@@ -7,4 +8,5 @@ spec:
     - name: {{ .Release.Name }}-event-service
   peers:
     - mtls:
-        mode: "PERMISSIVE"
\ No newline at end of file
+        mode: "PERMISSIVE"
```
k15r: this policy should not be needed anymore due to the fix with the host header

components/application-connectivity-validator/internal/validationproxy/handler.go
antoineco: Yes I get that, I made that change originally, but the application validator serves more than just the eventing endpoints. Hence my question: can we safely replace the original header ("...xip.io") with the internal host name? Most likely yes, but it's worth asking if we really checked every scenario.

components/application-connectivity-validator/internal/validationproxy/handler.go
antoineco: There is a method for that: https://golang.org/pkg/net/http/#Header.Del

components/application-connectivity-validator/internal/validationproxy/handler.go
```diff
@@ -318,8 +316,18 @@ func withRewriteBaseURL(path string) requestOption {
...
 }
+
+// withHTTPScheme sets the URL scheme to HTTP
+func withHTTPScheme(req *http.Request) {
+	req.URL.Scheme = "http"
+}
+
+// withEmptyXFwdClientCert clears the value of X-Forwarded-Client-Cert header
+func withEmptyXFwdClientCert(req *http.Request) {
+	delete(req.Header, "X-Forwarded-Client-Cert")
```
k15r:
```suggestion
	req.Header.Del("X-Forwarded-Client-Cert")
```

components/application-connectivity-validator/internal/validationproxy/handler.go
sayanh: Right, it works for all the other scenarios.

components/application-operator/README.md
sayanh: Will keep it as is as @montaro's PR is not going in and there is no visible benefit in using env var.

components/application-connectivity-validator/internal/validationproxy/handler.go
sayanh: Changed

components/application-connectivity-validator/internal/validationproxy/handler.go
sayanh: Changed for all other sneaky changes as well.

components/application-operator/charts/application/templates/authorization-policy.yaml
sayanh: Parameterized the best I could but release `core` cannot be known to `application-operator` chart.

components/application-operator/charts/application/templates/policy.yaml
sayanh: Right, policy is not needed for non-strict mode as well.

components/application-connectivity-validator/internal/validationproxy/handler.go
sayanh: Fixed

components/application-operator/README.md
bszwarc: `adapter` not `adaptor` ;)  more below :)

bszwarc:
```suggestion
 - **strictMode** is a toggle used to enable or disable Istio authorization policy for validator and HTTP source adapter. The default value is `disabled`.
```

components/application-operator/charts/application/templates/authorization-policy.yaml
```diff
@@ -0,0 +1,121 @@
...
+    heritage: {{ .Release.Service }}
+spec:
+  selector:
+    matchLabels:
+      app: {{ .Release.Name }}-event-service
+  rules:
+    - from:
+        - source:
+            principals:
+              - cluster.local/ns/kyma-integration/sa/{{ .Release.Name }}-connectivity-validator
```
Szymongib: I think `event-service-tests` are also calling the event service.

components/application-operator/charts/application/templates/authorization-policy.yaml
```diff
@@ -0,0 +1,121 @@
...
+              - cluster.local/ns/kyma-system/sa/{{ .Values.authorizationPolicy.serviceAccount.consoleBackendService }}
+      to:
+      - operation:
+          methods:
+            - GET
+          paths:
+            - /v1/health
+  selector:
+    matchLabels:
+      app: newapp-event-service
```
Szymongib: Shouldn't it be?
```suggestion
      app: {{ .Release.Name }}-event-service
```

components/application-operator/charts/application/templates/authorization-policy.yaml
```diff
@@ -0,0 +1,121 @@
...
+              - POST
+            paths:
+              - /{{ .Release.Name }}/v1/events
+              - /{{ .Release.Name }}/v2/events
+              - /{{ .Release.Name }}/events
+        - operation:
+            methods:
+              - GET
+              - PUT
+              - POST
```
Szymongib: There should also be a `DELETE` method.

sayanh: Added

components/application-operator/charts/application/templates/authorization-policy.yaml
sayanh: Changed

components/application-operator/charts/application/templates/authorization-policy.yaml
sayanh: Added `- cluster.local/ns/kyma-integration/sa/{{ .Release.Name }}-event-service-acceptance`

sayanh:
@Szymongib 
> I see on the image bump PR, that the strictMode is disabled, why is that?

This is on purpose. Right now we wanna roll out `strictMode` as `disabled`. There are other components that need to be in strict mode. Details are on this [issue](https://github.com/kyma-project/kyma/issues/7241).

We do have a [plan](https://github.com/kyma-project/kyma/issues/7374) to create jobs in Prow which will test the whole integration in `strictMode`. 

> Have you verified that the tests are passing if the strictMode is enabled?
Yes, I have verified the tests are running fine.

PR #7397, opened 2020-02-28
Refactor Gateway Tests

Szymongib:
**Description**

- Refactor Gateway Tests to work with new Gateway

**Related issue(s)**


tests/application-gateway-tests/test/gateway/suite.go
```diff
@@ -0,0 +1,235 @@
...
+func (ts *TestSuite) Setup(t *testing.T) {
+	ts.WaitForAccessToAPIServer(t)
+
+	ts.appMockServer.Start()
+
+	err := ts.createNamespace()
+	require.NoError(t, err)
+
+	err = ts.createServiceInstance()
+	require.NoError(t, err)
```
Maladie: Maybe we can perform error checking inside those methods like you do in CheckApplicationGatewayHealth. I think it's more consistent.

Szymongib: I kind of prefer not to pass `*testing.T` to functions that do not really need it, as it makes them easier to reuse in the future.

Issue #7303, opened 2020-02-22
Have grafana dashboards in view-mode by default


a-thaler:
**Description**

When watching dashboards in grafana not created by kubernetes, they are opened in edit mode usually. With that any change like selecting a namespace in a selector is causing a "change" and when leaving the dashboard grafana is asking you if you want to save the changes.
An example is the dashboard called "Prometheus".

Please have all bundled dashboard deployed in view mode so that you explicitly need to enable editing when you want to edit them.

**Reasons**
As a customer it is annoying to get confronted with the "save" dialog when exploring dashboards.

**Attachments**

PR #7408, opened 2020-02-28
setting "editable" property to false in grafana dashboards

shorim:
**Description**

- setting the dashboards to be in view mode by default

**Related issue(s)**
#7303


a-thaler:
I would like to avoid changing the original dashboards coming from the upstream prometheus-operator chart as much as possible. Have you checked if the "editable" flags on row lovel are really making a change if the root editable flag is already set to false?

shorim:
@a-thaler I will check if only setting  the "editable" flag on the root level is enough

Issue #7343, opened 2020-02-26
Not enough admins permissions


dariadomagala:
**Description**

Runtime and Namespace Admins don't have required permissions.

Missing permissions for Runtime Admin:

```
apiGroups: ["applicationconnector.kyma-project.io"]
resources: ["applications"]
verbs: ["get", "list", "watch"]

apiGroups: ["applicationconnector.kyma-project.io"]
resources: ["applicationmappings"]
verbs: ["create", "list"]
```

Missing permissions for Namespace Admin:

```
apiGroups: ["addons.kyma-project.io"]
resources: ["addonsconfigurations", "addonsconfigurations/status", "addonsconfigurations/finalizers"]
verbs: ["get", "list", "watch", "create", "update", "delete"]

apiGroups: ["rbac.authorization.k8s.io"]
resources: ["roles", "rolebindings"]
verbs: ["get", "list", "create", "update", "delete"]
```

Additionally, Namespace Admin should be able to create/update/delete all namespace-wide resources like pods etc.

**Expected result**

Runtime Admin should be able to manage also applications, Namespace Admin should be able to manage all namespace-wide resources in his namespace.

**Actual result**

Runtime Admin doesn't have permissions to manage applications. Namespace Admin doesn't have permission to see addons and permissions or to create pods etc.

**Steps to reproduce**

**Troubleshooting**

**Related**
https://github.com/kyma-project/kyma/issues/7803

Tomasz-Smelcerz-SAP:
Unable to verify.

I understand the ticket as follows

- the term "Runtime Admin" refers to the subject mapped to the group "runtimeAdmin". This value is visible in "groups" claim in JWT used to authenticate.

- the term "Namespace Admin" refers to the subject mapped to the group "runtimeNamespaceAdmin".
This value is visible in "groups" claim in JWT used to authenticate.

Naming of groups / XSUAA subjects has changed recently, that's why I am making this clarification.

Assuming the above, I have extended our tests to cover cases described in the ticket, all tests are passing.
I have also created a cluster and set up XSUAA connector manually

Manual verification confirmed that:

As a "Runtime Admin" I can
- create and list applications
- create and list applicationmappings


As a "Namespace Admin" I can
- create a role (from kubectl, I don't see any option to do it using UI)
- list role and see it's details
- delete a role in the namespace I've created

- create rolebinding in the namespace I've created
- list rolebindings in the namespace I've created
- delete rolebinding in the namespace I've created

[...]

PR #7428, opened 2020-03-02
Extend admin permissions test

Tomasz-Smelcerz-SAP:
**Description**

Extend admin permissions

- Fixed syntax errors in sar-test.sh
- Extended test coverage for Kyma namespace admin permissions (common k8s objects)
- Introduced function to test global (non-namespaced) permissions

**Related issue(s)**
See also #7343

### DIFF
--- a/tests/integration/cluster-users/sar-test.sh
+++ b/tests/integration/cluster-users/sar-test.sh
@@ -109,6 +109,8 @@ function testPermissions() {
 	TEST_NS="$3"
 	EXPECTED="$4"
 
+	sleep 0.1
+
 	set +e
 	TEST=$(kubectl auth can-i "${OPERATION}" "${RESOURCE}" -n "${TEST_NS}")
 	set -e
@@ -139,6 +141,8 @@ function testPermissionsClusterScoped() {
 	OPERATION="$1"
 	RESOURCE="$2"
 	EXPECTED="$4"
+
+	sleep 0.1
 
 	set +e
 	TEST=$(kubectl auth can-i "${OPERATION}" "${RESOURCE}" --all-namespaces)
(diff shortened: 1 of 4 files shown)


tests/integration/cluster-users/sar-test.sh
```diff
@@ -48,31 +48,111 @@ function __deleteTestNamespace() {
...
+	local RESOURCE="$2"
+	local TEST_NS="$3"
+	local EXPECTED="$4"
+	local TEST="not-set-yet"
+
+	if [[ "${TEST_NS}" != "--all-namespaces" ]]; then
+		TEST_NS="-n${TEST_NS}"
+	fi
+
+	sleep 0.1
```
Demonsthere: Is this really required?

tests/integration/cluster-users/sar-test.sh
```diff
@@ -48,31 +48,111 @@ function __deleteTestNamespace() {
...
+
+	local result=1
+	set +e
+	kubectl describe "${RESOURCE}" "${TEST_NS}" > /dev/null
+	result=$?
+	set -e
+
+	local IS_OK="no"
```
Demonsthere: Why not `IS_OK=true`?

kubadz: I would rather suggest `IS_OK=false`, as in case of some different `EXPECTED` value that is not handled by IFs below, the script should not pass

tests/integration/cluster-users/sar-test.sh
```diff
@@ -226,23 +290,53 @@ function runTests() {
...
+	echo "--> ${ADMIN_EMAIL} should be able to get Nodes in the cluster"
+	testPermissionsClusterScoped "get" "nodes" "yes"
+
+	echo "--> ${ADMIN_EMAIL} should be able to describe Pods in ${CUSTOM_NAMESPACE}"
+	testDescribe "pods" "${CUSTOM_NAMESPACE}" "yes"
+
+	echo "--> ${ADMIN_EMAIL} should be able to describe Pods in ${SYSTEM_NAMESPACE}"
+	testDescribe "pods" "${SYSTEM_NAMESPACE}" "yes"
+
+  echo "--> ${ADMIN_EMAIL} should be able to describe Nodes in the cluster"
```
kubadz:
```suggestion
	echo "--> ${ADMIN_EMAIL} should be able to describe Nodes in the cluster"
```

tests/integration/cluster-users/sar-test.sh
```diff
@@ -276,30 +371,192 @@ function runTests() {
 	echo "--> ${NAMESPACE_ADMIN_EMAIL} should NOT be able to delete system namespace"
 	testPermissions "delete" "namespace" "${SYSTEM_NAMESPACE}" "no"
 
-	# namespace admin should not be able to create clusterrolebindings - if they can't create it in one namespace, 
+	# namespace admin should not be able to create clusterrolebindings - if they can't create it in one namespace,
 	# that means they can't create it in any namespace (resource is non namespaced and RBAC is permissive)
```
kubadz: now that permissions are tested cluster-wide, the comment is obsolete and should be removed

tests/integration/cluster-users/sar-test.sh
```diff
@@ -276,30 +371,192 @@ function runTests() {
...
+	echo "--> ${NAMESPACE_ADMIN_EMAIL} should be able to update RoleBinding in the namespace they created"
+	testPermissions "update" "rolebinding.rbac.authorization.k8s.io" "${CUSTOM_NAMESPACE}" "yes"
+
+	echo "--> ${NAMESPACE_ADMIN_EMAIL} should be able to delete RoleBinding in the namespace they created"
+	testPermissions "delete" "rolebinding.rbac.authorization.k8s.io" "${CUSTOM_NAMESPACE}" "yes"
 
 	echo "--> ${NAMESPACE_ADMIN_EMAIL} should be able to create rolebindings to kyma-developer clusterrole in the namespace they created"
 	createRoleBindingForNamespaceDeveloper
 
+	echo "--> ${NAMESPACE_ADMIN_EMAIL} should NOT be able to delete any namespace in the cluster"
```
kubadz: I would suggest rephrasing that to `[..] to delete all namespaces in the cluster`. It may be misleading, as one of next assertions is testing that the `NAMESPACE_ADMIN` actually **CAN** delete some namespaces (specifically namespaces they created)

tests/integration/cluster-users/sar-test.sh
```diff
@@ -276,30 +371,192 @@ function runTests() {
...
+	echo "--> ${NAMESPACE_ADMIN_EMAIL} should be able to delete addonsconfigurations.addons.kyma-project.io in the namespace they created"
+	testPermissions "delete" "addonsconfigurations/finalizers.addons.kyma-project.io" "${CUSTOM_NAMESPACE}" "yes"
+
+	echo "--> ${NAMESPACE_ADMIN_EMAIL} should be able to describe Pods in ${CUSTOM_NAMESPACE}"
+	testDescribe "pods" "${CUSTOM_NAMESPACE}" "yes"
+
+	echo "--> ${NAMESPACE_ADMIN_EMAIL} should NOT be able to describe Pods in ${SYSTEM_NAMESPACE}"
+	testDescribe "pods" "${SYSTEM_NAMESPACE}" "no"
+
+  echo "--> ${NAMESPACE_ADMIN_EMAIL} should NOT be able to describe Nodes in the cluster"
```
kubadz:
```suggestion
	echo "--> ${NAMESPACE_ADMIN_EMAIL} should NOT be able to describe Nodes in the cluster"
```

tests/integration/cluster-users/sar-test.sh
Tomasz-Smelcerz-SAP: I would rather avoid using `false` and `true`. These are tricky and misleading in bash. Sometimes bash treats it is a builtin, sometimes as a command (/usr/bin/false), sometimes as a string. Take a look below. Surprise?
```
condition=false
if [ $condition ]; then
  echo 'Surprise!'
fi
```

tests/integration/cluster-users/sar-test.sh
Tomasz-Smelcerz-SAP: I added this to reduce the stress on api-server during test execution. It extends test duration by just few seconds in total.

tests/integration/cluster-users/sar-test.sh
Tomasz-Smelcerz-SAP: How about: "should NOT be able to delete arbitrary namespaces in the cluster" ?

Issue #7425, opened 2020-03-02
Update event-service docker image


Abd4llA:
**Description**

event-service [Dockerfile](https://github.com/kyma-project/kyma/blob/master/components/event-service/Dockerfile) docker image needs to be updated with either a scratch image or a more up-to-date alpine image.

**AC**

1. Event service docker image passes Protocode scan.

PR #7429, opened 2020-03-02
Upgrade event-service Docker base image to `From scratch` [Component]

montaro:
See also #7425


montaro:
Protecode security check is green.
image

PR #7439, opened 2020-03-03
Feature toggle for Application Operator

Maladie:
**Description**

- Application operator have now possibility to work in two modes: old one and Service Instance mode.

**Related issue(s)**

#7043

### DIFF
--- a/components/application-operator/charts/gateway/templates/deployment.yaml
+++ b/components/application-operator/charts/gateway/templates/deployment.yaml
@@ -1,3 +1,4 @@
+{{- if .Values.global.deployGatewayOncePerNamespace }}
 apiVersion: apps/v1
 kind: Deployment
 metadata:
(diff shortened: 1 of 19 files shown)


components/application-operator/charts/gateway/templates/deployment.yaml
```diff
@@ -1,3 +1,4 @@
+{{- if .Values.global.deployGatewayOncePerNamespace }}
```
Szymongib: I think in the case of disabling Gateway per namespace we can just not create the Service Instance Controller.

resources/application-connector/charts/application-operator/values.yaml
```diff
@@ -9,6 +9,7 @@ controller:
     tillerTLSInsecure: true
     syncPeriod: 30
     installationTimeout: 240
+    gatewayOncePerNamespace: false
```
Szymongib: I would suggest moving this value to global values and rename it to something more general, as this override will be used by other components as well.

PR #7443, opened 2020-03-03
cacert in compass to separate secret

zralt:
In order to use external managed certificates (in MPS), the cacert needs to be in a different secret than the crt/key, because when the certificates are rotated, the cacert key will get deleted. Istio automatically checks for a cacert in the [SECRET]-cacert resource.

### DIFF
--- a/resources/compass/charts/gateway/templates/gateway-certs-mtls.yaml
+++ b/resources/compass/charts/gateway/templates/gateway-certs-mtls.yaml
@@ -8,5 +8,4 @@ type: Opaque
 data:
   "key": {{ .Values.global.ingress.tlsKey }}
   "cert": {{ .Values.global.ingress.tlsCrt }}
-  "cacert": "" # This value is populated by Connectivity Certs Setup Job
 {{- end -}}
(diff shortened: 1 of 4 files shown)


resources/compass/charts/gateway/templates/gateway-certs-mtls.yaml
```diff
@@ -8,5 +8,4 @@ type: Opaque
 data:
   "key": {{ .Values.global.ingress.tlsKey }}
   "cert": {{ .Values.global.ingress.tlsCrt }}
-  "cacert": "" # This value is populated by Connectivity Certs Setup Job
```
Szymongib: I wonder if we cannot just remove the `key` and `cert` from this secret as they are already specified in `gateway-certs.yaml`

zralt: I don't know why there are 2 secrets for this chart (gateway-certs.yaml and gateway-certs-mtls.yaml)

Both of these use the same vars:
- {{ .Values.global.ingress.tlsKey }}
- {{ .Values.global.ingress.tlsCrt }} 

So no different values can be specified. But the connector chart refers to the mtls secret name ({{ .Values.global.connector.secrets.rootCA.name }}), and also the mtls gateway is configured to use this.  
Is this some future requirement, or can it be modified to just use one secret?

Demonsthere: It is recommended to use separate secrets per gateway, as in the case of a faulty secret (pending refresh, update, etc) all gateways using this secret are down. This separation allows us to keep gateways as separate entities

Szymongib: But we do use the single secret in Kyma and only have `cacert` in a separate one.
Isn't it harder to then renew certs?

PR #7447, opened 2020-03-03
Fix RBAC for event-service

sayanh:
**Description**

- Fix RBAC for event-service

### DIFF
--- a/components/application-operator/charts/application/templates/role-binding.yaml
+++ b/components/application-operator/charts/application/templates/role-binding.yaml
@@ -88,7 +88,10 @@ metadata:
     heritage: {{ .Release.Service }}
 rules:
 - apiGroups: ["*"]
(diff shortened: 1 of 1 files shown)


components/application-operator/charts/application/templates/role-binding.yaml
```diff
@@ -88,7 +88,10 @@ metadata:
     heritage: {{ .Release.Service }}
 rules:
 - apiGroups: ["*"]
-  resources: ["namespaces", "subscriptions", "triggers"]
```
montaro: you dropped `subscriptions` deliberately?

sayanh: yes, it's not used anymore

sayanh:
Fixed by https://github.com/kyma-project/kyma/pull/7349

Issue #7370, opened 2020-02-27
Docker registry development setup


michal-hudy:
**Description**

 - It must be possible to download images when Kyma is installed with a self-signed certificate
   - On production created solution shouldn't be executed
   - >**HINT**: Use DaemonSet
   - Check on AKS, GKE, Minikube and Gardener

**Reasons**

**Attachments**

aerfio:
After #7451 is merged fun-controller will work on GKE, Azure, Minikube and Gardener. Pipelines with tests will be run after function-controller replaces kubeless as default solution.

After some tests pipelines with minikube unfortunately timeout, it may be due to too low resources. To enable those tests one would have to either add `{function-name}.kyma.local` to `/etc/hosts` file, or make a request to $(minikube ip):{correct ingress port} with appropriate headers -> the problem that I've experienced is that it redirects to https, which fails afterwards

PR #7451, opened 2020-03-03
Enable docker registry on clusters with self signed certs

aerfio:
**Description**

- add daemonset which patches k8s internal docker, so that it can connect to itself, while we use self signed certs (hack from https://docs.docker.com/registry/insecure/#use-self-signed-certificates)
- change order of installation so that we can access `kyma-gateway-certs` from `core` (I've added fun-ctrl after permission-controller so that we are sure everything is already there)
- add two missing crds for cert-manager
- increase PVC volume to 20GB as @michal-hudy asked

**Related issue(s)**

See also #7370


components/function-controller/pkg/utils/serving_utils.go
```diff
@@ -63,9 +63,6 @@ func GetServiceSpec(imageName string, imagePullSecretName string, rnInfo *Runtim
 						Image: imageName,
 						Env:   envVarsForRevision,
 					}},
-					ImagePullSecrets: []corev1.LocalObjectReference{
```
aerfio: not needed, I added imagePullSecret to proper service account instead

resources/function-controller/charts/docker-registry/values.yaml
```diff
@@ -57,7 +57,7 @@ podLabels: {}
 persistence:
   accessMode: 'ReadWriteOnce'
   enabled: true
-  size: 10Gi
+  size: 20Gi
```
aerfio: as per @michal-hudy's plea

resources/function-controller/templates/daemonset.yaml
```diff
@@ -0,0 +1,65 @@
...
+            secretName: kyma-gateway-certs
+            items:
+              - key: cert
+                path: cert.txt
+        - configMap:
+            defaultMode: 420
+            name: {{ template "registry-fullname" }}-cert-patch
+          name: scripts
+      containers:
+        - name: init-container-did-the-work
```
magicmatatjahu: Maybe change name? :P `pause`

resources/function-controller/templates/daemonset-configmap.yaml
```diff
@@ -0,0 +1,41 @@
+# in case of changes please test this on GKE, Azure, Gardener and Minikube
+kind: ConfigMap
+metadata:
+  name: {{ template "registry-fullname" }}-cert-patch
+  namespace: istio-system
+apiVersion: v1
+data:
+  patch-docker.sh: |
+    #!/usr/bin/env bash
```
magicmatatjahu: Maybe inject script by Helm template?

resources/function-controller/templates/daemonset.yaml
```diff
@@ -0,0 +1,65 @@
...
+      app: {{ template "registry-fullname" }}-cert-update
+  template:
+    metadata:
+      labels:
+        app: {{ template "registry-fullname" }}-cert-update
+      annotations:
+        sidecar.istio.io/inject: "false"
+    spec:
+      initContainers:
+        - name: init-inject-certs
```
magicmatatjahu: `inject-certs`? :P

resources/function-controller/templates/daemonset-configmap.yaml
```diff
@@ -0,0 +1,41 @@
+# in case of changes please test this on GKE, Azure, Gardener and Minikube
+kind: ConfigMap
+metadata:
+  name: {{ template "registry-fullname" }}-cert-patch
+  namespace: istio-system
```
magicmatatjahu: Add `commonLabels`.

resources/function-controller/templates/daemonset.yaml
```diff
@@ -0,0 +1,65 @@
+apiVersion: apps/v1
+kind: DaemonSet
+metadata:
+  name: "{{ template "registry-fullname" }}-self-signed-cert"
+  namespace: istio-system
```
magicmatatjahu: Add `commonLabels`.

resources/function-controller/templates/daemonset-configmap.yaml
aerfio: this configmap is almost just this script, there's no point in extracting this to some kind of helper, it would be convoluted

resources/function-controller/templates/serverless-service-account.yaml
```diff
@@ -9,3 +9,5 @@ metadata:
 secrets:
 - name: {{ template "fullname" . }}-registry-credentials
 {{ end }}
```
magicmatatjahu: Add this `end` statement to end of file

aerfio:
`'oct-tp-testsuite-all-2020-03-14-17-11-function-controller-0' failed due to too long Running status` I might consider disabling fun-controller tests on minikube, because we have too low resources for it

PR #7461, opened 2020-03-04
Restrict access to Rafter, Tekton and Knative Serving resources

michal-hudy:
**Description**

- Readonly access to Rafter Resources
- No access to Tekton resources
- No access to Knative Serving resources

**Related issue(s)**


tests/integration/cluster-users/sar-test.sh
```diff
@@ -232,6 +309,8 @@ function runTests() {
 	echo "--> ${ADMIN_EMAIL} should be able to patch Installation CR in ${NAMESPACE}"
 	testPermissions "patch" "installation" "${NAMESPACE}" "yes"
```
piotrmsc: TBH this `testRafter` function should also contain expected authZ result because IMO admin should have access to CRUD operations on rafter resources. Currently, it's hardcoded for all type of users

magicmatatjahu: We want to prevent the creation/update/deletion of any resources by all type of users. Only person who has access to kubeconfig from cluster can create resources.

PR #7465, opened 2020-03-04
Add a security doc on Serverless

kazydek:
**Description**

- Document possible security risks in Serverless v2


docs/serverless-v2/03-02-security.md
```diff
@@ -0,0 +1,16 @@
+---
+title: Security
+type: Details
+---
+
+To eliminate potential security risks when using lambdas, bear in mind these few facts:
+
+- Kyma doesn't run any security scans against lambdas and their images. Before you store any sensitive data in lambdas, consider the potential risk of data leakage.
+
+- JSON Web Tokens (JWTs) used in Kyma are cluster-wide as they don't have the **scope** parameter defined. This means that if you expose your lambda and secure it with a JWT issued by Dex, you can use the token to validate access to all lambdas within the cluster.
```
michal-hudy: Some scopes are defined 😄, but we don't provide scope definition for lambda by default. Users can use different Identity Providers where they may restrict access to lambdas. @piotrmsc can you confirm this?

piotrmsc: Well, TBH there's no such things as cluster or namespace wide JWT. When you enable JWT authentication in your lambda it means ID Token issued from given IDP in JWT format is required. Yes, by default ID Token issued by dex does not provide `scp` or `scope` claim,however, other identity poroviders  may provide it

docs/serverless-v2/03-02-security.md
```diff
@@ -0,0 +1,16 @@
+---
+title: Security
+type: Details
+---
+
+To eliminate potential security risks when using lambdas, bear in mind these few facts:
+
+- Kyma doesn't run any security scans against lambdas and their images. Before you store any sensitive data in lambdas, consider the potential risk of data leakage.
```
bszwarc:
```suggestion
- Kyma does not run any security scans against lambdas or their images. Before you store any sensitive data in lambdas, consider the potential risk of data leakage.
```
Just because you don't use contractions later on. I would also use `or` with negation 🤔

kazydek: I'll remove contraction here 👍 but I would leave `and` as I think that is crucial that both are not scanned.

PR #7490, opened 2020-03-06
Update images of components that use kubectl

strekm:
**Description**

- new images are based on the alpine-kubectl image

**Changes in Dockerfiles**
#7510 
#7487


resources/cluster-essentials/values.yaml
```diff
@@ -2,7 +2,4 @@ global:
   isLocalEnv: false
 
 containerRegistry:
-  path: eu.gcr.io/kyma-project
-xip_patch:
-  dir: pr/
-  version: "PR-5226"
+  path: eu.gcr.io/kyma-project
```
suleymanakbas91: Please put a new line at the end ;)

resources/helm-broker/charts/etcd-stateful/values.yaml
```diff
@@ -1,6 +1,6 @@
 tlsSetup:
```
mszostok: whole that section can be removed because I see that we are not using the tls job in Helm Broker

resources/backup-init/values.yaml
```diff
@@ -2,5 +2,5 @@
 kubectl:
   image:
     repository: eu.gcr.io/kyma-project/pr/etcd-tls-setup
```
mszostok: such image `eu.gcr.io/kyma-project/pr/etcd-tls-setup:a12c148d` does not exist. Why CI pipelines didn't catch that?

PR #7492, opened 2020-03-06
Testing logging from Loki

shorim:
**Description**

- Testing logging from Loki

**Related issue(s)**


tests/integration/logging/Gopkg.toml
```diff
@@ -22,4 +22,8 @@
 
 required = [
   "golang.org/x/lint/golint"
-]
\ No newline at end of file
+]
+
+[[constraint]]
+  name = "github.com/kyma-project/kyma"
+  version = "1.10.0"
```
clebs: I am not sure if this is correct. Shouldn't it be master branch? Or are we going to manually bump this on every release?

tests/integration/logging/testCounterPod.yaml
```diff
@@ -15,6 +15,6 @@ spec:
       requests:
         memory: 2Mi
       limits:
-        memory: 10Mi
+        memory: 50Mi
     args: [sh, -c,
-            'for i in `seq 50` ; do echo "$i: logTest-$(date)"; sleep 2; done']
+            'let i=1; while true; do echo "$i: logTest-$(date)"; let i++; sleep 2; done']
```
clebs: An infinite loop does look dangerous here. How is this pod terminated?

shorim: Isn't it the case that all pods are terminated after the integration tests are finished?
Anyways, I was making it an infinite loop mainly for debugging. So, I have removed now the infinite loop.

tests/integration/logging/Gopkg.toml
shorim: As suggested by @hisarbalik, it is better to make it point to a stable version. Because if the authentication package is broken in the master branch, then this test will break and all PRs to the kyma repo will be blocked

PR #7502, opened 2020-03-09
Fix empty slack notifications by switching to .Annotations.message attribute

lauraanddola:
Consider below scenarios which only contains the  .Annotations.message ....

```
Alert: KuberJobCompletion
    Annotations:
         Message: Job ****
         Runbook URL: ***
    Expr:    ****
```


a-thaler:
That is a good point you encountered there. It looks like the standard is "Annotations.message" and we should use it as well.
For that, our own rules needs to be adjusted to use "message" as well. Can you please go over all the rules and change them all to "message", mainly the files in https://github.com/kyma-project/kyma/tree/master/resources/monitoring/templates/prometheus/kyma-rules

And remove the "summary" on that kyma-rules, they are just duplicating the stuff

lauraanddola:
Hi @a-thaler  , i found the summary may be called as below in the  ./resources/monitoring/templates/kyma-additions/alertmanager.config.yaml

`state_message: 'Alert: {{`{{ .CommonLabels.alertname }}`}}. Summary:{{`{{ .CommonAnnotations.summary }}`}}. RawData: {{`{{ .CommonLabels }}`}}'`

Do we really need to remove the summary? if yes, then the `.CommonAnnotations.summary` also need to change to `.CommonAnnotations.message`, am i right ?pls correct me if i am wrong.

resources/monitoring/templates/prometheus/kyma-rules/alert-rules-backup.yaml
```diff
@@ -15,21 +15,21 @@ spec:
       labels:
         severity: critical
       annotations:
-        description: "Backup failed"
+        message: "Backup failed"
         summary: "Backup failed"
```
a-thaler: Please remove the "summary" in all rules as well, I don't see any value

Issue #7062, opened 2020-01-30
Application Broker calls MPS via Agent to request and delete credentials


PK85:
**Description**

Application Broker calls MPS via Agent to request and delete credentials for given api-package

**Reasons**

We need credentials per request for service instances of Application API.

**AC**
- extends credentials to contain a type of credentials, credentials + GATEWAY_URL. It will be up to user if he wants to go thru gateway or call application directly.
- adjust GATEWAY_URL value in connection with requirements of refactored gateway. Gateway is deployed in the same namespace where service binding was created and works with secret provided by service catalog
- do a call to agent proxy for requesting and deleting credentials requests
-

mszostok:
**Raw diagram**
<p align="center">
  <img src="https://user-images.githubusercontent.com/17568639/73759752-acd67f80-476c-11ea-84e9-3201b6517027.png" alt="drawing" width="400px" />

</p>


**NOTES:**
- Sync with Frog team about GATEWAY_URL pattern
- Sync with Frog team about the Agent Proxy
- Sync with X-Team about the Director API graphql mutations for Auth

**Additional AC:** 
- add documentation

mszostok:
**Question:**
Generate the Auths per instance or per binding?

mszostok:
fetch additional info about auth like req headers etc. 

sync with @Szymongib

mszostok:
Call to Director:

```graphql
query getPackage($appID: ID!, $pkgID: ID!) {
  application(id: $appID) {
    package(id: $pkgID) {
      instanceAuths {
        context
        auth {
          additionalHeaders # wolanie API
          additionalQueryParams
          requestAuth {
            csrf {
              tokenEndpointURL # wolanie po CSRF token
            }
          }
          credential {
            __typename
            ... on OAuthCredentialData {
              clientId
              clientSecret
              url
            }
            
            ... on BasicCredentialData {
              username
              password
            }
          }
        }
        status {
          condition
          timestamp
          message
          reason
        }
      }
    }
  }
}

```

mszostok:
new agreement about the GATAWAY URL 

App Broker consume such env variable `GATAWAY_BASE_URL_FORMAT=http://[NAMESPACE]-some-hardcoded-name/` and then inject for lambda such env `{API_NAME}_GATEWAY_URL=<rendered_gateway_base_url>/secret/<secret_name>/api/<api_name>`

PR #7504, opened 2020-03-09
Add Director proxy in Runtime Agent

mszostok:
**Description**

- [x] Add Director proxy in Runtime Agent
- [x] Simplify`main.go` by adding **ExitOnError**
- [x] Simply the startup of components (this is already implemented no need to do it by custom logic), see logger service changes.
- [x] Remove redundant comments in main -`Info`log statement is enough
- [x] Execute `make fmt-local` and `make imports-local`
- [x] Change channel from `bool` to `struct` which [occupies zero bytes of storage](https://dave.cheney.net/2014/03/25/the-empty-struct) (there was no need to use the bool channel).
- [x] Change build pack version to use Go 1.13 version (in 1.12 type *http.Transport does not have Clone method) btw. the Dockerfile already uses 1.13 for building binary so test should also be executed against that version
- [x] To be discussed: Is there a bug where the input certs are outdated?

Istio RBAC will be added in this task: https://github.com/kyma-project/kyma/issues/7062

**Related issue(s)**

- https://github.com/kyma-incubator/compass/issues/752


components/compass-runtime-agent/internal/compassconnection/supervisor.go
```diff
@@ -121,6 +124,15 @@ func (s *crSupervisor) SynchronizeWithCompass(connection *v1alpha1.CompassConnec
 		return s.updateCompassConnection(connection)
 	}
 
+	// TODO(mszostok): bug with certs input - outdated?
```
mszostok: we need to discuss if there is a bug with outdated `credentials`.  In line 120 there is a method which can refresh the credentials but it does not return the new one if refreshed.

components/compass-runtime-agent/internal/compass/director/proxy_test.go
```diff
@@ -0,0 +1,158 @@
...
+		// given
+		var (
+			expectedBody = "Buenos Aires from Director!"
+			expectedCode = http.StatusOK
+			spyResponse  = httptest.NewRecorder()
+			fixCert      = MustLoadFixCert(t)
+		)
+
+		var gotPeerCertificates []*x509.Certificate
+		ts := NewTestTLSServer(func(w http.ResponseWriter, r *http.Request) {
```
Szymongib: I think it would be good to close the server after the test, as mentioned in the docs:
> // The caller should call Close when finished, to shut it down.

components/compass-runtime-agent/internal/compassconnection/supervisor.go
Szymongib: If credentials are refreshed, they are saved to the Secret. We could return them, but current credentials still need to be valid to be refreshed so unless credentials expired just after a call to refresh and before the call to Director, there should be no issue.

components/compass-runtime-agent/internal/compass/director/proxy_test.go
mszostok: added

mszostok:
>I also wonder if it wouldn't be good to add an acceptance test that calls the Director through the Proxy 
I agree, but can we do that as a follow up? Currently, the App Broker will use that proxy and testing that functionality will be covered by the external-solution tests.

Having dedicated test only for testing that Director proxy can be added in near future

PR #7527, opened 2020-03-10
Remove 'master' prefix from kyma version

parostatkiem:
**Description**

- ...
- ...
- ...

**Related issue(s)**


components/console-backend-service/internal/domain/k8s/versioninfo_converter_test.go
```diff
@@ -51,7 +51,7 @@ func TestVersionInfoConverter_ToGQL(t *testing.T) {
 	t.Run("Master version", func(t *testing.T) {
 		image := "eu.gcr.io/test/12345678"
```
dariadomagala: We should probably have the `master-12345678` since this is what the installer image actually looks like right now. Being picky, I know, sorry 😞

PR #7533, opened 2020-03-10
replace istio kubectl with alpine-kubectl

strekm:
**Description**

- replace istio kubectl with alpine-kubectl

**Related issue(s)**


Demonsthere:
Maybe for an easier future update we should have a single global value with the image and tag? Right now it seems to be more troublesome to update/ more distributed

PR #7540, opened 2020-03-10
Move out apiserver proxy and iam kubeconfig service from release core

piotrmsc:
**Description**

- Move apiserver-proxy to separate chart after core
- Provide apiserver-proxy-init subchart to apiserver-proxy
- Move iam-kubeconfig-service to separate chart after apiserver-proxy

**Related issue(s)**


resources/iam-kubeconfig-service/OWNERS
```diff
@@ -0,0 +1,11 @@
+approvers:
```
Demonsthere: I think you forgot @colunira  ;)

PR #7541, opened 2020-03-10
Specify namespace for Istio gateway for Grafana virtual service

sisiliao:
**Description**

- Added this line " - {{ .Values.global.istio.gateway.name }}.{{ .Values.global.istio.gateway.namespace }}.svc.cluster.local " to the Grafana virtual service yaml file. With this change, we can specify the namespace for the istio gateway, which solves the issue that the virtual service and the gateway are under 2 different namespaces.

**Related issue(s)**


a-thaler:
Thanks for your contribution. Indeed we should respect the namespace of the gateway.

a-thaler:
Hi @sisiliao, please sign the CLA, see the pending github check

ebensom:
Hi @sisiliao , please sign the CLA, that is the only pending check left.

Issue #7347, opened 2020-02-26
knative-eventing update fails with "xyz already exists"


antoineco:
**Description**

The update of the `knative-eventing` component from Kyma 1.10 to `master` (currently @ 0b8573a92ad22154c7922795127de852a3deb6c3) fails with the following errors:

* `CustomResourceDefinition with the name "subscriptions.messaging.knative.dev" already exists`
* `ConfigMap with the name "knative-eventing-event-mesh-dashboard-delivery" already exists`

Ref. Prow job: [post-master-kyma-gke-upgrade #1232589111890546690](https://status.build.kyma-project.io/view/gcs/kyma-prow-logs/logs/post-master-kyma-gke-upgrade/1232589111890546690)

The first one is surprising since we do delete that CRD in the pre-upgrade job:

https://github.com/kyma-project/kyma/blob/0b8573a92ad22154c7922795127de852a3deb6c3/resources/knative-eventing/charts/knative-eventing/templates/pre-upgrade-job.yaml#L70-L73

Another surprising aspect to this issue is that the same test succeeded in the corresponding pull request: [pre-master-kyma-gke-upgrade #1232273782467465216](https://status.build.kyma-project.io/view/gcs/kyma-prow-logs/pr-logs/pull/kyma-project_kyma/7297/pre-master-kyma-gke-upgrade/1232273782467465216)

**Troubleshooting**

I built the installer from `master` via

```
~/go/src/github.com/kyma-project/kyma/tools/kyma-installer master
❯ make DOCKER_PUSH_REPOSITORY=antoineco DOCKER_TAG=latest build-image
```

After swapping the installer image in a fresh Kyma 1.10 cluster and labeling the `Installation` object with `action: install`, the upgrade goes through and `knative-eventing` gets updated without error (before: 0.10, after: 0.12).

_Can not reproduce_ 😞

antoineco:
I think I found something interesting: there can be a long delay between the moment the `delete` request is acknowledged by the API server and the moment the CRD is actually deleted.
```
  - lastTransitionTime: "2020-02-27T15:58:20Z"
    message: CustomResource deletion is in progress
    reason: InstanceDeletionInProgress
    status: "True"
    type: Terminating
```

antoineco:
The error about the `subscriptions` CRD eventually goes away, but the following error persists after multiple runs:
```
kind ConfigMap with the name "knative-eventing-event-mesh-dashboard-delivery" already exists in the cluster
```

[...]

PR #7571, opened 2020-03-12
Fix pre-upgrade hook for knative-eventing

sayanh:
**Description**

- Fixed RBAC for pre-upgrade hook in knative-eventing chart
- Bumped kubectl alpine image

See #7347


antoineco:
@sayanh this doesn't solve the problem, we should look more closely at this issue with dashboard ConfigMaps (#7347).

PR #7632, opened 2020-03-16
Disable App Broker test when packages are enabled

mszostok:
**Description**

- Disable App Broker test when packages are enabled

**Related issue(s)**

- https://github.com/kyma-project/kyma/issues/7062


mszostok:
not yet valid

Issue #7615, opened 2020-03-16
Option to disable discovery of prometheus resources of non-system namespace


a-thaler:
**Description**
The monitoring stack is optimized to the default setup assuring a stable memory footprint. Workload of non-system namespaces can bring prometheus resources and with that introduce new metrics potentially harming the monitoring stack. New prometheus resources should be introduced very conscious by also adjusting the footprint of prometheus.
In some scenarios you might want to not allow any new metrics at all, so having an option to disable non-system namespaces for discovery will be good.
```
resourceSelector.namespaces=kyma-system,...
```

**Reasons**
Protection of monitoring stability for specific scenarios

**Attachments**

PR #7640, opened 2020-03-17
Disable monitoring non-system namespace discovery

suleymanakbas91:
**Description**

- Disable monitoring non-system namespace discovery

**Related issue(s)**

See https://github.com/kyma-project/kyma/issues/7615

### DIFF
--- a/resources/monitoring/templates/prometheus/prometheus.yaml
+++ b/resources/monitoring/templates/prometheus/prometheus.yaml
@@ -108,6 +108,7 @@ spec:
     - key: name
       operator: In
       values:
+      {{- $namespaces := splitList "," .Values.resourceSelector.namespaces -}}
       {{- range $namespaces }}
       - {{ . }}
       {{- end }}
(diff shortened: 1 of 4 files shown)


resources/monitoring/values.yaml
```diff
@@ -40,6 +40,14 @@ global:
   imagePullSecrets: []
   # - name: "image-pull-secret"
 
+resourceSelector:
```
hisarbalik: this value by default should be empty, configuration should be applied only for production profile via config overrides

resources/monitoring/templates/prometheus/prometheus.yaml
```diff
@@ -90,6 +90,39 @@ spec:
...
+
+{{- if .Values.resourceSelector.namespaces }}
+  serviceMonitorSelector: {}
+  serviceMonitorNamespaceSelector:
+    matchExpressions: 
+    - key: name
+      operator: In
+      values:
+      {{- $namespaces := splitList "," .Values.resourceSelector.namespaces -}}
+      {{- range $namespaces -}}
```
a-thaler:
```suggestion
      {{- range $namespaces }}
```

Issue #7623, opened 2020-03-16
Simplify ORY database upgrade and migration process


Demonsthere:
**Description**

The current process of database migration (`in memory -> local postgres -> external db`) is error-prone and depends on user input (given overrides), which can lead to unwanted password regeneration issues. 

**Reasons**
We need to simplify and streamline the process to be more idempotent and easy to manage. User-provided input (overrides) need to be considered as the only configuration option, and if not provided, generated and updated. 
Any additional, or manual steps to the process need to be documented 

**Attachments**

PR #7644, opened 2020-03-17
Simplify ORY database upgrade and migration process

jakkab:
**Description**

- use "ory-hydra-credentials" as a unique point of reference for all persistence modes; 
- add a pre-install/pre-upgrade job that updates the secret in case of database migration.

**Related issue(s)**
#7623


my-kyma.yaml
```diff
@@ -0,0 +1,331 @@
+---
```
Demonsthere: please remove :)

PR #7647, opened 2020-03-17
Remove special characters from compass labels

kfurgol:
**Description**

- Adjusted compass-runtime-agent to new Director labeling constraints
- Bumped images

**Related issue(s)**
[#845](https://github.com/kyma-incubator/compass/issues/845)

### DIFF
--- a/components/compass-runtime-agent/internal/compass/director/client.go
+++ b/components/compass-runtime-agent/internal/compass/director/client.go
@@ -12,8 +12,8 @@ import (
 const (
 	TenantHeader = "Tenant"
 
-	eventsURLLabelKey  = "runtime/event_service_url"
-	consoleURLLabelKey = "runtime/console_url"
+	eventsURLLabelKey  = "runtime_eventServiceUrl"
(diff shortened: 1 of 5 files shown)


components/compass-runtime-agent/internal/compass/director/client.go
```diff
@@ -12,8 +12,8 @@ import (
 const (
 	TenantHeader = "Tenant"
 
-	eventsURLLabelKey  = "runtime/event_service_url"
-	consoleURLLabelKey = "runtime/console_url"
+	eventsURLLabelKey  = "runtime_eventServiceUrl"
+	consoleURLLabelKey = "runtime_consoleUrl"
```
akgalwas: Question: is Kyma Environment Broker adjusted to this change? Compass Runtime Agent labels a runtime and KEB later reads label's value.

kfurgol: Yes, it is adjusted here https://github.com/kyma-incubator/compass/pull/984

PR #7654, opened 2020-03-17
Make knative-serving pre-install hook more robust

antoineco:
**Description**

* Allow polling instead of using `force` to ensure a deleted CRD is cleanly finalized.
* Add `list/watch` permissions, required for polling.
* Execute script directly instead of via `/bin/sh`.

**Related issue(s)**

Should mitigate kyma-incubator/compass#980


antoineco:
@k15r look at the sha-bang. If you execute a script via `sh myscript` the options that script define get bypassed. Just a good practice.

PR #7675, opened 2020-03-18
Fix RBAC issues for eventsource config folder

sayanh:
**Description**

- Fix RBAC issues for eventsource config folder

### DIFF
--- a/components/event-sources/config/200-rbac.yaml
+++ b/components/event-sources/config/200-rbac.yaml
@@ -66,13 +66,13 @@ rules:
       - watch
       - create
       - update
+  # Istio policies
   - apiGroups:
       - authentication.istio.io
     resources:
       - policies
     verbs:
       - create
-      - patch
       - update
       - get
       - list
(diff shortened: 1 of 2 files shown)


components/event-sources/config/200-rbac.yaml
```diff
@@ -66,6 +66,17 @@ rules:
       - watch
       - create
       - update
+  - apiGroups:
+      - authentication.istio.io
+    resources:
+      - policies
+    verbs:
+      - create
+      - patch
```
radufa: Is here "patch" necessary?

components/event-sources/config/200-rbac.yaml
```diff
@@ -66,6 +66,17 @@ rules:
       - watch
       - create
       - update
+  - apiGroups:
```
radufa: Please add here a comment, as in the other cases...

components/event-sources/config/200-rbac.yaml
radufa: Is "patch" necessary

PR #7683, opened 2020-03-18
Upgrade kiwigrid image

suleymanakbas91:
**Description**

- Upgrade kiwigrid image


a-thaler:
e2e-upgrade-test is failing. kiwigrid image is used in grafana only which cannot be affected by the test

PR #7700, opened 2020-03-19
Add proxy service ports for status checks

fhivemind:
**Description**

This push adds proxy service ports for health probing to following services:
* ory-hydra
* application-broker

**Related issue(s)**

### DIFF
--- a/resources/ory/charts/hydra/templates/service-admin.yaml
+++ b/resources/ory/charts/hydra/templates/service-admin.yaml
@@ -13,6 +13,10 @@ spec:
       targetPort: http-admin
       protocol: TCP
       name: http
+    - port: 15020
+      targetPort: 15020
+      protocol: TCP
+      name: status-port
   selector:
     app.kubernetes.io/name: {{ include "hydra.name" . }}
     app.kubernetes.io/instance: {{ .Release.Name }}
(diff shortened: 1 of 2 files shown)


fhivemind:
I have signed the CLA, but seems that there are some issues with it. I believe you should be able to verify this on your side.

Demonsthere:
@fhivemind The problem with the CLA is that the commit is [not signed with an email address](https://help.github.com/articles/why-are-my-commits-linked-to-the-wrong-user/#commits-are-not-linked-to-any-user), could you fix that?

resources/ory/charts/hydra/templates/service-admin.yaml
```diff
@@ -13,6 +13,10 @@ spec:
       targetPort: http-admin
       protocol: TCP
       name: http
+    - port: 15020
```
piotrmsc: AFAIR Hydra does not expose port 15020, what's the reason for this change?

ebensom: The istio sidecar exposes that port, and [Istio probe rewrite](https://istio.io/docs/ops/configuration/mesh/app-health-check/#probe-rewrite) is configured so the health endpoint for hydra looks like this:
`#> kubectl get pod -n kyma-system ory-hydra-c8879d574-tjvr8 -o yaml`
```
    livenessProbe:
      failureThreshold: 5
      httpGet:
        path: /app-health/hydra/livez
        port: 15020
        scheme: HTTP
```

piotrmsc: after offline discussion, this aims to provide accessibility for 3rd party service, that needs to query enpoints. From kubelet POV it is not required at all. I have verified that exposing pilot-agent port, therefore exposing `/quitquitquit` enpoint will not be an issue due to check for localhost requests only.

PR #7742, opened 2020-03-23
Remove Event-Bus sources

radufa:
**Description**

- Remove "event-bus" component
- Update "event-service"component
- Update Installation yaml files
- Update Application Operator yaml files

**Related issue(s)**

kyma-project/kyma#7730


components/application-operator/charts/application/templates/deployment.yaml
```diff
@@ -97,8 +97,6 @@ spec:
         - "/eventservice"
         - "--externalAPIPort={{ .Values.eventService.deployment.args.externalAPIPort }}"
         - "--eventMeshURL=http://{{ .Release.Name }}.{{ .Release.Namespace }}"
-        - "--eventsTargetURLV1={{ .Values.eventService.deployment.args.eventsTargetURLV1 }}"
```
Abd4llA, line 100: as far as I can see here https://github.com/kyma-project/kyma/blob/master/components/event-service/cmd/eventservice/options.go#L21 this cmd arg is still valid and used, and if it's not sent, the default is the event mesh new endpoint . am I missing something?

resources/core/charts/kubeless/values.yaml
```diff
@@ -15,4 +15,4 @@ config:
     secret: ""
 
 tests:
-  enabled: true
+  enabled: false
```
Abd4llA, line 18: is there a reason for disabling the tests?

components/application-operator/charts/application/templates/deployment.yaml
k15r: yeah, during refactoring eventmesh we might have forgotten this one
i will remove it

resources/core/charts/kubeless/values.yaml
k15r: yes:

k15r: This is the PR i am talking about: https://github.com/kyma-project/kyma/pull/7800

Issue #7731, opened 2020-03-23
Delete/update event-bus test components


anishj0shi:
Clean up event-bus test Components

**AC**
- [x] Remove Integration tests for event-bus
- [x] Update core-test-external-solution test, if needed.
- [ ] Check if there's a reference to event-bus publish-service in application operator test, update it if needed.
- [x] Check if Backup/Restore tests have event-bus references,  if yes, lets remove them
- [x] Check if Upgrade tests have event-bus references, if yes, lets remove them
- [x] Check if e2e-kubeless tests have event-bus references, if yes, lets remove them
- [x] Check and delete other references of event-bus components in test directory

k15r:
kubeless will be removed in this PR: #7800

PR #7755, opened 2020-03-23
remove event-bus subtest from upgrade test

k15r:
**Description**

- remove event-bus test

**Related issue(s)**

#7731


tests/end-to-end/upgrade/chart/upgrade/values.yaml
```diff
@@ -3,7 +3,7 @@ containerRegistry:
 
 image:
   dir:
-  version: 55ccf207
+  version: PR-7755
```
marcobebway: Is this for verification purposes?

As this change should be in another PR with the final image SHA.

k15r: yes that will happen ... but as all those updated tests create new images i want to wait until all evidence of event-bus has been wiped from the test folder and then create one pr to update all of them

PR #7756, opened 2020-03-23
remove event-bus subtest from backup test

k15r:
**Description**

- remove event-bus subtest 

**Related issue(s)**

#7731


tests/end-to-end/backup/pkg/common/common.go
```diff
@@ -72,9 +71,6 @@ func RunTest(t *testing.T, mode TestMode) {
 	helmBrokerTest, err := servicecatalog.NewHelmBrokerTest()
 	fatalOnError(t, err, "while creating structure for HelmBroker test")
 
-	myEventBusTest, err := eventbus.NewEventBusTest()
-	fatalOnError(t, err, "while creating structure for EventBus test")
-
```
sayanh: https://github.com/kyma-project/kyma/pull/7756/files#diff-7c8754c94a647df85f1b987aa53d14ddL84
This does not test the compatibility layer which is used by most of the clients.

k15r: i cannot see what you want to show me with that link. But i agree: this does not test the compatibility layer anymore.
Lets update the eventmesh test then so that it covers both scenarios.

k15r: done

tests/end-to-end/backup/pkg/tests/eventmesh/event.go
```diff
@@ -17,7 +21,51 @@ import (
...
-func SendEvent(target, payload, eventType, eventTypeVersion string) error {
+type LegacyEvent struct {
+	EventType        string    `json:"event-type"`
+	EventTypeVersion string    `json:"event-type-version"`
+	EventID          string    `json:"event-id"`
+	EventTime        time.Time `json:"event-time"`
+	Data             string    `json:"data"`
+}
+
+func SendLegacyEvent(target, payload, eventType, eventTypeVersion string) error {
```
nachtmaar: (nit) mhh ... for people who are not aware of that legacy is, a name like SendEventCompatibilityLayer would help I think. It clearly states where the event is sent to. Same for `LegacyEvent`. What about `EventBusV1` ? Not sure which version you are using.

```suggestion
func SendEventToCompatibilityLayer(target, payload, eventType, eventTypeVersion string) error {
```

sayanh: This is unused. Let's remove this.

tests/end-to-end/backup/pkg/tests/eventmesh/event.go
```diff
@@ -17,7 +21,51 @@ import (
...
+	event := LegacyEvent{
```
nachtmaar: (nit)
```suggestion
	event := EventBusV1{
```

tests/end-to-end/backup/pkg/tests/eventmesh/eventmesh.go
```diff
@@ -219,8 +225,12 @@ func (f *eventMeshFlow) WaitForBroker() error {
 	return WaitForBroker(f.messaging, f.brokerName, f.namespace, retry.Delay(10*time.Second), retry.DelayType(retry.FixedDelay), retry.Attempts(10))
 }
 
-func (f *eventMeshFlow) PublishTestEvent() error {
-	return SendEvent(fmt.Sprintf("http://%s.%s.svc.cluster.local", f.applicationName, integrationNamespace), "Dumbidu", f.eventType, f.eventTypeVersion)
+func (f *eventMeshFlow) PublishTestEventToMesh() error {
```
nachtmaar: Should we make the event publishing resilient? Like adding a few retries. I see old `PublishTestEvent ` didn't have it as well.  Anyways, nothing we have to do in this PR.

k15r: might be a good way to improve the test once we refactor it into a standalone test 
see https://github.com/kyma-project/community/pull/442

tests/end-to-end/backup/pkg/tests/eventmesh/event.go
k15r: i chose legacy explicitly, as it states that this one is the old way to send stuff

tests/end-to-end/backup/pkg/tests/eventmesh/event.go
k15r: the target can be specified ... compatibilitylayer is just one target that works with this function

PR #7769, opened 2020-03-24
Disabling Application Connectivity components for Compass scenario 

akgalwas:
**Description**

- disableLegacyConnectivity flag introduced to Application Connector chart
- enableAPIPackages flag replaced with  disableLegacyConnectivity


resources/application-connector/charts/application-registry/templates/role-binding.yaml
```diff
@@ -1,3 +1,4 @@
+{{- if not .Values.global.disableLegacyConnectivity }}
```
Szymongib, line 1: Just a question
Wouldn't it be easier to use `requirements.yaml` in case of disabling whole charts?

akgalwas: Well, I was thinking about it. It is possible for Application Registry but not for other components (e.g. Connector Service). I though it would be easier to comprehend when done this way, but I'm open to discussion.

PR #7775, opened 2020-03-24
istio-telemetry service monitor metric filter configuration

hisarbalik:
Istio-telemetry service configuration with new version changed, this will update service monitor metric filter.


hisarbalik:
removed, fix will be applied after #7904

PR #7787, opened 2020-03-25
Pre-upgrade job to remove Event-Bus resources

radufa:
**Description**

- Create a pre-upgrade job for Event-sources to remove all Event-Bus K8S resources
- After job termination, all Event-Bus resources are deleted

**Related issue(s)**

kyma-project/kyma#7730


components/event-sources/config/100-pre-upgrade-job.yaml
```diff
@@ -0,0 +1,97 @@
+# This job is a kyma custom resource.
```
sayanh: Why did you include the manifests here?

resources/event-sources/templates/100-pre-upgrade-job.yaml
```diff
@@ -0,0 +1,97 @@
+# This job is a kyma custom resource.
+# Please take caution during upgrade to not to remove it
```
sayanh: Why `helm del --purge <>` is not apt here?

The problem with this approach is: if I do `helm ls --all`, `event-bus` is still listed here. Then taking out the `event-bus` chart in the next release will be difficult cause if a chart is missing from resources but is installed, kyma-installer does not uninstall the chart.

resources/event-sources/templates/100-pre-upgrade-job.yaml
```diff
@@ -0,0 +1,136 @@
...
+    resources: ["pods/portforward"]
+    verbs: ["create", "get", "delete"]
+  - apiGroups: [""]
+    resources: ["secrets"]
+    verbs: ["get"]
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
+metadata:
+  name: event-sources-pre-upgrade-helm
```
sayanh: Following annotations are missing:
```
helm.sh/hook: "pre-upgrade"
helm.sh/hook-weight: "1"
 ```

File: resources/event-sources/templates/100-pre-upgrade-job.yaml [hunk 4]
```diff
@@ -0,0 +1,136 @@
...
+  apiGroup: rbac.authorization.k8s.io
+  kind: Role
+  name: event-sources-pre-upgrade
+subjects:
+  - kind: ServiceAccount
+    name: event-sources-pre-upgrade
+---
+kind: ClusterRole
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
```
sayanh: Following annotations are missing:
```
helm.sh/hook: "pre-upgrade"
helm.sh/hook-weight: "0"
 ```

File: resources/event-sources/templates/100-pre-upgrade-job.yaml [hunk 5]
```diff
@@ -0,0 +1,136 @@
...
+    verbs: ["get", "delete"]
+  - apiGroups: [""]
+    resources: ["serviceaccounts"]
+    verbs: ["get", "delete"]
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: RoleBinding
+metadata:
+  name: event-sources-pre-upgrade
+  namespace: kyma-system
```
sayanh: namespace is not required

resources/event-sources/templates/100-pre-upgrade-job.yaml
```diff
@@ -0,0 +1,136 @@
+---
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: event-sources-pre-upgrade
+  namespace: kyma-system
```
sayanh: namespace is not required

resources/event-sources/templates/100-pre-upgrade-job.yaml
```diff
@@ -0,0 +1,136 @@
...
+  kind: ClusterRole
+  name: event-sources-pre-upgrade-helm
+  apiGroup: rbac.authorization.k8s.io
+---
+apiVersion: batch/v1
+kind: Job
+metadata:
+  name: event-sources-pre-upgrade-job
+  namespace: kyma-system
+  annotations:
```
sayanh: Let's add the following annotations.
```
annotations:
  "helm.sh/hook-delete-policy": before-hook-creation,hook-succeeded
```

resources/event-sources/templates/100-pre-upgrade-job.yaml
```diff
@@ -0,0 +1,136 @@
...
+      restartPolicy: Never
+      containers:
+      - name: job
+        image: eu.gcr.io/kyma-project/test-infra/alpine-kubectl:v20200312-fc2a4147
+        terminationMessagePolicy: "FallbackToLogsOnError"
+        command:
+          - /bin/bash
+          - -c
+          - |
+            set -e
```
sayanh: Let's add the shebang `#!/usr/bin/env bash`

resources/event-sources/templates/100-pre-upgrade-job.yaml
```diff
@@ -0,0 +1,136 @@
...
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: event-sources-pre-upgrade
+  namespace: kyma-system
+  annotations:
+    helm.sh/hook: pre-upgrade
+    helm.sh/hook-weight: "0"
+---
+apiVersion: rbac.authorization.k8s.io/v1
```
sayanh: Since `helm del --purge` is used, is `kubectl delete` for each resource needed?

resources/event-sources/templates/100-pre-upgrade-job.yaml
```diff
@@ -0,0 +1,136 @@
+---
```
sayanh: General comments:
- Let's add the following templating everywhere:
```
name: {{ template "controller-manager.fullname" . }}-pre-upgrade-helm
```
- namespace is not required
- annotation: `helm.sh/hook-delete-policy: before-hook-creation,hook-succeeded` needs to be added

resources/event-sources/templates/100-pre-upgrade-job.yaml
```diff
@@ -0,0 +1,136 @@
...
+    resources: ["deployments"]
+    verbs: ["get", "delete"]
+  - apiGroups: [""]
+    resources: ["services"]
+    verbs: ["get", "delete"]
+  - apiGroups: [""]
+    resources: ["serviceaccounts"]
+    verbs: ["get", "delete"]
+---
+apiVersion: rbac.authorization.k8s.io/v1
```
sayanh: Do we still need this `Rolebinding`?

components/event-sources/config/100-pre-upgrade-job.yaml
```diff
@@ -0,0 +1,136 @@
+---
```
sayanh: Pasting from my old comment:

> Why did you include the manifests here?

resources/event-sources/templates/100-pre-upgrade-job.yaml
radufa: before-hook-creation is default.

components/event-sources/config/100-pre-upgrade-job.yaml
radufa: removd

resources/event-sources/templates/100-pre-upgrade-job.yaml
radufa: We have two roles defined "event-sources-pre-upgrade" and "event-sources-pre-upgrade-helm" and only the last one is a ClusterRole. This is why it is more restricted. Both roles are bound the the same service account: "event-sources-pre-upgrade"

resources/event-sources/templates/100-pre-upgrade-job.yaml
radufa: The job can be run multiple times and it tries to delete the Event-Bus resources, also if only some of them partially exists. This could be the case if something happened during upgrade. From the perspective of this job, it could be started again cause it works on incomplete deleted Event-Bus resources too.

resources/event-sources/templates/100-pre-upgrade-job.yaml
radufa: This job will be removed in the next release. This why is very simple now, but completely functional.

resources/event-sources/templates/100-pre-upgrade-job.yaml
radufa: added

resources/event-sources/templates/100-pre-upgrade-job.yaml
radufa: added

resources/event-sources/templates/100-pre-upgrade-job.yaml
```diff
@@ -0,0 +1,136 @@
...
+
+            echo "+ Deleting Event-Bus Helm package"
+            kubectl get -n kyma-installer secret helm-secret \
+              -o jsonpath="{.data['global\.helm\.ca\.crt']}" | base64 --decode > "$HOME/.helm/ca.pem"
+            kubectl get -n kyma-installer secret helm-secret \
+              -o jsonpath="{.data['global\.helm\.tls\.crt']}" | base64 --decode > "$HOME/.helm/cert.pem"
+            kubectl get -n kyma-installer secret helm-secret \
+              -o jsonpath="{.data['global\.helm\.tls\.key']}" | base64 --decode > "$HOME/.helm/key.pem"
+            if [[ $(helm list --tls event-bus) ]]; then
+              helm del --purge --tls event-bus
```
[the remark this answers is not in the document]
radufa: Only to be sure, that if something is happened during the upgrade or if this job is running multiple time it will cleanup the resources. No error will be returned if nothing has to be removed. And it can be used with "ko" too.

resources/event-sources/templates/100-pre-upgrade-job.yaml
radufa: My intention was to have the same code in both "config/" and "templates/" so that the job works with "ko" too. Cause this job will be removed in the next release, I made it as  simple as possible. From this perspective, making "name" configurable can be left out. Is also much simpler if something is happen on the customer side, to analyze the issue, I meant.
I've added the ```helm.sh/hook-delete-policy```.

components/event-sources/config/100-pre-upgrade-job.yaml
radufa: removed

resources/event-sources/templates/100-pre-upgrade-job.yaml
sayanh: What do you think can go wrong in `helm del --purge`? It is not clear to me why `kubectl delete` is needed here. It would make the job more concise and the Role/RoleBinding is not needed as well.

This job is meant to be executed once(`restartPolicy: Never`). https://github.com/kyma-project/kyma/pull/7787/files#diff-48334421faca654acd48a0ada21d19d0R102

resources/event-sources/templates/100-pre-upgrade-job.yaml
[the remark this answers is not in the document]
sayanh: I think ko is not valuable here as the upgrade process is complex and there is no container creation involved here. The upgrade process is a [complex](https://kyma-project.io/docs/#installation-upgrade-kyma) one and helm upgrade and Kyma installer add complexity to it. Simulating with ko will not add value while development in this case.

resources/event-sources/templates/100-pre-upgrade-job.yaml
sayanh: It is not the default value with helm 2.x. More details are [here](https://v2.helm.sh/docs/charts_hooks/#automatically-delete-hook-from-previous-release)ö. Sadly we don't use helm 3.x where it is default.

resources/event-sources/templates/100-pre-upgrade-job.yaml
sayanh: Again, it not clear to me the value of `ko` here.
Templating is considered one of the best practices of Helm. So we should do it.
Namespaces are redundant in the chart as kyma-installer take care of it.

resources/event-sources/templates/100-pre-upgrade-job.yaml
radufa: Done

resources/event-sources/templates/100-pre-upgrade-job.yaml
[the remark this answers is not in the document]
radufa: Only Helm is used now, but I like like to use ko during development phase, cause it reduce the time for a update, deploy, test significantly (from 20 minutes with Helm to 20 seconds...). And the pod generated bu ko contains all the functionality of "classic pod".   Of course, at the end, all should be tested inside of Helm environment.

resources/event-sources/templates/100-pre-upgrade-job.yaml
radufa: OK, done.

resources/event-sources/templates/100-pre-upgrade-job.yaml
radufa: See my early comments. Only Helm is used now.

resources/event-sources/templates/100-pre-upgrade-job.yaml
radufa: See my early comment. The required changes are done.

resources/event-sources/templates/100-pre-upgrade-job.yaml
radufa: Changed as proposed, using only Helm now.

k15r:
event-bus is gone after running the upgrade. 🎈

PR #7796, opened 2020-03-25
Update Director with mocked automatic scenarios assignment feature

crabtree:
**Description**

- Update Director with mocked automatic scenarios assignment feature

**Related issue(s)**


crabtree:
> Queries and mutations included in this director version won't work without adding scopes to that file:
> https://github.com/kyma-project/kyma/blob/master/resources/compass/charts/director/config.yaml

Good catch, definitely that is needed. Fixed.

Issue #5283, opened 2019-08-12
Migration from API.gateway.kyma-project.io/v1alpha2 to v2alpha1


piotrmsc:
**Description**

With new version of API Gateway and crd - api.gateway.kyma-project.io/v2alpha1 we need to migrate old v1alpha2 APIs. This process should be seamless for users, run on the upgrade phase.

Goal:
- Migrate all existing APIs to new version APIs 
- Add troubleshooting guide if the migration fails with action needed to proceed further
- Remove old api-controller tests
- Refactor e2e-external-solution tests to create API v2
- Remove e2e-function tests but align this with wookies team
- For APIs created by old function controller please skip them
- Update docs
- Create proper release notes for this change

**Reasons**

Old API is deprecated and soon UI will be replaced with new one working on APIv2. UI cannot be enabled before migration to new CR.

**Attachments**

PR #7810, opened 2020-03-26
Introduce changes necessary for Api to ApiRule migration

kubadz:
**Description**

A newer version of Oathkeeper supports the usage of `negative lookahead` in path regex, which is necessary to allow complete migration from Api CRD to ApiRule CRD.

Path validation pattern in ApiRule CRD have to be adjusted to support defining prefix, suffix and negative lookahead.

- bump oathkeeper version
- change validation pattern for paths in ApiRule CRD

**Related issue(s)**

Related to #5283

### DIFF
--- a/resources/api-gateway/files/crd-apirules.yaml
+++ b/resources/api-gateway/files/crd-apirules.yaml
@@ -433,7 +433,7 @@ spec:
                     type: array
                   path:
                     description: Path to be exposed
-                    pattern: ^([0-9a-zA-Z./*(]+)
+                    pattern: ^([0-9a-zA-Z./*()?!\\_-]+)
                     type: string
                 required:
                   - path
--- a/resources/api-gateway/files/crd-apirules.yaml
+++ b/resources/api-gateway/files/crd-apirules.yaml
@@ -433,7 +433,7 @@ spec:
                     type: array
                   path:
                     description: Path to be exposed
-                    pattern: ^/([0-9a-zA-Z./*]+)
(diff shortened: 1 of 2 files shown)


resources/api-gateway/files/crd-apirules.yaml
```diff
@@ -433,7 +433,7 @@ spec:
                     type: array
                   path:
                     description: Path to be exposed
-                    pattern: ^/([0-9a-zA-Z./*]+)
+                    pattern: ^([0-9a-zA-Z./*(]+)
```
Tomasz-Smelcerz-SAP, line 436: I know it wasn't there before, but I would add a dash and underscore.
It is used here and there, for example: Some httpbin endpoints contain a dash (`response-headers`)

Tomasz-Smelcerz-SAP, line 436: Why is the opening parenthesis `(` in the list?

kubadz: You are right. I will add it.

kubadz: The parenthesis is on the list because it is required for negative lookahead. For example:
```
- path: (?!.*/suffix.ico\s)/.*
```

Tomasz-Smelcerz-SAP: OK, but then we also need some other characters in the character class: `)`, `?`, `!`, `\`
Perhaps the pattern should be then: `^([0-9a-zA-Z./*()?!\\_-]+)` ?

PR #7822, opened 2020-03-27
Bump director image with automatic scenarios assignment feature

crabtree:
**Description**

- Bump director image with automatic scenarios assignment feature

**Related issue(s)**

### DIFF
--- a/resources/compass/values.yaml
+++ b/resources/compass/values.yaml
@@ -16,7 +16,7 @@ global:
       version: "109f4dc2"
     connectivity_adapter:
       dir:
-      version: "6dc7acc9"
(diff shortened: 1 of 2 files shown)


resources/compass/values.yaml
```diff
@@ -22,7 +22,7 @@ global:
       version: "ffe1a7c3"
     director:
       dir:
-      version: "95fc393e"
+      version: "56884fe1"
```
pkosiec: What about schema migrator? I think it hasn't been bumped before

crabtree: Well, it has all required migrations for this component

PR #7833, opened 2020-03-30
Enable external mTLS cacert management

zralt:
The setup-certs-job only creates a cacert that's valid for 3 months, the cacert needs to be frequently rotated on the MPS via cert-manager. Disable running the certs-setup-job so that cert-manager can handle this.

### DIFF
--- a/resources/compass/charts/connector/templates/certs-setup-job.yaml
+++ b/resources/compass/charts/connector/templates/certs-setup-job.yaml
@@ -1,3 +1,4 @@
+{{- if (.Values.certsSetupJob.enable) }}
 apiVersion: batch/v1
 kind: Job
 metadata:
@@ -30,3 +31,4 @@ spec:
       securityContext:
 {{ toYaml . | indent 12 }}
       {{- end }}
+{{- end -}}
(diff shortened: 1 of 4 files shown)


resources/compass/charts/connector/values.yaml
```diff
@@ -20,6 +20,7 @@ deployment:
     allowPrivilegeEscalation: false
 
 certsSetupJob:
+  enable: true
```
Demonsthere:
```suggestion
  enabled: true
```

resources/compass/charts/connector/templates/certs-setup-role-binding.yaml
```diff
@@ -1,3 +1,4 @@
+{{- if (.Values.certsSetupJob.enable) }}
```
Demonsthere:
```suggestion
{{- if (.Values.certsSetupJob.enabled) }}
```

resources/compass/charts/connector/templates/certs-setup-job.yaml
```diff
@@ -1,3 +1,4 @@
+{{- if (.Values.certsSetupJob.enable) }}
```
Demonsthere:
```suggestion
{{- if (.Values.certsSetupJob.enabled) }}
```

resources/compass/charts/gateway/templates/gateway-certs-mtls-cacert.yaml
```diff
@@ -1,4 +1,4 @@
-{{- if and (.Values.gateway.enabled) }}
+{{- if and (.Values.gateway.enabled) (eq .Values.gateway.manageCerts true) }}
```
Demonsthere:
```suggestion
{{- if and (.Values.gateway.enabled) (.Values.gateway.manageCerts) }}
```

PR #7844, opened 2020-03-30
Clean up old lambda UI view - kyma part

magicmatatjahu:
**Description**

- Clean up old lambda UI view - kyma part:
  - update `serverless` microfrontend: change category from Experimental to Development.
  - PR only support very easy switch from old microfrontend to new. Chart of `lambda-ui` will be removed with main PR for switching to a new serverless.

**Related issue(s)**
https://github.com/kyma-project/console/issues/1639


resources/function-controller/templates/cluster-microfrontend-functions.yaml
```diff
@@ -1,28 +1,32 @@
...
-      navigationPath: functions-preview
+    - label: Functions
+      navigationPath: functions
       viewUrl: '/lambdas'
       order: 1
-    - label: Lambda Details
-      navigationPath: functions-preview/details
+      viewGroup: "_core_ui_"
+      requiredBackendModules:
+        - serverless
```
aerfio, line 21: is it the only one?

magicmatatjahu: Yes, another like `service-catalog` etc is optional, because we handle that in code :)

magicmatatjahu: Yes, another like servicecatalog, application etc are optional. We handle existing them on UI in code :)

Issue #3943, opened 2019-05-07
Upgrade test for Loki


a-thaler:
**Description**
Have a test assuring that logging via Loki is still working after Kyma upgrade.
See https://github.com/kyma-project/kyma/tree/master/tests/end-to-end/upgrade

**Reasons**
We need to assure that an upgrade will not harm existing functionality

**Suggestion**
*Pre upgrade*
Set up a pod causing some logs
Ensure logs are generated (via Loki API)

*Post Upgrade*
Ensure new logs are still generated (via Loki API)
Ensure that old logs are available (via Loki API)

AC:
 There is a test to verify logging works post upgrade

PR #7848, opened 2020-03-31
Upgrade & Integration logging test

shorim:
**Description**

- Integration logging test:
    - refactored test into packages, so that it can be easily reused
    - used kubernetes API instead of running kubectl commands

- Implemented upgrade logging test with the following flow:
    - _pre upgrade_:    deploy a dummy pod which causes some logs 
   - _pre upgrade_:     ensure logs are generated (via Loki API)
   - _post upgrade_:   ensure new logs are still generated (via Loki API)
    - _post upgrade_:  Delete the dummy pod

**Related issue(s)**

#3943


tests/integration/logging/pkg/jwt/jwt.go
```diff
@@ -31,7 +33,13 @@ func getJWT() string {
...
+		return "", errors.Wrap(err, "cannot get JWT token")
 	}
-	return token
+	return token, nil
+}
+
+// SetAuthHeader sets authorization header with JWT
+func SetAuthHeader(token string) string {
+	authHeader := fmt.Sprintf("Authorization: Bearer %s", token)
+	return authHeader
```
suleymanakbas91, line 44: You can directly return fmt.Sprintf without creating this var :)

PR #7849, opened 2020-03-31
Bump gce-proxy images to 1.16

ebensom:
**Description**

Bump gce-proxy images to 1.16 in all deployments throughout Kyma. This ensures we only maintain and assess one gce-proxy image version as part of Protecode scans.


pkosiec:
@ebensom Could you please also propagate these changes to `compass` repo? Thanks a lot!

Issue #7507, opened 2020-03-09
Install ServiceEntry inside Knative-Eventing-Kafka chart


k15r:
**Description**

With `knative-eventing-kafka` on gke and kafka being hosted on azure kafka channels do not work.

**Expected result**

Events get sent

**Actual result**

Events do not get sent:

the `kantive-kafka-channel`s logs show:

```
{"level":"warn","ts":"2020-03-09T08:40:10.404Z","caller":"producer/producer.go:142","msg":"Kafka Error","error":"sasl_ssl://nachtmaar-test-2.servicebus.windows.net:9093/bootstrap: SSL handshake failed: SSL transport error: Connection reset by peer (after 120ms in state CONNECT)"}
```

**Fix**
Install a ServiceEntry for azure:

```yaml
apiVersion: networking.istio.io/v1alpha3
kind: ServiceEntry
metadata:
  name: azure-eventhubs
  namespace: kyma-system
spec:
  hosts:
  - '*.servicebus.windows.net'
  location: MESH_EXTERNAL
  ports:
  - name: kafka
    number: 9093
    protocol: HTTPS
  - name: https
    number: 443
    protocol: HTTPS
```

**AC**
- Add a Troubleshooting Guide which addresses this problem and provides user to add the *ServiceEntry* CR if they face this issue.

nachtmaar:
It seems to work on Minikube and Gardener Azure without ServiceEntry, but on GKE and AKS it seems to be needed.

Idea: We could let `helm` install the `ServiceEntry`. The only dynamic field is `spec.hosts[0]`. This value is available at helm templating time, it is already used [here](https://github.com/kyma-project/kyma/blob/488953a850a9d3438cfcc6df8b1283d4e61259f0/resources/knative-eventing-kafka/templates/broker-secret.yaml#L5).

nachtmaar:
Current approach on how to solve:
1. Change helm template to require hostname and port instead of brokers variable
2. Add ServiceEntry to template and use hostname and port as input for templating
3. Change documentation to reflect new helm template values

PR #7865, opened 2020-04-01
7507 Install service entry inside knative eventing kafka chart

montaro:
See also #7507

### DIFF
--- a/resources/knative-eventing-kafka/templates/broker-secret.yaml
+++ b/resources/knative-eventing-kafka/templates/broker-secret.yaml
@@ -2,7 +2,8 @@
 {{- if hasKey .Values.kafka "brokers" }}
 apiVersion: v1
 data:
-  brokers: {{ .Values.kafka.brokers | b64enc }}
+{{/*  brokers: {{ .Values.kafka.brokers | b64enc }}*/}}
+  brokers: {{ include "knative-kafka.brokers" . | b64enc }}
   {{- if hasKey .Values.kafka "namespace" }}
   namespace: {{ .Values.kafka.namespace | b64enc }}
   {{- end }}
--- a/resources/knative-eventing-kafka/templates/broker-secret.yaml
+++ b/resources/knative-eventing-kafka/templates/broker-secret.yaml
@@ -2,7 +2,6 @@
 {{- if hasKey .Values.kafka "brokers" }}
 apiVersion: v1
 data:
-{{/*  brokers: {{ .Values.kafka.brokers | b64enc }}*/}}
   brokers: {{ include "knative-kafka.brokers" . | b64enc }}
   {{- if hasKey .Values.kafka "namespace" }}
   namespace: {{ .Values.kafka.namespace | b64enc }}
--- a/resources/knative-eventing-kafka/templates/serviceentry.yaml
+++ b/resources/knative-eventing-kafka/templates/serviceentry.yaml
@@ -2,7 +2,7 @@
 apiVersion: networking.istio.io/v1alpha3
 kind: ServiceEntry
 metadata:
-  name: {{ template "knative-kafka.fullname" . }}
+  name: {{ $.Release.Name }}
   namespace: {{ $.Release.Namespace }}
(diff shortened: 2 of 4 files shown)


resources/knative-eventing-kafka/values.yaml
```diff
@@ -16,11 +16,21 @@ image:
 
 # Networking Configuration for the Pods container and the K8S Service
 kafka:
+  #TODO Delete
```
montaro, line 19: @nachtmaar will be deleted

resources/knative-eventing-kafka/values.yaml
```diff
@@ -16,11 +16,21 @@ image:
...
+  secretName: secretz
+  #TODO End Delete
   networking:
     container:
       metricsPort: 8081 # Port exposed by container for metrics
     service:
       metricsPort: 8081 # Port exposed by service for metrics
+  # Kafka Broker URL's (Comma Separated)
+  brokers:
+    hostname: namespace.servicebus.windows.net
```
anishj0shi, line 32: isn't it ***servicebus.windows.net**

resources/knative-eventing-kafka/templates/broker-secret.yaml
```diff
@@ -2,7 +2,8 @@
 {{- if hasKey .Values.kafka "brokers" }}
 apiVersion: v1
 data:
-  brokers: {{ .Values.kafka.brokers | b64enc }}
+{{/*  brokers: {{ .Values.kafka.brokers | b64enc }}*/}}
```
anishj0shi, line 5: this secret is azure specific, why are we templating it?

resources/knative-eventing-kafka/values.yaml
montaro: This is just a placeholder value, should be overriden by the charts values

resources/knative-eventing-kafka/templates/broker-secret.yaml
montaro: to concatenate the hostname with the port

resources/knative-eventing-kafka/templates/serviceentry.yaml
```diff
@@ -0,0 +1,16 @@
+---
+apiVersion: networking.istio.io/v1alpha3
+kind: ServiceEntry
+metadata:
+  name: {{ $.Release.Name }}
```
sayanh, line 5: IMO `knative-kafka.fullname` is the safest thing to use because of the added logic used to derive the value. But I see we haven't used it anywhere.

resources/knative-eventing-kafka/templates/serviceentry.yaml
```diff
@@ -0,0 +1,16 @@
...
+  labels:
+{{ include "knative-kafka.labels" . | indent 4 }}
+spec:
+  hosts:
+    - {{ .Values.kafka.brokers.hostname }}
+  location: MESH_EXTERNAL
+  ports:
+    - name: kafka
+      number: {{ .Values.kafka.brokers.port | int }}
+      protocol: HTTPS
```
sayanh, line 16: I think templating the `protocol:` makes it more flexible.

resources/knative-eventing-kafka/values.yaml
sayanh: Right, can we add a comment to let the users know that, this is a dummy?

montaro: OK

resources/knative-eventing-kafka/templates/serviceentry.yaml
```diff
@@ -0,0 +1,16 @@
+---
+apiVersion: networking.istio.io/v1alpha3
+kind: ServiceEntry
+metadata:
+  name: {{ $.Release.Name }}
+  namespace: {{ $.Release.Namespace }}
```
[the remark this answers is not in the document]
montaro: We agreed to leave it

PR #7873, opened 2020-04-01
Update migration guide for 1.11

bszwarc:
**Description**

- Update migration guide for 1.11 with upgrade procedure and information on Gardener certificates

### DIFF
--- a/docs/migration-guides/1.10-1.11.md
+++ b/docs/migration-guides/1.10-1.11.md
@@ -2,29 +2,35 @@
 
 ## Kyma upgrade
 
-Kyma release 1.11 introduces changes to the upgrade procedure. Follow these steps to perform the upgrade:
+Kyma release 1.11 introduces changes to the upgrade procedure. 
 
-1. Delete the existing `kyma-installer` deployment:
+>**NOTE**: Following the upgrade instruction is enough to start working with Kyma version 1.11. You do not need to perform steps listed in [this](https://kyma-project.io/docs/root/kyma/#installation-upgrade-kyma-upgrade-kyma-to-a-newer-version) document.
+
+To upgrade Kyma to version 1.11, follow these steps:
+
+1. Delete the existing `kyma-installer` deployment.
 
 ```bash
 kubectl delete deployment kyma-installer -n kyma-installer
 ```
-2. Upgrade Tiller:
+2. Once the deployment is deleted, upgrade Tiller:
 
 ```bash
 kubectl apply -f https://raw.githubusercontent.com/kyma-project/kyma/{NEW_KYMA_VERSION}/installation/resources/tiller.yaml
 ```
 3. Verify that Tiller is running:
+
 ```bash
 kubectl get po -n kube-system -l name=tiller -w
 ```
 
-5. Apply the release file to the cluster to trigger the upgrade:
+4. Apply the release file to the cluster to trigger the upgrade:
 ```bash
 kubectl apply -f {KYMA-INSTALLER-CLUSTER-FILE}
 ```
 
-6. To watch the upgrade process, run:
+5. To watch the upgrade process, run:
+
 ```bash
 while true; do \
 kubectl -n default get installation/kyma-installation -o jsonpath="{'Status: '}{.status.state}{', description: '}{.status.description}"; echo; \
--- a/docs/migration-guides/1.10-1.11.md
+++ b/docs/migration-guides/1.10-1.11.md
@@ -2,9 +2,9 @@
 
 ## Kyma upgrade
 
-Kyma release 1.11 introduces changes to the upgrade procedure. 
+In this release, we introduced changes to the upgrade procedure. 
 
->**NOTE**: Following the upgrade instruction is enough to start working with Kyma version 1.11. You do not need to perform steps listed in [this](https://kyma-project.io/docs/root/kyma/#installation-upgrade-kyma-upgrade-kyma-to-a-newer-version) document.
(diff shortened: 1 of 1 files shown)


docs/migration-guides/1.10-1.11.md
```diff
@@ -1,5 +1,63 @@
 # Migrate from v1.10 to v1.11
 
+## Kyma upgrade
+
+Kyma release 1.11 introduces changes to the upgrade procedure. 
+
+>**NOTE**: Following the upgrade instruction is enough to start working with Kyma version 1.11. You do not need to perform steps listed in [this](https://kyma-project.io/docs/root/kyma/#installation-upgrade-kyma-upgrade-kyma-to-a-newer-version) document.
```
kazydek, line 7:
```suggestion
>**NOTE:** Following the upgrade instruction is enough to start working with Kyma version 1.11. You do not need to perform steps listed in [this](https://kyma-project.io/docs/root/kyma/#installation-upgrade-kyma-upgrade-kyma-to-a-newer-version) document.
```

docs/migration-guides/1.10-1.11.md
```diff
@@ -1,5 +1,63 @@
...
+```bash
+while true; do \
+kubectl -n default get installation/kyma-installation -o jsonpath="{'Status: '}{.status.state}{', description: '}{.status.description}"; echo; \
+sleep 5; \
+done
+```
+
+## Clusters provisioned on Gardener
+
+Starting from release 1.11, Gardener issues certificates required for cluster provisioning. Before 1.11, the approach was not unified, and `apiserver-proxy` had certificates issued under the xip.io domain with a self-signed certificate. We unified this behavior to allow Gardener manage all certificates. We have also modified TLS certificate handling to allow Kyma components react on the TLS rotation Gardener provides.
```
kazydek, line 43:
```suggestion
Starting from release 1.11, Gardener issues certificates required for cluster provisioning. Before 1.11, the approach was not unified, and `apiserver-proxy` had certificates issued under the xip.io domain with a self-signed certificate. We unified this behavior to allow Gardener to manage all certificates. We have also modified TLS certificate handling to allow Kyma components to react to the TLS rotation Gardener provides. 
```

docs/migration-guides/1.10-1.11.md
```diff
@@ -1,5 +1,63 @@
...
+kubectl -n default get installation/kyma-installation -o jsonpath="{'Status: '}{.status.state}{', description: '}{.status.description}"; echo; \
+sleep 5; \
+done
+```
+
+## Clusters provisioned on Gardener
+
+Starting from release 1.11, Gardener issues certificates required for cluster provisioning. Before 1.11, the approach was not unified, and `apiserver-proxy` had certificates issued under the xip.io domain with a self-signed certificate. We unified this behavior to allow Gardener manage all certificates. We have also modified TLS certificate handling to allow Kyma components react on the TLS rotation Gardener provides. 
+
+Bear in mind that the domain you request the certificate for is subject to character restritions. This means that if the domain name you provide in the override exceeds 54 characters, applying an override to the cluster will result in the following error:
```
kazydek, line 45:
```suggestion
Bear in mind that the domain you request the certificate for is subject to character restrictions. This means that if the domain name you provide in the override exceeds 54 characters, applying an override to the cluster will result in the following error:
```

docs/migration-guides/1.10-1.11.md
```diff
@@ -1,5 +1,63 @@
...
+
+```bash
+Step error:  Details: Helm install error: rpc error: code = Unknown desc = Job failed: BackoffLimitExceeded
+```
+This happens because the certificate issued for `apiserver-proxy` follows the `apiserver.$DOMAIN` naming convention, and this entire string cannot exceed 64 characters. 
+
+For example, the `abcdefghij.myproject1.shoot.canary.k8s-hana.ondemand.com` domain name has 56 characters which together with the `apiserver` prefix is 65 characters.
```
kazydek, line 52:
```suggestion
For example, the `abcdefghij.myproject1.shoot.canary.k8s-hana.ondemand.com` domain name has 56 characters which gives a total of 65 characters together with the `apiserver` prefix.
```

docs/migration-guides/1.10-1.11.md
```diff
@@ -1,5 +1,63 @@
...
+To avoid any potential issues with the upgrade, make sure that the domain provided in the override does not exceed 54 characters. Run:
+
+```bash
+kubectl get cm -n kyma-installer net-global-overrides -o jsonpath="{.data['global\.\domainName']}" | wc -c
+```
+If the output value exceeds 54 characters, the installation of `apiserver-proxy` will fail. In such a case, you must create a new cluster with shorter name.
```
kazydek, line 58:
```suggestion
If the output value exceeds 54 characters, the installation of `apiserver-proxy` will fail. In such a case, you must create a new cluster with a shorter name.
```

docs/migration-guides/1.10-1.11.md
bszwarc: 👍

docs/migration-guides/1.10-1.11.md
bszwarc: 👍

docs/migration-guides/1.10-1.11.md
bszwarc: 👍

docs/migration-guides/1.10-1.11.md
bszwarc: 👍

PR #7904, opened 2020-04-06
Disable istio-ingressgateway service monitor

hisarbalik:
Service istio-ingressgateway has broken monitoring port configuration, prometheus configure service monitors for istio via auto-discovery fix should be done on istio side


resources/monitoring/values.yaml
```diff
@@ -1779,8 +1779,9 @@ prometheus:
       matchNames:
         - istio-system
     endpoints:
-      - port: prometheus
-        interval: 30s
+# disabled to fix wrong istio-ingressgateway service port configuration
```
a-thaler, line 1782: please link the istio ticket in the comment and name the file to be removed after the issue got fixed

resources/monitoring/templates/kyma-additions/istio-telemetry-prometheus-svc-monitor-patch.yaml
```diff
@@ -0,0 +1,23 @@
+# remove after fix
+apiVersion: monitoring.coreos.com/v1
+kind: ServiceMonitor
+metadata:
+  name: istio-telemetry-svc-monitor
```
a-thaler, line 5: Is that the name of the upstream monitor or self-invented?
If self-invented please align it with the naming of the other monitors of the chart

resources/monitoring/templates/kyma-additions/istio-telemetry-prometheus-svc-monitor-patch.yaml
```diff
@@ -0,0 +1,23 @@
...
+  namespace: {{ $.Release.Namespace }}
+  labels:
+    app: istio-telemetry-svc-monitor-patch
```
a-thaler, line 8: Same for the labelling

PR #7929, opened 2020-04-06
Bump Kyma version and KEB version 

montaro:
See also #7507

- This PR bumps the KEB version and updates the Kyma version
We needed to update the Kyma version in order to reflect the changes to the knative-kafka helm template https://github.com/kyma-project/kyma/pull/7865

-  We needed to update KEB code/version because the installation overrides format changed in knative-kafka helm template https://github.com/kyma-incubator/compass/pull/1120

- Merging this PR will update the MPS-DEV


resources/compass/charts/kyma-environment-broker/values.yaml
```diff
@@ -51,7 +51,7 @@ additionalRuntimeComponents:
   - name: "knative-eventing-kafka"
     namespace: "knative-eventing"
 
-kymaVersion: "1.11.0-rc2"
+kymaVersion: "master-347cca4d"
```
jasiu001, line 54: To use different kyma version you can pass it as a parameter during provisioning. Changing verison in chart is not neccessary.

PR #7930, opened 2020-04-06
Helm 3 installation PoC with k3s (k3d)

pbochynski:
**Description**

The idea is to install Kyma components without tiller. For quick feedback testing is done locally on k3s cluster running in docker (provided by k3d). 

Improvements in comparison to minikube:
- much faster (cluster is up in few second, you need about 6 minutes to have minimal Kyma ready)
- lower CPU and memory consumption (~4GB for Kyma cluster)
- local registry support (you can push images to your local machine without passwords) 
- you can keep docker registry volume even if you delete the cluster (caching images will be my next topic)
- use local.kyma.pro domain which redirects to 127.0.01 (no more minikubeIP)
- no changes in your /etc/hosts
- TLS certificate created in the cluster with cert-manager (and job to replicate it from istio-system to kyma-system)

To test:
```
git clone git@github.com:kyma-project/kyma.git
cd kyma
git checkout -b pbochynski-helm3 master
git pull git://github.com/pbochynski/kyma.git helm3
cd resources
# check dependencies (docker, kubectl, helm3, k3d)
./kyma-k3d.sh
```


ebensom:
Hi @pbochynski,

It is not a good idea to drop helm as a resource manager and solely rely on templating + kubectl apply. While this might be enough for simple POCs for installation scenarios, it completely lacks:

- ability to delete resources which were renamed/deleted from chart
- ability to remove deleted resource attributes (due to the 3-way merge of kubectl apply)
- revision history
- manual or automated rollbacks

I don't even mention support for some advanced, orchestasion-level deployment features, like canary rollout, or chaos engineering, which completely lacks in current kyma-operator.

If you want orchestration of kyma upgrades/patches/reconfigurations in 100-cluster scale with confidence, you should not follow this direction.

pbochynski:
Hi @ebensom 
I considered removing helm on the early stage of my PoC. With helm template I was able to see better the output and spot problems and fix them. You can check latest version which uses helm install again. But to be honest I don't agree with your arguments and still consider removing helm as deployer. I think we need to consolidate some resources, make them independent and simpler. I definitely do not see Kyma as a complex deployment. It should be simple but powerful.

valentinvieriu:
Would be great to have a blog post about this too 
[https://github.com/kyma-project/kyma/pull/7930](https://github.com/kyma-project/kyma/pull/7930)

Issue #7888, opened 2020-04-02
Event Service Dumped on sending events from commerce


anishj0shi:
On Sending Events from CCv2, event publishing failed. 

**This was fixed on restarting the Event-service pod**
**This issue is intermittent.**

```
kubectl logs -n kyma-integration commerce-event-service-79bb55d79-kwjxt commerce-event-service
2020/04/02 05:43:42 http: panic serving 127.0.0.1:55032: invalid WriteHeader code 0
goroutine 117957 [running]:
net/http.(*conn).serve.func1(0xc0005ea140)
	/usr/local/go/src/net/http/server.go:1769 +0x139
panic(0x1199920, 0xc000240070)
	/usr/local/go/src/runtime/panic.go:522 +0x1b5
net/http.checkWriteHeaderCode(...)
	/usr/local/go/src/net/http/server.go:1093
net/http.(*response).WriteHeader(0xc00008a1c0, 0x0)
	/usr/local/go/src/net/http/server.go:1127 +0x607
github.com/kyma-project/kyma/components/event-service/internal/externalapi.writeJSONResponse(0x15341e0, 0xc00008a1c0, 0xc0000ace70)
	/workspace/go/src/github.com/kyma-project/kyma/components/event-service/internal/externalapi/genericeventhandler.go:138 +0x250
github.com/kyma-project/kyma/components/event-service/internal/externalapi.getEventsHandler.func1(0x15341e0, 0xc00008a1c0, 0xc000444800)
	/workspace/go/src/github.com/kyma-project/kyma/components/event-service/internal/externalapi/genericeventhandler.go:97 +0x565
net/http.HandlerFunc.ServeHTTP(0xc000491700, 0x15341e0, 0xc00008a1c0, 0xc000444800)
	/usr/local/go/src/net/http/server.go:1995 +0x44
github.com/kyma-project/kyma/components/event-service/internal/externalapi.(*maxBytesHandler).ServeHTTP(0xc0002c9b80, 0x15341e0, 0xc00008a1c0, 0xc000444800)
	/workspace/go/src/github.com/kyma-project/kyma/components/event-service/internal/externalapi/genericeventhandler.go:34 +0xe0
github.com/kyma-project/kyma/components/event-service/vendor/github.com/gorilla/mux.(*Router).ServeHTTP(0xc0001b8000, 0x15341e0, 0xc00008a1c0, 0xc000444600)
[...]

PR #7951, opened 2020-04-08
Refactored error handling in event-service

sayanh:
**Description**

- Refactored error handling while sending a CE to eventmesh from compatibility layer in event service

**Related issue(s)**

See #7888


components/event-service/internal/externalapi/genericeventhandler_test.go
```diff
@@ -20,101 +20,116 @@ func TestNewEventsHandler(t *testing.T) {
...
 	tests := []struct {
-		name           string
-		givenPayload   string
-		wantCode       int
-		wantEventID    string
-		wantErrorType  string
-		wantErrorField string
+		name                      string
+		givenPayload              string
+		withEventMeshResponseCode int
```
nachtmaar, line 27: (nit) I think the original authors wanted to follow [this principe](https://github.com/uber-go/guide/blob/master/style.md#test-tables) there all input data for the test is prefixed with `given`, and the expected results are prefixed with `want`.

```suggestion
		givenEventMeshResponseCode int
```

components/event-service/internal/externalapi/genericeventhandler.go
```diff
@@ -41,6 +41,7 @@ func NewEventsHandler(config *mesh.Configuration, maxRequestSize int64) http.Han
 
 type permanentRedirectionHandler struct {
 	location string
+	handler  http.Handler
```
marcobebway, line 44: Why do we need this http handler here?

components/event-service/internal/externalapi/genericeventhandler_test.go
```diff
@@ -20,101 +20,116 @@ func TestNewEventsHandler(t *testing.T) {
...
+			wantEventID:               "8954ad1c-78ed-4c58-a639-68bd44031de0",
 		},
 	}
 
+	var mux *http.ServeMux
+	var closeFn meshtesting.CloseFunction
 	for _, test := range tests {
+
 		t.Run(test.name, func(t *testing.T) {
+			mux, closeFn = setupEventMeshServer(source, v1Endpoint, requestSize, test.withEventMeshResponseCode, t)
```
marcobebway, line 130: Can this run in parallel? Also applies for the test itself **TestNewEventsHandler**.

components/event-service/internal/externalapi/genericeventhandler_test.go
marcobebway: Same comment!

sayanh: Fixed

components/event-service/internal/externalapi/genericeventhandler.go
sayanh: Found this is unused. Removed.

components/event-service/internal/externalapi/genericeventhandler_test.go
sayanh: I tried to execute it in parallel(`t.Parallel()`) which works intermittently. Executing 100 runs makes the code run into panics. 

Parallelizing the tests is a great idea. Since the tests are indeterministic if we execute them in parallel, I think it is better if we take care of this in a separate issue and tackle the problem holistically for the entire test-suite.

sayanh: The panic was due to the sharing of variables that we fixed and executed the tests again. To our surprise, the gain was not evident.
```
➜ go test -count 1000 github.com/kyma-project/kyma/components/event-service/internal/externalapi (not in parallel)
ok      github.com/kyma-project/kyma/components/event-service/internal/externalapi      10.071s
~ at ☸️  gke-release-1-11-1-rc1.a.build.kyma-project.io took 13s
➜ go test -count 1000 github.com/kyma-project/kyma/components/event-service/internal/externalapi (in parallel)
ok      github.com/kyma-project/kyma/components/event-service/internal/externalapi      34.754s
```

Hence we decided to stay away from parallel for now.

PR #7958, opened 2020-04-08
Api-gateway test improvements

Demonsthere:
**Description**

- Add more verbosity to logs
- Add a Resource.Get function for verification

**Related issue(s)**


tests/integration/api-gateway/gateway-tests/pkg/resource/batch.go
```diff
@@ -17,6 +17,7 @@ func (b *Batch) CreateResources(k8sClient dynamic.Interface, resources ...unstru
 	for _, res := range resources {
 		resourceSchema, ns, _ := GetResourceSchemaAndNamespace(res)
 		b.ResourceManager.CreateResource(k8sClient, resourceSchema, ns, res)
+		b.ResourceManager.GetResource(k8sClient, resourceSchema, ns, res.GetName())
```
Tomasz-Smelcerz-SAP, line 20: I think we need a bit more sophisticated approach. I will put a general comment on the PR.

Issue #7955, opened 2020-04-08
Enable creation of AuthorizationPolicies out of the box


sayanh:
**Description**

Right now, the creation of AuthorizationPolicy is [disabled](https://github.com/kyma-project/kyma/blob/master/resources/application-connector/values.yaml#L11). In order to enable the security feature either one of the following steps can be done:
- Change the value [here](https://github.com/kyma-project/kyma/blob/master/resources/application-connector/values.yaml#L11) (Con: development environment will get very restricted)
- Pass `enabled` as an override to Kyma-installer (See https://github.com/kyma-incubator/compass/tree/master/docs/kyma-environment-broker)

**Reasons**

Security reasons

See https://github.com/kyma-project/kyma/issues/7240 for more details

anishj0shi:
> Con: development environment will get very restricted

I'm fine with a restricted environment, because we do not support in-cluster event-dispatch. 
And If someone really needs to send events from a microservice inside the cluster. they should follow the application-connector flow. or we will support them in editing the existing authorization policy.

anishj0shi:
We cannot enable istio Authorization Policy by default, as this causes unknown errors at the time of backup, We'll revisit this with Knative Open Policy at event-mesh level.

PR #7961, opened 2020-04-08
Enable Istio authorization policy for the validator and the HTTP source adapter

marcobebway:
**Description**

Enable Istio authorization policy for the validator and the HTTP source adapter.

**Related issue(s)**
- https://github.com/kyma-project/kyma/issues/7955

### DIFF
--- a/resources/application-connector/values.yaml
+++ b/resources/application-connector/values.yaml
@@ -8,7 +8,7 @@ global:
   disableLegacyConnectivity: false
   isLocalEnv: false
   namespace: kyma-integration
-  strictMode: disabled
+  strictMode: enabled
   istio:
     namespace: istio-system
     tls:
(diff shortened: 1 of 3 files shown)


components/application-operator/charts/application/templates/authorization-policy.yaml
```diff
@@ -60,6 +60,7 @@ spec:
     - from:
         - source:
             principals:
+              - cluster.local/ns/e2e-upgrade-test/sa/test-e2e-upgrade-test
```
k15r, line 63: i know that this is necessary with the current implementation of the test, but i really dont like that we reference test resources in a production component. Isn't there another solution?
e.g. Use one of the production service accounts for the test?

marcobebway: Yes, you're correct. I created an authorization policy for the upgrade test under its charts.

PR #7962, opened 2020-04-08
Enhance logging test

shorim:
**Description**

- Increased the time interval between retrials of querying loki to avoid overwhelming loki server
- Query logs starting only from current time, so that the execution of query would be faster on loki side
- use GO HTTP client instead of executing curl commands

### DIFF
--- a/tests/end-to-end/upgrade/pkg/tests/logging/pkg/logstream/logstream.go
+++ b/tests/end-to-end/upgrade/pkg/tests/logging/pkg/logstream/logstream.go
@@ -67,7 +67,7 @@ func WaitForDummyPodToRun(namespace string, coreInterface kubernetes.Interface) 
 }
 
 // Test querys loki api with the given label key-value pair and checks that the logs of the dummy pod are present
-func Test(labelKey string, labelValue string, authHeader string) error {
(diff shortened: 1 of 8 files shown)


tests/end-to-end/upgrade/pkg/tests/logging/pkg/logstream/logstream.go
```diff
@@ -103,3 +103,28 @@ func Cleanup(namespace string, coreInterface kubernetes.Interface) error {
...
 	return nil
 }
+
+func doGet(url string, authHeader string) (string, error) {
+	req, err := http.NewRequest("GET", url, nil)
+	if err != nil {
+		return "", errors.Wrap(err, "cannot create a new HTTP request")
+	}
+	req.Header.Add("Authorization", authHeader)
+	client := &http.Client{
```
piotrmiskiewicz, line 113: It is not a good idea to create a new client for every call. Can you just share it between calls?

shorim: yes, I agree it is better to share the client.
The changes are pushed now

PR #7987, opened 2020-04-09
add nats-init component

k15r:
**Description**

- nats-init was part of event bus, with event-bus gone we need to put it in its own component
**Related issue(s)**

#7730

### DIFF
--- a/components/nats-init/Dockerfile
+++ b/components/nats-init/Dockerfile
@@ -1,0 +1,8 @@
+FROM alpine:3.11
+LABEL source=git@github.com:kyma-project/kyma.git
+
+WORKDIR /nats-streaming/
+
+COPY scripts/prepare-config.sh .
+
+ENTRYPOINT ["/nats-streaming/prepare-config.sh"]
--- a/components/nats-init/Makefile
+++ b/components/nats-init/Makefile
@@ -1,0 +1,11 @@
+APP_NAME = nats-init
+APP_PATH = components/$(APP_NAME)
+BUILDPACK = eu.gcr.io/kyma-project/test-infra/buildpack-golang-toolbox:v20190930-d28d219
+SCRIPTS_DIR = $(realpath $(shell pwd)/../..)/common/makefiles
+
+include $(SCRIPTS_DIR)/docker.mk
+
+.PHONY: path-to-referenced-charts
+path-to-referenced-charts:
+	@echo "resources/core"
+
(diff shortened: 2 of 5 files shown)


components/nats-init/Dockerfile
```diff
@@ -0,0 +1,8 @@
+FROM alpine:3.11
+LABEL source=git@github.com:kyma-project/kyma.git
+
+WORKDIR /nats-streaming/
+
+COPY scripts/prepare-config.sh .
```
sayanh, line 6: Why don't we need the licences?
```
COPY licenses ./licenses
```

k15r: which licenses would you copy in?

sayanh: The licences which are pulled [here](https://github.com/kyma-project/kyma/blob/master/common/makefiles/generic-make-go.mk#L84).

k15r: `docker.mk` does not pull any licenses. This target is only valid in `generic-make-go.mk`

sayanh: The question is don't we need the licenses anymore? Cause we had licenses in the previous image of nats-init.

sayanh: Please feel free to ignore my comments if needed. I don't wanna block this PR. It is just not clear to me why we don't need the licenses anymore.

k15r: it was empty in the previous images as well

k15r: so, i checked our license puller script https://github.com/kyma-project/test-infra/blob/master/prow/scripts/license-puller.sh
there is just no code that would pull a license for a bash only image.

components/nats-init/Makefile
```diff
@@ -0,0 +1,11 @@
+APP_NAME = nats-init
+APP_PATH = components/$(APP_NAME)
+BUILDPACK = eu.gcr.io/kyma-project/test-infra/buildpack-golang-toolbox:v20190930-d28d219
```
nachtmaar, line 3: (nit) If we only use `docker.mk`, is `BUILDPACK` and `APP_PATH` still required? From my understanding only `APP_NAME` and `SCRIPTS_DIR` is required.

nachtmaar, line 3:
```suggestion
```

PR #7993, opened 2020-04-10
Add documentation for migration from Api to ApiRule

kubadz:
**Description**

- add documentation for migration from Api to ApiRule
- remove API Gateway documentation

**Related issue(s)**

#5283


docs/api-gateway-v2/03-04-migration.md
```diff
@@ -0,0 +1,144 @@
+---
+title: Migration from the previous Api resources
+type: Details
+---
+
+The migration is done automatically by a job. During the migration the old Api object is being migrated to an APIRule object, what may cause temporary unavailability of the exposed service. After the successful migration the old resource is removed. The migration might be skipped for some Api configuration, in which case the manual migration process may be done. It won't break the existing service exposure, but any changes to the Api object won't be effective.
```
jakkab: > The migration is done automatically by a job.

I suggest you describe the job and explain when it is triggered.

> During the migration the old Api object is being migrated to an APIRule object, what may cause temporary unavailability of the exposed service.

Avoid repetitions (migration, migrated). The "temporary unavailability" sounds awkward. My suggestion: "During the migration, the old Api object is being translated to an APIRule object, which may result in a temporary downtime of the exposed service.

> After the successful migration the old resource is removed.

"The original resource is deleted as a part of the migration process."

> The migration might be skipped for some Api configuration, in which case the manual migration process may be done.

Let's turn it into a warning, shall we? Then you can provide a link to the relevant section that describes manual migration. 

> It won't break the existing service exposure, but any changes to the Api object won't be effective.

This sentence is confusing and I can't quite figure out what it refers to.

docs/api-gateway-v2/03-04-migration.md
```diff
@@ -0,0 +1,144 @@
...
+
+As Api resources are removed after a successful migration, all the resources that are left should be considered as not migrated. To get more details fetch logs from the migrator job
+
+```shell script
+kubectl logs api-gateway-api-migrator-job -n kyma-system
+```
+
+You will find logs saying which Api was not migrated and a reason for that.
+
+If the Api was not migrated due to the invalid status or a blacklisted label, you either shouldn't or will not be able to do the migration.
```
jakkab, line 22: Now this part here is begging for a header. Without a proper introduction (details on the migration job, see my previous comment), it's just out of context. My position is that you should spell it out to the user: the migration is a part of Kyma upgrade. You can verify your APIs once the upgrade process is over.

docs/api-gateway-v2/03-04-migration.md
jakkab: this is by no means outdated!

docs/api-gateway-v2/03-04-migration.md
```diff
@@ -0,0 +1,144 @@
...
+      config:
+        trusted_issuers:
+        - "https://dex.kyma.local"
+        - "https://auth.kyma.local"
+        jwks_urls:
+        - "http://dex-service.kyma-system.svc.cluster.local:5556/keys"
+        - "http://auth-service.kyma-system.svc.cluster.local:5556/keys"
+```
+
+The important part to notice is the last path configuration containing regex with a negative lookahead. It is used to exclude paths handled by other path settings from the `/.*` path, as there can't be two different configurations for a single path. There is one exception to that, which applies only if the same `excludedPaths` element is present on all `authentication` settings, so the specific path doesn't require any authentication at all. In that case, a rule with `handler: allow` should be created, and the path doesn't have to be excluded using a negative lookahead.
```
jakkab, line 108:
```suggestion
Note that the last path configuration contains a regular expression with a negative lookahead. It is used to exclude paths handled by other path settings from the `/.*` path, as there can't be two different configurations for a single path. There is one exception to that, which applies only if the same `excludedPaths` element is present on all `authentication` settings, so the specific path doesn't require any authentication at all. In that case, a rule with `handler: allow` should be created, and the path doesn't have to be excluded using a negative lookahead.
```

> It is used to exclude paths handled by other path settings from the `/.*` path, as there can't be two different configurations for a single path.

This sentence might sound super confusing to regex greenhorns (like me). I think I get your point, but I'm not sure.

docs/api-gateway-v2/03-04-migration.md
```diff
@@ -0,0 +1,144 @@
+---
+title: Migration from the previous Api resources
+type: Details
+---
+
+The migration is done automatically by a job. During the migration, the old Api object is being migrated to an APIRule object, what may cause temporary unavailability of the exposed service. After the successful migration, the old resource is removed. The migration might be skipped for some Api configuration, in which case the manual migration process may be done. It won't break the existing service exposure, but any changes to the Api object won't be effective.
```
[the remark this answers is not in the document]
Tomasz-Smelcerz-SAP: It is described in the api-migrator README.md file. Perhaps a link will be enough? Or we can copy-paste this information here.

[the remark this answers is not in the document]
kubadz: I described below which cases should not be migrated manually (invalid status or a blacklisted label), which means that all other cases (exactly one - `API is too complex for automatic translation`) may be migrated.

docs/api-gateway-v2/03-04-migration.md
```diff
@@ -0,0 +1,144 @@
...
+
+The basic difference between Apis and APIRules authentication configuration is that while Api allows to enable the authentication and disable it on specific paths, the APIRule has an approach where you specifies what authentication should be used per specific path (including the possibility to set it for all paths) and the paths might not cross each other. Another important difference is that Api supports a list of issuers and jwks URIs, but excluded paths are set independently on both. It means that in the example below to access `/exact/path/to/resource.jpg` path the token issued from `https://auth.kyma.local` is required, to access any path starting with `/pref/` the token issued from `https://dex.kyma.local` is required, to access the `/no/auth/needed/resource.html` no token is required because it is excluded for both settings and to access all other paths the token from one of the issuers is required.
```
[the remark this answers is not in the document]
kubadz: The actual behaviour for both resources:

**Api** = enable authentication on all paths + disable on some paths (blacklist)
**APIRule** = enable authentication on specific paths (whitelist)

Anyway, I revised it a little. Please take a look if it is more understandable now.

docs/api-gateway-v2/03-04-migration.md
```diff
@@ -0,0 +1,148 @@
...
+
+You will find logs saying which Api was not migrated and a reason for that.
+
+If the Api was not migrated due to the invalid status or a blacklisted label, you either shouldn't or will not be able to do the migration.
+
+## Manual migration
+
+The manual migration process consists of the following steps:
+
+1. Get Api resource you want to migrate. It would be helpful to have it open during the whole migration process.
```
jakkab, line 32:
```suggestion
1. Fetch the Api resource you want to migrate:
```

docs/api-gateway-v2/03-04-migration.md
```diff
@@ -0,0 +1,148 @@
...
+
+```shell script
+kubectl get api <NAME> -n <NAMESPACE>
```
jakkab, line 35:
```suggestion
kubectl get api <NAME> -n <NAMESPACE> -o yaml
```

docs/api-gateway-v2/03-04-migration.md
```diff
@@ -0,0 +1,148 @@
...
+```
+
+2. Create APIRule resource with a different host
```
jakkab, line 38:
```suggestion
2. Create an APIRule resource based on the original Api object
```

docs/api-gateway-v2/03-04-migration.md
```diff
@@ -0,0 +1,148 @@
...
+
+Here is the documentation for both custom resources, which may be useful during the migration:
+
+https://kyma-project.io/docs/1.11/components/api-gateway/#custom-resource-api-sample-custom-resource
+https://kyma-project.io/docs/1.11/components/api-gateway-v2#custom-resource-api-rule
+
+Only the `spec` part of the CRD will be described. `metadata` field can be adjusted in any way and `status` field should not be copied.
+
+Starting from the `service` field of APIRule, copy `name` and `port` fields from the `service` field of Api object. Please set `host` value to any temporary value including the domain. For example, if the `hostname` value of Api was set to `sample-service.kyma.local` you can change it to `temp-sample-service.kyma.local`. Just make sure that the hostname is not used by other services on your cluster. That value will be changed in a later step, but for now it needs to be different than the original value. Set `gateway` field of APIRule to the Istio Gateway used to expose services. By default it will be `kyma-gateway.kyma-system.svc.cluster.local`.
```
jakkab, line 47: I suggest using bullet points for clarity here.

docs/api-gateway-v2/03-04-migration.md
```diff
@@ -0,0 +1,148 @@
...
+
+The basic difference between Apis and APIRules authentication configuration is that while Api allows to enable the authentication for the whole service and disable it on specific paths only, the APIRule has an approach where you specify what authentication should be used per specific path (including the possibility to set it for all paths) and the paths must not cross each other. Another important difference is that Api supports a list of issuers and jwks URIs, but excluded paths are set independently on both. It means that in the example below to access `/exact/path/to/resource.jpg` path the token issued from `https://auth.kyma.local` is required, to access any path starting with `/pref/` the token issued from `https://dex.kyma.local` is required, to access the `/no/auth/needed/resource.html` no token is required because it is excluded for both settings and to access all other paths the token from one of the issuers is required.
```
jakkab, line 51: > ... and the paths must not cross each other.

Perhaps "must not overlap" would be more accurate? You could also provide an example of different yet overlapping paths.

jakkab: > It means that in the example below to access `/exact/path/to/resource.jpg` path the token issued from `https://auth.kyma.local` is required, to access any path starting with `/pref/` the token issued from `https://dex.kyma.local` is required, to access the `/no/auth/needed/resource.html` no token is required because it is excluded for both settings and to access all other paths the token from one of the issuers is required.

Again, bullet points would make this section transparent. - Path, authentication method (if any), commentary.

docs/api-gateway-v2/03-04-migration.md
Tomasz-Smelcerz-SAP: I would replace the part: "as there can't be two different configurations for a single path" with: "as the path definitions in an APIRule **must** be unique."

docs/api-gateway-v2/03-04-migration.md
Tomasz-Smelcerz-SAP: I prefer more strict approach: "must not overlap" -> "must be unique"

docs/api-gateway-v2/03-04-migration.md
```diff
@@ -0,0 +1,148 @@
+---
+title: Migration from the previous Api resources
+type: Details
+---
+
+The migration is done automatically by a job, which runs during the Kyma upgrade. During the migration, the old Api object is being translated to an APIRule object, which may result in a temporary downtime of the exposed service. The original resource is deleted as a part of the migration process. The tool used by the job is described [in this document](https://github.com/kyma-project/kyma/blob/master/components/api-gateway-migrator/README.md#api-gateway-migrator).
+
+>**CAUTION:** The migration might be skipped for some Api configuration, in which case the [manual migration process](#manual-migration) may be done. Not proceeding with the manual migration won't break the existing service exposure, but any further changes or removal of the Api resource won't affect how the service is exposed - the original configuration will be used.
```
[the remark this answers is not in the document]
kubadz: I do not fully agree with the beginning. 

> Some Api configurations are too complex to be automatically migrated.

It is true, but there are also other cases, where the complexity doesn't matter (invalid status or a blacklisted label). 
Maybe something like this?

> Some Api configurations are too complex or do not meet all requirements to be automatically migrated.

docs/api-gateway-v2/03-04-migration.md
kubadz: IMHO "must be unique" is not sufficient here. You can say that `/.*` and `/something/.*` are two unique paths, but you can't use that configuration, as they overlap.

docs/api-gateway-v2/03-04-migration.md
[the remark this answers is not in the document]
jakkab: purrrrfect!

docs/api-gateway-v2/01-01-api-gateway.md
```diff
@@ -2,7 +2,7 @@
 title: Overview
 ---
 
->**CAUTION:** This implementation replaces the API Gateway that is based on the Api custom resource. The services you exposed and secured using the deprecated implementation require no action, as the API Gateway Controller co-exists with the API Controller in the cluster. Expose and secure new services and functions secured with OAuth2 using the `v2` implementation described in this documentation topic.
+>**CAUTION:** This implementation replaces the API Gateway that is based on the Api custom resource. The services you exposed and secured using the deprecated implementation are automatically converted to use the new implementation. Read [this](/components/api-gateway-v2#details-migration-from-the-previous-api-resources) document to learn how to check if a service was properly migrated.
```
Tomasz-Smelcerz-SAP, line 5:
```suggestion
>**CAUTION:** This implementation replaces the API Gateway that is based on the Api custom resource. The services you exposed and secured using the deprecated implementation are automatically migrated to use the new implementation. Read [this](/components/api-gateway-v2#details-migration-from-the-previous-api-resources) document to learn how to check if a service was properly migrated.
```

docs/api-gateway-v2/03-04-migration.md
```diff
@@ -0,0 +1,154 @@
+---
+title: Migration from the previous Api resources
+type: Details
+---
+
+The migration is done automatically by a job, which runs during the Kyma upgrade. During the migration, an old Api object is being translated to an APIRule object, which may result in a temporary downtime of the exposed service. The original resource is deleted as a part of the migration process. The tool used by the job is described [in this document](https://github.com/kyma-project/kyma/blob/master/components/api-gateway-migrator/README.md#api-gateway-migrator).
+
+>**CAUTION:** Some Api configurations are too complex or do not meet all requirements to be automatically migrated. To make sure all your services are exposed via APIRules, [verify the outcome](#verify-automatic-migration) of the migration job and follow the [manual migration process](#manual-migration) guide if necessary. Not proceeding with the manual migration won't break the existing service exposure, but any further changes or removal of the Api resource won't affect how the service is exposed - the original configuration will be used.
```
jakkab, line 8:
```suggestion
>**CAUTION:** Some Api configurations are too complex or do not meet all requirements to be automatically migrated. To make sure all your services are exposed via APIRules, [verify the outcome](#verify-automatic-migration) of the migration job and follow the [manual migration process](#manual-migration) guide if necessary. Skipping the migration won't break the existing service exposure, but any further changes or removal of the Api resource won't affect how the service is exposed - the original configuration will be used.
```

docs/api-gateway-v2/03-04-migration.md
```diff
@@ -0,0 +1,154 @@
...
+
+## Verify automatic migration
+
+List the existing Api objects after the Kyma upgrade is finished. Run this command:
+
+```shell script
+kubectl get apis --all-namespaces
+```
+
+As Api objects are removed after a successful migration, all the listed resources should be considered not migrated. Fetch logs from the job's pod to learn the reason for skipping the migration.
```
jakkab, line 18:
```suggestion
As Api objects are removed after a successful migration, all the listed resources are considered not migrated. Fetch logs from the job's pod to learn the reason for skipping the migration. 
```

docs/api-gateway-v2/03-04-migration.md
```diff
@@ -0,0 +1,154 @@
...
+```shell script
+kubectl get api <NAME> -n <NAMESPACE> -o yaml
+```
+
+2. Create an APIRule resource based on the original Api object
+
+Here is the documentation for both custom resources, which may be useful during the migration:
+
+https://kyma-project.io/docs/1.11/components/api-gateway/#custom-resource-api-sample-custom-resource
+https://kyma-project.io/docs/1.11/components/api-gateway-v2#custom-resource-api-rule
```
jakkab, line 42: >**TIP:**

maybe?

docs/api-gateway-v2/03-04-migration.md
```diff
@@ -0,0 +1,154 @@
...
+ * Set `gateway` field of APIRule to the Istio Gateway used to expose services. By default it will be `kyma-gateway.kyma-system.svc.cluster.local`.
+
+The configuration of the `rules` field of APIRule is more complex and depends on the `authentication` configuration of APIRule. As all the basic scenarios are covered by automatic migration, this explanation will only concern the configurations that are not handled automatically.
+
+The basic difference between Apis and APIRules authentication configuration is that while Api allows to enable the authentication for the whole service and disable it on specific paths only, the APIRule has an approach where you specify what authentication should be used per specific path (including the possibility to set it for all paths) and the paths must not overlap. Another important difference is that Api supports a list of issuers and jwks URIs, but excluded paths are set independently on both. In the example below:
```
Tomasz-Smelcerz-SAP, line 53:
```suggestion
The most important difference between Apis and APIRules authentication configuration is that Api is configured with a list of JWT issuers and APIRule is configured with a list of path definitions.
In the Api you can define a list of excluded paths for an issuer. Requests for these paths will not require authentication by this issuer but will require authentication from other configured issuers.
The APIRule has an approach where you specify what authentication should be used per the path definition.
Because paths are defined using regular expressions, you can define an expression that matches every path, or a subset of possible paths, or a single one.
Remember that path definitions in the APIRule must not overlap - if they do, requests to such paths are rejected. That's why translating an Api with multiple issuers having different excluded paths requires sophisticated regex definitions in a corresponding APIRule.
In the example below:
```

docs/api-gateway-v2/03-04-migration.md
```diff
@@ -0,0 +1,154 @@
...
+Below is the list showing how ApiRule path value corresponds to the excludedPaths values from Api resource, and what negative lookahead value should be added to the `/.*` path:
+
+| excludedPaths expression type | description | excludedPaths sample value | path value | negative lookahead value |
+|---|---|---|---|---|
+|**exact**| no changes to be made | /exact/path/to/resource.jpg | /exact/path/to/resource.jpg | (?!/exact/path/to/resource.jpg) |
+|**prefix**| add `.*` to the end | /pref/ | /pref/.* | (?!/pref/.*) |
+|**suffix**| add `.*` to the beginning; in negative lookahead add `\s` to the end | /suffix.ico | .*/suffix.ico | (?!.*/suffix.ico\s) |
+|**regex**| no changes to be made | /anything.* | /anything.* | (?!/anything.*) |
+
+When the configuration is ready, please create the APIRule object and test if the service is working as expected on the new host. It should work the same on both hosts.
```
Tomasz-Smelcerz-SAP, line 129: At this point, the host is still temporary and you can't test it, unless it points to something existing.

docs/api-gateway-v2/03-04-migration.md
```diff
@@ -0,0 +1,154 @@
...
+https://kyma-project.io/docs/1.11/components/api-gateway-v2#custom-resource-api-rule
+
+Only the `spec` part of the CRD will be described. `metadata` field can be adjusted in any way and `status` field should not be copied.
+
+Starting from the `service` field of APIRule:
+ * Copy `name` and `port` fields from the `service` field of Api object. 
+ * Set `host` value to any temporary value including the domain. For example, if the `hostname` value of Api was set to `sample-service.kyma.local` you can change it to `temp-sample-service.kyma.local`. Just make sure that the hostname is not used by other services on your cluster. That value will be changed in a later step, but for now it needs to be different than the original value. 
+ * Set `gateway` field of APIRule to the Istio Gateway used to expose services. By default it will be `kyma-gateway.kyma-system.svc.cluster.local`.
+
+The configuration of the `rules` field of APIRule is more complex and depends on the `authentication` configuration of APIRule. As all the basic scenarios are covered by automatic migration, this explanation will only concern the configurations that are not handled automatically.
```
strekm, line 51:
```suggestion
The configuration of the `rules` field of APIRule is more complex and depends on the `authentication` configuration of Api. As all the basic scenarios are covered by automatic migration, this explanation will only concern the configurations that are not handled automatically.
```

docs/api-gateway-v2/03-04-migration.md
jakkab: it is pointing to an existing service:

- old API exposes a service at https://service-a.my.kyma
- new APIRule exposes the same service at https://temporary-service-a.my.kyma

Both hosts should be accessible at this point:
>Copy `name` and `port` fields from the `service` field of Api object.

docs/api-gateway-v2/03-04-migration.md
```diff
@@ -0,0 +1,154 @@
...
+Below is the list showing how ApiRule path value corresponds to the excludedPaths values from Api resource, and what negative lookahead value should be added to the `/.*` path:
+
+| excludedPaths expression type | description | excludedPaths sample value | path value | negative lookahead value |
+|---|---|---|---|---|
+|**exact**| no changes to be made | /exact/path/to/resource.jpg | /exact/path/to/resource.jpg | (?!/exact/path/to/resource.jpg) |
+|**prefix**| add `.*` to the end | /pref/ | /pref/.* | (?!/pref/.*) |
+|**suffix**| add `.*` to the beginning; in negative lookahead add `\s` to the end | /suffix.ico | .*/suffix.ico | (?!.*/suffix.ico\s) |
+|**regex**| no changes to be made | /anything.* | /anything.* | (?!/anything.*) |
+
+When the configuration is ready, please create the APIRule object and test if the service is working as expected on the new host. It should work the same on both hosts.
```
strekm, line 129:
```suggestion
When the configuration is ready, create the APIRule object and test if the service is working as expected on the new host. It should work the same on both hosts.
```

docs/api-gateway-v2/03-04-migration.md
```diff
@@ -0,0 +1,154 @@
...
+
+3. Remove Api resource and dependant resources
+
+```shell script
+kubectl delete api <API_NAME> -n <NAMESPACE>
+kubectl delete virtualservice <API_NAME> -n <NAMESPACE>
```
strekm, line 135: why do we need to manually delete vs? At this point, we don't have api-controller already?

docs/api-gateway-v2/03-04-migration.md
```diff
@@ -0,0 +1,154 @@
...
+```
+
+Please make sure that the Virtual Service resource is deleted before proceeding to the next instruction.
```
strekm, line 138:
```suggestion
Make sure that the Virtual Service resource is deleted before proceeding to the next instruction.
```

docs/api-gateway-v2/03-04-migration.md
```diff
@@ -0,0 +1,154 @@
...
+| excludedPaths expression type | description | excludedPaths sample value | path value | negative lookahead value |
+|---|---|---|---|---|
+|**exact**| no changes to be made | /exact/path/to/resource.jpg | /exact/path/to/resource.jpg | (?!/exact/path/to/resource.jpg) |
+|**prefix**| add `.*` to the end | /pref/ | /pref/.* | (?!/pref/.*) |
+|**suffix**| add `.*` to the beginning; in negative lookahead add `\s` to the end | /suffix.ico | .*/suffix.ico | (?!.*/suffix.ico\s) |
+|**regex**| no changes to be made | /anything.* | /anything.* | (?!/anything.*) |
+
+When the configuration is ready, please create the APIRule object and test if the service is working as expected on the new host. It should work the same on both hosts.
+
+3. Remove Api resource and dependant resources
```
jakkab, line 131:
```suggestion
3. Remove Api resource and dependent resources
```

docs/api-gateway-v2/03-04-migration.md
```diff
@@ -0,0 +1,154 @@
...
```
Tomasz-Smelcerz-SAP, line 131: I would move this to the end, last step after successful migration. This object is still a blueprint for the migration even if it doesn't actually work.

docs/api-gateway-v2/03-04-migration.md
kubadz: Yes. The manual migration should occur after the upgrade and an api-controller is removed during the upgrade.

docs/api-gateway-v2/03-04-migration.md
jakkab: This is why @kubadz suggests to fetch it at the very beginning od the process. It is not possible to swap points 3 and 4: you need to delete the original API in order to avoid conflicting hosts in Virtual Services. Am I correct @kubadz?

docs/api-gateway-v2/03-04-migration.md
kubadz: @jakkab is right

docs/api-gateway-v2/03-04-migration.md
kubadz: I do not agree.

1. It can't be removed as the last step, because it occupies the original host. It has to be removed before the last step, otherwise it won't work.
2. In step 1 of manual migration, a user fetches the Api configuration, so it can be used as a blueprint. During the migration nothing changes in the Api resource, so there is no need to have another "blueprint"

docs/api-gateway-v2/03-04-migration.md
jakkab: disambiguation: **maybe** you could use the `tip` preset here. 😆

docs/api-gateway-v2/03-04-migration.md
Tomasz-Smelcerz-SAP: OK, my fault - I wasn't precise enough:)
I was meaning to move to the end only removal of the Api object. VS must be removed at this point (because it occupies host), Api can be removed at the end.

docs/api-gateway-v2/03-04-migration.md
kubadz: done

docs/api-gateway-v2/03-04-migration.md
kubadz: done

docs/api-gateway-v2/03-04-migration.md
kubadz: done

docs/api-gateway-v2/03-04-migration.md
```diff
@@ -0,0 +1,164 @@
...
+
+Starting from the `service` field of APIRule:
+ * Copy `name` and `port` fields from the `service` field of Api object. 
+ * Set `host` value to any temporary value including the domain. For example, if the `hostname` value of Api was set to `sample-service.kyma.local` you can change it to `temp-sample-service.kyma.local`. Just make sure that the hostname is not used by other services on your cluster. That value will be changed in a later step, but for now it needs to be different than the original value. 
+ * Set `gateway` field of APIRule to the Istio Gateway used to expose services. By default it will be `kyma-gateway.kyma-system.svc.cluster.local`.
+
+The configuration of the `rules` field of APIRule is more complex and depends on the `authentication` configuration of Api. As all the basic scenarios are covered by automatic migration, this explanation will only concern the configurations that are not handled automatically.
+
+The most important difference between Apis and APIRules authentication configuration is that Api is configured with a list of JWT issuers set for all paths of a service and APIRule is configured with a list of per path definitions.
```
Tomasz-Smelcerz-SAP, line 52:
```suggestion
The most important difference between Apis and APIRules authentication configuration is that Api is configured with a list of JWT issuers set for all paths of a service and APIRule is configured with a list of path definitions.
```

docs/api-gateway-v2/03-04-migration.md
```diff
@@ -0,0 +1,164 @@
+---
+title: Migration from the previous Api resources
+type: Details
+---
+
+The migration is done automatically by a job, which runs during the Kyma upgrade. During the migration, an old Api object is being translated to an APIRule object, which may result in a temporary downtime of the exposed service. The original resource is deleted as a part of the migration process. The tool used by the job is described [in this document](https://github.com/kyma-project/kyma/blob/master/components/api-gateway-migrator/README.md#api-gateway-migrator).
```
jakkab, line 6:
```suggestion
The migration is done automatically by a job that runs during the Kyma upgrade. During the migration, an old Api object is being translated to an APIRule object, which may result in a temporary downtime of the exposed service. The original resource is deleted as a part of the migration process. The tool used by the job is described [in this document](https://github.com/kyma-project/kyma/blob/master/components/api-gateway-migrator/README.md#api-gateway-migrator).
```

docs/api-gateway-v2/03-04-migration.md
jakkab, line 52:
```suggestion
The most important difference in authentication configuration between Apis and APIRules is that an Api object is configured with a list of JWT issuers set for all paths of a service, while an APIRule is configured with a list of path definitions along with their respective authentication methods.
```

docs/api-gateway-v2/03-04-migration.md
```diff
@@ -0,0 +1,197 @@
...
+
+>**NOTE:** This step focuses on the`spec` part of the CRD. You can `metadata` field in a preferred way and do not copy the [`status`](components/api-gateway-v2#custom-resource-api-rule-additional-information) field from the original API resource.
+
+a. Replace the default value (`kyma-gateway.kyma-system.svc.cluster.local`) for the **gateway** parameter with Istio Gateway used to expose services. 
+
+b. Copy the values for **service.name** and **service.port** parameters from the corresponding fields of the API object. 
+
+c. Set the value for **service.host** parameter to any temporary value which includes the domain. For example, if the value for the **hostname** parameter of the API object was set to `sample-service.kyma.local`, change it to `temp-sample-service.kyma.local` so that differs from the original value. Make sure other services on your cluster do not use this hostname. 
+
+d. Configure the **rules** parameter of APIRule.
```
bszwarc, line 57: I was wondering - maybe we can have an example with such a configuration here - I mean we have this big section explaining all the differences and comparing API but maybe an example would be good?

docs/api-gateway-v2/03-04-migration.md
```diff
@@ -0,0 +1,144 @@
...
+| **spec.service.name**| Copy the value of the corresponding parameter of the Api CR.|
+| **spec.service.port**| Copy the value of the corresponding parameter of the Api CR.|
+| **spec.service.host**| Set the value for **service.host** parameter to any temporary value which includes the domain. For example, if the value for the **hostname** parameter of the Api CR was set to `sample-service.kyma.local`, change it to `temp-sample-service.kyma.local` so that differs from the original value. Make sure other services on your cluster do not use this hostname. |
+| **spec.rules**| The **spec.rules** parameter allows you to provide authentication by defining a list of paths along with the authentication methods. The way of handling authentication differs for Api CR and APIRule CR. Consider the section below as a point of reference when defining this part of specification.|
+
+Use the **spec.rules** parameter to specify the authentication type per path and use a regular expression to match every path, a subset of possible paths, or a single one.  
+
+See the example:
```
bszwarc, line 62: If we need the Api CR example for comparison, we can rewrite it to:

See the example of an Api CR:
```
authentication:
- type: JWT
  jwt:
    issuer: https://dex.kyma.local
    jwksUri: http://dex-service.kyma-system.svc.cluster.local:5556/keys
    triggerRule:
      excludedPaths:
      - exact: /exact/path/to/resource.jpg
      - exact: /no/auth/needed/resource.html
- type: JWT
  jwt:
    issuer: https://auth.kyma.local
    jwksUri: http://auth-service.kyma-system.svc.cluster.local:5556/keys
    triggerRule:
      excludedPaths:
      - prefix: /pref/
      - exact: /no/auth/needed/resource.html
```
In this configuration:

To access the /exact/path/to/resource.jpg path, you need a token issued by https://auth.kyma.local.
To access any path starting with /pref/ you need a token issued by https://dex.kyma.local.
To access the /no/auth/needed/resource.html you don't need any token because the path is excluded for both configurations.
To access all other paths you need a token from one of the issuers.

The APIRule corresponding to this example would look as follows:

.... and the rest follows.

docs/api-gateway-v2/03-04-migration.md
```diff
@@ -44,105 +44,22 @@ Follow these steps:
...
-
-## Authentication of Apis and APIRules
+2. Create an APIRule CR based on the Api CR's specification.
 
-To properly configure the authentication mechanism for the new APIRule, learn more about the differences in authentication configuration for Api and APIRule CRs:
+>**NOTE:** Do not copy the [**status**](components/api-gateway-v2#custom-resource-api-rule-additional-information) parameter from the original Api CR.
 
-| Api CR       | APIRule CR| 
-|--------------| ----------|
-|Configured with a list of JWT issuers set for all paths of a service.| Configured with a list of path definitions along with their respective authentication methods.|
```
kubadz: IMHO the information in this table is important - it explains the most important difference between both Custom Resources.

docs/api-gateway-v2/03-04-migration.md
```diff
@@ -44,105 +44,22 @@ Follow these steps:
...
-| You can define a list of excluded paths for a given issuer. Requests for these paths won't require authentication by this issuer but will require authentication from other configured issuers.| You can specify what authentication to use per path definition and use a regular expression to match every path, or a subset of possible paths, or a single one. |
+| Parameter | Action  |
+|-----------|---------|
+| **metadata.name**| Specify the name of the API Rule.| 
+| **spec.gateway**| Replace the default value (`kyma-gateway.kyma-system.svc.cluster.local`) with Istio Gateway used to expose services.|
```
kubadz, line 54: The user should not replace the default value. Instead, most probably they should use the default value, as it is the default Istio Gateway in Kyma used to expose services. 

There was no field specifying gateway in an Api CR, as it was set by default to the value `kyma-gateway.kyma-system.svc.cluster.local`. In APIRule the user can use a different Isito Gateway, but it is not something we suggest.

docs/api-gateway-v2/03-04-migration.md
```diff
@@ -180,18 +97,48 @@ rules:
...
-However, if the same `excludedPaths` element is present throughout the authentication settings, a particular path doesn't require any authentication. In that case, create a rule with `handler: allow`, so that you don't have to exclude the path using a negative lookahead.
+>**NOTE:** Path definitions in the APIRule CR cannot overlap, otherwise the requests to such paths are rejected. 
 
-  </details>
-</div>
+The **spec.rules** parameter corresponds to **spec.authentication** defined in the Api CR where you can specify a [list multiple JWT token issuers](https://kyma-project.io/docs/components/api-gateway/#details-security-specify-multiple-jwt-token-issuers) to allow secured access and use the **triggerRule.excludedPaths** parameter to exclude paths that don't require any authentication. 
 
+If you want to exclude paths in APIRule definition, for example to avoid two different configurations for a single path, you can use a regular expression with a negative lookahead. In the example, such an expression is used to in the last path exclude the paths already handled by other configuration settings from the `/.*` path.
```
kubadz, line 104: Just to be sure. Shouldn't this part:
`[...] used to in the last path exclude [...]`
be `[...] used in the last path to exclude [...]` instead? Or was that intended?

docs/api-gateway-v2/03-04-migration.md
```diff
@@ -180,18 +97,48 @@ rules:
...
         - "http://auth-service.kyma-system.svc.cluster.local:5556/keys"
 ```
-The value for the last **rules.path** parameter is a regular expression with a negative lookahead. Its purpose is to exclude the paths already handled by other configuration settings from the `/.*` path to avoid two different configurations for a single path. 
-However, if the same `excludedPaths` element is present throughout the authentication settings, a particular path doesn't require any authentication. In that case, create a rule with `handler: allow`, so that you don't have to exclude the path using a negative lookahead.
+>**NOTE:** Path definitions in the APIRule CR cannot overlap, otherwise the requests to such paths are rejected. 
 
-  </details>
-</div>
+The **spec.rules** parameter corresponds to **spec.authentication** defined in the Api CR where you can specify a [list multiple JWT token issuers](https://kyma-project.io/docs/components/api-gateway/#details-security-specify-multiple-jwt-token-issuers) to allow secured access and use the **triggerRule.excludedPaths** parameter to exclude paths that don't require any authentication.
```
kubadz, line 102: IMHO this seem to suggest that the **spec.rules** in the example corresponds to **spec.authentication** in the linked Api CR, but those are two separate examples. It would make sense if you used an API example, like you described in your comment. https://github.com/kyma-project/kyma/pull/7993#discussion_r413574223

docs/api-gateway-v2/03-04-migration.md
```diff
@@ -180,18 +97,48 @@ rules:
...
+3. When the configuration is ready, create the APIRule object and test if the service is working as expected on the new host. It should work the same on both hosts.
+
+4. Remove the dependent resources of the Api CR:
+
+   ```shell script
+     kubectl delete virtualservice {API_NAME} -n {NAMESPACE}
+    ```
+Make sure that the Virtual Service resource is deleted before proceeding.
+
+5. If the Api CR was secured with an authentication mechanism, delete the Policy resource:
```
kubadz, line 126: The Policy CR is also a dependent resource - I would suggest merging it with point 4 or use 
```
4. Remove the VirtualService resource:
```
instead of
```
4. Remove the dependent resources of the Api CR:
```

docs/api-gateway-v2/03-04-migration.md
```diff
@@ -1,38 +1,38 @@
...
     ```
 
->**NOTE:** If the migration process skipped the Api resources due to the invalid status or a blacklisted label, you either should not or will not be able to migrate them.
-
 ## Manual migration
 
-This guide shows how you can manually migrate Api resources to API Rules.
+This guide shows how you can manually migrate Api CRs to APIRule CRs.
 
 >**NOTE:** Before you start the manual migration process, see the [Api CR](/components/api-gateway/#custom-resource-api-sample-custom-resource) and [APIRule CR](/components/api-gateway-v2#custom-resource-api-rule) documents for the custom resource detailed definition.
```
kubadz, line 37: The problem with the link `[Api CR](/components/api-gateway/#custom-resource-api-sample-custom-resource)` is that it will not work after merging this PR, as we are removing that part of the documentation. Previously I used links to specific version of the documentation (https://kyma-project.io/docs/1.11/components/api-gateway#custom-resource-api-sample-custom-resource) but I see that you changed it. It applies to all links referring to the api-controller (api-gateway v1) documentation.

docs/api-gateway-v2/03-04-migration.md
kubadz: I think that it would be better to have the example.

docs/api-gateway-v2/03-04-migration.md
```diff
@@ -0,0 +1,171 @@
+---
+title: Migration from Api to APIRule custom resources
+type: Details
+---
+
+
+The migration from Api to APIRule custom resources (CRs) is performed automatically by a job that runs during the Kyma upgrade. During this process, the [API Gateway Migrator tool](https://github.com/kyma-project/kyma/blob/master/components/api-gateway-migrator/README.md#api-gateway-migrator) translates the existing Api CRs to APIRule CRs and deletes original resources.
+
+**CAUTION:** Migrating resources may result in a temporary downtime of the exposed service.
```
bszwarc, line 9:
```suggestion
>**CAUTION:** Migrating resources may result in a temporary downtime of the exposed service. 
```

docs/api-gateway-v2/03-04-migration.md
bszwarc: It's there, just at a later point :) After both examples.

Issue #7624, opened 2020-03-16
Strict mtls settings for ingress gateway


piotrmsc:
**Description**

During istio installation we apply Destination Rule for ingress gateway with mtls set to `permissive`, take a look [here](https://github.com/kyma-project/kyma/blob/master/resources/istio/files/destination-rules.yaml). When we enabled strict mTLS mesh wide, this DR was created for testing pods. Currently, we need to check if it can be deleted, ingress will have strict mTLS settings.

**Reasons**

From a security perspective it should be aligned. We have mesh wide strict mTLS settings but ingress is set with permissive mode. 

**Attachments**

Demonsthere:
https://github.com/kyma-project/kyma/pull/8010

Demonsthere:
Results of the tasks:
- For cluster operations, we can safely remove the DR after an improvement in the `api-gateway` tests.  These tests, as well as `dex`, are using an outdated ingress client which is required for minikube ip address discovery. 
- As the `dex` tests are run only in the minikube environment, we have decided to leave the DR for minikube and switch the client for cluster operations

PR #8010, opened 2020-04-14
Move ingress-gateway DR to minikube env only

Demonsthere:
**Description**

- Move ingress DestinationRule to localenv
- Add migration job for DR deletion
- Update base image for api-gateway tests

**Related issue(s)**

- Implements #7624


resources/istio/files/destination-rules-minikube.yaml
```diff
@@ -0,0 +1,23 @@
+---
+# DestinationRule to disable (m)TLS when talking to helm tiller, as it doesn't have a sidecar.
+apiVersion: networking.istio.io/v1alpha3
+kind: DestinationRule
+metadata:
+  name: tiller-deploy
+spec:
+  host: tiller-deploy.kube-system.svc.cluster.local
```
piotrmsc, line 8: tiller should be applied always :)

resources/istio/templates/configmap-extra-manifests.yaml
```diff
@@ -5,4 +5,8 @@ metadata:
   name: kyma-extra-manifests
 data:
   destination-rules.yaml: |-
-{{tpl (.Files.Get "files/destination-rules.yaml" | printf "%s" | indent 4) .}}
+{{- if .Values.global.isLocalEnv }}
+{{ tpl (.Files.Get "files/destination-rules-minikube.yaml" | printf "%s" | indent 4) . }}
+{{- else }}
```
piotrmsc, line 10: and then this if else can be simplified to single if containing additional DR for ingress :)

resources/istio/templates/rbac-migration-remove-dr.yaml
```diff
@@ -0,0 +1,39 @@
+---
+apiVersion: v1
+kind: ServiceAccount
```
piotrmsc, line 3: Do we need that? this DR will be removed on upgrade by tiller?

resources/istio/templates/job-migration-remove-dr.yaml
```diff
@@ -0,0 +1,27 @@
...
+apiVersion: batch/v1
+kind: Job
+metadata:
+  annotations:
+    helm.sh/hook-delete-policy: "hook-succeeded,before-hook-creation"
+    helm.sh/hook: "pre-upgrade"
+    helm.sh/hook-weight: "5"
+  name: remove-dr
+spec:
+  activeDeadlineSeconds: 1000
```
piotrmsc, line 11: See below

resources/istio/templates/rbac-migration-remove-dr.yaml
Demonsthere: The DR is not managed by tiller, we create by with kubectl in the install job, so we need to remove it during upgrades

piotrmsc: Yeah, I forgot that this was changed and is applied as a part of istio-installer job :)

resources/istio/templates/rbac-migration-remove-dr.yaml
```diff
@@ -0,0 +1,39 @@
+---
```
piotrmsc, line 1: Please create a follow up for us to clean this after next release :)

resources/istio/templates/job-migration-remove-dr.yaml
```diff
@@ -0,0 +1,27 @@
...
+  template:
+    spec:
+      containers:
+        -
+          command:
+            - bash
+            - "-c"
+            - |
+              set -e
+              kubectl delete dr istio-ingressgateway -n istio-system
```
piotrmsc, line 22: nit :  we can prepare ourselves for the future :D by addidn --ignore-not-found  flag. That will prevent upgrade plan from failing after 1.12 before we remove this delete

resources/istio/templates/rbac-migration-remove-dr.yaml
Demonsthere: Followup created: https://github.com/kyma-project/kyma/issues/8066

PR #8013, opened 2020-04-14
Exposing deployment strategy + fix ORY Hydra resources

lumi017:
**Description**

- Exposing deployment strategy in dex, ory and compass helm charts
- Fix ORY Hydra resources

**Related issue(s)**


resources/ory/values.yaml
```diff
@@ -80,13 +80,14 @@ hydra:
...
     dangerousAllowInsecureRedirectUrls: false
     existingSecret: "ory-hydra-credentials"
-  resources:
-    limits:
-      cpu: 100m
-      memory: 256Mi
-    requests:
-      cpu: 50m
-      memory: 128Mi
+    deployment:
```
ebensom, line 83: resources: are now under hydra.hydra., it should be moved one level upper to hydra.

resources/compass/charts/cockpit/values.yaml
```diff
@@ -19,3 +19,9 @@ service:
 deployment:
   image:
     pullPolicy: IfNotPresent
+  # By default deploymentStrategy is set to rollingUpdate with maxSurge of 25% and maxUnavailable of 25% . you can change type to `Recreate` or can uncomment `rollingUpdate` specification and adjust them to your usage.
+  strategy: {}
+    # rollingUpdate:
+    #   maxSurge: 25%
+    #   maxUnavailable: 25%
+    # type: RollingUpdate
```
pkosiec, line 27: Hmm I wonder if we need these two comments here.
I would suggest something more condensed, like:

```yaml
strategy: {} # Read more: https://kubernetes.io/docs/concepts/workloads/controllers/deployment/#strategy
```

If you like this idea, please change it in every chart.

resources/ory/values.yaml
lumi017: ooups, fixed

resources/compass/charts/cockpit/values.yaml
lumi017: sure, changed

resources/compass/charts/director/values.yaml
```diff
@@ -10,6 +10,14 @@ deployment:
   dbPool:
     maxOpenConnections: 30
     maxIdleConnections: 2
+  liveness:
```
ebensom, line 13: Please remove liveness: and readiness:, I think those should have been added by a faulty merge from upstream.

lumi017: cleaned up

Issue #7449, opened 2020-03-03
Refactor event-service informer


sayanh:
**Description**

https://github.com/kyma-project/kyma/blob/master/components/event-service/internal/events/subscribed/events.go#L45

needs to be ruled out by an informer sync timeout and the app needs to fail fast if the informers are not synced. Without the informer sync timeout, the informer sync may fail but the app will still run and will be non functional.

Reference:
https://github.com/kyma-project/kyma/blob/master/components/event-bus/internal/knative/util/kn-eventing.go#L170

montaro:
Reference link is not available anymore

montaro:
Reference:
```
// waitForInformersSyncOrDie blocks until all informer caches are synced, or panics after a timeout.
func waitForInformersSyncOrDie(f evinformers.SharedInformerFactory) {
	ctx, cancel := context.WithTimeout(context.Background(), informerSyncTimeout)
	defer cancel()

	err := hasSynced(ctx, f.WaitForCacheSync)
	if err != nil {
		log.Fatalf("Error waiting for caches sync: %s", err)
	}
}
```

PR #8018, opened 2020-04-14
Refactor event-service informer

montaro:
Applying a get or die pattern when waiting for the informers cache sync.

See also #7449

### DIFF
--- a/components/event-service/internal/events/subscribed/events.go
+++ b/components/event-service/internal/events/subscribed/events.go
@@ -45,6 +45,7 @@ type eventsClient struct {
 
 // waitForInformersSyncOrDie blocks until all informer caches are synced, or panics after a timeout.
 func waitForInformersSyncOrDie(f externalversions.SharedInformerFactory) {
+	log.Println("waiting for informers caches sync...")
 	ctx, cancel := context.WithTimeout(context.Background(), informerSyncTimeout)
 	defer cancel()
 
@@ -67,11 +68,13 @@ func hasSynced(ctx context.Context, fn waitForCacheSyncFunc) error {
 	stopWait := make(chan struct{})
 	defer close(stopWait)
 
+	// close the synced channel after the `WaitForCacheSync()` to announce that it finished the execution
 	go func() {
 		fn(stopWait)
 		close(synced)
 	}()
 
+	// wait for closure of the goroutine or return an error if it timed out
 	select {
 	case <-ctx.Done():
 		return ctx.Err()
@@ -87,6 +90,7 @@ func NewEventsClient(client versioned.Interface) EventsClient {
 	lister := informerFactory.Eventing().V1alpha1().Triggers().Lister()
 	ctx := signals.NewContext()
 	informerFactory.Start(ctx.Done())
+	informerFactory.WaitForCacheSync(ctx.Done())
 	waitForInformersSyncOrDie(informerFactory)
 
 	return &eventsClient{
--- a/components/event-service/internal/events/subscribed/events.go
+++ b/components/event-service/internal/events/subscribed/events.go
@@ -1,6 +1,11 @@
 package subscribed
 
 import (
+	"context"
+	"log"
+	"reflect"
+	"time"
+
 	"k8s.io/apimachinery/pkg/labels"
 
 	knv1alpha1 "knative.dev/eventing/pkg/apis/eventing/v1alpha1"
@@ -14,6 +19,8 @@ const (
 	keySource           = "source"
 	keyEventType        = "type"
 	keyEventTypeVersion = "eventtypeversion"
+
+	informerSyncTimeout = time.Second * 5
 )
 
 //EventsClient interface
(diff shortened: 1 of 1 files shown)


components/event-service/internal/events/subscribed/events.go
```diff
@@ -36,13 +43,51 @@ type eventsClient struct {
...
+}
+
 //NewEventsClient function creates client for retrieving all active events
 func NewEventsClient(client versioned.Interface) EventsClient {
 	informerFactory := externalversions.NewSharedInformerFactory(client, 0)
 	lister := informerFactory.Eventing().V1alpha1().Triggers().Lister()
 	ctx := signals.NewContext()
 	informerFactory.Start(ctx.Done())
-	informerFactory.WaitForCacheSync(ctx.Done())
+	waitForInformersSyncOrDie(informerFactory)
```
Abd4llA, line 90: can we add a log that we're waiting for cache to sync, ideally it'd be good to see in the log if this service is blocked on waiting for cache to sync

components/event-service/internal/events/subscribed/events.go
```diff
@@ -36,13 +43,51 @@ type eventsClient struct {
...
+	// gets canceled
+	stopWait := make(chan struct{})
+	defer close(stopWait)
+
+	go func() {
+		fn(stopWait)
+		close(synced)
+	}()
+
+	select {
```
Abd4llA, line 75: I think this code needs more of an explanation in the comment

components/event-service/internal/events/subscribed/events.go
```diff
@@ -36,13 +43,51 @@ type eventsClient struct {
...
+	// synced gets closed as soon as fn returns
+	synced := make(chan struct{})
+
+	// closing stopWait forces fn to return, which happens whenever ctx
+	// gets canceled
+	stopWait := make(chan struct{})
+	defer close(stopWait)
+
+	go func() {
+		fn(stopWait)
```
Abd4llA, line 71: do you know if fn (i.e `externalversions.SharedInformerFactory.WaitForCacheSync` ) will return an error ? and what will happen if it does? does it get propagated to `waitForInformersSyncOrDie` ?

Abd4llA:
@montaro can you also please add a full description to the PR ?

components/event-service/internal/events/subscribed/events.go
montaro: Actually the `WaitForCacheSync()` doesn't return any errors from the signature, also checking the implementation, it doesn't throw any errors to be caught.

Abd4llA: after tracing the code here https://github.com/kubernetes/client-go/blob/master/informers/factory.go#L158 which eventually calls this https://github.com/kubernetes/client-go/blob/4529075edf97f1129ef696f95a4bc67c4798f334/tools/cache/shared_informer.go#L254 I think you need to handle the boolean result stored in the `map[type]` . Because according to the last link, this is set to false when an error happens and the controller needs to exit.

Abd4llA: even though I think the only condition where it will be set to false is if `stopWait` was closed, but I'd say it's better to handle the false result from a contract POV regardless of the implementation internal.

components/event-service/internal/events/subscribed/events.go
```diff
@@ -36,13 +43,55 @@ type eventsClient struct {
...
 
+// waitForInformersSyncOrDie blocks until all informer caches are synced, or panics after a timeout.
+func waitForInformersSyncOrDie(f externalversions.SharedInformerFactory) {
+	log.Println("waiting for informers caches sync...")
+	ctx, cancel := context.WithTimeout(context.Background(), informerSyncTimeout)
+	defer cancel()
+
+	err := hasSynced(ctx, f.WaitForCacheSync)
+	if err != nil {
+		log.Fatalf("Error waiting for caches sync: %s", err)
```
radufa, line 54: Is the call to ```log.Fatalf``` the best solution here?

Abd4llA: I think the idea is to die early if the sync didn't work.

radufa: Yes, this was clear for me, but it implies a pod restarting. Should the issue be solved after this restart? And if not? As I remember, the informer has a "resync" logic, with a resync duration. Could this be used somehow?
It remembers me to STONITH, which should be the last solution... :)

components/event-service/internal/events/subscribed/events.go
```diff
@@ -36,13 +43,62 @@ type eventsClient struct {
...
+		return ctx.Err()
+	case <-synced:
+	}
+
+	return nil
+}
+
 //NewEventsClient function creates client for retrieving all active events
 func NewEventsClient(client versioned.Interface) EventsClient {
 	informerFactory := externalversions.NewSharedInformerFactory(client, 0)
```
montaro, line 97: @radufa I think with passing the Default Resync Duration to the factory the resync may happen automatically.

radufa:
lgtm

PR #8024, opened 2020-04-15
Fix overrides CM in KEB

mszostok:
**Description**

- Fix overrides CM in KEB

The problem is that in KEB we are sending the override but in master-91a18c32 version the override name was changed from `enableAPIPackages` to `disableLegacyConnectivity` (https://github.com/kyma-project/kyma/pull/7769)


mszostok:
The following changes are also done in compass repo: https://github.com/kyma-incubator/compass/pull/1176

PR #8026, opened 2020-04-15
Update image for api-gateway tests

Demonsthere:
**Description**

- Change image to alpine-kubectl
- Add ca-certificates

**Related issue(s)**

- Implements #7624


tests/integration/api-gateway/gateway-tests/main_test.go
```diff
@@ -46,29 +48,46 @@ const jwtAndOauthOnePathApiruleFile = "jwt-oauth-one-path-strategy.yaml"
...
-	ReqDelay   uint   `envconfig:"TEST_REQUEST_DELAY,default=5"`
-	Domain     string `envconfig:"DOMAIN"`
+	HydraAddr     string        `envconfig:"TEST_HYDRA_ADDRESS"`
+	User          string        `envconfig:"TEST_USER_EMAIL"`
+	Pwd           string        `envconfig:"TEST_USER_PASSWORD"`
+	ReqTimeout    uint          `envconfig:"TEST_REQUEST_TIMEOUT,default=180"`
+	ReqDelay      uint          `envconfig:"TEST_REQUEST_DELAY,default=5"`
+	Domain        string        `envconfig:"DOMAIN"`
+	ClientTimeout time.Duration `envconfig:"TEST_CLIENT_TIMEOUT,default=10s"` //Don't forget the unit!
+	IngressClient bool          `envconfig:"USE_INGRESS_CLIENT,default=false"`
```
piotrmsc, line 65: Maybe we should have an ENV that indicates local(minikube) environment instead of ingress client because later if needed we can perform more actions, not only configuring proper HTTP client. The `ingress client` is not straightforward, without looking into the implementation of the client you will not know why it is used (this is old implementation drawback)

tests/integration/api-gateway/gateway-tests/main_test.go
```diff
@@ -46,29 +48,46 @@ const jwtAndOauthOnePathApiruleFile = "jwt-oauth-one-path-strategy.yaml"
...
 
-	var conf Config
 	if err := envconfig.Init(&conf); err != nil {
 		t.Fatalf("Unable to setup config: %v", err)
 	}
 
-	httpClient, err := ingressgateway.FromEnv().Client()
-	if err != nil {
-		t.Fatalf("Unable to initialize ingress gateway client: %v", err)
+	if conf.IngressClient {
```
piotrmsc, line 76: Maybe we could name that will indicate that is being used internally on local envs?

tests/integration/api-gateway/Dockerfile
```diff
@@ -20,15 +20,12 @@ RUN mkdir /test &&\
...
+FROM eu.gcr.io/kyma-project/test-infra/alpine-kubectl:v20200310-5f52f407
 LABEL source=git@github.com:kyma-project/kyma.git
 WORKDIR /test
 
 RUN apk update &&\
-	apk add curl bash grep &&\
-	curl -LO https://storage.googleapis.com/kubernetes-release/release/$(curl -s https://storage.googleapis.com/kubernetes-release/release/stable.txt)/bin/linux/amd64/kubectl &&\
-	chmod +x kubectl &&\
-	mv kubectl /usr/local/bin/kubectl
+	apk add curl bash grep ca-certificates
```
jakkab, line 28: AFAIK, the alpine-kubectl image features curl and bash by default, no need to add them here.

tests/integration/api-gateway/gateway-tests/main_test.go
```diff
@@ -46,29 +48,46 @@ const jwtAndOauthOnePathApiruleFile = "jwt-oauth-one-path-strategy.yaml"
...
+		if err != nil {
+			t.Fatalf("Unable to initialize ingress gateway client: %v", err)
+		}
+	} else {
+		log.Printf("Fallback to default http client")
+		httpClient = &http.Client{
+			Transport: &http.Transport{
+				TLSClientConfig: &tls.Config{InsecureSkipVerify: true},
+			},
+			Timeout: time.Second * conf.ClientTimeout,
```
jakkab, line 89: `conf.ClientTimeout` field is of type `time.Duration`, why do we need to multiply it  by`time.Second`?

PR #8031, opened 2020-04-15
Print Installation Overrides to a logfile

anishj0shi:
Currently installer was logging installation overrides in verbose mode to Standard output. 
Adding additional flag to control this behavior. and logging overrides to a log file in a container

### DIFF
--- a/components/kyma-operator/cmd/operator/main.go
+++ b/components/kyma-operator/cmd/operator/main.go
@@ -4,9 +4,12 @@ import (
 	"flag"
 	"fmt"
 	"log"
+	"os"
 	"strconv"
 	"strings"
 	"time"
+
+	"github.com/sirupsen/logrus"
 
 	"github.com/kyma-project/kyma/components/kyma-operator/pkg/actionmanager"
 	"github.com/kyma-project/kyma/components/kyma-operator/pkg/conditionmanager"
@@ -34,6 +37,8 @@ import (
 
 var gitCommitHash string
 
+const STDOUT = "/dev/stdout"
+
 func main() {
 
 	log.Println("Starting operator...")
@@ -51,7 +56,7 @@ func main() {
 	tlsCrt := flag.String("tillerTLSCrt", "/etc/certs/tls.crt", "Path to TLS cert file")
 	TLSInsecureSkipVerify := flag.Bool("tillerTLSInsecureSkipVerify", false, "Disable verification of Tiller TLS cert")
 	backoffIntervalsRaw := flag.String("backoffIntervals", "10,20,40,60,80", "Number of seconds to wait before subsequent retries")
-	printOverrides := flag.Bool("printOverrides", false, "Print Installation Overrides")
+	overrideLogFile := flag.String("overrideLogFile", "/tmp/installer.log", "Log File to Print Installation overrides. (Default: /dev/stdout)")
 
 	flag.Parse()
 
@@ -75,7 +80,12 @@ func main() {
 		log.Fatalf("Unable to create internal client. Error: %v", err)
 	}
 
-	helmClient, err := kymahelm.NewClient(*helmHost, *tlsKey, *tlsCrt, *TLSInsecureSkipVerify, *printOverrides)
+	overridesLogger, closeFn, err := setupLogrus(*overrideLogFile)
+	if err != nil {
+		log.Fatalf("Unable to create logrus Instance. Error: %v", err)
+	}
+
+	helmClient, err := kymahelm.NewClient(*helmHost, *tlsKey, *tlsCrt, *TLSInsecureSkipVerify, overridesLogger)
 	if err != nil {
 		log.Fatalf("Unable create helm client. Error: %v", err)
 	}
@@ -105,6 +115,10 @@ func main() {
 	internalInformerFactory.Start(stop)
 
 	installationController.Run(stop)
+
+	if closeFn != nil {
+		closeFn(stop)
+	}
 }
 
 func getClientConfig(kubeconfig string) (*rest.Config, error) {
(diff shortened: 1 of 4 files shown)


anishj0shi:
> mhh ... why do we log data of secret overrides at all? My idea would be to only log the overrides coming from a ConfigMap, but not from a Secret. What do you think?

@nachtmaar  there's no way to distinguish this. and it seems we need this data somewhere. but not in stdout.

Tomasz-Smelcerz-SAP:
@nachtmaar @anishj0shi We log all overrides for debugging purposes. Overrides do change in time. If a component installation produces a secret that fulfills the overrides contract, it's added to all the other overrides for the installation of the next component.
Because it's so dynamic, when looking at the cluster state after the installation you can't really tell which overrides were used for the installation of a certain component. That's why we log it.

anishj0shi:
> @nachtmaar @anishj0shi We log all overrides for debugging purposes. Overrides do change in time. If a component installation produces a secret that fulfills the overrides contract, it's added to all the other overrides for the installation of the next component.
> Because it's so dynamic, when looking at the cluster state after the installation you can't really tell which overrides were used for the installation of a certain component. That's why we log it.

@Tomasz-Smelcerz-SAP  Makes sense, that's why with this PR, we log all the overrides. only overrides to a log file in the container rather then stdout. this will clean up with the installation logs.  and for watching overrides, we can exec into the container and view it.

components/kyma-operator/cmd/operator/main.go
```diff
@@ -51,6 +56,7 @@ func main() {
 	tlsCrt := flag.String("tillerTLSCrt", "/etc/certs/tls.crt", "Path to TLS cert file")
 	TLSInsecureSkipVerify := flag.Bool("tillerTLSInsecureSkipVerify", false, "Disable verification of Tiller TLS cert")
 	backoffIntervalsRaw := flag.String("backoffIntervals", "10,20,40,60,80", "Number of seconds to wait before subsequent retries")
+	printOverridesToSTDOut := flag.Bool("printOverridesToSTDOut", false, "Print Installation Overrides")
```
Tomasz-Smelcerz-SAP, line 59: I suggest to provide a filename as a flag. Default value should be a standard alias /dev/stdout. In this way the Installer behaves in the same way unless you explicitly change the log destination.
Stretch: After some time, when we ensure it works correctly in all environments, we can consider a follow-up task that changes the default to /dev/null. Then you **have to** explicitly tell the Installer where to log this stuf.

components/kyma-operator/cmd/operator/main.go
```diff
@@ -74,7 +80,12 @@ func main() {
 		log.Fatalf("Unable to create internal client. Error: %v", err)
 	}
 
-	helmClient, err := kymahelm.NewClient(*helmHost, *tlsKey, *tlsCrt, *TLSInsecureSkipVerify)
+	logger, err := setupLogrus(*printOverridesToSTDOut)
+	if err != nil {
+		log.Fatalf("Unable to create logrus Instance. Error: %v", err)
+	}
+
+	helmClient, err := kymahelm.NewClient(*helmHost, *tlsKey, *tlsCrt, *TLSInsecureSkipVerify, logger)
```
Tomasz-Smelcerz-SAP, line 88: please change the name to "overridesLogger"

components/kyma-operator/pkg/kymahelm/kyma-helm-client.go
```diff
@@ -26,19 +28,21 @@ type ClientInterface interface {
 
 // Client .
 type Client struct {
-	helm *helm.Client
+	helm   *helm.Client
+	logger *logrus.Logger
```
Tomasz-Smelcerz-SAP, line 32: please change the name to "overridesLogger"

jakkab:
@anishj0shi please add the `--overrideLogFile` flag in `components/kyma-operator/Dockerfile`

components/kyma-operator/cmd/operator/main.go
```diff
@@ -128,3 +143,28 @@ func parseBackoffIntervals(backoffIntervals string) ([]uint, error) {
 
 	return backoffIntervalsParsed, nil
 }
+
+func setupLogrus(logFile string) (*logrus.Logger, func(ch <-chan struct{}), error) {
+	// create the logger
+	logger := logrus.New()
+	// with Json Formatter
+	logger.Formatter = &logrus.JSONFormatter{}
```
Demonsthere, line 151: Hi there and thanks for the PR!
I think that a nice-to-have feature would be to set the log format with an env/flag, as the logs in a json format are quite unreadable

anishj0shi: done

components/kyma-operator/cmd/operator/main.go
```diff
@@ -104,6 +116,10 @@ func main() {
 	internalInformerFactory.Start(stop)
 
 	installationController.Run(stop)
+
+	if closeFn != nil {
```
Tomasz-Smelcerz-SAP, line 120: I recommend to **always** return a function, instead of having `nil` check. In the case when the function shouldn't do anything, then - well, just provide the function that doesn't do anything.NOOP.

components/kyma-operator/cmd/operator/main.go
```diff
@@ -128,3 +144,34 @@ func parseBackoffIntervals(backoffIntervals string) ([]uint, error) {
...
+		logger.SetFormatter(&logrus.JSONFormatter{})
+	}
+
+	//open the log file
+	file, err := os.OpenFile(logFile, os.O_WRONLY|os.O_CREATE|os.O_APPEND, 0755)
+	if err != nil {
+		return nil, nil, err
+	}
+	logger.SetOutput(file)
+	if logFile == STDOUT { //if the output file is /dev/stdout, we should not return a function to close the file
```
Tomasz-Smelcerz-SAP, line 162: I think we should just return a "NOOP" function.

components/kyma-operator/pkg/kymahelm/kyma-helm-client.go
```diff
@@ -26,19 +28,21 @@ type ClientInterface interface {
...
 // Client .
 type Client struct {
-	helm *helm.Client
+	helm            *helm.Client
+	overridesLogger *logrus.Logger
 }
 
 // NewClient .
-func NewClient(host string, TLSKey string, TLSCert string, TLSInsecureSkipVerify bool) (*Client, error) {
+func NewClient(host string, TLSKey string, TLSCert string, TLSInsecureSkipVerify bool, logger *logrus.Logger) (*Client, error) {
```
Tomasz-Smelcerz-SAP, line 36: rename to overridesLogger. It's a good name :)

anishj0shi: done

components/kyma-operator/cmd/operator/main.go
anishj0shi: done

components/kyma-operator/cmd/operator/main.go
anishj0shi: done

components/kyma-operator/cmd/operator/main.go
```diff
@@ -128,3 +142,39 @@ func parseBackoffIntervals(backoffIntervals string) ([]uint, error) {
...
+	logger := logrus.New()
+	logger.SetFormatter(formatter)
+
+	//open the log file
+	file, err := os.OpenFile(logFile, os.O_WRONLY|os.O_CREATE|os.O_APPEND, 0755)
+	if err != nil {
+		return nil, nil, err
+	}
+	logger.SetOutput(file)
+	if logFile == STDOUT { //if the output file is /dev/stdout, we should not return a function to close the file
```
Tomasz-Smelcerz-SAP, line 157: Please remove the comment. The code is clear enough :)

Demonsthere:
LGTM

Issue #7948, opened 2020-04-08
Remove Tekton chart and its dependencies


michal-hudy:
**Description**

Tekton has been introduced in Kyma to build Function images. It brings an additional component that we need to maintain and keep up-to-date. The effort required to maintain it is too big when you check how it is used in Kyma. We simply create one `TaskRun` even without `Task` and it is not in `PipelineRun` or `Pipeline`.  This is something that can be achieved with a pure Kubernetes `Job`.

So we should remove `Tekton` chart and replace `TaskRun` with `Job`

**Reasons**

Kyma should be as simple as possible and we should keep the number of dependencies really low. We should avoid adding new dependencies when possible. Also, it will reduce the amount of required resources for Kyma.

michal-hudy:
TODO:
 - Documentation

PR #8035, opened 2020-04-15
Remove TaskRun and Tekton references from docs

kazydek:
**Description**

- Removed TaskRun and Tekton references from docs.

**Related issue(s)**
See also #7948


components/function-controller/README.md
```diff
@@ -25,7 +25,7 @@ The Function Controller uses these environment variables:
...
 | **APP_CONFIG_WATCHER_BASE_NAMESPACE**             | Name of the Namespace with the serverless configuration (such as runtime, Secret and service account for the Docker Registry) propagated to other Namespaces                                                                                                                                                                    | `kyma-system`                                                                                                                                              |
-| **APP_CONFIG_WATCHER_EXCLUDED_NAMESPACES**        | List of Namespaces to which serverless configuration is not propagated                                                                                                                                                | `istio-system;knative-eventing;knative-serving;kube-node-lease;kube-public;kube-system;kyma-installer;kyma-integration;kyma-system;tekton-pipelines;natss` |
```
michal-hudy, line 28: Could you also update it here: https://github.com/kyma-project/kyma/blob/master/components/function-controller/internal/configwatcher/configwatcher.go#L21 ?

docs/serverless-v2/02-01-serverless-v2.md
```diff
@@ -2,27 +2,25 @@
 title: Architecture
 ---
 
-Serverless v2 relies on [Knative Serving](https://knative.dev/docs/serving/) for deploying and managing functions, and [Tekton Pipelines](https://github.com/tektoncd/pipeline) for creating Docker images. See how these and other resources process a lambda within a Kyma cluster:
+Serverless v2 relies on [Knative Serving](https://knative.dev/docs/serving/) for deploying and managing functions and [Kubernetes Jobs](https://kubernetes.io/docs/concepts/workloads/controllers/jobs-run-to-completion/) for creating Docker images. See how these and other resources process a lambda within a Kyma cluster:
 
-![Serverless v2 architecture](./assets/serverless-v2-architecture.svg)
+![Serverless architecture](./assets/serverless-architecture.svg)
```
michal-hudy, line 7: Hm... before the creation of Job there is a step that creates ConfigMap - it is described in Details topic as a separate section so this diagram should be aligned with the next sections.

docs/serverless-v2/01-01-serverless-v2.md
```diff
@@ -2,7 +2,7 @@
...
+>**CAUTION:** This implementation will soon replace Serverless that is based on [Kubeless](https://github.com/kubeless/kubeless). Serverless v2 relies on [Knative Serving](https://knative.dev/docs/serving/) for deploying and managing functions, and [Kubernetes Jobs](https://kubernetes.io/docs/concepts/workloads/controllers/jobs-run-to-completion/) for creating Docker images. Consider this implementation as experimental as it is still under development.
 
 "Serverless" refers to an architecture in which the infrastructure of your applications is managed by cloud providers. Contrary to its name, serverless applications do require a server but they don't require you to run and manage it on your own. Instead, you subscribe to a given cloud provider, such as AWS, Azure or GCP, and pay a subscription fee only for the resources you actually use. Since the resource allocation can be dynamic and depends on your current needs, the serverless model is particularly cost-effective when you want to implement a certain logic that is triggered on demand. Simply, you get your things done and don't pay for the infrastructure that sits idle.
```
bszwarc, line 7: I know that this is not the scope of the review :D but I have a question about this part:
`Contrary to its name, serverless applications do require a server but they don't require you to run and manage it on your own.` Shouldn't it be `contrary to their name`? as you are speaking about the `applications` later on.
Or `contrary to its name, a serverless application does...`

docs/serverless-v2/02-01-serverless-v2.md
```diff
@@ -2,23 +2,23 @@
...
-4. Tekton Pipelines Controller (TPC) detects the new TaskRun CR and reads its definition.
+4. Based on the ConfigMap, FC creates a Kubernetes Job that triggers the creation of a lambda image.
 
-5. Based on the TaskRun CR definition, TPC triggers a pipeline that uses [Kaniko](https://github.com/GoogleContainerTools/kaniko/blob/master/README.md) to create a Docker image with lambda definition and publish this image in a Docker registry.
+5. The Job creates a Pod with the Docker image containing the lambda definition. It also pushes the image to a Docker registry.
 
-6. FC monitors the TaskRun CR. When the image creation finishes successfully, FC creates a Service CR (KService) that points to the Pod with the image.
+6. FC monitors the Job status. When the image creation finishes successfully, FC creates a Service CR (KService) that points to the Pod with the image.
 
 7. Knative Serving controller (KSC) detects the new KService and reads its definition.
```
bszwarc, line 23: controller or Controller? Looking at the abbreviation (KSC) and also Function Controller I would capitalize it.

docs/serverless-v2/03-01-api-rules.md
```diff
@@ -3,7 +3,7 @@ title: Exposing lambdas
...
+By default, lambdas in Kyma are not exposed outside the cluster. The Knative Serving Controller normally creates two virtual services, one for the internal cluster communication and the other one for exposing the lambda outside the cluster. To restrict the access, the Function Controller automatically sets the default `serving.knative.dev/visibility=cluster-local` label on a KService CR that is added to its metadata when a Function CR is processed. This limits the Knative Serving Controller to create only one, local virtual service that allows only for cluster-wide access to lambda services. Such a lambda is accessible for other resources within the cluster under the `{service-name}.{namespace}.svc.cluster.local` endpoint, such as `test-lambda.default.svc.cluster.local`.
```
bszwarc, line 6:
```suggestion
By default, lambdas in Kyma are not exposed outside the cluster. The Knative Serving Controller normally creates two virtual services, one for the internal cluster communication and the other one for exposing the lambda outside the cluster. To restrict the access, the Function Controller automatically sets the default `serving.knative.dev/visibility=cluster-local` label on a KService CR that is added to its metadata when a Function CR is processed. This limits the Knative Serving Controller to create only one, local virtual service that allows only for cluster-wide access to lambda services. Other resources can access such a lambda within the cluster under the `{service-name}.{namespace}.svc.cluster.local` endpoint, such as `test-lambda.default.svc.cluster.local`.
```

the suggestion doesn't show it, but I changed the last line a bit. Also In the architecture doc, the controller is called `Knative Serving controller ` but I think you are right to capitalize it, just change the other entry :)

docs/serverless-v2/01-01-serverless-v2.md
kazydek: True, I'll change that :)

docs/serverless-v2/02-01-serverless-v2.md
kazydek: ok

Issue #8037, opened 2020-04-15
Support for TLS configuration of external logging systems with fluentbit


a-thaler:
**Description**
The fluentbit daemon coming with kyma is using an istio side-car for now.
When using the logging/fluentbit chart to integrate with 3party logging systems, there can be scenarios where TLS configuration will not work out of the box as the default istio side-car config must be adjusted.

This can be achieved by defining an istio ServiceEntry resource. Please bundle a ServiceEntry resource as kyma addition to the fluent-bit chart so that you can configure it  via installer overrides.

**Reasons**

**Attachments**

PR #8052, opened 2020-04-16
External service configuration with HTTPS for fluent bit

hisarbalik:
#8037

### DIFF
--- a/resources/logging/charts/fluentbit/templates/serviceentry.yaml
+++ b/resources/logging/charts/fluentbit/templates/serviceentry.yaml
@@ -1,0 +1,20 @@
+{{- if .Values.serviceEntry }}
+apiVersion: networking.istio.io/v1alpha3
+kind: ServiceEntry
+metadata:
+  name: {{ template "fluent-bit.fullname" . }}-service-entry
+  namespace: {{ .Release.Namespace }}
+  labels:
+    {{- include "fluent-bit.metaLabels" . | trim | nindent 4 }}
+  annotations:
+    {{- if .Values.globalAnnotations }}
+      {{- toYaml .Values.globalAnnotations | trim | nindent 4 }}
+      {{- end }}
+spec:
+  hosts:
+  {{- toYaml .Values.serviceEntry.hosts | trim | nindent 4 }}
+  location: MESH_EXTERNAL
+  ports:
+  {{- toYaml .Values.serviceEntry.ports | trim | nindent 4 }}
+  resolution: DNS
+{{- end }}
--- a/resources/logging/charts/fluentbit/templates/serviceentry.yaml
+++ b/resources/logging/charts/fluentbit/templates/serviceentry.yaml
@@ -1,8 +1,8 @@
-{{- if .Values.serviceEntry }}
+{{- if .Values.externalServiceEntry }}
 apiVersion: networking.istio.io/v1alpha3
 kind: ServiceEntry
 metadata:
-  name: {{ template "fluent-bit.fullname" . }}-service-entry
+  name: {{ template "fluent-bit.fullname" . }}
   namespace: {{ .Release.Namespace }}
   labels:
     {{- include "fluent-bit.metaLabels" . | trim | nindent 4 }}
@@ -12,9 +12,9 @@ metadata:
       {{- end }}
 spec:
   hosts:
-  {{- toYaml .Values.serviceEntry.hosts | trim | nindent 4 }}
+  {{- toYaml .Values.externalServiceEntry.hosts | trim | nindent 4 }}
   location: MESH_EXTERNAL
   ports:
-  {{- toYaml .Values.serviceEntry.ports | trim | nindent 4 }}
-  resolution: DNS
+  {{- toYaml .Values.externalServiceEntry.ports | trim | nindent 4 }}
+  resolution: {{ .Values.externalServiceEntry.resolutionMode }}
 {{- end }}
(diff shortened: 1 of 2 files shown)


resources/logging/charts/fluentbit/templates/serviceentry.yaml
```diff
@@ -0,0 +1,20 @@
...
+    {{- if .Values.globalAnnotations }}
+      {{- toYaml .Values.globalAnnotations | trim | nindent 4 }}
+      {{- end }}
+spec:
+  hosts:
+  {{- toYaml .Values.serviceEntry.hosts | trim | nindent 4 }}
+  location: MESH_EXTERNAL
+  ports:
+  {{- toYaml .Values.serviceEntry.ports | trim | nindent 4 }}
+  resolution: DNS
```
rakesh-garimella, line 19: why dont you add resolution mode value from values.yaml just like others ? Because it looks redundant.

resources/logging/charts/fluentbit/templates/serviceentry.yaml
```diff
@@ -0,0 +1,20 @@
+{{- if .Values.serviceEntry }}
+apiVersion: networking.istio.io/v1alpha3
+kind: ServiceEntry
+metadata:
+  name: {{ template "fluent-bit.fullname" . }}-service-entry
```
a-thaler, line 5: please don't put a suffix containing the resource type, it is redundant

resources/logging/charts/fluentbit/values.yaml
```diff
@@ -315,3 +315,13 @@ tolerations:
 
 # Pod affinity constraint
 affinity: {}
+
+# external service entry configuration, this config add an external service to the service mesh
```
a-thaler, line 319: that is a very istio specific configuration.. I thought we maybe can formulate it a bit more use case specific. So I want to enable a 3party integration. For that I might need to configure the Protocol settings explicit. Why the relevant key in the values.yaml is ServiceEntry? If I have no istio background I will not get this..

resources/logging/charts/fluentbit/values.yaml
```diff
@@ -315,3 +315,15 @@ tolerations:
...
+# If your pod istio enabled (istio proxy injected), in some cases istio doesn't allow applications to access external services through https and use none standard ports (not 443 or 8443)
```
a-thaler, line 320:
```suggestion
# By default, the fluent-bit daemon will be part of the service-mesh (having istio sidecar injection enabled). In some cases the sidecar doesn't allow access to external services through https using none standard ports (not 443 or 8443)
```

resources/logging/charts/fluentbit/values.yaml
```diff
@@ -315,3 +315,15 @@ tolerations:
...
+# Typical error message look like 'routines:ssl3_get_record:wrong version number', in this case remove comment for values below and change it accordingly for your service
```
a-thaler, line 321:
```suggestion
# A typical error message will look like 'routines:ssl3_get_record:wrong version number'. In such cases, add an entry for the external service like below
```

resources/logging/charts/fluentbit/values.yaml
```diff
@@ -315,3 +315,15 @@ tolerations:
 
 # Pod affinity constraint
 affinity: {}
+
+# external service entry configuration, this config add an external service to the service mesh
```
a-thaler, line 319:
```suggestion
# Defines an entry to add an external service to the service mesh
```

resources/logging/charts/fluentbit/templates/serviceentry.yaml
```diff
@@ -0,0 +1,20 @@
...
+  namespace: {{ .Release.Namespace }}
+  labels:
+    {{- include "fluent-bit.metaLabels" . | trim | nindent 4 }}
+  annotations:
+    {{- if .Values.globalAnnotations }}
+      {{- toYaml .Values.globalAnnotations | trim | nindent 4 }}
+      {{- end }}
+spec:
+  hosts:
+  {{- toYaml .Values.externalServiceEntry.hosts | trim | nindent 4 }}
```
rakesh-garimella, line 15: I tested it and the overrides are not passed as array rather as string.
```suggestion
  {{- tpl (.Values.externalServiceEntry.hosts | nindent 4) . }}
  ```

File: resources/logging/charts/fluentbit/templates/serviceentry.yaml [hunk 8]
```diff
@@ -0,0 +1,20 @@
...
+  location: MESH_EXTERNAL
+  ports:
+  {{- toYaml .Values.externalServiceEntry.ports | trim | nindent 4 }}
```
rakesh-garimella, line 18: Same reason as before.
```suggestion
  {{- tpl (.Values.externalServiceEntry.ports | nindent 4) . }}
```

resources/logging/charts/fluentbit/values.yaml
```diff
@@ -315,3 +315,15 @@ tolerations:
...
+# By default, the fluent-bit daemon will be part of the service-mesh (having istio sidecar injection enabled). In some cases the sidecar doesn't allow access to external services through https using none standard ports (not 443 or 8443)
+# A typical error message will look like 'routines:ssl3_get_record:wrong version number'. In such cases, add an entry for the external service like below
+#externalServiceEntry:
+#  hosts:
+#  - my.example.host.dns
+#  ports:
+#    - number: 8081
+#      name: https
+#      protocol: TLS
+#  resolutionMode: DNS
```
rakesh-garimella, line 329: why is this called `resolutionMode` rather than `resolution` as the key is called `resolution`.

Issue #7933, opened 2020-04-07
Helm charts clean up


pbochynski:
**Description**
Improve kyma components charts by:
- [x] make charts compatible with helm 3
- [x] fix errors in schemas
- [x] remove deprecated APIs
- [ ] ~~do not install resources in multiple namespaces within one chart~~
- [x] move custom resource definitions to the first step (cluster-essentials)
- [ ] ~~enable parallel installation of charts by removing dependencies and waiting steps where possible (e.g. waiting for certificates or DNS should be avoided)~~

Acceptance criteria:
- each kyma chart can be installed by `helm template | kubectl apply -f -`

**Reasons**
High-quality charts are the first step into simpler, faster and better installation and upgrade processes.

**Attachments**

See the PoC: https://github.com/kyma-project/kyma/pull/7930

pbochynski:
The most important changes are applied.

PR #8057, opened 2020-04-16
fix linter errors

m00g3n:
Fix linter errors pahse1

- add required `apiVersion` elements
- regenerate outdated jsonschema
- remove `Release.Time.Seconds` to have `helm3` chart compatibility

**Related issue(s)**
See also : https://github.com/kyma-project/kyma/issues/7933


resources/cluster-essentials/charts/pod-preset/Chart.yaml
```diff
@@ -1,4 +1,6 @@
 name: pod-preset
-description: A Helm chart for PodPreset resource that's implemented via a CRD that is paired with a mutating webhook in order to modify pod specs on the fly
+description: A Helm chart for PodPreset resource that's implemented via a CRD that
```
Demonsthere, line 2: For a multiline string shouldn't this look like

```yaml
description: |
  foo
```

m00g3n: thx it looks like autoformatter

PR #8061, opened 2020-04-17
Refactor application-connector charts to be compatible with helm3

m00g3n:
**Description**
Refactor `application-connector` charts to be compatible with helm3.

-  remove `crd-install` helm hook ([docs](https://helm.sh/docs/chart_best_practices/custom_resource_definitions/#method-1-let-helm-do-it-for-you))
- move `CRD`s to `cluster-essentials`

**Related issue(s)**

See also: https://github.com/kyma-project/kyma/issues/7933


resources/cluster-essentials/files/crd-applications.applicationconnector.yaml
```diff
@@ -2,8 +2,6 @@ apiVersion: apiextensions.k8s.io/v1beta1
 kind: CustomResourceDefinition
```
Szymongib, line 2: All the other CRDs were copied to the files directory and suppose to be installed only on upgrade. As I understand it was done so that Helm would not delete them, why is that not the case with this CRD as this one has been moved?

Issue #7579, opened 2020-03-12
Define a Production Profile for Serverless v2


michal-hudy:
**Description**

Define a Production Profile for Serverless that will speed up the creation of lambda. Currently, simple lambda is built in more than 3min, which is not acceptable. Simple lambda (without dependencies) should be built in less than the `30s`.

Production Profile should be documented.

**Reasons**

Improve the UX

PR #8079, opened 2020-04-19
Adjust build resources

michal-hudy:
**Description**

- Adjusted default build resources (function is built in less than 40s)
- Adjusted build resources for minikube (function is built in  ~2min)
- Disabled Knative Serving tests - it is tested by `function-controller` tests

**Related issue(s)**

#7579

### DIFF
--- a/installation/resources/installer-config-local.yaml.tpl
+++ b/installation/resources/installer-config-local.yaml.tpl
@@ -191,8 +191,9 @@ metadata:
     component: function-controller
     kyma-project.io/installation: ""
 data:
-  containers.manager.envs.buildRequestsCPU.value: "350m"
-  containers.manager.envs.buildRequestsMemory.value: "350Mi"
-  containers.manager.envs.buildLimitsCPU.value: "550m"
-  containers.manager.envs.buildLimitsMemory.value: "550Mi"
-  tests.enabled: "true"
+  containers.manager.envs.buildRequestsCPU.value: "100m"
+  containers.manager.envs.buildRequestsMemory.value: "200Mi"
+  containers.manager.envs.buildLimitsCPU.value: "200m"
+  containers.manager.envs.buildLimitsMemory.value: "400Mi"
+  # TODO: Solve a problem with DNS
+  tests.enabled: "false"
--- a/installation/resources/installer-config-local.yaml.tpl
+++ b/installation/resources/installer-config-local.yaml.tpl
@@ -184,11 +184,11 @@ data:
 apiVersion: v1
 kind: ConfigMap
 metadata:
-  name: function-controller-overrides
+  name: serverless-overrides
   namespace: kyma-installer
   labels:
     installer: overrides
-    component: function-controller
+    component: serverless
     kyma-project.io/installation: ""
 data:
   containers.manager.envs.buildRequestsCPU.value: "100m"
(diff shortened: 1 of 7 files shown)


resources/function-controller/templates/tests/test.yaml
```diff
@@ -16,12 +16,6 @@ spec:
       annotations:
         sidecar.istio.io/inject: "true"
     spec:
-      {{ if .Values.global.isLocalEnv }}
```
magicmatatjahu, line 19: This is not needed for minikube?

resources/function-controller/values.yaml
```diff
@@ -66,15 +63,41 @@ containers:
...
+      imagePullAccountName:
+        value: "{{ template \"fullname\" . }}"
+      functionRequeueDuration:
+        value: 1m
+      functionBuildExecutorImage:
+        value: gcr.io/kaniko-project/executor:v0.17.1
+      functionBuildCredsInitImage:
+        value: gcr.io/tekton-releases/github.com/tektoncd/pipeline/cmd/creds-init:v0.10.1
+      functionDockerAddress:
+        # TODO: I don't know why {{ .Values.global.dockerServicePort }} doesn't work. It is rendered as {{\n  .Values.global.dockerServicePort }}
```
magicmatatjahu, line 96: Probably this is caused by `global` context in templating.

resources/function-controller/templates/tests/test.yaml
michal-hudy: Currently, we don't know how to handle minikube and it points to storage that is completely not used in this test... so I removed this to not confuse others.

resources/function-controller/templates/deployment.yaml
```diff
@@ -82,21 +82,21 @@ spec:
...
+            {{ include "createEnv" ( dict "name" "APP_FUNCTION_BUILD_RUNTIME_CONFIG_MAP_NAME" "value" .Values.containers.manager.envs.runtimeConfigMapName "context" . ) | nindent 12 }}
+            {{ include "createEnv" ( dict "name" "APP_FUNCTION_BUILD_REQUESTS_CPU" "value" .Values.containers.manager.envs.buildRequestsCPU "context" . ) | nindent 12 }}
+            {{ include "createEnv" ( dict "name" "APP_FUNCTION_BUILD_REQUESTS_MEMORY" "value" .Values.containers.manager.envs.buildRequestsMemory "context" . ) | nindent 12 }}
+            {{ include "createEnv" ( dict "name" "APP_FUNCTION_BUILD_LIMITS_CPU" "value" .Values.containers.manager.envs.buildLimitsCPU "context" . ) | nindent 12 }}
+            {{ include "createEnv" ( dict "name" "APP_FUNCTION_BUILD_LIMITS_MEMORY" "value" .Values.containers.manager.envs.buildLimitsMemory "context" . ) | nindent 12 }}
+            {{ include "createEnv" ( dict "name" "APP_CONFIG_WATCHER_NAMESPACE_REQUEUE_DURATION" "value" .Values.containers.manager.envs.namespaceRequeueDuration "context" . ) | nindent 12 }}
```
magicmatatjahu, line 92: Where is env for excluded namespaces? :)

resources/function-controller/templates/tests/test.yaml
aerfio: but it "works" if you add that specific lambda name to /etc/hosts

maybe we could keep it for posterity?

michal-hudy: hm... I will check as I'm waiting for rename merge ;)

michal-hudy: @aerfio ok, I'm still thinking that it doesn't make sense to keep it, the function has generated FQDN (like `test-function-1234566.kyma.local`), not static.

PR #8090, opened 2020-04-20
Revamp of Kiali

a-thaler:
**Description**

- update kiali to latest version
- restructured chart to use kiali-operator
- update keycloak proxy to keycloak-gatekeeper

**Related issue(s)**


resources/cluster-essentials/templates/kiali.yaml
```diff
@@ -0,0 +1,27 @@
...
+    plural: kialis
+    singular: kiali
+  scope: Namespaced
+  subresources:
+    status: {}
+  version: v1alpha1
+  versions:
+  - name: v1alpha1
+    served: true
+    storage: true
```
suleymanakbas91, line 27: new line please :)

resources/kiali/templates/operator/clusterrole.yaml
```diff
@@ -0,0 +1,295 @@
...
+  - experiments
+  verbs:
+{{- if .Values.kiali.spec.deployment.view_only_mode }}
+  - create
+  - delete
+  - patch
+{{- end }}
+  - get
+  - list
+  - watch
```
suleymanakbas91, line 295: new line

resources/kiali/templates/operator/deployment.yaml
```diff
@@ -0,0 +1,51 @@
...
+              fieldPath: metadata.namespace
+        - name: OPERATOR_NAME
+          value: "kiali-operator"
+        - name: ANSIBLE_DEBUG_LOGS
+          value: "True"
+        resources:
+{{ toYaml .Values.kiali.operator.resources | indent 10 }}
+      volumes:
+      - name: runner
+        emptyDir: {}
```
suleymanakbas91, line 51: new line

resources/kiali/templates/operator/deployment.yaml
```diff
@@ -0,0 +1,51 @@
...
+        - name: POD_NAME
+          valueFrom:
+            fieldRef:
+              fieldPath: metadata.name
+        - name: POD_NAMESPACE
+          valueFrom:
+            fieldRef:
+              fieldPath: metadata.namespace
+        - name: OPERATOR_NAME
+          value: "kiali-operator"
```
suleymanakbas91, line 44: It's not very important, but it'd be better if this is the same as the deployment name `{{ template "kiali.name" . }}-operator`.

resources/kiali/values.yaml
```diff
@@ -1,82 +1,699 @@
+
```
suleymanakbas91, line 1: unnecessary space

resources/kiali/values.yaml
```diff
@@ -1,82 +1,699 @@
...
-  # These namespaces have service mesh components that are to be observed by Kiali.
-  # You can provide names using regex expressions matched against all namespaces the operator can see.
-  # The default makes all namespaces accessible except for some internal namespaces that typically should be ignored.
-  # NOTE! If this has an entry with the special value of "**" (two asterisks), that will denote you want
-  # Kiali to be given access to all namespaces via a single cluster role (if using this special value of "**",
-  # you are required to have already granted the operator permissions to create cluster roles and cluster role bindings).
-  accessibleNamespaces: "**"
+  inPort: 3000
+  clientId: kiali
+  clientSecret: hiFWLWqIxw5d3gl
```
suleymanakbas91, line 21: Is it really ok to have this information public?

a-thaler: They are introduced in the dex config, also for other clients like console.
I found no better way of re-using the setting so that it could be also just generated in the dex config and referenced here.
So I at least made it configurable so that an admin could override both places with custom values.
It was like that before... Open for finding a better solution in a different PR solving the problem for all clients consistently.

a-thaler:
/test pre-master-kyma-gke-compass-integration
--

PR #8108, opened 2020-04-20
Bump images due to event bus removal

k15r:
**Description**

- multiple images bumps

**Related issue(s)**

#7730


tests/end-to-end/backup/chart/backup-test/values.yaml
```diff
@@ -7,4 +7,4 @@ image:
   dir:
   namePrefix: end-to-end-backup
   nameSuffix: tests
-  tag: PR-7800
+  tag: PR-7800
```
sayanh, line 10: Is the new line removal intended? I think the newline is good to have.

k15r: currently there is no rule enforced, so as long as it is valid yaml id say it is ok

PR #8112, opened 2020-04-21
consistent kyma-gateway reference in virtualservices

a-thaler:
**Description**

- fixed gateway reference in virtualservice of ory and dex
- aligned approach of referencing gateway in virtualservice for all charts
- ...

**Related issue(s)**

### DIFF
--- a/resources/compass/charts/cockpit/templates/mf-virtualservice.yaml
+++ b/resources/compass/charts/cockpit/templates/mf-virtualservice.yaml
@@ -13,7 +13,7 @@ spec:
   hosts:
     - {{ .Values.mf.host }}.{{ .Values.global.ingress.domainName }}
   gateways:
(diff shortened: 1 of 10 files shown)


resources/rafter/charts/controller-manager/charts/minio/templates/apirule.yaml
```diff
@@ -9,7 +9,7 @@ metadata:
     release: "{{ .Release.Name }}"
     heritage: "{{ .Release.Service }}"
 spec:
-  gateway: {{ .Values.global.istio.gateway.name }}.{{ .Values.global.istio.gateway.namespace }}.svc.cluster.local
```
crabtree, line 12: I'm not entirely sure if this resource expects the gateway in this format

a-thaler: I checked with @Tomasz-Smelcerz-SAP and indeed it is not supported. He will think about how to make it more convenient.
Anyway, I reverted that change, good finding.

PR #8123, opened 2020-04-21
Add runtimeOperator group scopes (L2 operator access)

ebensom:
**Description**

- Add runtimeOperator group scopes (L2 operator access)

### DIFF
--- a/resources/compass/charts/director/values.yaml
+++ b/resources/compass/charts/director/values.yaml
@@ -16,4 +16,5 @@ adminGroupNames:
   - "runtimeAdmin"
   - "runtimeNamespaceAdmin"
 
-operatorGroupName: "runtimeOperator"
+operatorGroupNames:
+  - "runtimeOperator"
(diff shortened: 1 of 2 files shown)


resources/compass/charts/director/values.yaml
```diff
@@ -11,7 +11,9 @@ deployment:
     maxOpenConnections: 30
     maxIdleConnections: 2
 
-staticGroupNames:
+adminGroupNames:
   - "mps-superadmin"
   - "runtimeAdmin"
   - "runtimeNamespaceAdmin"
+
+operatorGroupName: "runtimeOperator"
```
pkosiec, line 19: Hmm, maybe we could also transform it to an array? It would be consistent with admin group names defined above.

Like:

```yaml
operatorGroupNames: 
   - ""runtimeOperator"
```

What do you think?

pkosiec:
Oh, and it would be great if you could sync back the change to the `compass` repo. Thanks a lot!

PR #8142, opened 2020-04-22
core: add service ports to core console components

fhivemind:
**Description**

Exposes health endpoint ports of `core-console` services, annotated by `core-SERVICE-NAME`, as services share the same annotations, we didn't want to introduce breaking changes.

Following services were updated
- `core-console` - exposes port `status-port-console`
- `core-console-core-ui` - exposes port `status-port-core-ui`
- `core-console-compass-mfs` - exposes port `status-port-compass-mfs`

**Related issue(s)**


sjanota:
Looks ok, but to understand better, why do we need these ports in service definition?

PR #8152, opened 2020-04-22
Display runtime labels

Wawrzyn321:
**Description**

- ...
- ...
- ...

**Related issue(s)**
[Compass](https://github.com/kyma-incubator/compass/pull/1219)
[Console](https://github.com/kyma-project/console/pull/1725/files)


resources/core/charts/console/values.yaml
```diff
@@ -17,7 +17,7 @@ core_ui:
   statusPort: 6081
   image:
     dir: 
-    tag: "477b7708"
+    tag: 5a8b31e2
```
dariadomagala, line 20: Shouldn't we also bump compass_mfs?

sjanota: We will have also labels display in applications today, so I'll wait for that. @Wawrzyn321 do you agree?

PR #8155, opened 2020-04-22
Set default domain name in API Gateway

kubadz:
**Description**

- set default domain name parameter
- bump api-gateway

**Related issue(s)**

See also [api-gateway/#95](https://github.com/kyma-incubator/api-gateway/pull/95)

### DIFF
--- a/resources/api-gateway/templates/_helpers.tpl
+++ b/resources/api-gateway/templates/_helpers.tpl
@@ -74,8 +74,6 @@ Get a default domain from values if set or use the default domain name for Kyma
 {{ if .Values.config.defaultDomain }}
 {{- printf "%s" .Values.config.defaultDomain -}}
 {{ else }}
-{{- with .Values.global.ingress.domainName }}
-{{- printf "%s" . -}}
-{{- end }}
+{{- printf "%s" .Values.global.ingress.domainName -}}
 {{- end }}
 {{- end -}}
--- a/resources/api-gateway/templates/_helpers.tpl
+++ b/resources/api-gateway/templates/_helpers.tpl
@@ -54,3 +54,28 @@ Create a list of namespaced services to blacklist
 {{- end }}
 {{- end }}
 {{- end -}}
+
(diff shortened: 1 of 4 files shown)


resources/api-gateway/templates/deployment.yaml
```diff
@@ -39,6 +39,7 @@ spec:
             - --generated-objects-labels={{ .Values.config.generatedObjects.labels }}
             {{- with .Values.global.ingress.domainName }}
             - --domain-whitelist={{ . | join ", " }}
+            - --default-domain-name={{ . | join ", " }}
```
Demonsthere, line 42: I think we can refactor this a bit. 
Currently in `values.yaml` we have a `domainWhiteList` field that is not used, as we only use the value from `global.ingress.domainName`. 

```yaml
config:
  domainWhiteList:
    - kyma.local
```

Imho we can improve this by defining 2 helm functions:
- getDomainWhiteList: which does "`domainWhiteList + global.ingress.domainName | uniqe`" 
- getDefaultDomain: which uses a new `defaultDomain` variable if set, and default to `global.ingress.domainName` if empty

kubadz: done like suggested

resources/api-gateway/templates/_helpers.tpl
```diff
@@ -53,4 +53,29 @@ Create a list of namespaced services to blacklist
...
+{{/*
+Get a default domain from values if set or use the default domain name for Kyma
+*/}}
+{{- define "api-gateway.defaultDomain" -}}
+{{ if .Values.config.defaultDomain }}
+{{- printf "%s" .Values.config.defaultDomain -}}
+{{ else }}
+{{- with .Values.global.ingress.domainName }}
+{{- printf "%s" . -}}
+{{- end }}
```
Demonsthere, line 79:
```suggestion
{{- printf "%s" .Values.global.ingress.domainName -}}
```

Issue #4501, opened 2019-06-11
Extended component test for monitoring


a-thaler:
**Description**
Assure basic functionality of monitoring stack in an automated way:

- Check if all targets are up and healthy
- Check if all alerts are up and healthy

**Reasons**
A non working servicemonitor might indicate non-working component

**Workaround**
Make a port-foward for the prometheus-dashboard and check if all targets are healthy
```
kubectl port-forward -n kyma-system svc/monitoring-prometheus 9090:9090

Go to http://localhost:9090/targets
```

shorim:
**Merged Changes**

For monitoring integration test:
- Check that all targets are healthy
- Check that all rules are healthy
- Refactor HTTP client, so that it would be created only once

For monitoring upgrade test:
- Check that all targets are healthy before and after upgrade. Checking for target with job=istio-ingressgateway is excluded for now, because it is not healthy in release 1.11.1 and this check will be added in the next release
- Check that all rules are healthy before and after upgrade

PR #8156, opened 2020-04-22
Extending monitoring upgrade test

shorim:
**Description**

- Check that all targets are healthy before and after upgrade. Checking for target with job=istio-ingressgateway is excluded for now, because it is not healthy in release 1.11.1 and this check will be added in the next release
- Check that all rules are healthy before and after upgrade

**Related issue(s)**

#4501

### DIFF
--- a/tests/end-to-end/upgrade/pkg/tests/monitoring/promAPI/targets.go
+++ b/tests/end-to-end/upgrade/pkg/tests/monitoring/promAPI/targets.go
@@ -1,0 +1,24 @@
+package promAPI
+
+// To parse JSON response from http://monitoring-prometheus.kyma-system:9090/api/v1/targets
+type TargetsResponse struct {
+	Status    string      `json:"status"`
+	Data      TargetsData `json:"data"`
+	ErrorType string      `json:"errorType"`
+	Error     string      `json:"error"`
+}
+
+type TargetsData struct {
+	ActiveTargets []ActiveTarget `json:"activeTargets"`
+}
+
+type ActiveTarget struct {
+	Labels    Labels `json:"labels"`
+	LastError string `json:"lastError"`
+	Health    string `json:"health"`
+}
+
+type Labels struct {
+	Instance string `json:"instance"`
+	Job      string `json:"job"`
+}
--- a/tests/end-to-end/upgrade/pkg/tests/monitoring/targets_rules.go
+++ b/tests/end-to-end/upgrade/pkg/tests/monitoring/targets_rules.go
@@ -1,0 +1,285 @@
+package monitoring
+
+import (
+	"bytes"
+	"encoding/json"
+	"fmt"
+	"io"
+	"net/http"
+	"strings"
+	"time"
+
+	"github.com/kyma-project/kyma/tests/end-to-end/upgrade/pkg/tests/monitoring/promAPI"
+	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+)
+
+const prometheusURL = "http://monitoring-prometheus.kyma-system:9090"
+const namespace = "kyma-system"
+const expectedAlertManagers = 1
+const expectedPrometheusInstances = 1
+const expectedKubeStateMetrics = 1
+const expectedGrafanaInstance = 1
+
+// TargetsAndRulesTest checks that all targets and rules are healthy
+type TargetsAndRulesTest struct {
+	k8sCli     kubernetes.Interface
+	httpClient *http.Client
+}
+
+// NewTargetsAndRulesTest creates a new instance of TargetsAndRulesTest
+func NewTargetsAndRulesTest(k8sCli kubernetes.Interface) TargetsAndRulesTest {
+	return TargetsAndRulesTest{
+		k8sCli:     k8sCli,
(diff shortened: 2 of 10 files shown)


tests/end-to-end/upgrade/pkg/tests/monitoring/targets_rules.go
```diff
@@ -0,0 +1,285 @@
...
+		return err
+	}
+	return nil
+}
+
+// CreateResources checks that all targets and rules are healthy after upgrade
+func (t TargetsAndRulesTest) TestResources(stop <-chan struct{}, log logrus.FieldLogger, namespace string) error {
+	log.Println("checking that monitoring pods are ready")
+	err := t.testPodsAreReady()
+	if err != nil {
```
clebs, line 65:
```suggestion
	if err := t.testPodsAreReady(); err != nil {
```

tests/end-to-end/upgrade/pkg/tests/monitoring/targets_rules.go
```diff
@@ -0,0 +1,285 @@
...
+		return err
+	}
+	log.Println("checking that all targets are healthy after upgrade")
+	err = t.testTargetsAreHealthy()
+	if err != nil {
```
clebs, line 70:
```suggestion
	if err := t.testTargetsAreHealthy(); err != nil {
```

tests/end-to-end/upgrade/pkg/tests/monitoring/targets_rules.go
```diff
@@ -0,0 +1,285 @@
...
+		return err
+	}
+	log.Println("checking that all rules are healthy after upgrade")
+	err = t.testRulesAreHealthy()
+	if err != nil {
```
clebs, line 75:
```suggestion
	if err := t.testRulesAreHealthy(); err != nil {
```

tests/end-to-end/upgrade/pkg/tests/monitoring/targets_rules.go
```diff
@@ -0,0 +1,285 @@
...
+			return errors.Errorf(timeoutMessage)
+		case <-tick.C:
+			var resp promAPI.TargetsResponse
+			url := fmt.Sprintf("%s/api/v1/targets", prometheusURL)
+			respBody, statusCode, err := t.doGet(url)
+			if err != nil {
+				return errors.Wrap(err, "cannot query targets")
+			}
+			err = json.Unmarshal([]byte(respBody), &resp)
+			if err != nil {
```
clebs, line 176:
```suggestion
			err := json.Unmarshal([]byte(respBody), &resp); err != nil {
```

tests/end-to-end/upgrade/pkg/tests/monitoring/targets_rules.go
```diff
@@ -0,0 +1,285 @@
...
+			return errors.Errorf(timeoutMessage)
+		case <-tick.C:
+			var resp promAPI.AlertResponse
+			url := fmt.Sprintf("%s/api/v1/rules", prometheusURL)
+			respBody, statusCode, err := t.doGet(url)
+			if err != nil {
+				return errors.Wrap(err, "cannot query rules")
+			}
+			err = json.Unmarshal([]byte(respBody), &resp)
+			if err != nil {
```
clebs, line 216:
```suggestion
			if err := json.Unmarshal([]byte(respBody), &resp); err != nil {
```

tests/end-to-end/upgrade/pkg/tests/monitoring/promAPI/targets.go
```diff
@@ -0,0 +1,24 @@
+package promAPI
```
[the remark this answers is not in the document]
shorim: package is renamed now to `prom`

tests/end-to-end/upgrade/pkg/tests/monitoring/targets_rules.go
```diff
@@ -0,0 +1,277 @@
...
+			if expectedAlertManagers == actualAlertManagers && expectedNodeExporter == actualNodeExporter && expectedPrometheusInstances == actualPrometheusInstances && expectedKubeStateMetrics == actualKubeStateMetrics && expectedGrafanaInstance == actualGrafanaInstance {
+				return nil
+			}
+		}
+	}
+}
+
+func (t TargetsAndRulesTest) testTargetsAreHealthy() error {
+	timeout := time.After(3 * time.Minute)
+	tick := time.NewTicker(30 * time.Second)
```
piotrmiskiewicz, line 155: do we need to wait 30 seconds for the first check?

piotrmiskiewicz:
General question: how log does it take the "testResource" method usually?

tests/end-to-end/upgrade/pkg/tests/monitoring/targets_rules.go
shorim: @piotrmiskiewicz the idea is that the scraping interval is 30 seconds, so if in the first tick one of the targets was down, then there is no meaning to check again in less than 30 seconds beacause promtheus would have not yet done another scraping and the check will still fail

shorim: but at the same time you are right, this makes the first check only happens after 30 seconds, so in total that makes the test runs in around 1 minute

piotrmiskiewicz: so please do the first check without waiting 30 seconds

PR #8159, opened 2020-04-22
Introduce our own runtime for functions

aerfio:
**Description**

- ...
- ...
- ...

**Related issue(s)**

### DIFF
--- a/components/function-controller/Makefile
+++ b/components/function-controller/Makefile
@@ -23,7 +23,7 @@ else
 GOBIN=$(shell go env GOBIN)
 endif
 
-verify:: mod-verify
+verify:: mod-verify build-function-base-image push-function-base-image
 
 resolve-local:
 	GO111MODULE=on go mod download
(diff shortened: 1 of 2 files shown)


components/function-controller/deploy/function-base-image/nodejs12/kubeless-npm-install.sh
```diff
@@ -0,0 +1,20 @@
...
+fi
+
+cd "$KUBELESS_INSTALL_VOLUME"
+npm config set "${scope}registry" "${registry}"
+
+if [ -n "${NPM_CONFIG_EXTRA}" ]; then
+  npm config set "${NPM_CONFIG_EXTRA}"
+fi
+
+npm install --production --prefix="$KUBELESS_INSTALL_VOLUME"
```
aerfio, line 20: I had to copy this file to adapt it to sh, original one from https://github.com/kubeless/runtimes/blob/master/stable/nodejs/kubeless-npm-install.sh works only with bash

PR #8160, opened 2020-04-22
Set proper restrictions for Rafter and Knative Serving resources

michal-hudy:
**Description**

- Set read-only access for all
- Set full access for admin

Rafter and Knative Serving are Kyma dependencies, not functionalities. Because of that, users shouldn't be able to modify/create/delete their resources. For debugging purposes, they should have view access to them.

### DIFF
--- a/resources/cluster-users/templates/clusterroles-knative-serving.yaml
+++ b/resources/cluster-users/templates/clusterroles-knative-serving.yaml
@@ -1,56 +1,56 @@
 ---
-# View access to rafter resources
+# View access to knative serving resources
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRole
 metadata:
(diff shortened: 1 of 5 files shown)


michal-hudy:
@piotrmsc Why users with `view` access cannot `watch` resources?
https://github.com/kyma-project/kyma/blob/master/resources/cluster-users/values.yaml#L66

piotrmsc:
> @piotrmsc Why users with `view` access cannot `watch` resources?
> https://github.com/kyma-project/kyma/blob/master/resources/cluster-users/values.yaml#L66

If you need it for kyma-viewers you can add it :) Mostly you do watch to observe changes you apply to so developer/admin scenarios imo :)

michal-hudy:
Np. I was simply curious ;)

Issue #8203, opened 2020-04-23
Injecting pod labels in Function CR


michal-hudy:
**Description**

Add the ability to define Pod labels in Function CR. I'm not sure if we should support full PodSpec definition, for now maybe not - some investigation is needed. In UI it should be visible in the configuration as something advanced, by default Function CR labels should be propagated. Internal labels like `function-name`, `uuid` still must be propagated to Pod and override user-provided values.

If `.spec.template.metadata.labels` approach will be chosen, then resources management should also be moved to `.spec.template`... and old resourced deprecated (not removed).

**Reasons**

Some functionalities like Service Bindings are based on Pod labels that are injected. Currently, UsageKind points to Knative Serving, because it is not possible to define labels for Pods on Function CR. It is not good as we need to wait for Serving resource that can be replaced in the future or even removed in some cases.

kazydek:
@magicmatatjahu - and pls update the [Function CR](https://kyma-project.io/docs/master/components/serverless/#custom-resource-custom-resource) definition in docs :)

PR #8287, opened 2020-04-28
Inject serviceBindings labels in Function CR

magicmatatjahu:
**Description**

- extract leftovers from controller to `build_resources`
- inject serviceBindings labels in Function CR

**Related issue(s)**
https://github.com/kyma-project/kyma/issues/8203


components/function-controller/internal/controllers/serverless/build_resources.go
```diff
@@ -12,6 +15,34 @@ import (
 	serverlessv1alpha1 "github.com/kyma-project/kyma/components/function-controller/pkg/apis/serverless/v1alpha1"
 )
 
+const (
+	serviceBindingUsagesTracingAnnotation = "servicebindingusages.servicecatalog.kyma-project.io/tracing-information"
```
michal-hudy, line 19: Do we still need to track that as we don't recreate/update Function CR (only status) in the controller?

michal-hudy: Ok, I see, the title confuses me as I was sure that you migrated also `UsageKind` and `ServiceBindingUsages` to Function CR. Just add a `TODO` note that it will be not needed after migration.

components/function-controller/internal/controllers/serverless/build_resources.go
```diff
@@ -172,6 +189,118 @@ func (r *FunctionReconciler) buildService(log logr.Logger, instance *serverlessv
...
+
+	for key, value := range functionLabels {
+		podLabels[key] = value
+	}
+	for key, value := range bindingLabels {
+		podLabels[key] = value
+	}
+
+	return podLabels
+}
```
michal-hudy, line 266: Instead of having a special if for empty podLabels map, better is to create a new map that you will return - code is much simpler
```suggestion
func (r *FunctionReconciler) servicePodLabels(log logr.Logger, instance *serverlessv1alpha1.Function, oldService *servingv1.Service) map[string]string {
	functionLabels := r.functionLabels(instance)
	bindingLabels := r.retrieveBindingLabels(log, instance, oldService)
	podLabels := instance.Spec.PodLabels

	// len(functionLabels) as we cannot predict the exact size of map
	result := make(map[string]string, len(functionLabels))
	for key, value := range podLabels {
		result[key] = value
	}
	for key, value := range functionLabels {
		result[key] = value
	}
	//TODO: remove after migration of Bindings to Function CR
	for key, value := range bindingLabels {
		result[key] = value
	}

	return result
}
```

https://play.golang.org/p/wbjfV82CXvH

components/function-controller/internal/controllers/serverless/build_resources.go
```diff
@@ -172,6 +189,118 @@ func (r *FunctionReconciler) buildService(log logr.Logger, instance *serverlessv
...
+
+func (r *FunctionReconciler) retrieveBindingLabels(log logr.Logger, instance *serverlessv1alpha1.Function, oldService *servingv1.Service) map[string]string {
+	bindingAnnotation := instance.GetAnnotations()[serviceBindingUsagesTracingAnnotation]
+	functionLabels := r.retrieveBindingLabelsFromAnnotation(log, bindingAnnotation)
```
michal-hudy, line 270: Rather `bindingLabels`

components/function-controller/pkg/apis/serverless/v1alpha1/function_types.go
```diff
@@ -24,6 +24,9 @@ type FunctionSpec struct {
 
 	// +kubebuilder:validation:Minimum:=0
 	MaxReplicas *int32 `json:"maxReplicas,omitempty"`
+
+	// +optional
+	PodLabels map[string]string `json:"podLabels,omitempty"`
```
michal-hudy, line 29: Hm... I'm not sure if `PodLabels` is a good name. Maybe just a `labels` as we have with `resources`, or create a new struct that will contain labels and resources as we don't want to use the `template` field (because we will need to block most of it...).

components/function-controller/internal/controllers/serverless/build_resources.go
michal-hudy: Also, it is not good to update `podLabels` because you updating `instance.Spec.PodLabels` field in such a case and some logic that uses that field may not work as expected. It is a good practice to not modify arguments of function inside the function.

michal-hudy: Hm... you can even create a method responsible for merging labels, as we do that in many places, something like that:

```go
func mergeLabels(labels ...map[string]string) map[string]string {
	result := make(map[string]string, 0)
	for _, labelsMap := range labels {
		for key, value := range labelsMap {
			result[key] = value
		}
	}

	return result
}
```
https://play.golang.org/p/Ud9XM3SPXMF

components/function-controller/internal/controllers/serverless/build_resources.go
magicmatatjahu: After migration is also will be needed, because I retrieved in code labels from tracingAnnotations and override from `spec.podLabels` field, just to be sure of logic.

components/function-controller/internal/controllers/serverless/build_resources.go
magicmatatjahu: I changed name to `bindingLabelsFromFunction`.

components/function-controller/pkg/apis/serverless/v1alpha1/function_types.go
magicmatatjahu: I changed to `labels`, but we can thinking about this on our daily on Monday :)

components/function-controller/internal/controllers/serverless/build_resources.go
michal-hudy: hm... the situation is slightly different. `Kservice` was created and updated by us and we needed to be sure to not remove these labels, but `Function` is something managed by the user. As we don't control `Function` spec, we shouldn't change its values during reconciliation. Also, what if SBU change logic, or will remove first labels later annotation, we will need to adjust logic in `Function` controller. As this is the functionality of SBU leaves it for SBU, if the user removes labels or annotations, SBU will recreate it later. `Function` controller shouldn't have that logic as it is no longer needed - it is a user responsibility to not remove labels/annotations created by SBU.

components/function-controller/internal/controllers/serverless/build_resources.go
```diff
@@ -172,6 +184,79 @@ func (r *FunctionReconciler) buildService(log logr.Logger, instance *serverlessv
...
+}
+
+func (r *FunctionReconciler) functionLabels(instance *serverlessv1alpha1.Function) map[string]string {
+	labels := make(map[string]string, len(instance.GetLabels())+3)
+	for key, value := range instance.GetLabels() {
+		labels[key] = value
+	}
+
+	return r.mergeLabels(labels, r.internalFunctionLabels(instance))
+}
```
michal-hudy, line 215: It doesn't make sense here to copy labels from instance to separate map, smiply:
```suggestion
func (r *FunctionReconciler) functionLabels(instance *serverlessv1alpha1.Function) map[string]string {
	return r.mergeLabels(instance.GetLabels(), r.internalFunctionLabels(instance))
}
```

components/function-controller/internal/controllers/serverless/function_controller_test.go
```diff
@@ -172,13 +185,14 @@ func newFixFunction(namespace, name string) *serverlessv1alpha1.Function {
...
 	suffix := rand.Int()
+
 	return &serverlessv1alpha1.Function{
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      fmt.Sprintf("%s-%d", name, suffix),
 			Namespace: namespace,
 		},
 		Spec: serverlessv1alpha1.FunctionSpec{
-			Source: "module.exports = {main: function(event, context) {return 'Hello World 321'}}",
+			Source: "module.exports = {main: function(event, context) {return 'Hello World. Epstein didnt kill himself.'}}",
```
aerfio, line 195: 😄

Issue #7629, opened 2020-03-16
POC Advanced retry in kyma-installer


piotrmsc:
**Description**

The latest change in the installer brings retries of single release instead of retrying the whole installation/upgrade process. This simplifies the process and shortens time, however, it brings up other issues like weak chart design of releases. Let's take release core as an example : 

It's an umbrella chart for multiple components and resources likes `console, cbs, istio rbac, cluster users` etc. When something will fail installer is retrying installation/upgrade of core release, however, ends up with an error `release core already exists` and eventually it reaches maximum number of retries(5) and stop the entire process. A similar issue happens when tiller has problems with connectivity to apiserver, for example `connection refused` happens, such error is propagated by installer and is a subject of retry.

We need to do a **POC**  for improved retries.

If an error occurs during : 

1. Installation - do helm delete before retrying
2. Upgrade - find last deployed revision of release and perform the rollback. If we find all revisions failed we need to do helm delete and then install

In case of `Installation` we can set `atomic` which from tiller docs says : ```If set, installation process purges chart on fail, also sets --wait flag```

On `Upgrade` we could also use `atomic`, tiller docs mentions : `If set, upgrade process rolls back changes made in case of failed upgrade, also sets --wait flag` or  

`cleanup-on-fail `      
` Allow deletion of new resources created in this upgrade when upgrade failed`

**Reasons**

Identify how we can improve retries handling for helm releases. Which will be base for future implementations.

**Attachments**

aerfio:
There were some changes on Gardener that make this issue with Tiller being unable to connect to apiserver really painful. While investigating https://github.com/kyma-incubator/compass/issues/980 and discussing things with Gardener guys they've explained that while they detect high load they scale up apiserver, which results in Tiller losing connection. 

On our side it manifests by kyma-installler trying to install release once again, which fails and prevents Kyma installation, and it's not really dependent on any component, because we've seen it fail on istio, knative serving, knative eventing, dex and logging (see that compass issue for Tiller logs, they all look the same).

BenceJ:
my suggestion would be to upgrade and use helm 3 instead of 2, which didn't use Tiller, could be a way (hopefully) to go around this problem

piotrmsc:
Helm 3 is planned for the future with in general installation topic, currently, we want to make installer bit more resilient for tiller issues.

PR #8293, opened 2020-04-29
delete relase on install error - draft

strekm:
**Description**

- delete relase on install error - draft

**Related issue(s)**
see #7629

### DIFF
--- a/components/kyma-operator/pkg/kymahelm/kyma-helm-client.go
+++ b/components/kyma-operator/pkg/kymahelm/kyma-helm-client.go
@@ -96,14 +96,11 @@ func (hc *Client) IsReleaseDeletable(rname string) (bool, error) {
 		retry.Attempts(uint(maxAttempts)),
 		retry.DelayType(func(attempt uint, config *retry.Config) time.Duration {
 			log.Printf("Retry number %d on getting release status.\n", attempt+1)
-			return time.Duration(maxAttempts) * time.Second
+			return time.Duration(3) * time.Second
 		}),
 	)
-	if err != nil {
-		return false, err
-	}
 
-	return isDeletable, nil
+	return isDeletable, err
 }
 
 // InstallReleaseFromChart .
--- a/components/kyma-operator/pkg/kymahelm/kyma-helm-client.go
+++ b/components/kyma-operator/pkg/kymahelm/kyma-helm-client.go
@@ -4,7 +4,9 @@ import (
 	"fmt"
 	"log"
 	"strings"
+	"time"
 
+	"github.com/avast/retry-go"
 	"github.com/sirupsen/logrus"
 
 	"k8s.io/helm/pkg/chartutil"
@@ -74,12 +76,29 @@ func (hc *Client) ReleaseStatus(rname string) (string, error) {
 
 //IsReleaseDeletable returns true for release that can be deleted
 func (hc *Client) IsReleaseDeletable(rname string) (bool, error) {
-	statusRes, err := hc.helm.ReleaseStatus(rname)
+	isDeletable := false
+	maxAttempts := 3
+
+	err := retry.Do(
+		func() error {
+			statusRes, err := hc.helm.ReleaseStatus(rname)
+			if err != nil {
+				return err
+			}
+			isDeletable = statusRes.Info.Status.Code != release.Status_DEPLOYED
+			return nil
+		},
+		retry.Attempts(uint(maxAttempts)),
+		retry.DelayType(func(attempt uint, config *retry.Config) time.Duration {
+			log.Printf("Retry numebr %d on getting release status.\n", attempt+1)
+			return time.Duration(maxAttempts) * time.Second
+		}),
+	)
 	if err != nil {
 		return false, err
 	}
 
-	return statusRes.Info.Status.Code != release.Status_DEPLOYED, nil
+	return isDeletable, nil
 }
 
 // InstallReleaseFromChart .
--- a/components/kyma-operator/pkg/kymahelm/kyma-helm-client.go
+++ b/components/kyma-operator/pkg/kymahelm/kyma-helm-client.go
@@ -18,6 +18,7 @@ import (
 type ClientInterface interface {
 	ListReleases() (*rls.ListReleasesResponse, error)
 	ReleaseStatus(rname string) (string, error)
+	IsReleaseDeletable(rname string) (bool, error)
 	InstallReleaseFromChart(chartdir, ns, releaseName, overrides string) (*rls.InstallReleaseResponse, error)
 	InstallRelease(chartdir, ns, releasename, overrides string) (*rls.InstallReleaseResponse, error)
 	InstallReleaseWithoutWait(chartdir, ns, releasename, overrides string) (*rls.InstallReleaseResponse, error)
@@ -69,6 +70,16 @@ func (hc *Client) ReleaseStatus(rname string) (string, error) {
 	}
 	statusStr := fmt.Sprintf("%+v\n", status)
 	return strings.Replace(statusStr, `\n`, "\n", -1), nil
+}
+
+//IsReleaseDeletable returns true for release that can be deleted
+func (hc *Client) IsReleaseDeletable(rname string) (bool, error){
+	statusRes, err := hc.helm.ReleaseStatus(rname)
+	if err != nil {
+		return false, err
+	}
+
+	return statusRes.Info.Status.Code == release.Status_DEPLOYED || statusRes.Info.Status.Code == release.Status_FAILED, nil
 }
 
 // InstallReleaseFromChart .
--- a/components/kyma-operator/pkg/kymahelm/kyma-helm-client.go
+++ b/components/kyma-operator/pkg/kymahelm/kyma-helm-client.go
@@ -13,6 +13,7 @@ import (
 	"k8s.io/helm/pkg/helm"
 	"k8s.io/helm/pkg/proto/hapi/release"
 	rls "k8s.io/helm/pkg/proto/hapi/services"
+	"k8s.io/helm/pkg/storage/errors"
 	"k8s.io/helm/pkg/tlsutil"
 )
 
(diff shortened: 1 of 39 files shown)


components/kyma-operator/pkg/kymahelm/kyma-helm-client.go
```diff
@@ -71,6 +72,16 @@ func (hc *Client) ReleaseStatus(rname string) (string, error) {
 	return strings.Replace(statusStr, `\n`, "\n", -1), nil
 }
 
+//IsReleaseDeletable returns true for release that can be deleted
+func (hc *Client) IsReleaseDeletable(rname string) (bool, error) {
+	statusRes, err := hc.helm.ReleaseStatus(rname)
```
piotrmsc, line 77: we will need retries here

components/kyma-operator/pkg/kymahelm/kyma-helm-client.go
```diff
@@ -71,6 +72,16 @@ func (hc *Client) ReleaseStatus(rname string) (string, error) {
...
+	if err != nil {
+		return false, err
+	}
+
+	return statusRes.Info.Status.Code == release.Status_DEPLOYED || statusRes.Info.Status.Code == release.Status_FAILED, nil
```
piotrmsc, line 82: on install failure we can check if status is != DEPLOYED, so other statuses like SUPERSEDED,FAILED,PENDING(..) we can ignore,however, in the future upgrade they will be important

components/kyma-operator/pkg/kymahelm/kyma-helm-client.go
```diff
@@ -71,6 +74,33 @@ func (hc *Client) ReleaseStatus(rname string) (string, error) {
...
 
+//IsReleaseDeletable returns true for release that can be deleted
+func (hc *Client) IsReleaseDeletable(rname string) (bool, error) {
+	isDeletable := false
+	maxAttempts := 3
+
+	err := retry.Do(
+		func() error {
+			statusRes, err := hc.helm.ReleaseStatus(rname)
+			if err != nil {
```
piotrmsc, line 85: probably we'll need to check here for not found error. `helm status xxx` reports `Error: getting deployed release "xxx": release: "xxx" not found` for nonexisting releases (which is the case when error is returned on prepare-release phase in tiller)

components/kyma-operator/pkg/kymahelm/kyma-helm-client.go
```diff
@@ -71,6 +75,37 @@ func (hc *Client) ReleaseStatus(rname string) (string, error) {
 	return strings.Replace(statusStr, `\n`, "\n", -1), nil
 }
 
+//IsReleaseDeletable returns true for release that can be deleted
+func (hc *Client) IsReleaseDeletable(rname string) (bool, error) {
```
piotrmsc, line 79: Nit: consider that in the future this method will be also used in upgrade scenarios which will have different predicate determining if release is ok for delete.

components/kyma-operator/pkg/kymainstallation/step.go
```diff
@@ -60,7 +60,27 @@ func (s installStep) Run() error {
...
+			_, err := s.helmClient.DeleteRelease(s.component.GetReleaseName())
+
+			if err != nil {
+				deleteErrMsg := fmt.Sprintf("Helm delete of %s failed with an error: %s", s.component.GetReleaseName(), err.Error())
+				log.Println(deleteErrMsg)
+				return errors.New(fmt.Sprintf("%s \n %s \n", installErrMsg, deleteErrMsg))
+			}
+		}
+
+		return errors.New(installErrMsg)
```
piotrmsc, line 83: Maybe let's enhance the error message with info that installation failed but cleanup was done. This will less confusing during debugging

components/kyma-operator/pkg/kymahelm/kyma-helm-client.go
strekm: i would go with different method for upgrade rather

components/kyma-operator/pkg/kymainstallation/step.go
```diff
@@ -60,7 +60,31 @@ func (s installStep) Run() error {
...
+			log.Println("Look at the proceeding logs to see the reason of failure")
+			_, err := s.helmClient.DeleteRelease(s.component.GetReleaseName())
+
+			if err != nil {
+				deleteErrMsg := fmt.Sprintf("Helm delete of %s failed with an error: %s", s.component.GetReleaseName(), err.Error())
+				log.Println(deleteErrMsg)
+				return errors.New(fmt.Sprintf("%s \n %s \n", installErrMsg, deleteErrMsg))
+			}
+
+			errorMsg = fmt.Sprintf("%s\nHelm delete of %s was successfull", installErrMsg, s.component.GetReleaseName())
```
piotrmsc, line 84: We may need to wait here after delete command. `helm del` reports status but it does not always mean that release is fully deleted. Release has status deleting and if we start installation quickly enough, we will have error that release is being deleted, another delete and another retry but we increase the delay between retries.

strekm: introduced hardcoded 5s sleep, in the future we might think of more sophisticated solution

components/kyma-operator/pkg/kymahelm/kyma-helm-client.go
```diff
@@ -71,6 +75,37 @@ func (hc *Client) ReleaseStatus(rname string) (string, error) {
...
+				}
+				return err
+			}
+			isDeletable = statusRes.Info.Status.Code != release.Status_DEPLOYED
+			return nil
+		},
+		retry.Attempts(uint(maxAttempts)),
+		retry.DelayType(func(attempt uint, config *retry.Config) time.Duration {
+			log.Printf("Retry numebr %d on getting release status.\n", attempt+1)
+			return time.Duration(maxAttempts) * time.Second
```
Tomasz-Smelcerz-SAP, line 99: I don't understand why the delay corresponds to maxAttempts. If we increase maxAttempts, the delay increases as well. Is that what we want? Why don't we have fixed delay?

components/kyma-operator/pkg/kymahelm/kyma-helm-client.go
```diff
@@ -71,6 +75,37 @@ func (hc *Client) ReleaseStatus(rname string) (string, error) {
...
+					return nil
+				}
+				return err
+			}
+			isDeletable = statusRes.Info.Status.Code != release.Status_DEPLOYED
+			return nil
+		},
+		retry.Attempts(uint(maxAttempts)),
+		retry.DelayType(func(attempt uint, config *retry.Config) time.Duration {
+			log.Printf("Retry numebr %d on getting release status.\n", attempt+1)
```
Tomasz-Smelcerz-SAP, line 98:
```suggestion
			log.Printf("Retry number %d on getting release status.\n", attempt+1)
```

components/kyma-operator/pkg/kymahelm/kyma-helm-client.go
```diff
@@ -71,6 +75,37 @@ func (hc *Client) ReleaseStatus(rname string) (string, error) {
...
+			return time.Duration(maxAttempts) * time.Second
+		}),
+	)
+	if err != nil {
```
Tomasz-Smelcerz-SAP, line 102: We could save a few lines just by:
```
return isDeletable, err
```

Reason:
isDeletable is true IFF err == nil.
If err != nil, isDeletable is always false.

components/kyma-operator/pkg/kymainstallation/step.go
Tomasz-Smelcerz-SAP: What if instead of this fixed 5s sleep we put there a function that gets the release status (we have such code already) and returns true when helm reports "release does not exist" (we have such code already)? Then we wrap this into a retry.Do and perform it 25 times with 5 sec fixed delay. This is a simple function, about 20 loc. and it gives us way more flexibility: 5 sec resolution up to 2 minutes of waiting.

components/kyma-operator/pkg/kymahelm/kyma-helm-client.go
```diff
@@ -71,6 +75,37 @@ func (hc *Client) ReleaseStatus(rname string) (string, error) {
...
+		func() error {
+			statusRes, err := hc.helm.ReleaseStatus(rname)
+			if err != nil {
+				if strings.Contains(err.Error(), errors.ErrReleaseNotFound(rname).Error()) {
+					isDeletable = false
+					return nil
+				}
+				return err
+			}
+			isDeletable = statusRes.Info.Status.Code != release.Status_DEPLOYED
```
Tomasz-Smelcerz-SAP, line 93: What about status "DELETING"? Is such a release deletable, or Helm will throw back an error if we try to delete it?

strekm: helm will throw an error and we will have retry, which would work if release was deleted

components/kyma-operator/pkg/kymahelm/kyma-helm-client.go
```diff
@@ -71,6 +75,34 @@ func (hc *Client) ReleaseStatus(rname string) (string, error) {
...
+			return nil
+		},
+		retry.Attempts(uint(maxAttempts)),
+		retry.DelayType(func(attempt uint, config *retry.Config) time.Duration {
+			log.Printf("Retry number %d on getting release status.\n", attempt+1)
+			return time.Duration(3) * time.Second
```
Tomasz-Smelcerz-SAP, line 99:
```suggestion
			return time.Duration(fixedDelay) * time.Second
```
Let's avoid "magic" constants - it should be named properly. Declare `fixedDelay` next to maxAttempts.

components/kyma-operator/pkg/kymainstallation/step.go
```diff
@@ -60,7 +61,34 @@ func (s installStep) Run() error {
...
+			_, err := s.helmClient.DeleteRelease(s.component.GetReleaseName())
+
+			if err != nil {
+				deleteErrMsg := fmt.Sprintf("Helm delete of %s failed with an error: %s", s.component.GetReleaseName(), err.Error())
+				log.Println(deleteErrMsg)
+				return errors.New(fmt.Sprintf("%s \n %s \n", installErrMsg, deleteErrMsg))
+			}
+
+			//waiting for release to be deleted
+			time.Sleep(time.Second * time.Duration(10))
```
Tomasz-Smelcerz-SAP, line 86:
```suggestion
			time.Sleep(time.Second * time.Duration(deleteWaitTime))
```
and declare `deleteWaitTime` as 10

Issue #8283, opened 2020-04-28
Fix core-test-external-solution-event-mesh test on Gardener


marcobebway:
**Description**

The `kyma-integration-gardener-azure` prow job is failing.

**Job**

https://status.build.kyma-project.io/view/gcs/kyma-prow-logs/logs/kyma-integration-gardener-azure/1255029318887673856.

**Logs**
```
time="2020-04-28T07:51:33Z" level=info msg="Step: 'Send Cloud Event to Mesh'"
time="2020-04-28T07:51:34Z" level=info msg="Step: 'Check counter pod'"
time="2020-04-28T07:58:01Z" level=error msg="Error in 'Check counter pod': the counter pod is not updated: All attempts fail:
#1: counter different then expected value: Got: 0 but expected 1
#120: counter different then expected value: Got: 0 but expected 1"
```

**AC**

- The issue is fixed and the `kyma-integration-gardener-azure` is green.

sayanh:
The cluster creation was broken with missing components which was fixed [here](https://github.com/kyma-project/kyma/pull/8276). The [test](https://status.build.kyma-project.io/view/gcs/kyma-prow-logs/logs/kyma-integration-gardener-azure/1255099159380234240) was successful after the fix. Let's keep an eye on this issue.

sayanh:
As it turns out, this test is flaky. It failed again https://storage.googleapis.com/kyma-prow-logs/logs/kyma-integration-gardener-azure/1255401508933996544/build-log.txt.

sayanh:
I did a quick test, it seems like the timeout is too little for the event to traverse from Azure event hub to Kyma. The counter pod does get the counter incremented as the event reaches but after the test times out. I think increasing the timeout for this step, `CheckCounterPod` can render robustness to this test.

a-thaler:
The test still has an high error rate for https://status.build.kyma-project.io/?job=kyma-integration-gardener-azure

radufa:
During my tests, I've found other errors too:
- timeout at Post request:
```
  time="2020-05-06T09:19:48Z" level=info msg="Step: 'Register test service'"
  time="2020-05-06T09:19:48Z" level=info msg="Step: 'Create service instance: e2e-event-mesh'"
  time="2020-05-06T09:19:55Z" level=info msg="Step: 'Create service binding'"
  time="2020-05-06T09:19:55Z" level=info msg="Step: 'Create service binding usage'"
[...]

PR #8355, opened 2020-05-04
Make retry configurable and fix "core-test-external-solution-event-mesh" test

radufa:
**Description**

- Make "retry" configurable per default and refactor code
- Extend CheckCounterPod to allow configurable "options" 
- Fix the E2E Event Mesh  scenario to make it more robust 

**Related issue(s)**

See also #8283

### DIFF
--- a/resources/core/templates/tests/test-external-solution.yaml
+++ b/resources/core/templates/tests/test-external-solution.yaml
@@ -1,4 +1,5 @@
 {{- if .Values.test.external_solution.event_mesh.enabled }}
+{{- $val := .Values.test.external_solution -}}
 ---
 apiVersion: "testing.kyma-project.io/v1alpha1"
 kind: TestDefinition
--- a/resources/core/templates/tests/test-external-solution.yaml
+++ b/resources/core/templates/tests/test-external-solution.yaml
@@ -4,11 +4,11 @@
 apiVersion: "testing.kyma-project.io/v1alpha1"
 kind: TestDefinition
 metadata:
-  name: {{ $.Chart.Name }}-test-external-solution-{{ $suffix | replace "_" "-" }}
+  name: {{ $.Chart.Name }}-test-external-solution
   labels:
     kyma-project.io/test-kind: e2e
     app: {{ $.Chart.Name }}-test-external-solution
-    app.kubernetes.io/name: {{ $.Chart.Name }}-test-external-solution-{{ $suffix | replace "_" "-" }}
+    app.kubernetes.io/name: {{ $.Chart.Name }}-test-external-solution
     app.kubernetes.io/managed-by: {{ $.Release.Service }}
     app.kubernetes.io/instance: {{ $.Release.Name }}
     helm.sh/chart: {{ $.Chart.Name }}-{{ $.Chart.Version | replace "+" "_" }}
--- a/resources/core/templates/tests/test-external-solution.yaml
+++ b/resources/core/templates/tests/test-external-solution.yaml
@@ -1,5 +1,5 @@
-{{- if .Values.test.external_solution.event_mesh.enabled }}
-{{- $val := .Values.test.external_solution -}}
+{{- range $suffix, $val := .Values.test.external_solution }}
+  {{- if and ( $val.enabled ) ( $.Capabilities.APIVersions.Has "testing.kyma-project.io/v1alpha1" ) }}
 ---
 apiVersion: "testing.kyma-project.io/v1alpha1"
 kind: TestDefinition
@@ -54,4 +54,5 @@ spec:
               curl -XPOST http://127.0.0.1:15020/quitquitquit
               sleep 4
               exit $exit_code
+  {{- end}}
 {{- end }}
--- a/resources/core/templates/tests/test-external-solution.yaml
+++ b/resources/core/templates/tests/test-external-solution.yaml
@@ -1,5 +1,4 @@
-{{- range $suffix, $val := .Values.test.external_solution }}
-  {{- if and ( $val.enabled ) ( $.Capabilities.APIVersions.Has "testing.kyma-project.io/v1alpha1" ) }}
+{{- if .Values.test.external_solution.event_mesh.enabled }}
 ---
 apiVersion: "testing.kyma-project.io/v1alpha1"
 kind: TestDefinition
@@ -54,5 +53,4 @@ spec:
               curl -XPOST http://127.0.0.1:15020/quitquitquit
               sleep 4
               exit $exit_code
-  {{- end}}
 {{- end }}
--- a/tests/end-to-end/external-solution-integration/internal/http/cloudevent_client.go
+++ b/tests/end-to-end/external-solution-integration/internal/http/cloudevent_client.go
@@ -27,7 +27,7 @@ func NewWrappedCloudEventClient(ceClient cloudevents.Client, opts ...retrygo.Opt
 }
 
 func (c *WrappedCloudEventClient) Send(ctx context.Context, event cloudevents.Event) (ct context.Context, evt *cloudevents.Event, err error) {
-	err = retry.WithCustomOpts(func() error {
+	err = retry.Do(func() error {
 		ct, evt, err = c.underlying.Send(ctx, event)
 		return err
 	}, c.options...)
--- a/tests/end-to-end/external-solution-integration/internal/http/cloudevent_client.go
+++ b/tests/end-to-end/external-solution-integration/internal/http/cloudevent_client.go
@@ -2,12 +2,21 @@ package http
 
 import (
 	"context"
+	"time"
 
 	retrygo "github.com/avast/retry-go"
 	cloudevents "github.com/cloudevents/sdk-go"
 
 	"github.com/kyma-project/kyma/tests/end-to-end/external-solution-integration/pkg/retry"
 )
+
+const retryAttemptsCount = 20
+const retryDelay = 250 * time.Millisecond
+
+var defaultOpts = []retrygo.Option{
+	retrygo.Attempts(retryAttemptsCount),
+	retrygo.Delay(retryDelay),
+}
 
 type WrappedCloudEventClient struct {
 	underlying cloudevents.Client
@@ -21,7 +30,7 @@ type ResilientCloudEventClient interface {
 func NewWrappedCloudEventClient(ceClient cloudevents.Client, opts ...retrygo.Option) *WrappedCloudEventClient {
 	var client = &WrappedCloudEventClient{
 		underlying: ceClient,
-		options:    opts,
+		options:    append(defaultOpts, opts...),
 	}
 	return client
 }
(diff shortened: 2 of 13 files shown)


resources/core/templates/tests/test-external-solution.yaml
```diff
@@ -4,11 +4,11 @@
 apiVersion: "testing.kyma-project.io/v1alpha1"
 kind: TestDefinition
 metadata:
-  name: {{ $.Chart.Name }}-test-external-solution-{{ $suffix | replace "_" "-" }}
+  name: {{ $.Chart.Name }}-test-external-solution
```
nachtmaar, line 7: Can we also remove the for loop above? We only have one test at the moment. The loop is only used to get `$val.testsuite` which evaluates to `e2e-event-mesh`. So we still use the name inside the TestDefiniton.

tests/end-to-end/external-solution-integration/internal/scenario/event_mesh/event_mesh.go
```diff
@@ -28,6 +29,7 @@ const (
 
 var (
 	apiRuleRes = schema.GroupVersionResource{Group: "gateway.kyma-project.io", Version: "v1alpha1", Resource: "apirules"}
+	sleep      = func(s *Scenario) time.Duration { return time.Duration(s.waitTime) * time.Second }
```
nachtmaar, line 32: (nit) What is the benefit of assigning it to a variable? We could just have a function named `sleep` 🤔

tests/end-to-end/external-solution-integration/internal/scenario/event_mesh/event_mesh.go
```diff
@@ -67,16 +69,15 @@ func (s *Scenario) Steps(config *rest.Config) ([]step.Step, error) {
 				s.applicationGroup, appOperatorClientset.ApplicationconnectorV1alpha1().Applications(),
 				httpSourceClientset.HTTPSources(kymaIntegrationNamespace)),
 		),
-		step.Parallel(
```
nachtmaar, line 70: What is the reasoning behind removing Parallel?

tests/end-to-end/external-solution-integration/internal/scenario/event_mesh/event_mesh.go
```diff
@@ -67,16 +69,15 @@ func (s *Scenario) Steps(config *rest.Config) ([]step.Step, error) {
...
-			testsuite.NewConnectApplication(connector, state, s.applicationTenant, s.applicationGroup),
-		),
+		testsuite.NewCreateMapping(s.testID, appBrokerClientset.ApplicationconnectorV1alpha1().ApplicationMappings(s.testID)),
+		testsuite.NewDeployFakeLambda(s.testID, helpers.LambdaPayload, helpers.LambdaPort,
+			coreClientset.AppsV1().Deployments(s.testID),
+			coreClientset.CoreV1().Services(s.testID),
+			coreClientset.CoreV1().Pods(s.testID),
+			true),
+		testsuite.NewStartTestServer(testService),
+		testsuite.NewSleep(sleep(s)),
```
nachtmaar, line 79: For what condition to be fulfilled are we sleeping? Instead of just waiting a specified time, can we wait for a specific condition to happen? Something get's ready?

tests/end-to-end/external-solution-integration/internal/scenario/event_mesh/wrapper_counter_pod.go
```diff
@@ -0,0 +1,22 @@
...
+const retryAttemptsCount = 240
+const retryDelay = 1 * time.Second
+
+var opts = []retrygo.Option{
+	retrygo.Attempts(retryAttemptsCount),
+	retrygo.Delay(retryDelay),
+}
+
+// NewCheckCounterPod returns a new CheckCounterPod
+func NewWrappedCounterPod(testService *testkit.TestService, expectedValue int) *testsuite.CheckCounterPod {
```
nachtmaar, line 20: (nit) Why don't we move the `opts` to `testsuite.NewCheckCounterPod` ? So we file would not be required at all. Otherwise the only reason for this file is to set a default value which could be set in the original file as well :) You can compare for empty struct and use the defaults then :)

tests/end-to-end/external-solution-integration/pkg/testkit/events.go
```diff
@@ -42,7 +42,7 @@ func NewEventSender(httpClient *http.Client, domain, application string) *EventS
 	return &EventSender{
 		httpClient: resilient.WrapHttpClient(httpClient),
 		domain:     domain,
-		ceClient:   ceClient,
+		ceClient:   http2.NewWrappedCloudEventClient(ceClient),
```
nachtmaar, line 45: (nit) I know you didn't introduce the import name `http2`, but can you rename it to something more meaningful? :)

tests/end-to-end/external-solution-integration/internal/scenario/event_mesh/event_mesh.go
radufa: This is why I've want to have the second step to solve all these issue, see: https://app.zenhub.com/workspaces/kyma---all-repositories-5b6d5985084045741e744dea/issues/kyma-project/kyma/8409

tests/end-to-end/external-solution-integration/internal/scenario/event_mesh/wrapper_counter_pod.go
radufa: The "check_counter_pod" continue to use as default value, like in the past (see in "pkg/retry/retry.go"): 
```
const retryAttemptsCount = 120
const retryDelay = 1 * time.Second
```
But a user of CheckCounterPod  can set new values from outside. I didn't want to change the old behavior of  CheckCounterPod. The Wrapper could set in the future other config values too, overwriting some default values, if necessary.
It's the same "pattern" like in NewWrappedCloudEventClient, which sets its default values too, replacing the default values contained in "pkg/retry/retry.go"

tests/end-to-end/external-solution-integration/internal/scenario/event_mesh/scenario.go
```diff
@@ -13,6 +13,7 @@ type Scenario struct {
 	skipSSLVerify     bool
 	applicationTenant string
 	applicationGroup  string
+	waitTime          int
```
marcobebway, line 16: Does it make sense to change `int` to `time.Duration`.

tests/end-to-end/external-solution-integration/internal/scenario/event_mesh/event_mesh.go
radufa: To let the code to be more compact, like in the past. But it's used as a function:
```
testsuite.NewSleep(sleep(s))
```

tests/end-to-end/external-solution-integration/internal/scenario/event_mesh/event_mesh.go
radufa: It was not reliable under load constraints. This should be also better solved in: https://app.zenhub.com/workspaces/kyma---all-repositories-5b6d5985084045741e744dea/issues/kyma-project/kyma/8409

tests/end-to-end/external-solution-integration/internal/scenario/event_mesh/wrapper_counter_pod.go
marcobebway: Also, maybe have something like inside:

```golang
func GetDefaultRetryOpts() []retrygo.Option {
	return []retrygo.Option{
		retrygo.Attempts(retryAttemptsCount),
		retrygo.Delay(retryDelay),
	}
}
```

and put it inside `kyma/tests/end-to-end/external-solution-integration/pkg/testsuite/check_counter_pod.go`.

tests/end-to-end/external-solution-integration/pkg/testkit/events.go
radufa: Done

tests/end-to-end/external-solution-integration/internal/scenario/event_mesh/scenario.go
radufa: Done

resources/core/templates/tests/test-external-solution.yaml
radufa: Done

tests/end-to-end/external-solution-integration/internal/scenario/event_mesh/wrapper_counter_pod.go
radufa: The caller, in our case NewWrappedCounterPod overwrites the default values of  CheckCounterPod (which gets its own default values from "retry"). This is similar with NewWrappedCloudEventClient implementation. I've used the same "pattern" and the wrapper can always set its config values, overriding the default values of used components if necessary. Of course it can add new values too. I did'n want to change to much now.

tests/end-to-end/external-solution-integration/pkg/testsuite/check_counter_pod.go
```diff
@@ -12,15 +13,17 @@ import (
 type CheckCounterPod struct {
 	testService   *testkit.TestService
 	expectedValue int
+	retry_opts    []retrygo.Option
```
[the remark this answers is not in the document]
radufa: done

tests/end-to-end/external-solution-integration/internal/scenario/event_mesh/wrapper_counter_pod.go
```diff
@@ -0,0 +1,22 @@
...
+
+import (
+	"time"
+
+	retrygo "github.com/avast/retry-go"
+	"github.com/kyma-project/kyma/tests/end-to-end/external-solution-integration/pkg/testkit"
+	"github.com/kyma-project/kyma/tests/end-to-end/external-solution-integration/pkg/testsuite"
+)
+
+const retryAttemptsCount = 240
```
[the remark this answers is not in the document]
radufa: I've used the same constants as in the most deeper retry component used, see "pkg/retry/retry.go":
```
const retryAttemptsCount = 120
const retryDelay = 1 * time.Second
```
to be more clear that it will overwrite them

tests/end-to-end/external-solution-integration/internal/http/cloudevent_client.go
```diff
@@ -2,13 +2,22 @@ package http
...
 	"github.com/kyma-project/kyma/tests/end-to-end/external-solution-integration/pkg/retry"
 )
 
+const retryAttemptsCount = 20
+const retryDelay = 250 * time.Millisecond
+
+var defaultOpts = []retrygo.Option{
+	retrygo.Attempts(retryAttemptsCount),
+	retrygo.Delay(retryDelay),
+}
```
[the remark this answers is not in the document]
radufa: I've used the same code structure like in the our "basic" retry code, see: "pkg/retry/retry.go". I hope that the "retry" code is more understandable.

tests/end-to-end/external-solution-integration/pkg/testsuite/check_counter_pod.go
```diff
@@ -1,6 +1,7 @@
 package testsuite
 
 import (
+	retrygo "github.com/avast/retry-go"
```
k15r, line 4: (nit) `github.com/avast/retry-go` uses the package name `retry`. Why do we want to rename this to the imho uglier `retrygo`?

radufa: To have the same import alias like in "internal/http/cloudevent_client.go".

Issue #8259, opened 2020-04-27
Test functions with bindings and triggers


aerfio:
**Description**

Throught tests that validate whole solution need to be created - right now we lack tests for service bindings and triggers, also some of the features of serverless-controller-manager, like deleting old completed jobs, need to be tested

**Reasons**

**Attachments**

michal-hudy:
Epic: https://github.com/kyma-project/kyma/issues/8256

PR #8364, opened 2020-05-05
Add integration tests for servicebindingusages in serverless-controller-manager testsuite

aerfio:
**Description**

- fix tests dockerfile - it no longer downloads go modules every time 🎉 
- add docs for used envs
- add tests for the servicebindingusage, which in turn needs addonsconfiguration, servicebinding and service instance 💣 
- clean up a bit
- move function body from fmt.Sprintf to template, so it's easier to read 

**Related issue(s)**

#8259


tests/function-controller/pkg/servicebindingusage/servicebindingusage.go
```diff
@@ -0,0 +1,168 @@
...
+	"github.com/kyma-project/kyma/tests/function-controller/pkg/servicebindingusage/types/v1alpha1"
+	"github.com/kyma-project/kyma/tests/function-controller/pkg/shared"
+
+	"github.com/pkg/errors"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+)
+
+const (
+	usedByName = "knative-service"
```
aerfio, line 22: make it an env

tests/function-controller/testsuite/testsuite.go
```diff
@@ -106,62 +133,94 @@ func (t *TestSuite) Run() {
...
+	err = t.servicebinding.Create(t.cfg.ServiceInstanceName)
+	failOnError(t.g, err)
+
+	t.t.Log("Waiting for service binding to have ready phase...")
+	err = t.servicebinding.WaitForStatusRunning()
+	failOnError(t.g, err)
+
+	t.t.Log("Creating service binding usage...")
+	// we are deliberately creating servicebindingusage HERE, to test how it behaves after function update
+	err = t.servicebindingusage.Create(t.cfg.ServiceBindingName, t.cfg.FunctionName, t.cfg.UsageKindName, redisEnvPrefix)
```
magicmatatjahu, line 169: You can put UsageKind as one of field in SBU struct, and not pass UsageKInd in Create function.

tests/function-controller/testsuite/testsuite.go
```diff
@@ -171,7 +230,17 @@ func (t *TestSuite) Run() {
 	failOnError(t.g, err)
 
 	t.t.Log("Testing local connection through the service")
-	err = t.pollForAnswer(inClusterURL, gotEventMsg)
+	err = t.pollForAnswer(inClusterURL, "", gotEventMsg)
+	failOnError(t.g, err)
+
+	answerForEnvPing := "Redis port: 6379"
```
magicmatatjahu, line 236: It could be a one of const defined above in code.

tests/function-controller/testsuite/testsuite.go
```diff
@@ -189,6 +258,18 @@ func (t *TestSuite) Cleanup() {
 	err = t.broker.Delete()
 	failOnError(t.g, err)
 
+	err = t.servicebindingusage.Delete()
+	failOnError(t.g, err)
+
+	err = t.servicebinding.Delete()
```
magicmatatjahu, line 264: ServiceBindingUsage has as ownerReference ServiceBinding, so you can delete only SB, but this is not a problem :)

aerfio: I create that manually, so I there's any owner reference

aerfio: ok, I just learned that sbu controller does it for us, nice

still, I think we could leave it here, just to be explicit

PR #8369, opened 2020-05-05
Enable caching in Serverless and disable colors in logs 

michal-hudy:
**Description**

- Enable caching - when there are no changes in `package.json` lambda is built in `9s`, when there are no changes at all then `5s` and sometimes less
- Disable colors in output as it is looking ugly in logs

   Before:
   image

   After:
   image

    Unfortunately, the timestamp column is added by UI...


aerfio:
have a look at https://github.com/GoogleContainerTools/kaniko#--log-format

michal-hudy:
Yup, I know. I created this PR to generate image :P screens are from changes in `kaniko` arguments

michal-hudy:
I'm not sure if we should configure `--cache-ttl`, default it is two weeks: https://github.com/GoogleContainerTools/kaniko/blob/master/cmd/executor/cmd/root.go#L157

components/function-controller/internal/controllers/serverless/build_resources.go
```diff
@@ -88,7 +88,7 @@ func (r *FunctionReconciler) buildJob(instance *serverlessv1alpha1.Function, con
 						{
 							Name:  "executor",
 							Image: r.config.Build.ExecutorImage,
-							Args:  []string{fmt.Sprintf("--destination=%s", imageName), "--insecure", "--skip-tls-verify"},
+							Args:  append(r.config.Build.ExecutorArgs, fmt.Sprintf("--destination=%s", imageName), "--context=dir:///workspace"),
```
michal-hudy, line 91: `--destination` and `--context` should be configurable as it is strictly related to volumes and docker repository...

components/function-controller/internal/controllers/serverless/config.go
```diff
@@ -24,7 +24,8 @@ type BuildConfig struct {
 	LimitsMemory         string            `envconfig:"default=1Gi"`
 	LimitsMemoryValue    resource.Quantity `envconfig:"-"`
 	RuntimeConfigMapName string            `envconfig:"default=dockerfile-nodejs-12"`
-	ExecutorImage        string            `envconfig:"default=gcr.io/kaniko-project/executor:v0.19.0"`
+	ExecutorArgs         []string          `envconfig:"default=--insecure;--skip-tls-verify;--skip-unused-stages;--log-format=text;--cache=true"`
```
aerfio, line 27: --insecure and --skip-tls-verify shouldn't be set from env

components/function-controller/internal/controllers/serverless/function_controller.go
```diff
@@ -147,8 +147,16 @@ func (r *FunctionReconciler) onJobChange(ctx context.Context, log logr.Logger, i
...
-	return existing.Spec.Template.Spec.Containers[0].Args[0] == expected.Spec.Template.Spec.Containers[0].Args[0]
+	// Compare arguments as it contains image tag
+	if len(existing.Spec.Template.Spec.Containers[0].Args) != len(expected.Spec.Template.Spec.Containers[0].Args) {
+		return false
+	}
+	for i, arg := range existing.Spec.Template.Spec.Containers[0].Args {
+		if expected.Spec.Template.Spec.Containers[0].Args[i] != arg {
+			return false
+		}
+	}
```
aerfio, line 158: isn't there a bug somewhere? what if those slices have the same elements but in different order

resources/serverless/values.yaml
```diff
@@ -100,8 +100,10 @@ containers:
         value: 5m
       functionRequeueDuration:
         value: 5m
+      functionBuildExecutorArgs:
+        value: "--insecure,--skip-tls-verify,--skip-unused-stages,--log-format=text,--cache=true"
```
aerfio, line 104: see previous comment

components/function-controller/internal/controllers/serverless/config.go
michal-hudy: Not exactly, this is only valid when our internal docker registry is used

components/function-controller/internal/controllers/serverless/function_controller.go
michal-hudy: It shouldn't be a problem as Job is created by the controller, but in fact, we should only compare `--destination` arg as it contains image tag. Other args shouldn't trigger build.

michal-hudy: So, I will change it.

PR #8370, opened 2020-05-05
Move ory crds to cluster essentials

m00g3n:
**Description**

- Move ory custom resource definitions to cluster essentials

**Related issue(s)**

See also: #7933


resources/cluster-essentials/templates/crd-init-oathkeeper.yaml
```diff
@@ -22,9 +22,28 @@ spec:
         - name: crd-rules
           mountPath: /etc/ory/crd
           readOnly: true
-        command: ["kubectl",  "apply", "-f", "/etc/ory/crd/rules.yaml"]
+        command:
+        - /bin/bash
+        - -c
+        - |-
+          set -e
```
piotrmsc, line 29: In general we should consider retries in all kubectl operastions in jobs. Cool example can be found here with timeout & until https://github.com/kyma-project/kyma/blob/master/installation/resources/tiller.yaml#L129
crafted by our bash hero @Demonsthere

resources/cluster-essentials/templates/crd-init-hydra-master.yaml
```diff
@@ -22,9 +22,31 @@ spec:
         - name: crd-oauth2clients
           mountPath: /etc/crd
           readOnly: true
-        command: ["kubectl",  "apply", "-f", "/etc/crd/oauth2clients.yaml"]
+        command:
+        - timeout
```
Demonsthere, line 26: Since we are doing bash in bash here, maybe to simplify this structure a little bit, move it to a separate file in the files directory, and include it with Files.Get? The current form is very hard to read

m00g3n: Good point. I think the best idea would be to do it in a follow-up task that we could start to work on once we will have all `CRDs` in `cluster-essentials`. We could also come up with some generic mechanism (e.g. base on directory structure) that would have a logic to install/update all the `CRDs` .

piotrmsc:
Please also update README of CE repo https://github.com/kyma-project/kyma/tree/master/resources/cluster-essentials. 

With all your pr's moving CRDs we need to properly describe what is now `cluster-essentials` repo, what is being installed :) /cc @bszwarc

PR #8379, opened 2020-05-06
Add tests for revision controller in serverless-controller-manager

aerfio:
**Description**

- as in title, add e2e tests for this controller

**Related issue(s)**

See #8259


tests/function-controller/pkg/revision/revision.go
```diff
@@ -0,0 +1,131 @@
...
+		r.log.Logf("There are still %d revisions, need only 1", len(revs.Items))
+	} else {
+		r.log.Logf("Old revisions are deleted, only 1 left")
+	}
+	if r.verbose {
+		r.log.Logf("%+v", revs)
+	}
+}
+
+func (r *Revision) VerifyConfigurationGeneration(generation int) error {
```
magicmatatjahu, line 108: This function should return (bool, error)

Issue #7862, opened 2020-03-31
Assure that namespace admin can follow backup guide


a-thaler:
**Description**
With https://github.com/kyma-project/kyma/issues/7861 there will be a new guide on how to create volume snapshots and restore them. Please assure that the namespace-admin role can execute that guide

**Reasons**

**Attachments**

clebs:
The tested role `kyma-namespace-admin-essentials` does not have the necessary permissions to do anything other than creating namespaces.
The following permissions are missing in order to be able to create a volume snapshot (optional permissions in brackets):
- `StorageClass`: get, create, (delete)
- `PVC`: get, create, (delete)
- `Deployment`: get, create, (delete)
- `Pods`: get, (delete)
- `Snapshot Class`: get, create, (delete)
- `Snapshot`: get, create, (delete)

clebs:
Retested with the correct setup as the previous role information was incomplete.
Currently the namespace admin role has 2 issues preventing the backups:
1. Can't create a `Snapshot Class` since it is a non-namespaced resource: Gardener team will add a default `Snapshot Class` out of the box that customers can use when provisioning.
2. Can't create, get, list, delete or update `Snapshot` resources on their namespace: #8382 adds those permissions to the namespace admin.

With both things in place, customers will be able to create persistent volume snapshots by following the guide.

PR #8382, opened 2020-05-06
Volume snapshots role

clebs:
**Description**

- Add rights to manage volume snapshots for admins:
  - Add a new cluster role to grant permissions over volume snapshots and append it to admin adn ns admin roles.
  - This is needed for admins to create and maintain backups for volumes .

**Related issue(s)**
#7862

### DIFF
--- a/resources/cluster-users/templates/rbac-roles.yaml
+++ b/resources/cluster-users/templates/rbac-roles.yaml
@@ -216,6 +216,33 @@ aggregationRule:
 rules: []
 
 ---
+################################################################################
+# kyma-snapshots = manage volume snapshots for backups.
+################################################################################
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: kyma-snapshots
+  labels:
+    app: kyma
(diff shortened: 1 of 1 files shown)


resources/cluster-users/templates/rbac-roles.yaml
```diff
@@ -215,6 +215,33 @@ aggregationRule:
...
+    chart: {{ .Chart.Name }}-{{ .Chart.Version }}
+    rbac.authorization.kyma-project.io/aggregate-to-kyma-snapshots: "true"
+  annotations:
+    helm.sh/hook-weight: "0"
+rules:
+- apiGroups:
+    - "snapshot.storage.k8s.io"
+  resources:
+    - "volumesnapshots"
+  verbs:
```
Tomasz-Smelcerz-SAP, line 237: There is a predefined list of "edit" verbs in values. You can use it here like this:
```
{{ toYaml .Values.clusterRoles.verbs.edit | indent 4 }}
```

clebs: Applied your suggestion, thanks!

PR #5176, opened 2019-08-02
Label eventing resources

antoineco:
**Description**

Changes proposed in this pull request:

- Use recommended Kubernetes labels in eventing components

This is an attempt to start defining a more standard labeling pattern in Kyma eventing components. Touching labels of resources borrowed from external projects (knative-eventing, knative-provisioner-natss) is error prone so I left these untouched until we discuss internally what to do with these.

The first few commits are purely cosmetic (automated changes using sed).

**Related issue(s)**

#5025

PR #8430, opened 2020-05-11
Remove event-bus removal pre-upgrade job

nachtmaar:
**Description**

- event-bus was removed with this pre-upgrade job in migration to Kyma 1.12. Since event-bus is gone, the script is not required anymore

**Related issue(s)**


sayanh:
Would it make sense to delete this [job](https://github.com/kyma-project/kyma/blob/master/resources/nats-streaming/templates/migration-jobs.yaml) as well?

nachtmaar:
> Would it make sense to delete this [job](https://github.com/kyma-project/kyma/blob/master/resources/nats-streaming/templates/migration-jobs.yaml) as well?

From what I understand it was only required once to handle the selector updates. See the PR which introduced it: https://github.com/kyma-project/kyma/pull/5176/files
I would like to remove it in another small PR :)

sayanh:
> selector updates

Right, as the selectors are good now, the pre-upgrade hook is irrelevant. 

> I would like to remove it in another small PR :)

Sure, this works.

PR #8438, opened 2020-05-12
Installer retries last deployed rev

strekm:
**Description**

- release step enhanced with info about last deployed rev 

**Related issue(s)**
see #7629


components/kyma-operator/pkg/kymahelm/kyma-helm-client.go
```diff
@@ -104,6 +105,24 @@ func (hc *Client) IsReleaseDeletable(rname string) (bool, error) {
 	return isDeletable, err
 }
 
+func (hc *Client) ReleaseDeployedRevision(rname string) (int32, error) {
+	maxHistory := 10
```
piotrmsc, line 109: it matches history set in tiller however, should be configurable

piotrmsc:
temporary approve to unblock upgrade handling but in general listing releases, finding its revision should be done different.

PR #8439, opened 2020-05-12
change the key for audit log

rakesh-garimella:
**Description**

- Since the key in configmap is called `auditlog-url-basic` update the deployment so.
**Related issue(s)**


a-thaler:
Force-merged as it is one config line reviewed by three people

Issue #8408, opened 2020-05-08
monitoring test fails - knative-eventing-event-mesh-dashboard-{broker,httpsource} targets not healthy


nachtmaar:
**Description**
Happened on ci-force https://sap-cx.slack.com/archives/CBMPTMCV9/p1588856758372800

https://status.build.kyma-project.io/view/gcs/kyma-prow-logs/logs/post-master-kyma-gke-compass-integration/1258474546667720705:

```
Target with job=knative-eventing/knative-eventing-event-mesh-dashboard-broker and instance=10.32.1.35:9090 is not healthy
Target with job=knative-eventing/knative-eventing-event-mesh-dashboard-httpsource and instance=10.32.2.66:9091 is not healthy
```

**Expected result**

Targets are up.

**Actual result**
Targets are down.

**Steps to reproduce**

**Troubleshooting**

How to get all cases where the same problem happened: 
```shell
gcloud logging read 'textPayload:"Target with job=knative-eventing/knative-eventing-event-mesh-dashboard-" AND textPayload:"is not healthy" AND receiveTimestamp > "2020-05-07T00:00:00.00Z"' --project sap-kyma-prow-workloads 
```

[event-mesh-targets-down.log](https://github.com/kyma-project/kyma/files/4597769/event-mesh-targets-down.log)

PR #8446, opened 2020-05-12
Fix monitoring test

sayanh:
**Description**

- It is found that while running the tests(`monitoring` and `core-test-external-solution-event-mesh`) in parallel, the targets created in `core-test-external-solution-event-mesh` were getting ready while the `monitoring` test was checking the targets. Hence, those targets were not ready causing the monitoring test to fail.
- The solution is to exclude the targets created by the test. 

**Related issue(s)**

See https://github.com/kyma-project/kyma/issues/8408

### DIFF
--- a/tests/integration/monitoring/test-monitoring.go
+++ b/tests/integration/monitoring/test-monitoring.go
@@ -170,7 +170,7 @@ func testTargetsAreHealthy() {
 
 }
 
-func hasLabels(targetLabels promAPI.Labels, labelsToBeIgnored promAPI.Labels) bool {
+func hasLabels(targetLabels, labelsToBeIgnored promAPI.Labels) bool {
 	for l, _ := range targetLabels {
 		if labelsToBeIgnored[l] == targetLabels[l] {
 			return true
--- a/tests/integration/monitoring/test-monitoring.go
+++ b/tests/integration/monitoring/test-monitoring.go
@@ -153,7 +153,7 @@ func testTargetsAreHealthy() {
 			timeoutMessage = ""
 			for _, target := range activeTargets {
 				// Ignoring the targets with certain labels
-				if hasAnyLabels(target.Labels, labelsToBeIgnored) {
+				if hasAnyLabel(target.Labels, labelsToBeIgnored) {
 					continue
 				}
 				if target.Health != "up" {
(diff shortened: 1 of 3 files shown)


tests/integration/monitoring/test-monitoring.go
```diff
@@ -128,6 +128,9 @@ func testPodsAreReady() {
 func testTargetsAreHealthy() {
 	timeout := time.After(3 * time.Minute)
 	tick := time.NewTicker(30 * time.Second)
+	var labelsToBeIgnored = promAPI.Labels{
+		"namespace": "e2e-event-mesh", // e2e test pod
```
nachtmaar, line 132: I am wondering. If the test causes problems with the core-test-external-solution test, would it make sense to `not` run the monitoring test in `parallel` ? Maybe other tests are also creating dynamic targets and have the same problem.

tests/integration/monitoring/test-monitoring.go
```diff
@@ -163,6 +170,15 @@ func testTargetsAreHealthy() {
 
 }
 
+func hasLabels(targetLabels, labelsToBeIgnored promAPI.Labels) bool {
```
nachtmaar, line 173: (nit) Wouldn't `hasAnyLabel` be a better name 🤔, because the function does not check if all `labelsToBeIgnored ` are part of `targetLabels`. Instead it returns, when one key/value pair from `labelsToBeIgnored` is found in `targetLabels`. 
```suggestion
func hasAnyLabel(targetLabels, labelsToBeIgnored promAPI.Labels) bool {
```

tests/integration/monitoring/test-monitoring.go
sayanh: If that's the case, then running monitoring tests, `not` in parallel will not help as the targets created by the other tests get deleted once they are done. Ignoring those targets is a better option, IMO which doesn't impact the total execution time of the test suite.

tests/integration/monitoring/test-monitoring.go
k15r: (nit) i think the name is correct, just the code is the wrong way around....
loop over labels to be ignored and check if the target has those ... not the other way round

sayanh: changed to `hasAnyLabel `

PR #8450, opened 2020-05-13
Define codeowners for microfrontend definitions

kwiatekus:
**Description**

- define the only contributors of the MF definitions as code-owners. code owners of the charts usually don't know about the micro front-ends and how they build the UI navigation. Changes in those files should not be merged unnoticed and unreviewed by the owners.


CODEOWNERS
```diff
@@ -388,3 +382,26 @@ milv.config.yaml @michal-hudy @m00g3n @aerfio @pPrecel @magicmatatjahu @kazydek
 
 # performance tests
 /tests/perf/ @a-thaler @hisarbalik @lilitgh @suleymanakbas91 @clebs @rakesh-garimella
+
+# The `api-gateway ui` micro front-end definition
+/resources/api-gateway/templates/api-rules-microfrontend.yaml @dariadomagala @parostatkiem @akucharska @Wawrzyn321 @kwiatekus
+
+# add-on ui microfrontend definitions
+/resources/helm-broker/charts/addons-ui/templates/addons-clustermicrofrontend.yaml @dariadomagala @parostatkiem @akucharska @Wawrzyn321 @kwiatekus
+/resources/helm-broker/charts/addons-ui/templates/addons-microfrontend.yaml @dariadomagala @parostatkiem @akucharska @Wawrzyn321 @kwiatekus
```
akucharska, line 391: Should we add also files with images? For example /resources/helm-broker/values.yaml

sjanota: We should extract it to sub-chart so ownership is clear

kwiatekus: done

PR #8452, opened 2020-05-13
Add appropriate owners to serverless and rafter domain

sjanota:
**Description**

Give @michal-hudy @aerfio @pPrecel @magicmatatjahu access to rafter and serverless-related parts of CBS.

**Related issue(s)**


magicmatatjahu:
@sjanota Thanks for this idea! Can I ask you, about adding us to another two files `schema_gen.go` and `models_gen.go`? They also are changed on every time if you run gqlgen script. Thanks!

sjanota:
It is included in `gqlschema` package

a-thaler:
I don't like that you have CODEowners defined on file level as any renaming or splitting into multiple files will cause problems. Maybe put the single file into a folder at least to have more flexibility?

PR #8453, opened 2020-05-13
Fix error handling in certhelpers

### LINKED ISSUE #8447: tiller-certhelper script does not handle errors
**Description**

The [tiller-certhelper](https://github.com/kyma-project/kyma/blob/master/installation/resources/tiller.yaml#L58-L131) script has a bug:

```
---> Generate CA
Generating RSA private key, 4096 bit long modulus (2 primes)
...........................++++
............................................................................++++
e is 65537 (0x010001)
---> Generate Tiller key
Generating RSA private key, 4096 bit long modulus (2 primes)
..++++
................................++++
e is 65537 (0x010001)
Signature ok
subject=C = PL, ST = Gliwice, L = Gliwice, O = Tiller Server, CN = tiller-server
Getting CA Private Key
---> Create secrets in k8s
error: unable to recognize "STDIN": Get https://100.64.0.1:443/api?timeout=32s: dial tcp 100.64.0.1:443: i/o timeout
tiller-secret configured.
```

as you can see from the above logs Secret was not created because of timeout on api server but in logs, we see that `tiller-secret configured.` and no error is returned. As a result, the tiller and kyma-installer were crashing.

**Expected result**

If an error occurred, the retry is executed

**Actual result**

If an error occurred, the operation is marked as succeeded

Demonsthere:
**Description**

- Add retry for kubectl operations
- Add failonerror

**Related issue(s)**

- Fixes #8447

### DIFF
--- a/installation/resources/tiller.yaml
+++ b/installation/resources/tiller.yaml
@@ -126,7 +126,7 @@ spec:
             type: Opaque
             EOF
 
-            timeout 35s bash -c 'until kubectl create -f /tmp/secret.yaml ; do sleep 5; done'
+            timeout 60s bash -c 'until kubectl create -f /tmp/secret.yaml ; do sleep 2; done'
             echo "tiller-secret configured."
         volumeMounts:
         - mountPath: /etc/certs
--- a/installation/resources/tiller.yaml
+++ b/installation/resources/tiller.yaml
@@ -46,7 +46,6 @@ spec:
         name: tiller
     spec:
       serviceAccountName: tiller
-
       initContainers:
       - name: init-tiller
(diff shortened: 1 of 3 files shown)


installation/resources/tiller.yaml
```diff
@@ -125,11 +125,9 @@ spec:
               namespace: kube-system
             type: Opaque
             EOF
-            )
 
-            echo "${COMBO_YAML}" | kubectl apply -f -
+            timeout 35s bash -c 'until kubectl create -f /tmp/secret.yaml ; do sleep 5; done'
```
piotrmsc, line 129: will this work with upgrade? kubectl create returns error if object already exists

PR #8485, opened 2020-05-15
[PoC] Installer singleton

Demonsthere:
**Description**

- Change Deployment to StatefulSet

**Conclusions**:
- Swapping deployment for statefulset won't implement all wanted features, but will work as the `Recreate` strategy for deployments
- We don't need PVC/PV when working with a statefullset
- Statefulsets have static names (`kyma-installer-0`) which can make some docs simpler
- We cannot forbid the use of `kubectl scale` on the resource, but we can limit its effect
- By adding an HPA resource with min and max pods set 1, we can force that only 1 replica can be active, and other replicas are deleted if created. 
- In order to implement a full singleton mode in the installer, we need to add a `--enable-leader-election` option to the installer ensuring that only one instance is active

**Related issue(s)**


Demonsthere:
Verification: Forcing scaling with `kubectl scale`

Statefullset with default strategy:
Zrzut ekranu 2020-05-15 o 14 13 05

Deployment (current solution)
Zrzut ekranu 2020-05-15 o 14 25 46

Statefullset with `parallel` strategy:
Zrzut ekranu 2020-05-15 o 14 31 58

Demonsthere:
We close this PoC, as to have a full Singleton mode of operation, we would need to implement `leader-election` within the installer. As the changes in this PR don't give enough functionality and   a big refactor to the installer was deemed not important at the present.

Issue #8492, opened 2020-05-18
"Instance Connection Name" nomenclature is inconsistent with Google's Usage & Docs


24601:
**Description**

This issue describes a possible mix of user-error and vague documentation around a mismatch in what exactly is meant by "Instance Connection Name". Experienced users of Google Cloud SQL trained by Google's training materials, programs, or documentation (or simply using their UI and REST APIs) expect this value to be one thing, and Kyma expects a subset of this value in its configuration. Failure to do this results in a hard install failure that is difficult and troublesome to diagnose given cryptic error messages. A user skilled in both Cloud SQL and Kyma may still have trouble discerning this subtle yet important difference in vernacular. While it is unclear who is at "fault" or what is "broken" or to "blame" (or if anything really is), and Kyma is technically not wrong or incorrect, it is an issue that can be easily solved by clearing up documentation with a simple sentence or two. 

When using Google Cloud SQL as the persistent store for Ory/Hydra in the production configurations, configuring the desired Cloud SQL Instance requires setting the instance name of the Google Cloud SQL Instance so the Google Cloud SQL Proxy can connect to the appropriate instances. 

In Google's documentation & parlance, the "instance connection name" is a tuple/triplet style identifier following the GCP_PROJECT:GCP_REGION:INSTANCE_NAME format. Google's APIs and Cloud Console (web interface) provide and reference (as well as their attendant documentation) the "Instance Connection Name" as this _entire_ string. Kyma's documentation references the "Instance Connection Name" (or similar analogues such as the configuration item gcloud-sqlproxy.cloudsql.instance.instanceName).

https://github.com/kyma-project/kyma/blob/master/docs/security/05-05-production-profile.md#L94

However, Kyma's intent (and its expectation) in using this vernacular is to reference *only* the third element of the tuple/triplet of Google's definition/usage of the identifier. 

As this does not match Google (or the understanding of users familiar with Cloud SQL), it creates an ambiguity that results in a very frustrating and difficult to debug misconfiguration that presents itself as a bug at first until the user dives into the actual Helm chart to see that Kyma assembles this tuple/triplet on its own and intends/expects the user to provide the "Instance Connection Name" not in the format provided by the Google Cloud console, the REST API, or their other tooling, but having the leading two components of this identifier stripped. 

[...]

PR #8493, opened 2020-05-18
Improve possible ambiguity in gcloud-sqlproxy configuration

24601:
**Description**
Resolves an ambiguity in what "Instance Connection Name" means through more clear description of the difference in the value Google's Console/APIs provide and the value that Kyma expects to be entered into the configuration.

- Update docs to be precise on exactly the format of the substring to enter from the full string provided by Google's console/APIs

**Related issue(s)**

Proposed solution for #8492


24601:
Yay! Tiny little speck of sand on the beach, but first accepted PR to this project, looking forward to contributing more, thanks for the help improving it and getting it merged!

Issue #8498, opened 2020-05-18
Monitoring tests failed - Target with job=knative-eventing/knative-eventing-event-mesh-dashboard-broker is not healthy


nachtmaar:
**Description**
```
Fetching logs
  Start of logs from container "tests" in pod "oct-tp-testsuite-all-2020-05-18-00-01-monitoring-0" in status "Failed"
  2020/05/18 00:28:04 Test pods status: All pods are ready!
  2020/05/18 00:31:04 Target with job=knative-eventing/knative-eventing-event-mesh-dashboard-broker and instance=10.16.1.42:9090 is not healthy
  Target with job=knative-eventing/knative-eventing-event-mesh-dashboard-broker and instance=10.16.1.43:9090 is not healthy
  End of logs from container "tests" in pod "oct-tp-testsuite-all-2020-05-18-00-01-monitoring-0" in status "Failed"
  Start of logs from container "tests" in pod "oct-tp-testsuite-all-2020-05-18-00-01-monitoring-1" in status "Failed"
  2020/05/18 00:31:16 Test pods status: All pods are ready!
  2020/05/18 00:34:16 Target with job=knative-eventing/knative-eventing-event-mesh-dashboard-broker and instance=10.16.1.42:9090 is not healthy
  Target with job=knative-eventing/knative-eventing-event-mesh-dashboard-broker and instance=10.16.1.43:9090 is not healthy
  End of logs from container "tests" in pod "oct-tp-testsuite-all-2020-05-18-00-01-monitoring-1" in status "Failed"
- Fetching logs from testing pods in Unknown status...
```

Might be related to #8446

Issue #6457, opened 2019-11-28
Create a Proposal of E2E Metrics Tests 


sayanh:
**Description**

There need to be E2E tests for event-mesh metrics. But this needs to be solved at kyma level. Let's try to create a proposal to evaluate the right way to test metrics for a given component and roll it out in the community repo.

**Reasons**

To prevent regression, automated tests for  metrics should be part of the E2E test. pipeline. (Avoid Broken DASHBOARDS!!!)

**AC** 
- Create a proposal to evaluate the right technique to test Metrics Test.

nachtmaar:
@sayanh Can you add more information here? What should the e2e test exactly do? That are the steps required for manual testing?

nachtmaar:
I would say:
Test that there are metrics at the following dashboards and that they make sense:
- Kyma / Event Bus
- Kyma / Event Bus / Subscription

k15r:
closing for now, will recreate an issue for this improvement if we still think this is required in the future

PR #8505, opened 2020-05-18
Ignore dynamic targets for monitoring test

sayanh:
**Description**

- Following targets are only`up` when an application is created along with binding an event-service to a ns:
  - knative-eventing/knative-eventing-event-mesh-dashboard-broker
  - knative-eventing/knative-eventing-event-mesh-dashboard-httpsource
- When monitoring tests are executed in parallel to e2e tests, it does not make sense to check whether these targets are up or not cause there is no deterministic approach to that.
- These targets will eventually be tested here: https://github.com/kyma-project/kyma/issues/6457

**Related issue(s)**

See https://github.com/kyma-project/kyma/issues/8498

### DIFF
--- a/tests/integration/monitoring/test-monitoring.go
+++ b/tests/integration/monitoring/test-monitoring.go
@@ -169,7 +169,7 @@ func testTargetsAreHealthy() {
 
 func shouldIgnoreTarget(target promAPI.Labels) bool {
 	var jobsToBeIgnored = []string{
-		// TODO [sayanh]: These targets will be tested here: https://github.com/kyma-project/kyma/issues/6457
+		// Note: These targets will be tested here: https://github.com/kyma-project/kyma/issues/6457
 		"knative-eventing/knative-eventing-event-mesh-dashboard-broker",
 		"knative-eventing/knative-eventing-event-mesh-dashboard-httpsource",
 	}
--- a/tests/integration/monitoring/test-monitoring.go
+++ b/tests/integration/monitoring/test-monitoring.go
@@ -128,9 +128,7 @@ func testPodsAreReady() {
 func testTargetsAreHealthy() {
 	timeout := time.After(3 * time.Minute)
 	tick := time.NewTicker(30 * time.Second)
-	var labelsToBeIgnored = promAPI.Labels{
-		"namespace": "e2e-event-mesh", // e2e test pod
-	}
+
 	var timeoutMessage string
 	for {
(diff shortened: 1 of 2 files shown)


tests/integration/monitoring/test-monitoring.go
```diff
@@ -152,13 +150,12 @@ func testTargetsAreHealthy() {
...
 			for _, target := range activeTargets {
-				// Ignoring the targets with certain labels
-				if hasAnyLabel(target.Labels, labelsToBeIgnored) {
+				if shouldIgnoreTarget(target.Labels) {
 					continue
 				}
 				if target.Health != "up" {
 					allTargetsAreHealthy = false
-					timeoutMessage += fmt.Sprintf("Target with job=%s and instance=%s is not healthy\n", target.Labels["job"], target.Labels["instance"])
+					timeoutMessage += fmt.Sprintf("Target with job=%s and instance=%s is not healthy, with labels=%v\n", target.Labels["job"], target.Labels["instance"], target.Labels)
```
marcobebway, line 158: A good candidate for a `strings.Builder`.

tests/integration/monitoring/test-monitoring.go
```diff
@@ -170,9 +167,15 @@ func testTargetsAreHealthy() {
...
-		if target[l] == anyOf[l] {
+func shouldIgnoreTarget(target promAPI.Labels) bool {
+	var jobsToBeIgnored = []string{
+		// Note: These targets will be tested here: https://github.com/kyma-project/kyma/issues/6457
+		"knative-eventing/knative-eventing-event-mesh-dashboard-broker",
+		"knative-eventing/knative-eventing-event-mesh-dashboard-httpsource",
+	}
+
+	for _, j := range jobsToBeIgnored {
+		if target["job"] == j {
```
marcobebway, line 178: (nit) Does it make sense to use lower case `target["job"]`?

tests/integration/monitoring/test-monitoring.go
sayanh: I agree. There are other occurrences of this pattern that need to be handled holistically in the file. Hence I would keep it as is for the sake of consistency.

tests/integration/monitoring/test-monitoring.go
sayanh: I'd keep it as is for now. Let's take this offline.

PR #8554, opened 2020-05-21
Add missing Serverless info to docs

kazydek:
**Description**

- Clearly stated that the `allow` handler is not supported in access strategies for Functions.
- Added info that the Function's body must have `main` as a handler name.

### DIFF
--- a/docs/serverless/03-01-api-rules.md
+++ b/docs/serverless/03-01-api-rules.md
@@ -13,7 +13,7 @@ To expose a Function outside the cluster, you must create an [APIRule custom res
 
 1. Create the APIRule CR where you specify the Function to expose, define a [Oathkeeper Access Rule](/components/api-gateway/#details-available-security-options) to secure it, and list which HTTP request methods you want to enable for it.
 
-> **CAUTION:** If you decide to expose your Function on an unsecured endpoint, use the `noop` **handler** for the **accessStrategy** you define in the APIRule CR. The `allow` value for the **handler** is not supported in the current Serverless implementation.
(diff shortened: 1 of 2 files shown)


docs/serverless/06-01-function-cr.md
```diff
@@ -69,6 +69,8 @@ spec:
         type: ConfigurationReady
 ```
 
+> **CATION:** When you create a Function, the exported object in the Function's body must have `main` as a handler name.
```
klaudiagrz, line 72: Shouldn't **handler** be put in bold here (just as in the other document)? 🤔

docs/serverless/03-01-api-rules.md
```diff
@@ -13,6 +13,8 @@ To expose a Function outside the cluster, you must create an [APIRule custom res
 
 1. Create the APIRule CR where you specify the Function to expose, define a [Oathkeeper Access Rule](/components/api-gateway/#details-available-security-options) to secure it, and list which HTTP request methods you want to enable for it.
 
+> **CAUTION:** If you decide to expose your Function on an unsecured endpoint, use the `noop` **handler** for the **accessStrategy** you define in the APIRule CR. The `allow` value for the **handler** is not supported in the current Serverless implementation.
```
majakurcius, line 16:
```suggestion
> **CAUTION:** If you decide to expose your Function through an unsecured endpoint, use the `noop` **handler** for the **accessStrategy** you define in the APIRule CR. The `allow` value for the **handler** is not supported in the current Serverless implementation.
```
I'm wondering whether `through` would not be better here.

docs/serverless/06-01-function-cr.md
majakurcius, line 72:
```suggestion
> **CATION:** When you create a Function, the exported object in the Function's body must have `main` as the handler name.
```

kazydek: As discussed, I will leave it as it is.

Issue #8180, opened 2020-04-23
APP Broker after restart is not triggering the interrupted provisioning/deprovisioning


mszostok:
**Description**

The App Broker after restart is not triggering the interrupted provisioning/deprovisioning. To fix that we can introduce similar process as already implemented in KEB, see: 
https://github.com/kyma-incubator/compass/blob/22072888dbd4e6774a28c282a9d7f716aaacb400/components/kyma-environment-broker/cmd/broker/main.go#L252-L255

**AC**
- App Broker retriggers not finished operations after pod restart

PK85:
When done, patch kyma 1.12 and contact @adamwalach

PR #8565, opened 2020-05-25
Upgrade service instance populator

jasiu001:
**Description**

During provisioning/deprovisioning ServiceInstance if an interruption occurs, AppBroker does not resume the process. The interruption may occur in four cases:

_1. provision process, before sending a response and execute asynchronous process_ 
In this case, after restoring the AppBroker, the ServiceCatalog will resume the provisioning process itself. ServiceInstance receives a failed status and will be processed at least one hour after interruption.

_2. provision process, after sending a response and execute asynchronous process_
ServiceInstance receives a failed status with `ErrorPollingLastOperation` condition state. ServiceCatalog will ask about the last operation state at least half an hour after starting the provisioning process. In this case, the AppBroker operation has to be restored with `in progress` status, AppBroker triggers the provisioning process again. 

_3. deprovision process, before sending a response and execute asynchronous process_ 
ServiceInstance receives a failed status with `DeprovisionCallFailed` condition state. ServiceCatalog will not resume the deprovisioning process itself (it waits for AppBroker connection and removes ServiceInstance). AppBroker triggers the deprovisioning process again. 

_4. deprovision process, after sending a response and execute asynchronous process_
ServiceInstance receives a failed status with `ErrorPollingLastOperation` condition state. 
AppBroker triggers the deprovisioning process again. 

Additionally, PR solves a restore issue with `true/ready` ServiceInstance by restoring the provision operation, without it removing ServiceInstance ends with error on the AppBroker side.

**Related issue(s)**
#8180


components/application-broker/internal/broker/provision.go
```diff
@@ -189,6 +196,33 @@ func (svc *ProvisionService) Provision(ctx context.Context, osbCtx osbContext, r
 	return resp, nil
 }
 
+// ProvisionProcess triggers provision process for other than broker (http) calls
+func (svc *ProvisionService) ProvisionProcess(req RestoreProvisionRequest) error {
+	var app *internal.Application
+	err := wait.Poll(500*time.Millisecond, 10*time.Second, func() (done bool, err error) {
```
polskikiel, line 202: Let's use here `PollImmediate` to speed up the execution

components/application-broker/internal/storage/populator/instances.go
```diff
@@ -69,11 +93,132 @@ func (p *Instances) Do(ctx context.Context) error {
...
+		if err != nil {
+			return errors.Wrap(err, "while add operation for failed instance (DeprovisionCallFailed)")
+		}
+		p.broker.DeprovisionProcess(broker.DeprovisionProcessRequest{
+			Instance:             p.converter.MapServiceInstance(si),
+			OperationID:          opID,
+			ApplicationServiceID: internal.ApplicationServiceID(si.Spec.ServiceClassRef.Name),
+		})
+	default:
+		p.log.Infof("ServiceInstance will to be populated, unsupported mode (%s - last state: %s)",
```
polskikiel, line 159: will to be -> will not be ?

components/application-broker/internal/storage/populator/instances.go
```diff
@@ -69,11 +93,132 @@ func (p *Instances) Do(ctx context.Context) error {
...
+			return readyMode
+		}
+	}
+
+	// case when AppBroker restarts before SC receives response with operation ID
+	if instance.Status.LastConditionState == "DeprovisionCallFailed" {
+		return deprovisionCallFailedMode
+	}
+
+	if instance.Status.LastOperation == nil {
```
polskikiel, line 187: shouldn't we check this at beginning?

components/application-broker/internal/broker/server.go
```diff
@@ -25,10 +26,12 @@ type (
 
 	provisioner interface {
 		Provision(ctx context.Context, osbCtx osbContext, req *osb.ProvisionRequest) (*osb.ProvisionResponse, *osb.HTTPStatusCodeError)
+		ProvisionProcess(request RestoreProvisionRequest) error
 	}
 
 	deprovisioner interface {
 		Deprovision(ctx context.Context, osbCtx osbContext, req *osb.DeprovisionRequest) (*osb.DeprovisionResponse, error)
+		DeprovisionProcess(req DeprovisionProcessRequest)
```
polskikiel, line 34: How about calling these functions `DeprovisionReprocess` and `ProvisionReprocess`?

components/application-broker/internal/storage/populator/instances.go
mszostok, line 187: could you add a comment on what does it mean?

components/application-broker/internal/storage/populator/instances.go
```diff
@@ -69,11 +93,132 @@ func (p *Instances) Do(ctx context.Context) error {
...
+			return errors.Wrap(err, "while add operation for failed instance (DeprovisionCallFailed)")
+		}
+		opID, err := p.addOperation(si, internal.OperationTypeRemove, internal.OperationStateInProgress, "")
+		if err != nil {
+			return errors.Wrap(err, "while add operation for failed instance (DeprovisionCallFailed)")
+		}
+		p.broker.DeprovisionProcess(broker.DeprovisionProcessRequest{
+			Instance:             p.converter.MapServiceInstance(si),
+			OperationID:          opID,
+			ApplicationServiceID: internal.ApplicationServiceID(si.Spec.ServiceClassRef.Name),
```
mszostok, line 156: **[BLOCKER] needs to be investigated before merging it to master**

did you test that with api-packages?  IMO may not work properly

Here you have a helper method to select a proper id:
https://github.com/kyma-project/kyma/blob/f2c3b3498f91c22250ddbc7a6a4449b679a40263/components/application-broker/internal/broker/broker.go#L201-L227

I was using it in other OSB API calls. The problem is that in the new approach:
- `appID` = `classID`
- `serviceID` = `planID`

in the old approach:
- `serviceID` = `classID`
- `hardcoded-id`  = `planID`


Please check that and if changes will be required please also add a test case to highlight that contract and to prevent breaking it in the future.

components/application-broker/internal/broker/provision.go
```diff
@@ -189,6 +196,33 @@ func (svc *ProvisionService) Provision(ctx context.Context, osbCtx osbContext, r
...
+		app, err = svc.appSvcFinder.FindOneByServiceID(req.ApplicationServiceID)
+		if err != nil {
+			return false, errors.Wrap(err, "while finding application based on service ID")
```
mszostok, line 205: returning error will abort wait loop, is that intended? or you want to also retry on error?

components/application-broker/internal/storage/populator/instances.go
jasiu001: in the beginning, instances with ready status are checked where `LastOperation` is not necessary, then instances with failed status which didn't get a response so it means last operation id was not responded. After that instances with failed status and `ErrorPollingLastOperation` reason are checked what means they got a response and wait for operation status.

components/application-broker/internal/broker/server.go
jasiu001: changed

components/application-broker/internal/storage/populator/instances.go
```diff
@@ -69,11 +93,132 @@ func (p *Instances) Do(ctx context.Context) error {
 	for _, si := range serviceInstances {
 		if si.Spec.ServiceClassRef != nil {
 			if _, ex := abClassNames[si.Spec.ServiceClassRef.Name]; ex {
-				if err := p.inserter.Insert(p.converter.MapServiceInstance(si)); err != nil {
```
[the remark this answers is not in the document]
jasiu001: sounds good to me, done

components/application-broker/internal/broker/provision.go
jasiu001: retry on error, fixed

components/application-broker/internal/storage/populator/instances.go
jasiu001: fixed, please take a look

PR #8566, opened 2020-05-25
Refactor rafter CRDs 

m00g3n:
**Description**

- Refactor rafter CRDs

**Related issue(s)**

See also: #7933
See also: https://github.com/kyma-project/kyma/issues/8609


m00g3n:
@michal-hudy 
- The transition from `{{- if .Values.installCRDs -}}` to `{{- if not (.Capabilities.APIVersions.Has` is proposition of standardization and will fullfill requirements yo've provided; personally I see no reason for custom logic if the [functionality](https://helm.sh/docs/chart_template_guide/builtin_objects/) is provided by package manager out of the box.
- I'm not sure if installation of rafter `CRDs` via `Job` is the best option; in this case we will end up with duplicated set of partially stubbed `CRDs` just to prevent package manager from removing `CRDs` after job is done in update scenario.

michal-hudy:
@m00g3n 
> The transition from {{- if .Values.installCRDs -}} to {{- if not (.Capabilities.APIVersions.Has is proposition of standardization and will fullfill requirements yo've provided; personally I see no reason for custom logic if the functionality is provided by package manager out of the box.

Please also align it in `kyma-project/rafter`, because as I said, we would like to have these charts in sync

> I'm not sure if installation of rafter CRDs via Job is the best option; in this case we will end up with duplicated set of partially stubbed CRDs just to prevent package manager from removing CRDs after job is done in update scenario.

I only noted that you are moving other CRDs to `resources/cluster-essentials/files` and then you are installing them with the job. So I would expect that all of our CRDs will be installed the same way.

resources/rafter/charts/controller-manager/templates/assetgroups.crd.yaml
```diff
@@ -1,4 +1,5 @@
-{{- if .Values.installCRDs -}}
+{{- if not (.Capabilities.APIVersions.Has "rafter.kyma-project.io/v1beta1/AssetGroup") }}
+{{- if .Release.IsUpgrade }}
```
michal-hudy, line 2: Anyway, it should be rather `or`? If the cluster already has `rafter.kyma-project.io/v1beta1/AssetGroup` then it will not upgrade, because the first `if` will be false.

michal-hudy: But hm... it will be still wrong. The first installation will skip that, but the next one will update to the version that is in a chart...

m00g3n: As agreed
- we will remove rafter `CRDs` from `kyma/rafter` chart
- we will leave rafter `CRDs` in `cluster-essentials` 
- we will create separate issue to install `assets`, `clusterassetgroups`, `assetgroups`, `clusterbuckets`, `buckets`, `clusterassets` in a `Job` after we migrate to `helm3`

PR #8581, opened 2020-05-25
Add sleep test step after Knative trigger creation

radufa:
**Description**

- Add a "sleep" test step after the creation of a Knative trigger in both Compass and Connectivity Adapter tests

**Related issue(s)**

#8283 and more specific:
```
 No guarantees are made regarding WHEN a newly starting trigger / subscription / dispatcher will START receiving events. In practice it should be a very short amount of time, on the order of seconds, but that will depend on cluster load and K8S scheduling and myriad other factors. 
```


tests/end-to-end/external-solution-integration/internal/scenario/compass/compass.go
```diff
@@ -67,6 +67,7 @@ func (s *Scenario) Steps(config *rest.Config) ([]step.Step, error) {
 			knativeEventingClientset.EventingV1alpha1().Brokers(s.testID), knativeEventingClientset.MessagingV1alpha1().Subscriptions(helpers.KymaIntegrationNamespace)),
 		testsuite.NewCreateKnativeTrigger(s.testID, helpers.DefaultBrokerName, functionEndpoint, knativeEventingClientset.EventingV1alpha1().Triggers(s.testID)),
 		testsuite.NewSendEventToMesh(s.testID, helpers.FunctionPayload, state),
+		testsuite.NewSleep(s.waitTime),
```
nachtmaar, line 70: Shouln't both sleeps be at the same position? So either both before or both after `NewSendEventToMesh`?

radufa: Done

tests/end-to-end/external-solution-integration/internal/scenario/connectivity_adapter/scenario.go
```diff
@@ -12,13 +14,15 @@ type Scenario struct {
 	domain        string
 	testID        string
 	skipSSLVerify bool
+	waitTime      time.Duration
```
Szymongib, line 17: I feel like `waitTime` is to general term and it is only used before sending the event. I would suggest renaming it to something more descriptive like `eventSendDelay`.

tests/end-to-end/external-solution-integration/internal/scenario/connectivity_adapter/scenario.go
```diff
@@ -12,13 +14,15 @@ type Scenario struct {
...
 }
 
 // AddFlags adds CLI flags to given FlagSet
 func (s *Scenario) AddFlags(set *pflag.FlagSet) {
 	pflag.StringVar(&s.domain, "domain", "kyma.local", "domain")
 	pflag.StringVar(&s.testID, "testID", "connectivity-adapter-e2e", "domain")
 	pflag.BoolVar(&s.skipSSLVerify, "skipSSLVerify", false, "Skip verification of service SSL certificates")
+	pflag.DurationVar(&s.waitTime, "waitTime", time.Duration(10)*time.Second, "Wait time in seconds, e.g. 5s")
```
Szymongib, line 25: Same for the flags I feel like `waitTime` is to general term and it is only used before sending the event. I would suggest renaming it to something more descriptive like `eventSendDelay`.

polskikiel: I agree, also the flag description can be changed: `Wait time in seconds, e.g. 5s`.  You can provide any duration there `1s, 1m, 1h`.

tests/end-to-end/external-solution-integration/internal/scenario/connectivity_adapter/scenario.go
radufa: Yes, it's more general, but it's used to make the general step:
```
testsuite.NewSleep(s.waitTime)
```
configurable. The name is the same as in the other e2e tests.
We have done different tuning attempts to use the ```Sleep test step``` after other KN related actions too, but remained now at this simple solution. It could be not the last one... The main idea is to have similar code related to KN in all our three e2e test scenarios. The main issue is here that no API exists in KN/Eventing (like a "ready status")  which reflect the readiness of the eventing related infrastructure.

tests/end-to-end/external-solution-integration/internal/scenario/connectivity_adapter/scenario.go
radufa: Cause it's a ```time.Duration``` you can configure it to have any "go" duration, of course. I wanted to constrain it to seconds, otherwise it could be a sign for a true error in Event Mesh.

tests/end-to-end/external-solution-integration/internal/scenario/connectivity_adapter/scenario.go
radufa: Done

tests/end-to-end/external-solution-integration/internal/scenario/connectivity_adapter/scenario.go
radufa: Make comment more explicit.

Issue #8576, opened 2020-05-25
Kyma-operator refactoring after implementing advanced retries


strekm:
**Description**

Implement the following:

[r] means in review :)

- [x] refactor on error handling for install / upgrade step (maybe it could be extracted to the OnError method on step)
- [ ] ~~review kyma helm client to check when retry operation could be implemented~~
- [x] write tests for stepsfactory
- [x] write tests for installation
- [ ] improve logging on installation / upgrade error
- [x] implement wait method which actively checks if helm release was deleted / rollbacked 

**Reasons**

**Attachments**

Tomasz-Smelcerz-SAP:
I am skipping 'review kyma helm client to check when retry operation could be implemented' to avoid conficts with ongoing migration to Helm3.

Tomasz-Smelcerz-SAP:
This task is mostly done. Improvement in logging will be continued in #8517

PR #8612, opened 2020-05-27
Installer refactoring

Tomasz-Smelcerz-SAP:
**Description**
Refactoring of kyma-operator.  Cleanup of the code. Making it less entagled, more maintainable, to prepare it for upcoming new features.
One change in logic: Steps are not created at the time components are installed in the long processing loop. Now the list of steps is created before the installation of the first component starts. The logic now iterates over this list of steps instead of the component list from CR.

- Reduced dependencies
- Improved separation
- Huge files split into smaller units
- Mainly mechanical refactoring.

**Related issue(s)**
See also #8576


components/kyma-operator/pkg/kymaoperation/steps/factory.go
```diff
@@ -0,0 +1,119 @@
...
+	errors "github.com/pkg/errors"
+
+	"github.com/kyma-project/kyma/components/kyma-operator/pkg/apis/installer/v1alpha1"
+	"github.com/kyma-project/kyma/components/kyma-operator/pkg/config"
+	"github.com/kyma-project/kyma/components/kyma-operator/pkg/kymahelm"
+	"github.com/kyma-project/kyma/components/kyma-operator/pkg/overrides"
+)
+
+const (
+	defaultDeleteWaitTimeSec   = 10
```
Demonsthere, line 13: I think those values should be configurable with envs/params

components/kyma-operator/pkg/kymaoperation/steps/creator.go
```diff
@@ -0,0 +1,107 @@
...
+
+// ForInstallation returns implementation of StepFactory interface used to install or upgrade Kyma
+func (sfc *stepFactoryCreator) ForInstallation(installationData *config.InstallationData, overrideData overrides.OverrideData) (StepLister, error) {
+
+	installedReleases, err := sfc.getInstalledReleases()
+	if err != nil {
+		return nil, err
+	}
+
+	sourceGetter := sfc.sourceGetterSupport.SourceGetterFor(installationData.URL, installationData.KymaVersion)
```
Demonsthere, line 72: Unnecessary variable, just use `sfc.sourceGetterSupport.SourceGetterFor(installationData.URL, installationData.KymaVersion)` directly below

components/kyma-operator/pkg/kymaoperation/steps/upgrade.go
```diff
@@ -0,0 +1,59 @@
...
+		return releaseOverridesErr
+	}
+
+	upgradeResp, upgradeErr := s.helmClient.UpgradeRelease(
+		chartDir,
+		s.component.GetReleaseName(),
+		releaseOverrides)
+
+	if upgradeErr != nil {
+		upgradeErrMsg := fmt.Sprintf("Helm upgrade error: %s", upgradeErr.Error())
```
Demonsthere, line 36: Do we really need 3 separate errormsg vars? I think it would be cleaners to use one, and update the value

components/kyma-operator/pkg/kymaoperation/steps/install.go
```diff
@@ -0,0 +1,74 @@
...
+
+	installResp, installErr := s.helmClient.InstallRelease(
+		chartDir,
+		s.component.Namespace,
+		s.component.GetReleaseName(),
+		releaseOverrides)
+
+	if installErr != nil {
+		installErrMsg := fmt.Sprintf("Helm install error: %s", installErr.Error())
+		errorMsg := installErrMsg
```
Demonsthere, line 42: Again, I don't think we need this duplication

components/kyma-operator/pkg/kymaoperation/steps/creator.go
```diff
@@ -0,0 +1,107 @@
...
+		return nil, errors.New("Helm error: " + err.Error())
+	}
+
+	if list != nil {
+		log.Println("Helm releases list:")
+
+		for _, release := range list.Releases {
+			var lastDeployedRev int32
+
+			statusCode := release.Info.Status.Code
```
Demonsthere, line 43: If we don't do anything with `release.Info.Status.Code`, maybe just use it directly?

components/kyma-operator/pkg/kymaoperation/steps/factory.go
Tomasz-Smelcerz-SAP: This PR focused on more high-level refactoring, I didn't want to add anything new in terms of features/flags/params/etc. Besides that, the sleep that is controlled by this value will be removed entirely in the subsequent PR, where a proper wait mechanism, based on release status, is implemented. I'd just leave it as it is for now.

components/kyma-operator/pkg/kymaoperation/steps/creator.go
Tomasz-Smelcerz-SAP: Sorry, after doing that I don't think it's more readable. I prefer unnecessary variable because it provides a good name for the object. It's not directly obvious what kind of object is returned from this: `sfc.sourceGetterSupport.SourceGetterFor(installationData.URL, installationData.KymaVersion)`

components/kyma-operator/pkg/kymaoperation/steps/upgrade.go
Tomasz-Smelcerz-SAP: I did not modify this code, just moved it. Nevertheless, I am not a big fan of reusing vars. It makes telling what a variable contains in a given point in the code almost impossible without careful inspection of every tiny bit of it.
I agree that it can be useful in some scenarios, though.
I will refactor this to use two vars - we need two anyway.

components/kyma-operator/pkg/kymaoperation/steps/install.go
Tomasz-Smelcerz-SAP: This was existing code that I did not care to refactor - these are problems local to this method and I was addressing more important (IMO) issues at package/module level. Having said that, I'll see what can be done here...

components/kyma-operator/pkg/kymaoperation/steps/creator.go
Tomasz-Smelcerz-SAP: Done.

Issue #8410, opened 2020-05-08
[Monitoring] iam-kubeconfig-service


strekm:
**Description**

Implement http health check endpoint for monitoring

**Reasons**

iam-kubeconfig-service is one of essentials Kyma services and needs to be monitored.

**Attachments**

PR #8615, opened 2020-05-27
Add health endpoint for iam-kubeconfig-service

colunira:
**Description**

- add health endpoint to iam-kubeconfig-service
- use istio probe rewrite

**Related issue(s)**
https://github.com/kyma-project/kyma/issues/8410


resources/iam-kubeconfig-service/templates/deployment.yaml
```diff
@@ -43,6 +43,13 @@ spec:
         ports:
           - name: http
             containerPort: 8000
+        readinessProbe:
```
piotrmsc, line 46: You need to provide PR image, your code changes will not work without it, k8s will try to do the probe but the image used in the deployment has older version, without this endpoint.

colunira: Yep, that's true, forgot about it, thanks!

PR #8620, opened 2020-05-27
Bump to gqlgen 0.11.3

sjanota:
**Description**

- Bump to new gqlgen version

**Related issue(s)**


sjanota:
My fork needs to stay in `go.mod` for now, because on master CBS is not a module yet. Because of that dependency to master ends with ambiguous imports. As soon as this PR is merged we need to prepare one more to use Kyma from the master.

components/console-backend-service/Dockerfile
```diff
@@ -1,43 +1,39 @@
-FROM golang:1.11.4-alpine3.8 as builder
+FROM eu.gcr.io/kyma-project/test-infra/buildpack-golang-toolbox:v20190930-d28d219 as builder
```
aerfio, line 1: Curiosity: why not `golang-alpine`?

components/console-backend-service/Dockerfile
```diff
@@ -1,43 +1,39 @@
...
+
+# Copy built app
+RUN mkdir /app
+RUN mv ./console-backend-service /app/main 
 
-RUN CGO_ENABLED=0 GOOS=linux go build -v -o main .
-RUN mkdir /app && mv ./main /app/main && mv ${BASE_APP_DIR}/licenses /app/licenses
+COPY ./licenses/ /app/licenses
 
 FROM alpine:3.10
```
aerfio, line 28: You could bump it to 3.11 to solve some security issues (treat it just as a comment, not "request changes")

components/console-backend-service/Dockerfile
aerfio: this image is quite old btw, you use go: 1.13 in go mod, here you have 1.12, and newest go version is 1.14
image

image
 is 1.14

sjanota: We use buildpack because it contains some tooling like goimports already vendored.

sjanota: And I have to revert to go 1.12 in go mod, that's true :)

Issue #8650, opened 2020-05-29
Add scaling to Function based on native k8s resources


aerfio:
**Description**

**Reasons**

Listed in https://github.com/kyma-project/kyma/issues/8356. Make sure to fine-tune HPAs spec.

**Attachments**

See https://github.com/kyma-project/kyma/issues/8356

PR #8651, opened 2020-05-29
Add HorizontalPodAutoScaler as function's subresource

aerfio:
**Description**

- add HPA as subresource to Function
- update console image from https://github.com/kyma-project/console/pull/1822

Based on https://github.com/kyma-project/kyma/pull/8633

**Related issue(s)**

See https://github.com/kyma-project/kyma/issues/8650


resources/serverless/templates/cluster-role.yaml
```diff
@@ -5,13 +5,61 @@ metadata:
   labels:
     {{- include "tplValue" ( dict "value" .Values.global.commonLabels "context" . ) | nindent 4 }}
 rules:
+- apiGroups:
```
magicmatatjahu, line 8: Duplication of configmap.

components/function-controller/internal/controllers/serverless/function_reconcile.go
```diff
@@ -75,33 +75,36 @@ func (r *FunctionReconciler) Reconcile(request ctrl.Request) (ctrl.Result, error
...
 		return ctrl.Result{}, err
 	}
 
+	var hpas autoscalingv1.HorizontalPodAutoscalerList
+	if err := r.client.ListByLabel(ctx, instance.GetNamespace(), r.functionLabels(instance), &hpas); err != nil {
+		log.Error(err, "Cannot list HorizotalPodAutoscalers")
+		return ctrl.Result{}, err
+	}
+
 	// log.Info("Listing HorizotalPodAutoscalers")
```
magicmatatjahu, line 108: Remove it

components/function-controller/internal/controllers/serverless/deployment.go
```diff
@@ -90,18 +82,21 @@ func (r *FunctionReconciler) equalDeployments(existing appsv1.Deployment, expect
...
+	// it's checked here in onDeploymentChange to prevent nasty data races, where somehow deployment becomes ready before we
+	// trigger next reconcile loop, in which we should create svc
+	case r.isDeploymentInCondition(deployment, appsv1.DeploymentAvailable):
 		log.Info(fmt.Sprintf("Deployment %s is ready", deployment.GetName()))
 		return r.updateStatus(ctx, ctrl.Result{}, instance, serverlessv1alpha1.Condition{
 			Type:               serverlessv1alpha1.ConditionRunning,
-			Status:             corev1.ConditionUnknown,
+			Status:             runningStatus,
 			LastTransitionTime: metav1.Now(),
 			Reason:             serverlessv1alpha1.ConditionReasonDeploymentReady,
```
magicmatatjahu, line 96: This is only a question/sugestion. In this case we should point to last reason - for example, if hpa will be the last reconciled item, then we should point to last reason from Function (probably pointed to HPA), not change to `ConditionReasonDeploymentReady`. Of course, if deployment has change, then we must force update status to appropriate from Deployment, but when we change status from main reconcile function (default case from last `switch`), then we should retrieve reason from Function instance.

resources/serverless/templates/cluster-role.yaml
aerfio: It's because I've fixed rbacs generated by `controllers/kubernetes`, and now they also generate an entry here, but K8s should be able to "squash" all of those, so that there are no duplicates. From it's perspective it's just one simple operation, and from ours it is beneficial in a sense that each controller ensures in has proper rbacs set - imagine a situation in which you remove one of the controllers we now have, removing also marker which generates the rbacs -> in turn generated rbacs could break.

One idea I have is to add `rbac.go` file with `// +build rbac` comment (so it's not complited), and put ALL rbac markers in there, preventing duplication through the codebase

or even somewhere in `main.go`

components/function-controller/internal/controllers/serverless/deployment.go
aerfio: But in this reconcile deployment IS the last resource reconciled, as it goes like this:
- ...
- create deployment
- create svc
- create hpa 
- wait for deployment to be ready, if it is set proper status

If we wanted hpa to be last reconciled item, then we'd need to add more states, which would make our external API look really confusing (then we'd have to ditch `Running` state and add something like `HPAcreated` and others, which is confusing)

components/function-controller/internal/controllers/serverless/horizontalpodautoscaler.go
```diff
@@ -0,0 +1,108 @@
...
+	case len(hpas) == 0:
+		return r.createHorizontalPodAutoscaler(ctx, log, instance, newHpa)
+	case len(hpas) > 1:
+		log.Info("Too many HorizontalPodAutoscalers for function")
+		return r.updateStatus(ctx, ctrl.Result{RequeueAfter: r.config.RequeueDuration}, instance, serverlessv1alpha1.Condition{
+			Type:               serverlessv1alpha1.ConditionRunning,
+			Status:             corev1.ConditionFalse,
+			LastTransitionTime: metav1.Now(),
+			Reason:             serverlessv1alpha1.ConditionReasonHorizontalPodAutoscalerError,
+			Message:            fmt.Sprintf("HorizontalPodAutoscaler %s failed", instance.GetName()),
```
michal-hudy, line 40: Please add also the message on why it failed. Otherwise, you will need to look into the logs for a reason...

components/function-controller/internal/controllers/serverless/service.go
```diff
@@ -33,19 +31,13 @@ func (r *FunctionReconciler) onServiceChange(ctx context.Context, log logr.Logge
 			Status:             corev1.ConditionFalse,
 			LastTransitionTime: metav1.Now(),
 			Reason:             serverlessv1alpha1.ConditionReasonServiceError,
-			Message:            fmt.Sprintf("Service %s is ready", instance.GetName()),
+			Message:            fmt.Sprintf("Service %s failed", instance.GetName()),
```
michal-hudy, line 34: Same as for HPA, please update in other places if needed

Issue #8562, opened 2020-05-24
Uninstallation of service management components


PK85:
**Description**
Uninstallation of components (helm install / helm delete / helm install should work)

**AC**
- Uninstallation and then installation of service management components works

PR #8657, opened 2020-06-01
Make management components uninstallationable

jasiu001:
**Description**

Proposed changes allow to install and uninstall management components repeatedly.
The management components consists of 4 elements:
1. ServiceCatalog
  This component did not require any modifications, installation and uninstallation work correctly.
2. ServiceBindingUsage
  The component is one of the sub chart of `service-catalog-addons` chart. The component required the following modifications:
  - add default value for key `global.ingress.domainName` in `values.yaml`
  - requirements correction (removes old dependencies, adding new ones)
  - adding pre-delete job which removes CR `usagekinds` "deployment" before removing SBU controller.
    Remove controller before CR, resulted in the finalizer `servicecatalog.kyma-project.io/usage-kind-protection` not being removed from the CR and CR was not removed at all.
3. HelmBroker
  The component required the following modifications:
  - add default value for keys `global.ingress.domainName`, `global.helm.tls.crt` and `global.helm.tls.key` in `values.yaml`
  - adding pre-delete job which removes CR `clusteraddonsconfigurations` "helm-repos-urls" before removing addons controller.
    Remove controller before CR, resulted in the finalizer `addons.kyma-project.io` not being removed from the CR and CR was not removed at all.
4. ApplicationBroker
  The component is one of the sub chart of `application-connector` chart. The component required the following modifications:
  - add default value for keys `global.ingress.domainName`, `global.helm.tls.crt` and `global.helm.tls.key` in `values.yaml`

  after reinstalling `application-connector` chart, pod `connection-token-handler` restarts with error:
  ```
  > if kind is a CRD, it should be installed before calling Start: no matches for kind "TokenRequest" in version "applicationconnector.kyma-project.io/v1alpha1"
  > unable to register the tokenrequest controller to the manager: no matches for kind "TokenRequest" in version "applicationconnector.kyma-project.io/v1alpha1"
  ```
  the chart itself is installed correctly with status `DEPLOYED`

**Related issue(s)**
#8562

### DIFF
--- a/resources/application-connector/values.yaml
+++ b/resources/application-connector/values.yaml
@@ -9,6 +9,12 @@ global:
   isLocalEnv: false
   namespace: kyma-integration
   strictMode: disabled
+  ingress:
+    domainName: "TBD"
+  helm:
+    tls:
+      crt: "TBD"
+      key: "TBD"
   istio:
     namespace: istio-system
     tls:
--- a/resources/helm-broker/templates/clear-cr-job.yaml
+++ b/resources/helm-broker/templates/clear-cr-job.yaml
@@ -1,0 +1,51 @@
+apiVersion: batch/v1
+kind: Job
+metadata:
+  name: {{ template "fullname" . }}-cleanup
+  labels:
+    app: {{ template "fullname" . }}
+    chart: {{ .Chart.Name }}-{{ .Chart.Version | replace "+" "_" }}
+    release: {{ .Release.Name }}
+    heritage: {{ .Release.Service }}
+  annotations:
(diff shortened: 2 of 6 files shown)


jasiu001:
Error in point 4 on the Pod `connection-token-handler` was caused by installing the chart from the master version on Kyma version 1.12.
In Kyma 1.12 CRD TokenRequest is included in the connection-token-handler sub chart. After delete helm release, CRD was removed, chart from master does not include TokenRequest CRD (now this CRD is part of cluster-essentials chart)

PR #8665, opened 2020-06-01
Optimize CBS build

Wawrzyn321:
**Description**

- Use GO111MODULE
- Cleanup Dockerfile & Makefile
- Format files

**Related issue(s)**


components/console-backend-service/Dockerfile
```diff
@@ -1,43 +1,52 @@
-FROM golang:1.11.4-alpine3.8 as builder
+FROM golang:1.13.11-alpine3.11 as builder
```
sjanota, line 1: Use golang buildpack instead. It is supposed to hold all the versions we rely upon

components/console-backend-service/Dockerfile
```diff
@@ -1,43 +1,52 @@
...
+COPY ./go.sum .
+
+COPY ./internal ./internal
+COPY ./pkg ./pkg
+COPY ./main.go .
+COPY ./Makefile .
+
+# cache deps before building and copying source so that we don't need to re-download as much
+# and so that source changes don't invalidate our downloaded layer
+RUN go mod download
```
sjanota, line 21: `go mod download` must be run before any sources are copied. Otherwise caching won't work

components/console-backend-service/Dockerfile
```diff
@@ -1,43 +1,52 @@
...
+RUN go mod vendor
```
sjanota, line 22: It's not necessary if go modules are on. It's necessary if you want to support environment without modules, like IDE.

components/console-backend-service/Dockerfile
```diff
@@ -1,43 +1,52 @@
...
 
-COPY ./vendor/ ${BASE_APP_DIR}/vendor/
-COPY ./internal/ ${BASE_APP_DIR}/internal/
-COPY ./pkg/ ${BASE_APP_DIR}/pkg/
-COPY ./main.go ${BASE_APP_DIR}/
-COPY ./licenses/ ${BASE_APP_DIR}/licenses/
+# check imports using makefile with custom ignore patterns
+RUN apk add make
+RUN go get golang.org/x/tools/cmd/goimports
```
sjanota, line 26: All dependencies are included in buildpack

components/console-backend-service/Dockerfile
```diff
@@ -1,43 +1,52 @@
...
+RUN make check-imports
```
sjanota, line 27: Let's use `make verify-local` instead

components/console-backend-service/Dockerfile
```diff
@@ -1,43 +1,52 @@
...
-#
-# Copy files
-#
+COPY ./go.mod .
+COPY ./go.sum .
+
+COPY ./internal ./internal
+COPY ./pkg ./pkg
+COPY ./main.go .
+COPY ./Makefile .
```
sjanota, line 17: Would it be possible to just copy everything?

components/console-backend-service/Dockerfile
```diff
@@ -1,43 +1,52 @@
...
-COPY ./licenses/ ${BASE_APP_DIR}/licenses/
+# check imports using makefile with custom ignore patterns
+RUN apk add make
+RUN go get golang.org/x/tools/cmd/goimports
+RUN make check-imports
 
-#
 # Build app
-#
+RUN go build -v -o main .
```
sjanota, line 30: Use `make build`

components/console-backend-service/Dockerfile
```diff
@@ -1,43 +1,52 @@
...
-RUN CGO_ENABLED=0 GOOS=linux go build -v -o main .
-RUN mkdir /app && mv ./main /app/main && mv ${BASE_APP_DIR}/licenses /app/licenses
+# Run tests
+RUN go test ./...
+
+# Copy built app
+RUN mkdir /app
+RUN mv ./main /app/main 
+
+COPY ./licenses/ /app/licenses
```
sjanota, line 39: I'd rather generate the licences here. there's target for it.

components/console-backend-service/Dockerfile
parostatkiem: It would be but would also make creating a `.dockerignore` file necessary. We suggest (and commited) another soultion - 4 `COPY` lines instad of 6 😃

akucharska:
Duplicate

PR #8669, opened 2020-06-02
Refine ClusterRoles for Kyma-Eventing

anishj0shi:
Certain Resources for kyma eventing were not aggregated into the eventing clsuter roles, as a consequence it was impossible to debug state of eventing unless a user was granted a cluster-admin role.

with this change, eventing resources are added into relevant kyma-roles.

### DIFF
--- a/resources/cluster-users/values.yaml
+++ b/resources/cluster-users/values.yaml
@@ -57,6 +57,11 @@ clusterRoles:
       - "caching.internal.knative.dev"
       - "autoscaling.internal.knative.dev"
       - "serving.knative.dev"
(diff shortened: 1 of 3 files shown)


resources/cluster-users/values.yaml
```diff
@@ -57,6 +57,11 @@ clusterRoles:
       - "caching.internal.knative.dev"
       - "autoscaling.internal.knative.dev"
```
piotrmsc, line 58: Please update tests for that https://github.com/kyma-project/kyma/tree/master/resources/cluster-users/files

Issue #8563, opened 2020-05-24
Service Management components works with helm3


PK85:
**Description**
As Kyma 1.14 will work with helm3 we need to make sure Service Management components works as expected

**AC**
- check all potential issues mentioned in the issue: https://github.com/kyma-project/kyma/issues/7933
- check helm install, upgrade and uninstall on all Service Management components

PR #8678, opened 2020-06-02
Adapt management components to Helm 3

jasiu001:
**Description**

- Adapt management components to work with Helm 3
- From all management components, ServiceBindingUsage (inside `service-catalog-addons`) needs fix

**Related issue(s)**
#8563

### DIFF
--- a/resources/service-catalog-addons/charts/service-binding-usage-controller/templates/deployment.yaml
+++ b/resources/service-catalog-addons/charts/service-binding-usage-controller/templates/deployment.yaml
@@ -8,10 +8,6 @@ metadata:
     release: {{ .Release.Name }}
     heritage: {{ .Release.Service }}
 spec:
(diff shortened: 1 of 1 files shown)


resources/service-catalog-addons/charts/service-binding-usage-controller/templates/deployment.yaml
```diff
@@ -29,6 +25,10 @@ spec:
         release: {{ .Release.Name }}
     spec:
       serviceAccountName: {{ template "fullname" . }}
+      {{- with .Values.securityContext }}
+      securityContext:
+        {{ toYaml . | indent 8 }}
```
polskikiel, line 30: is indent needed here?

polskikiel: I think not :)

PR #8681, opened 2020-06-02
Enable metrics of the serverless webhook

pPrecel:
**Description**

- Enable metrics of the serverless webhook with the default configuration


resources/serverless/charts/webhook/templates/deployment.yaml
```diff
@@ -48,6 +48,10 @@ spec:
               valueFrom:
                 fieldRef:
                   fieldPath: metadata.namespace
+            - name: METRICS_DOMAIN
```
michal-hudy, line 51: Add a comment that it is needed because webhook is based on knative

pPrecel: Good point. Done

PR #8679, opened 2020-06-02
[POC] Installer helm3

Demonsthere:
**Description**

Changes proposed in this pull request:

- implements helm3 changes

**Related issue(s)**
- see #8517
- Requires https://github.com/kyma-project/kyma/pull/8677 
- Requires https://github.com/kyma-project/kyma/pull/8678

Issue #8524, opened 2020-05-20
Conform knative-eventing charts with Helm3


anishj0shi:
**Dependency**
- kyma-operator needs to be updated with helm3 client.

**AC**
Improve kyma components charts by:
- [ ] make charts compatible with helm 3
- [ ] fix errors in schemas
- [ ] remove deprecated APIs
- [ ] do not install resources in multiple namespaces within one chart
- [ ] move custom resource definitions to the first step (cluster-essentials)
- [ ] each kyma chart can be installed by `helm template | kubectl apply -f -`

Issue #8517, opened 2020-05-19
Migrate kyma-operator to helm 3


pbochynski:
**Description**
Refactor kyma-operator to use helm 3.

**Reasons**
A lot of security vulnerabilities are assigned to Tiller (helm 2 server-side component). The best way to deal with them is just to upgrade to "tillerless" helm 3.

**Related issues**
- https://github.com/kyma-project/kyma/issues/7933
- https://github.com/kyma-project/helm-broker/issues/66
- https://github.com/kyma-project/kyma/issues/6552

piotrmsc:
Additionally we will need to :
- Update our prow jobs (remove tiller part)
- Update documentation (remove tiller part)

piotrmsc:
@adamwalach I think prow part could be done on your side?

PR #8689, opened 2020-06-03
Migrate Knative-Eventing to Helm3

radufa:
**Description**

- Migrate Knative-Eventing to Helm3
- Add "values.schema.json"

**Related issue(s)**

See #8524
See also #8517
Requires #8679


resources/knative-eventing/Chart.yaml
```diff
@@ -1,4 +1,4 @@
-apiVersion: v1
+apiVersion: v2
 name: knative-eventing
 version: 0.12.0
 tillerVersion: ">=2.7.2-0"
```
nachtmaar, line 4: I guess tillerVersion is not required anymore, since there is no tiller in helm3

radufa: Done

resources/knative-eventing/charts/knative-eventing/Chart.yaml
```diff
@@ -1,7 +1,6 @@
-apiVersion: v1
+apiVersion: v2
 name: knative-eventing
 version: 0.12.0
```
sayanh, line 3: version -> this is the Chart version
appVersion -> this is the version of the application
In this case, `0.12.0` should be the appVersion, IMO

resources/knative-eventing/Chart.yaml
```diff
@@ -1,9 +1,11 @@
-apiVersion: v1
+apiVersion: v2
```
sayanh, line 1: `v1` would make the existing Kyma operator happy as well as helm3 client. Let's go with that for now. Once the complete migration is done, we will switch to v2.

resources/knative-eventing/Chart.yaml
```diff
@@ -1,9 +1,11 @@
...
 name: knative-eventing
 version: 0.12.0
-tillerVersion: ">=2.7.2-0"
 description: Helm chart for installing Knative Eventing
 keywords:
   - knative
   - knative-eventing
   - kyma
+dependencies:
```
sayanh, line 9: But then if we use `apiVersion: v1`, `dependencies` won't work hence we need to remove this.

radufa: For v1, we have to restore the "requirements.yaml" which defines the dependencies for v1. For V2, the dependencies should be moved into the chart, under the new "dependencies". This is an incompatible change in v2.

resources/knative-eventing/charts/knative-eventing/Chart.yaml
radufa: I didn't change the version, it seems that it is the Knative Eventing version used.

resources/knative-eventing/Chart.yaml
radufa: For this PR, it means that the "requirements.yaml" file which is needed only by v1 should be restored. This is an incompatible change in v2, see also my comment below.

PR #8691, opened 2020-06-03
Adjust Kyma Operator and Api Gateway tests to work with independent Compass installation

kjaksik:
**Description**

- Add field selector to Kyma Operator so that it won't trigger on the Compass Installation CR
- Make the APIRule gateway configurable via envs in Api Gateway Tests

The changes allow having two separate Operators on the cluster at once. Apart from that it fixes the api gateway tests so that they will work with other gateways (eg. compass-istio-gateway).


components/kyma-operator/pkg/env/env.go
```diff
@@ -0,0 +1,21 @@
+package env
+
+import "github.com/vrischmann/envconfig"
+
+var (
+	Config EnvConfig
```
pkosiec, line 6: I wonder if this config should be global. But of course it's up to codeowners.

kjaksik:
[Here](https://github.com/kyma-project/kyma/pull/8703) are the docs for that

PR #8710, opened 2020-06-04
Update serverless to forbid scale-to-zero

aerfio:
**Description**

- HPA doesn't allow to have minReplicas set to 0:
```bash
"error":"HorizontalPodAutoscaler.autoscaling \"upbeat-bleah-5vntg\" is invalid: spec.minReplicas: Invalid value: 0: must be greater than 0"
```
so we need to disable that option entirely
- update envs which are forbidden
- update core-ui img with changes from https://github.com/kyma-project/console/pull/1829

**Related issue(s)**

https://github.com/kyma-project/kyma/issues/8650


michal-hudy:
Please also update the UI :)

PR #8714, opened 2020-06-04
Fix Grafana & Kiali for Service Catalog

mszostok:
Description:
- Fix Grafana & Kiali for Service Catalog
  - Also query for showing the OSB API request was fixed, now it shows req/min

image

Fixed Kiali: 
- https://kiali.schemas.kymcia.online/console/namespaces/kyma-system/services/service-catalog-catalog-webhook?duration=60
- https://kiali.schemas.kymcia.online/console/namespaces/kyma-system/services/service-catalog-catalog-controller-manager?duration=60

Fixed Grafana:
- https://grafana.schemas.kymcia.online/d/t9EmddWWzn/kyma-service-catalog-fixed

### DIFF
--- a/resources/service-catalog/charts/catalog/templates/controller-manager-service-monitor.yaml
+++ b/resources/service-catalog/charts/catalog/templates/controller-manager-service-monitor.yaml
@@ -12,10 +12,14 @@ metadata:
   name: {{ template "fullname" . }}-controller-manager
 spec:
   endpoints:
-  - port: secure
+  - port: https-secure
     scheme: https
     tlsConfig:
       insecureSkipVerify: true
+    metricRelabelings:
+      - sourceLabels: [ __name__ ]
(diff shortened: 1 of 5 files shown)


polskikiel:
I have 3 things: 
image
When you display nulls as zeros it looks better imo

There is also some inconsistency between memory and CPU graphs, for current memory usage you use prefix `Current:` and for CPU you don't:
image

On the above ss you can see that we also have now also a field: `Current: container_name`. If you want to get rid of it, add `,container!=""` to query filters.

mszostok:
Hi @polskikiel 
Thanks for the review, I've adjusted the CPU and Memory charts and also added the `,container!=""` arg to Prometheus query.

About nulls as zeros IMO is better to do not apply too much manipulation on our charts, if you see that chart is empty for some time that you know that there can be some problem pods restarts, wrong servicemonitor modification, etc. Without that, you will always see zero even if something will be broken

mszostok:
Issue for the Helm Broker: https://github.com/kyma-project/helm-broker/issues/101

Issue #7512, opened 2020-03-09
Upgrade tests should cover 'external-end-to-end' scenario


anishj0shi:
in continuation to https://github.com/kyma-project/kyma/issues/7277 

After an upgrade process, there are still a lot unclear issues which we might not be aware of for example: Certificate expired, Resources created before upgrade are non responsive, etc
Having a test in place which tests an end to end upgrade process, will help us illuminate such issues at development phase.

- Before Upgrade we need to prepare sources like: Creating Application, Connecting Application and storing the certificate to perform calls to application gateway.
- After Completing the Upgrade process, Fire Events at Application Gateway Level.

**AC** 
- Prepare  resources before upgrade, for example pairing an Application, Creating Service Instances in a namespace followed by Creating Knative-Triggers pointing to a subscriber.
- After Upgrading Kyma, Fire an Event to the above Application's Eventing Endpoints (/v1/events and /events)
- Successful Validation of the Event Delivery from the subscriber.
- This Upgrade test should be part of all the upgrade tests in the prow jobs (eg: GKE Upgrade)

PR #8734, opened 2020-06-08
Fix the E2E upgrade test

marcobebway:
**Description**

- Fix the E2E upgrade test skipSSLVerify flag.
- Fix the subscriber service URL endpoints.

**Related issue(s)**

#7512

### DIFF
--- a/tests/end-to-end/external-solution-integration/chart/external-solution/values.yaml
+++ b/tests/end-to-end/external-solution-integration/chart/external-solution/values.yaml
@@ -3,7 +3,7 @@ containerRegistry:
 
 image:
   dir:
-  version: PR-8511
+  version: PR-8734
   pullPolicy: "Always"
 
 testServiceImage:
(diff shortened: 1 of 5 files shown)


tests/end-to-end/external-solution-integration/chart/external-solution/values.yaml
```diff
@@ -3,7 +3,7 @@ containerRegistry:
 
 image:
   dir:
-  version: PR-8511
+  version: PR-8734
```
tgorgol, line 6: As this image is used in more than one chart and some of the changed code is shared between different e2e scenarios please bump also the versions in:

1. file: `resources/compass-runtime-agent/values.yaml`, value: `global.compassExternalSolutionTests.version`
2. file: `resources/core/values.yaml`, value: `global.e2e_external_solution.version`

to verify if nothing broke.

marcobebway: Done. Nice catch 👍

PR #8735, opened 2020-06-08
Implement go modules and k8s 1.18 for SBU controller

polskikiel:
**Description**

- Implemented go modules and k8s 1.18 for SBU controller

**Related issue(s)**


components/service-binding-usage-controller/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM golang:1.12.10-alpine3.10 as builder
+FROM golang:1.13.10-alpine3.10 as builder
```
jasiu001, line 1: Is there any reason we use golang 1.13.10 version instead for example 1.14.2 like in `buildpack-golang-toolbox` version `v20200423-1d9d6590`?

components/service-binding-usage-controller/cmd/controller/main.go
```diff
@@ -180,20 +180,21 @@ func runHTTPServer(stop <-chan struct{}, addr string, sbuClient *bindingUsageCli
 func informerAvailability(sbuClient *bindingUsageClientset.Clientset, log logrus.FieldLogger, namespace string) error {
 	deleteSample := func() {
 		err := sbuClient.ServicecatalogV1alpha1().ServiceBindingUsages(namespace).Delete(
+			context.TODO(),
```
jasiu001, line 183: there is a context created already in `runHTTPServer` method, you can send it as a parameter and reuse, here and in other places

PR #8740, opened 2020-06-08
Expose nodeSelector for compass components

sisiliao:
**Description**

- Expose nodeSelector for compass components

### DIFF
--- a/resources/compass/charts/metris/templates/deployment.yaml
+++ b/resources/compass/charts/metris/templates/deployment.yaml
@@ -24,6 +24,8 @@ spec:
     spec:
 {{- include "metris.imagePullSecrets" . | indent 6 }}
       serviceAccountName: {{ template "metris.fullname" . }}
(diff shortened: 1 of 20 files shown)


resources/compass/charts/metris/templates/deployment.yaml
```diff
@@ -24,6 +24,8 @@ spec:
     spec:
 {{- include "metris.imagePullSecrets" . | indent 6 }}
       serviceAccountName: {{ template "metris.fullname" . }}
+      nodeSelector: 
+        {{- toYaml .Values.deployment.nodeSelector | nindent 8 }}
```
maksd, line 28: It should be more `{{- toYaml .Values.nodeSelector | nindent 8 }}`, there is no `deployment` in the value file.

sisiliao: yes, sorry.. pasted too quick....

PR #8794, opened 2020-06-15
Update Luigi in Console

parostatkiem:
**Description**

- ...
- ...
- ...

**Related issue(s)**


resources/console/values.yaml
```diff
@@ -10,7 +10,8 @@ console:
 
 global:
   isLocalEnv: false
-  knative: false
+  knative:
+    false
```
sjanota, line 14:
```suggestion
  knative: false
```

PR #8818, opened 2020-06-16
Fix security issues in Rafter and Serveless domains

aerfio:
**Description**

- edit tests and components images to use scratch, to increase security

**Related issue(s)**

https://github.com/kyma-project/kyma/issues/8823

### DIFF
--- a/resources/knative-serving/values.yaml
+++ b/resources/knative-serving/values.yaml
@@ -57,10 +57,10 @@ global:
     path: eu.gcr.io/kyma-project
   test_knative_serving_acceptance:
     dir: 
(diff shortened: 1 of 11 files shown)


resources/knative-serving/values.yaml
```diff
@@ -57,10 +57,10 @@ global:
...
     dir: 
-    version: "PR-8769"
+    version: "PR-8818"
   ingress:
     domainName:
 
 test:
   target: "Test Target"
-  enabled: false
+  enabled: true
```
aerfio, line 66:
```suggestion
  enabled: false
```

Issue #8730, opened 2020-06-08
Create a long running test for Serverless


michal-hudy:
**Description**

Implement a test that will create a Function that will be not deleted after execution. When that test is executed again it should simply check if Function is still running and working.

**Reasons**

Kyma stability and quality

PR #8820, opened 2020-06-16
Add long running tests for serverless

m00g3n:
**Description**
Add long running tests for serverless.

- Serverless long running test added

**Related issue(s)**

See also: https://github.com/kyma-project/kyma/issues/8730

### DIFF
--- a/resources/serverless/templates/tests/long-test.yaml
+++ b/resources/serverless/templates/tests/long-test.yaml
@@ -65,8 +65,9 @@ spec:
 
               # test start
 
-              curl -XGET http://{{ .Values.tests.long.name }}.{{ .Values.tests.long.namespace }} \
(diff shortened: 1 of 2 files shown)


resources/serverless/templates/tests/long-test.yaml
```diff
@@ -0,0 +1,85 @@
...
+            - "/bin/bash"
+          args:
+            - "-c"
+            - |
+              echo 'TESTING has started'
+              sleep 20
+
+              # test start
+
+              curl -XGET http://{{ .Values.tests.long.name }}.{{ .Values.tests.long.namespace }} \
```
michal-hudy, line 68: Please also check if Function CR has all conditions set to `true`, as status can be invalid.

magicmatatjahu: Please also check that response is `OK` :)

m00g3n: @magicmatatjahu The result code is enough, why you want to check the response body?

Issue #8750, opened 2020-06-09
Update to Grafana 7.0.3


a-thaler:
**Description**
There is a new patch release available for grafana which should be used.
Please update to it.

PR #8824, opened 2020-06-17
Upgrade Grafana

suleymanakbas91:
**Description**

- Upgrade Grafana

**Related issue(s)**

See #8750


resources/serverless/templates/function-dashboard.yaml
```diff
@@ -1084,7 +1084,9 @@ data:
       "refresh": "10s",
       "schemaVersion": 25,
       "style": "dark",
-      "tags": [],
```
rakesh-garimella, line 1087: do we need tags for other dashboards too ?

suleymanakbas91: Yes, we already have tags for all the dashboards. This dashboard had a syntax problem. That's why we were not able to see it on grafana and not realized that it had a missing tag.

Issue #8837, opened 2020-06-17
Function stuck on Deploying after labels update


aerfio:
**Description**

As in description -> I can see that we have a lot of errors on unsuccesful service update
Also, on label update we create new deployment and attemp to create new service...

![image](https://user-images.githubusercontent.com/17271979/84902631-592a5a00-b0ad-11ea-8dbe-279527f2ef83.png)

**Expected result**

There's only 1 of each resources, and update succeeds.

**Actual result**

It does not

**Steps to reproduce**

**Troubleshooting**

Issue #8833, opened 2020-06-17
Function Controller creates resources multiple times when communication with API server fails


michal-hudy:
**Description**

Sometimes, when communication with the API server fails, FC creates resources that have generated names multiple times. Sometimes we have multiple ConfigMaps, sometimes Deployments, etc. The problem is that FC is not able to solve such an issue, and it changes Function status to failed and manual action is required.

**Expected result**

FC doesn't create multiple resources or at least is able to handle such a situation. Manual action shouldn't be required when that resources are created by FC - if a user creates such resources manually then Function should have failed status.

**Actual result**

The function has failed status when FC creates multiple resources and manual action is required.

**Steps to reproduce**

Hard to say, unstable communication with API server, sample logs from tests:
https://storage.googleapis.com/kyma-prow-logs/logs/kyma-integration-gardener-azure/1270702363258130432/build-log.txt
https://storage.googleapis.com/kyma-prow-logs/logs/post-master-kyma-gke-upgrade/1273161933649874944/build-log.txt

**Troubleshooting**

Issue #8809, opened 2020-06-15
Autoscaling is not working when cluster is deployed on AKS


michal-hudy:
**Description**

When the Kyma is deployed on AKS with a production profile, the HPA is not able to scale functions, even when it is using 100% CPU. According to events, HPA is not able to read metrics:
```
  Type     Reason                        Age                From                       Message
  ----     ------                        ----               ----                       -------
  Warning  FailedGetResourceMetric       13m (x3 over 14m)  horizontal-pod-autoscaler  unable to get metrics for resource cpu: no metrics returned from resource metrics API
  Warning  FailedComputeMetricsReplicas  13m (x3 over 14m)  horizontal-pod-autoscaler  Invalid metrics (1 invalid out of 1), last error was: failed to get cpu utilization: unable to get metrics for resource cpu: no metrics returned from resource metrics API
  Warning  FailedGetResourceMetric       12m (x6 over 14m)  horizontal-pod-autoscaler  did not receive metrics for any ready pods
  Warning  FailedComputeMetricsReplicas  12m (x6 over 14m)  horizontal-pod-autoscaler  Invalid metrics (1 invalid out of 1), last error was: failed to get cpu utilization: did not receive metrics for any ready pods
```

**Expected result**

Function is scaled under high load.

**Actual result**

The function is not scaled, even under high load:
![image](https://user-images.githubusercontent.com/26445970/84702794-7e0cb900-af57-11ea-9313-b2613cb17178.png)

**Steps to reproduce**

Expose the function with `allow` rule and execute performance tests:
```bash
ab -n 1000 -c 100 https://{ADDRESS_TO_EXPOSED_FUNCTION}
```

michal-hudy:
It is looking like it doesn't work on other platforms too - for example, GKE.

michal-hudy:
Ok, I have found the root cause of the problem. Deployment has selector that matches also pod created for the build's job. Because of that autoscaling is not working. As a quick solution, we can add a label with resource type and use it in deployment and service selectors.

michal-hudy:
Another problem is that the HPA takes the sum of resources assigned to all containers in a pod, so Istio Sidecar is also taken into account. It is a problem when Kyma is deployed with a production profile - HPA thinks that CPU usage is really low because Istio Sidecar has a pretty big limit. So, we need to find a different autoscaler that can monitor a particular container in a pod.

evilyeti:
[...]

PR #8838, opened 2020-06-17
Fix serverless bugs with underlying resources

aerfio:
**Description**

- in certain situation, when we have stale cache, we create second resource, like deployment, which led to function being stuck and needing manual action -> this was refactored so that if fn-controller encounters such a situation it'll delete all underlying resources. In case of services it'll leave that one which has proper name, as someone (person, not fn-controller) must have created second service using our internal labels, which is forbidden.
- also fix bug which prevented us from changing labels, as controller couldn't handle such a case -> this needed special care for job resource, as we need to have separate step just to label it, and not recreate.
- write a ton of tests.
- fix wrong labels/selector combo, which led service to target job's pods

Things that should work now:
- you can freely change labels from UI/kubectl, and they should properly propagate
- doubled resources should be automatically deleted (e.g. do `k expose...`-> created service should be immediately deleted)
- scaling should work without a problem now, no matter what's the platform

**Related issue(s)**

https://github.com/kyma-project/kyma/issues/8837
https://github.com/kyma-project/kyma/issues/8833
https://github.com/kyma-project/kyma/issues/8809


components/function-controller/internal/controllers/serverless/configmap.go
```diff
@@ -59,15 +60,19 @@ func (r *FunctionReconciler) isOnConfigMapChange(instance *serverlessv1alpha1.Fu
 	image := r.buildExternalImageAddress(instance)
 	configurationStatus := r.getConditionStatus(instance.Status.Conditions, serverlessv1alpha1.ConditionConfigurationReady)
 
-	if len(deployments) == 1 && deployments[0].Spec.Template.Spec.Containers[0].Image == image && configurationStatus != corev1.ConditionUnknown {
+	if len(deployments) == 1 &&
```
magicmatatjahu, line 63: Should we check that only one configMap exists?

components/function-controller/internal/controllers/serverless/job.go
```diff
@@ -20,16 +20,21 @@ func (r *FunctionReconciler) isOnJobChange(instance *serverlessv1alpha1.Function
...
+		r.mapsEqual(expectedJob.GetLabels(), jobs[0].GetLabels()) {
 		return false
 	}
 
-	expectedJob := r.buildJob(instance, "")
-
 	return len(jobs) != 1 ||
 		len(jobs[0].Spec.Template.Spec.Containers) != 1 ||
 		// Compare image argument
 		!r.equalJobs(jobs[0], expectedJob) ||
```
magicmatatjahu, line 36: We can check `r.mapsEqual(expectedJob.GetLabels(), jobs[0].GetLabels())` inside `r.equalJobs`.

components/function-controller/internal/controllers/serverless/service.go
```diff
@@ -92,3 +85,29 @@ func (r *FunctionReconciler) updateService(ctx context.Context, log logr.Logger,
...
+
+	for i, _ := range services {
+		svc := services[i]
+		if svc.GetName() == instance.GetName() {
+			continue
+		}
+
+		log.Info(fmt.Sprintf("Deleting Service %s", svc.GetName()))
+
+		err := r.client.Delete(ctx, &services[i])
```
magicmatatjahu, line 104: `r.client.Delete(ctx, &svc)`

components/function-controller/internal/controllers/serverless/function_reconcile_test.go
```diff
@@ -300,6 +310,588 @@ var _ = ginkgo.Describe("Function", func() {
...
+			ginkgo.By("updating configmap")
+			_, err = reconciler.Reconcile(request)
+			gomega.Expect(err).To(gomega.BeNil())
+
+			configMapList := &corev1.ConfigMapList{}
+			err = reconciler.client.ListByLabel(context.TODO(), function.GetNamespace(), fnLabels, configMapList)
+			gomega.Expect(err).To(gomega.BeNil())
+			gomega.Expect(configMapList.Items).To(gomega.HaveLen(1))
+
+			cmLabelVal, ok := configMapList.Items[0].Labels[addedLabelKey]
```
magicmatatjahu, line 640: Check also length of labels :)

components/function-controller/internal/controllers/serverless/function_reconcile_test.go
```diff
@@ -300,6 +310,588 @@ var _ = ginkgo.Describe("Function", func() {
...
+			gomega.Expect(resourceClient.Get(context.TODO(), request.NamespacedName, function)).To(gomega.Succeed())
+			gomega.Expect(function.Status.Conditions).To(gomega.HaveLen(conditionLen))
+			gomega.Expect(reconciler.getConditionReason(function.Status.Conditions, serverlessv1alpha1.ConditionBuildReady)).To(gomega.Equal(serverlessv1alpha1.ConditionReasonJobUpdated))
+
+			jobList = &batchv1.JobList{}
+			err = resourceClient.ListByLabel(context.TODO(), function.GetNamespace(), fnLabels, jobList)
+			gomega.Expect(err).To(gomega.BeNil())
+			gomega.Expect(jobList.Items).To(gomega.HaveLen(1))
+
+			jobLabelVal, ok := jobList.Items[0].Labels[addedLabelKey]
```
magicmatatjahu, line 658: Check also length of labels :)

components/function-controller/internal/controllers/serverless/function_reconcile_test.go
```diff
@@ -300,6 +310,588 @@ var _ = ginkgo.Describe("Function", func() {
...
+			ginkgo.By("updating deployment")
+			_, err = reconciler.Reconcile(request)
+			gomega.Expect(err).To(gomega.BeNil())
+
+			deployList := &appsv1.DeploymentList{}
+			err = reconciler.client.ListByLabel(context.TODO(), function.GetNamespace(), fnLabels, deployList)
+			gomega.Expect(err).To(gomega.BeNil())
+			gomega.Expect(deployList.Items).To(gomega.HaveLen(1))
+
+			deployLabelVal, ok := deployList.Items[0].Labels[addedLabelKey]
```
magicmatatjahu, line 680: Check also length of labels :)

components/function-controller/internal/controllers/serverless/function_reconcile_test.go
```diff
@@ -300,6 +310,588 @@ var _ = ginkgo.Describe("Function", func() {
...
+			ginkgo.By("updating service")
+			_, err = reconciler.Reconcile(request)
+			gomega.Expect(err).To(gomega.BeNil())
+
+			svcList := &corev1.ServiceList{}
+			err = reconciler.client.ListByLabel(context.TODO(), function.GetNamespace(), fnLabels, svcList)
+			gomega.Expect(err).To(gomega.BeNil())
+			gomega.Expect(svcList.Items).To(gomega.HaveLen(1))
+
+			svcLabelVal, ok := svcList.Items[0].Labels[addedLabelKey]
```
magicmatatjahu, line 693: Check also length of labels :)

components/function-controller/internal/controllers/serverless/function_reconcile_test.go
```diff
@@ -300,6 +310,588 @@ var _ = ginkgo.Describe("Function", func() {
...
+			ginkgo.By("updating hpa")
+			_, err = reconciler.Reconcile(request)
+			gomega.Expect(err).To(gomega.BeNil())
+
+			hpaList := &autoscalingv1.HorizontalPodAutoscalerList{}
+			err = reconciler.client.ListByLabel(context.TODO(), function.GetNamespace(), fnLabels, hpaList)
+			gomega.Expect(err).To(gomega.BeNil())
+			gomega.Expect(hpaList.Items).To(gomega.HaveLen(1))
+
+			hpaLabelVal, ok := hpaList.Items[0].Labels[addedLabelKey]
```
magicmatatjahu, line 706: Check also length of labels :)

components/function-controller/internal/controllers/serverless/service.go
aerfio: &svc adress is always the same, as we do a for loop here and svc gets always the same address, it's safer to use &services[i] as it'll get different pointer every time

components/function-controller/internal/controllers/serverless/job.go
michal-hudy: Hm... but it will trigger build when you change the labels. Labels don't affect generated docker image, so what is the point of building it again?

magicmatatjahu: Ah.. yes in second reconcile :)

PR #8865, opened 2020-06-19
Use Helm3 in Helm Broker

polskikiel:
**Description**

- Use Helm3 in Helm Broker

**Related issue(s)**

### DIFF
--- a/resources/helm-broker/values.yaml
+++ b/resources/helm-broker/values.yaml
@@ -25,10 +25,10 @@ ctrl:
 broker:
   resources:
     limits:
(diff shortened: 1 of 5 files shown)


resources/helm-broker/values.yaml
```diff
@@ -25,10 +25,10 @@ ctrl:
 broker:
   resources:
     limits:
-      memory: 76Mi
+      memory: 120Mi
```
polskikiel, line 28: It's much but it makes it more stable for the release, we should investigate the memory usage on grafana.

Issue #8526, opened 2020-05-20
Support Uninstallation of Eventing Components


anishj0shi:
kyma-operator triggers the uninstallation by adding relevant labels. 

Helm delete should remove the installed charts. 

**AC**
- Purging of components should be supported and compliant with helm3 
```shell
helm delete <eventing_chart> #Should delete the CRs. present in the eventing charts
```
- The idea behind uninstallation of the eventing components is to give back the cluster in it's original shape. i.e All the resources present due to installation of eventing chart such as knative-eventing, knative-kafka, natss, etc should be cleaned up.
- if the component is uninstalled **helm ls** should not show it in the list and a consequent **helm install** should go smoothly.
- Knative-eventing namespace should be removed after uninstallation

nachtmaar:
This ToDo is left after PR https://github.com/kyma-project/kyma/pull/8826
- [ ] add uninstall for nats namespace and knative-eventing namespace, hopefully in kyma-operator

PR #8870, opened 2020-06-22
Fix eventing uninstall rbac

nachtmaar:
**Description**

- pre-delete hooks didn't have rbac get permission. The consequence was that the script thought the API server doesn't know the CRD, hence CRs have not been deleted by the job
- Use static label  `kyma-project.io/uninstall: eventing` => easier watching of pre-delete job e.g. via `stern -l  kyma-project.io/uninstall: eventing --all-namespaces
`

**Related issue(s)**

#8526

### DIFF
--- a/resources/event-sources/templates/uninstall.yaml
+++ b/resources/event-sources/templates/uninstall.yaml
@@ -5,7 +5,7 @@
 apiVersion: v1
 kind: ServiceAccount
 metadata:
-  name: {{ .Release.Name }}-rc-delete
+  name: {{ .Release.Name }}-cr-delete
   namespace: {{ .Release.Namespace }}
   labels:
     app: {{ .Release.Name }}-cr-delete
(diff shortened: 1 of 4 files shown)


resources/event-sources/templates/uninstall.yaml
```diff
@@ -72,14 +72,14 @@ spec:
     metadata:
       annotations:
         sidecar.istio.io/inject: "false"
-      name: {{ .Release.Name }}-rc-delete
+      name: {{ .Release.Name }}-cr-delete
       labels:
-        app: "event-sources-delete"
+        kyma-project.io/app: "eventing-uninstall"
```
k15r, line 77: how about: kyma-project.io/uninstall: eventing

one could then get all uninstall jobs by checking for the existence of that label or get the specific uninstall job by also checking for its value

PR #8871, opened 2020-06-22
Fix the upgrade of knative-eventing-kafka chart with Helm 3 operator 

sayanh:
**Description**

- Fix the upgrade of knative-eventing-kafka chart with Helm 3 operator

Reason:
With the [change](https://github.com/kyma-project/kyma/pull/8713/files#diff-d848376d40fe767413ce5973e2e2d333R2) in chart name, the `matchLabel` `app.kubernetes.io/name` changed as well. As this field is immutable, the upgrade process failed. Hence it is imperative to delete the deployment before the Helm upgrade.

### DIFF
--- a/resources/knative-eventing-kafka/pre-upgrade.sh
+++ b/resources/knative-eventing-kafka/pre-upgrade.sh
@@ -1,0 +1,5 @@
+#!/usr/bin/env bash
+
+set -eux -o pipefail
+
+kubectl delete deployment -n knative-eventing -lapp.kubernetes.io/instance=knative-eventing-kafka --ignore-not-found
--- a/resources/knative-eventing-kafka/templates/pre-upgrade.yaml
+++ b/resources/knative-eventing-kafka/templates/pre-upgrade.yaml
@@ -1,0 +1,88 @@
+# About:
+# This pre-upgrade job delete the deployment knative-kafka-channel-controller
+
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: {{ .Release.Name }}-pre-upgrade
+  namespace: {{ .Release.Namespace }}
+  labels:
+    app: {{ .Release.Name }}-pre-upgrade
+  annotations:
+    helm.sh/hook: "pre-upgrade"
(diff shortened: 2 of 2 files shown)


resources/knative-eventing-kafka/pre-upgrade.sh
```diff
@@ -0,0 +1,5 @@
+#!/usr/bin/env bash
+
+set -eux -o pipefail
+
+kubectl delete deployment -n knative-eventing -lapp.kubernetes.io/instance=knative-eventing-kafka --ignore-not-found
```
nachtmaar, line 5: Should we template the script as well? 🤔 

```suggestion
kubectl delete deployment -n {{ .Release.Namespace }} -lapp.kubernetes.io/instance={{ .Release.Name }} --ignore-not-found
```

sayanh: Scripts cannot be templated. The placeholders are rendered as in. For example:
```
.....
          command:
            - "/bin/bash"
          args:
            - "-c"
            - |
              #!/usr/bin/env bash

              set -eux -o pipefail

              kubectl delete deployment -n {{ .Release.Namespace }} -lapp.kubernetes.io/instance={{ .Release.Name }} --ignore-not-found
          terminationMessagePolicy: "FallbackToLogsOnError"
          resources:
            requests:
              cpu: 200m
              memory: 128Mi
            limits:
              cpu: 200m
              memory: 128Mi
```

k15r: if we would inline this script into the yaml, we could template it... so the question would be. does it make sense?

sayanh: We could but that would make this inconsistent with the other helm hook jobs. E.g. uninstall

nachtmaar: Sad, I hoped it would work without inlining. We idea of having a script which is included in the job was that you can run it locally. This would not be the case anymore with templating. Don't have a strong opinion here, continue as you want from my side :)

PR #8882, opened 2020-06-23
Fix API Rules list query and subscription in cbs - bump image

### LINKED ISSUE #7988: [CBS] APIRules query fails with under very specific conditions
Query
```
APIRules(namespace: "<any-namespace>", serviceName: "<any-service>") {
  ...
}
```
sometimes (when count of returned rules would be >= 3) returns
```
{
  "errors": [
    {
      "message": "internal system error",
      "path": [
        "APIRules"
      ]
    }
  ],
  "data": null
}
```
while the same query without `serviceName` parameter always works properly.

**Troubleshooting**

Results can be filtered on recipient side, hurting performance.

magicmatatjahu:
**Description**

- Fix API Rules list query and subscription in cbs - bump image

**Related issue(s)**
Fixed #7988

### DIFF
--- a/resources/console/values.yaml
+++ b/resources/console/values.yaml
@@ -22,6 +22,6 @@ global:
   ui_acceptance_tests:
     version: e0146b08
   console_backend_service:
-    version: 5af71530
+    version: PR-8878
   console_backend_service_test:
     version: c86ce2c0


magicmatatjahu:
Fixed by https://github.com/kyma-project/kyma/pull/8900

PR #8890, opened 2020-06-24
Add Namespace selector for eventing-webhook

anishj0shi:
**Description**
Current Webhook Configurations are disrupting Gardeners' Cluster Operations. Hence, Proposing some measures to avoid that.

**Changes Proposed**
- Add a namespace-selector for eventing-webhook to exclude kube-system namespace in gardener clusters
- Reduce webhook timeout to 30 seconds


resources/knative-eventing/charts/knative-eventing/templates/pdb.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: policy/v1beta1
+kind: PodDisruptionBudget
+metadata:
+  name: eventing-webhook-pdb
```
sayanh, line 4: Let's template the name with chart details and add the default labels.

anishj0shi: As Discussed, kicking out PDB as we don't know how mutating webhooks behave on upscaling so we don't want to increment replcas of the mutating webhook

Issue #8756, opened 2020-06-09
[PoC] Local installation with k3s


pbochynski:
**Description**
Provide a way to easily (quickly) start Kyma on the local machine. The idea is to use k3s cluster (with k3d as a docker wrapper for k3s). The goal is to:
- start Kyma cluster in about 5-10 minutes from scratch
- avoid /etc/hosts updates (also for new lambda)
- have an air gap solution (cache docker images) 
- easier overrides mechanism

**Reasons**
Starting Kyma on minikube takes a lot of time and resources. People do not like the experience. This PoC is part of https://github.com/kyma-project/kyma/issues/8664

pbochynski:
Initial code is in this PR: https://github.com/kyma-project/kyma/pull/8806

sayanh:
I achieved the following [PR](https://github.com/kyma-project/kyma/pull/8950/files) by extending this [PR](https://github.com/kyma-project/kyma/pull/8806):
- Installed Kyma with other components that are necessary for an [end to end test](https://github.com/kyma-project/kyma/pull/8950/files?short_path=7648797#diff-7648797ff93200bf9b45a801402f191e) to work. Following are the steps:
  - Creating an application
  - Registered an eventing API to the application
  - Bind the application to a namespace
  - Enabled eventing in the namespace
  - Create a function to consume events sent to the gateway of the eventing API

Following are the observations:
- Time taken to install varies between 8 - 12 minutes
- More the local docker resources, the better. I worked with 5CPUs and 6Gi RAM which worked well.
- The tests could be executed sequentially to guarantee the success rate.
- Most of the tests are unstable although the manual execution of the workflow works fine. I'm still debugging on the reasons.
- k3d can only work with one container as the workers fail to pull images from any other registry except dockerhub.com
- `k3d stop -n kyma` doesn't work. When `k3d start -n kyma` is done, some of the pods don't recover from Crashbackloop.

Further actions:
- Create a prow pipeline to differentiate k3d and minikube
- Testing suite is still failing.

pbochynski:
The issue with k3d stop/start probably is related to DNS resolution inside the cluster. The core-dns patch is not permanent (after restart default configuration is restored). I just applied patch again and restarted affected services (usually console backing service or apiserver-proxy).

sayanh:
[...]

PR #8940, opened 2020-07-01
Add duration to minikube installation

sayanh:
**Description**

- Add duration to the minikube installation as we need to monitor the time taken to install Kyma using Minikube. This will in turn will help in choosing the tool which takes the least possible time to install.

**Related issue(s)**

Related https://github.com/kyma-project/kyma/issues/8756


a-thaler:
I know it is not in scope as first we need to migrate it to use the CLI, still want to mention it here. What about adding the duration to the CLI install command, then it will be usable in all jobs using the CLI? :)

sayanh:
> I know it is not in scope as first we need to migrate it to use the CLI, still want to mention it here. What about adding the duration to the CLI install command, then it will be usable in all jobs using the CLI? :)

@a-thaler Sure, this sounds good. Let's work on that in [this issue](https://github.com/kyma-project/cli/issues/490).

PR #8939, opened 2020-07-01
Update issue templates

kasiakepka:
**Description**

Changes proposed in this pull request:

- Update issue templates – add labels, and add new issue type: `consulting`.

**Related issue(s)**

PR #8957, opened 2020-07-03
Remove labels from issue templates

pbochynski:
**Description**

Remove labels from issue templates. The labels are added during the triage process by kyma developers. If the label is created automatically the issue is treated as already reviewed.

**Related issue(s)**
See also https://github.com/kyma-project/kyma/pull/8939


pbochynski:
@kazydek @kasiakepka Please do not add labels to issue templates. Our triage process is broken now as we process only issues without labels.

Issue #6346, opened 2019-11-21
Application Connector Custom API Auth handler


jcawley5:
The use cases supported by the application connector rely on know specifications.  In some cases systems perform authentication in an ad-hoc fashion not following any spec.  In effort to support these I would like to propose that we include an option to perform the authentication via a custom microservice.  During the application registration this could be defined as 

## Example Registration Object
```
api: {
	targetUrl: "http://myapiendpoint",
	customAuthHandler: {
		targetUrl: "custom-auth-handler.kyma-integration.svc.cluster.local",
                deploymentYamlUrl: "deployment.yaml for the microservice"
		customAuthHandlerParams: { //custom object to be sent to the customAuthHandler
			anyParamName1: anyParamName1Value,
			anyParamName2: anyParamName2Value
		}
	}
}
```

The service is then expected to return a json which conforms to the app connector registration which would be merged with any pre-existing definition

## Example Microservice Response
```
"requestParameters": {
	"headers": {
		"custom-header": ["foo"]
	},
	"queryParameters": {
		"param": ["bar"]
	}
}
```

PR #8969, opened 2020-07-06
Annotate and enrich outgoing request with SAP Passport headers

Poussinette:
**Description**

DO NOT MERGE THIS PR. This pull request is meant to build the application gateway image that is needed to test a local kyma PoC instance with SAP Passport handling.

- ...
- ...
- ...

**Related issue(s)**

### DIFF
--- a/components/application-gateway/cmd/applicationgateway/options.go
+++ b/components/application-gateway/cmd/applicationgateway/options.go
@@ -6,15 +6,18 @@ import (
 )
 
 type options struct {
-	externalAPIPort int
-	proxyPort       int
-	application     string
-	namespace       string
-	requestTimeout  int
-	skipVerify      bool
-	proxyTimeout    int
-	requestLogging  bool
-	proxyCacheTTL   int
+	externalAPIPort         int
+	proxyPort               int
+	application             string
+	namespace               string
+	requestTimeout          int
+	skipVerify              bool
+	proxyTimeout            int
+	requestLogging          bool
+	proxyCacheTTL           int
+	annotatePassportHeaders bool
+	storageKeyName          string
+	redisURL                string
 
 	namespacedGateway bool
 }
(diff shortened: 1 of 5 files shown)


components/application-gateway/cmd/applicationgateway/options.go
```diff
@@ -29,28 +32,33 @@ func parseArgs() *options {
 	proxyTimeout := flag.Int("proxyTimeout", 10, "Timeout for proxy call.")
 	requestLogging := flag.Bool("requestLogging", false, "Flag for logging incoming requests.")
 	proxyCacheTTL := flag.Int("proxyCacheTTL", 120, "TTL, in seconds, for proxy cache of Remote API information")
+	annotatePassportHeaders := flag.Bool("annotatePassportHeaders", false, "Flag to indicate if we should annotate passport headers to outgoing request")
```
akgalwas, line 35: I'm not convinced that Application Gateway is the right place to add this functionality. Application Gateway is designed to be a generic component that is responsible for authenticating requests. We should strive to keep it simple and not litter the design with specific use cases.

I would consider the following solutions instead:

- Using Service Catalog Bindings and implementing header annotation in Serverless Function. Service Catalog offers Redis binding. You could implement a Server-less Function that will fetch  key value from database and add headers before calling Application Gateway. 

- Extending Application Gateway with generic mechanism for enriching requests.There is already #6346 issue that proposes such a mechanism.  Adding passport headers is a specific case that could be implemented with the mentioned mechanism.

components/application-gateway/internal/proxy/proxy.go
```diff
@@ -57,6 +63,12 @@ func New(
 	csrfTokenStrategyFactory csrf.TokenStrategyFactory,
 	config Config,
 	configRepository proxyconfig.TargetConfigProvider) ProxyHandler {
+	var passportEnricher *passport.RequestEnricher
+	if config.AnnotatePassportHeaders {
```
akgalwas, line 67: I would rather introduce some new abstraction here that will be responsible for enriching requests (similar to Authorization Strategy Factory and CSRF Token Strategy Factory).

kwiatekus:
@Poussinette please check if you still need this PR

PR #8978, opened 2020-07-07
Fix for AKS Kubelet Servicemonitor

hisarbalik:
Kubelet service monitor for prometheus is down on AKS cluster, service monitor monitoring wrong port. Kubelet not working anymore on http port, this fix will switch to the https metric port.


resources/monitoring/templates/kyma-additions/aks-kubelet-monitoring-patch.yaml
```diff
@@ -25,20 +25,30 @@ data:
     metadata:
       labels:
         app: {{ template "prometheus-operator.name" . }}-kubelet
-        {{- include "prometheus-operator.labels" . | indent 8 }}
+    {{- include "prometheus-operator.labels" . | indent 8 }}
```
rakesh-garimella, line 28: i think this should be indented 2 spaces.

hisarbalik: fixed

resources/monitoring/templates/kyma-additions/aks-kubelet-monitoring-patch.yaml
```diff
@@ -32,13 +32,27 @@ data:
       endpoints:
         - bearerTokenFile: /var/run/secrets/kubernetes.io/serviceaccount/token
           honorLabels: true
-          port: http-metrics
-          scheme: http
+          port: https-metrics
```
a-thaler, line 35: After that changes, what is the difference to the original kubelet servicemonitor?

hisarbalik:
> When seeing the changes it look slike that they aligned the aks kubelet to the standard one making the hole custom aks handling obsolete. Am I right? If yes, the goal must be to remove the custom aks stuff

You are right, patch for AKS removed testet on cluster.

a-thaler:
As the fix seem to stabilize the upgrade job as well, I force merged it in order to bypass the upgrade job fail

PR #9013, opened 2020-07-10
A way to have additional addons repositories configured

a-thaler:
**Description**
I introduced a new value to define additional CLusterAddonsConfiguration. That can be handy if you want to get an installation with a default set of repos populated. As a customer you don't want to maintain the single default repo, instead you want to additional, that's why introduced the new value instead of extending the default one.
I wanted to use a list initially but that will be much more complicated to configure by overrides, so I sticked to the map approach (not that flexible for the future but much more handy to use).

- fixed naming for default repository as it had a spelling mistake anyway
- introduced new value to have additional repos
- ...

**Related issue(s)**


docs/helm-broker/05-01-helm-broker.md
```diff
@@ -22,6 +22,6 @@ This table lists the configurable parameters, their descriptions, and default va
...
 | **global.isDevelopMode** | Defines whether to accept URL prefixes from the **global.urlRepoPrefixes.additionalDevelopMode** list. If set to `true`, the Helm Broker accepts the prefixes from the list. | `false` |
-  **global.urlRepoPrefixes.default** | Defines a list of accepted prefixes for repository URLs. | `'https://', 'git::', 'github.com/', 'bitbucket.org/'` |
-  **global.urlRepoPrefixes.additionalDevelopMode** | Defines a list of accepted prefixes for repository URLs when develop mode is enabled. | `'http://'` |
-
+| **global.urlRepoPrefixes.default** | Defines a list of accepted prefixes for repository URLs. | `'https://', 'git::', 'github.com/', 'bitbucket.org/'` |
+| **global.urlRepoPrefixes.additionalDevelopMode** | Defines a list of accepted prefixes for repository URLs when develop mode is enabled. | `'http://'` |
+| **additionalAddonsRepositories.myRepo** | A map of additional ClusterAddonsConfiguration repositories to create by default. The key gets used as name while the value gets used as URL for the repository. | `github.com/myOrg/myRepo//addons/index.yaml` |
```
klaudiagrz, line 27:
```suggestion
| **additionalAddonsRepositories.myRepo** | Provides a map of additional ClusterAddonsConfiguration repositories to create by default. The key is used as a name and the value is used as a URL for the repository. | `github.com/myOrg/myRepo//addons/index.yaml` |
```

resources/helm-broker/values.yaml
```diff
@@ -36,7 +36,13 @@ broker:
   statusPort: 8071
   metricsPort: 8072
 
-defautlAddonsURL: "https://github.com/kyma-project/addons/releases/download/0.12.0/index.yaml"
+defaultAddonsRepository: "https://github.com/kyma-project/addons/releases/download/0.12.0/index.yaml"
+
+# Map of additional addons repositories to configure. The key gets used as the name of the created ClusterAddonsConfiguration
```
klaudiagrz, line 41:
```suggestion
# A map of additional addons repositories to configure. The key is used as the name of the created ClusterAddonsConfiguration.
```

PR #9035, opened 2020-07-14
Bump memory in Helm broker

polskikiel:
**Description**

- Bumped memory in Helm broker

**Related issue(s)**


polskikiel:
discussed that, we will try to fix the root issue and then we can revert those settings

PR #9040, opened 2020-07-15
disable tracing in event source adapter

k15r:
**Description**

- tracing was always enabled and led to enormous amounts of logs lines if no zipkin endpoint is reachable. 
- tracing is enabled by default and can be disabled using overrides

**Related issue(s)**


components/event-sources/adapter/http/adapter.go
```diff
@@ -24,7 +24,8 @@ type envConfig struct {
 	EventSource string `envconfig:"EVENT_SOURCE" required:"true"`
 
 	// PORT as required by knative serving runtime contract
-	Port int `envconfig:"PORT" required:"true" default:"8080"`
+	Port           int  `envconfig:"PORT" required:"true" default:"8080"`
+	TracingEnabled bool `envconfig:"TRACING_ENABLED" default:"false"`
```
nachtmaar, line 28: Proposal: let's keep tracing enabled by default and provide an installation-override to disable it if needed

Issue #8959, opened 2020-07-03
Define Repository CR in kyma that could serve as source reference for functions


kwiatekus:
**Description**

Define a Repository CR that would reference a GIT repo.

**AC** 
 - repo CR should not be installed in serverless, at it could be used outside serverless as well
 - repo CR should reference a secret, allowing access to private repos  

**Reasons**

We want to enable GitOps development experience for lambda functions developers.
Such Repo CR will serve this purpose, but also might be used for other domains in kyma (i.e addons, or overall gitpos experience for k8s resources)

**Attachments**
[POC](https://github.com/kyma-project/kyma/issues/5752)

PR #9111, opened 2020-07-24
Enhance serverless' git repository auth support

tgorgol:
**Description**

- Enhance serverless' git repository auth support

**Related issue(s)**
https://github.com/kyma-project/kyma/issues/8959

### DIFF
--- a/components/function-controller/config/crd/bases/serverless.kyma-project.io_gitrepositories.yaml
+++ b/components/function-controller/config/crd/bases/serverless.kyma-project.io_gitrepositories.yaml
@@ -54,7 +54,6 @@ spec:
                   description: Type is the type of authentication
                   enum:
                   - basic
-                  - token
                   - key
                   type: string
               required:
--- a/components/function-controller/deploy/jobinit/Dockerfile
+++ b/components/function-controller/deploy/jobinit/Dockerfile
@@ -25,16 +25,15 @@ RUN go build -ldflags "-s -w" -a -o jobinit cmd/jobinit/main.go \
 && mv ./jobinit /app/jobinit \
 && if [ -f ${BASE_APP_DIR}/licenses ]; then mv ${BASE_APP_DIR}/licenses /app/licenses; fi
 
-# get latest CA certs
-FROM alpine:latest as certs
-RUN apk --update add ca-certificates
-
 # result container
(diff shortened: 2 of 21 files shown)


components/function-controller/deploy/jobinit/Dockerfile
```diff
@@ -25,16 +25,15 @@ RUN go build -ldflags "-s -w" -a -o jobinit cmd/jobinit/main.go \
 && mv ./jobinit /app/jobinit \
 && if [ -f ${BASE_APP_DIR}/licenses ]; then mv ${BASE_APP_DIR}/licenses /app/licenses; fi
 
-# get latest CA certs
-FROM alpine:latest as certs
-RUN apk --update add ca-certificates
-
 # result container
-FROM scratch
+FROM alpine:3.11
```
m00g3n, line 29: The alpine 3.12 is, it would be good to bump the version

pPrecel: Why is `alpine` better than `scratch`?

components/function-controller/config/crd/bases/serverless.kyma-project.io_gitrepositories.yaml
```diff
@@ -54,7 +54,6 @@ spec:
                   description: Type is the type of authentication
                   enum:
                   - basic
-                  - token
```
pPrecel, line 57: Why did you remove the `token` option? Is it not one of supported options?

tgorgol: Custom OAuth implementations differ between popular git providers, I think it's better to add a note to documentation mentioning the possibility of using token auth in a way described by the specific provider.

components/function-controller/deploy/jobinit/Dockerfile
tgorgol: We needed ssh in the container.

components/function-controller/config/crd/bases/serverless.kyma-project.io_gitrepositories.yaml
pPrecel: Good point. Thanks!

PR #9113, opened 2020-07-25
consistent chart.yaml for all kyma components

a-thaler:
**Description**

- introduced a consistent content of chart.yaml for all kyma components
- added icon and home for kyma
- set chart version to 1.0.0 across all charts as putting the actual kyma version is not a trivial task for automation
- removed all other existing fields like maintainers (they can be looked up in CODEOWNERS) or appVersion (they are rarely maintained at all)

**Related issue(s)**

### DIFF
--- a/resources/monitoring/Chart.yaml
+++ b/resources/monitoring/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v1
-apiVersion: v1
 description: Kyma component 'monitoring'
 name: monitoring
 version: 1.0.0
+# version: 8.12.6
 home: https://kyma-project.io
 icon: https://github.com/kyma-project/kyma/blob/master/logo.png?raw=true
--- a/resources/monitoring/Chart.yaml
+++ b/resources/monitoring/Chart.yaml
@@ -1,13 +1,7 @@
 apiVersion: v1
+apiVersion: v1
+description: Kyma component 'monitoring'
 name: monitoring
-description: Monitoring chart based on kube-prometheus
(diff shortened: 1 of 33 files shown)


k15r:
Would it make sense to add a proper description? Currently the generic "Kyma component '$componentname'" does not add any additional value.

a-thaler:
@k15r as we cannot modify the name of the chart, it was important for me to indicate that it is an official kyma component. We might add a link to the docu for more info. I'm afraid that any description will not be maintained in long run.. That might change if we switch to real helm charts hosted in a helm repo..

a-thaler:
@Szymongib I'm not sure. Sub-charts often contain 3party charts like for fluentbit, containing all the meta-data for the upstream chart. Sub-charts metadata will be not present on the cluster anyway, or?

resources/monitoring/Chart.yaml
```diff
@@ -1,13 +1,7 @@
 apiVersion: v1
+apiVersion: v1
```
suleymanakbas91, line 2: `apiVersion` is duplicated.

a-thaler: fixed

resources/ory/Chart.yaml
```diff
@@ -1,5 +1,6 @@
 apiVersion: v1
-appVersion: "1.0"
```
piotrmsc, line 2: With ory we in general want to put it 1:1 as much as possible with upstream chart

Issue #8179, opened 2020-04-23
Create specific directory for serverless function runtimes


aerfio:
**Description**

Right now the only function runetime is in `components/function-controller` directory, with rest of the files. This is not a good thing, as we build those images every time we make uncorrelated change in function-controller. I propose to split those function-base-images to separate folder in `components` folder

**Reasons**

Separation of concerns, less time spent on job when you only care for changes for fun-controller. Also, we do not want to spam with base images, as they will change very very rarely

**Attachments**

Issue #9055, opened 2020-07-16
Dockerfile presets for functions 


kwiatekus:
**Description**

Allow developers to skip dockerfile definition for their lambda functions. 
Lambda Function developer should only specify the source of the function code (git repository) and select one of the languages (with versions), ie:
 - **nodejs 12**
 - **nodejs 10**
 - **nodejs 8**
 - python 3.8 
 - ...

For those three runtimes we should provide a dockerfile out of the box, so that developer doesnt need to care about it.
If the repository contains dockerfile, it should overwrite the predefined one.

**Reasons**

Convenience for true serverless DevX
"Just run my node/go/pyhton code, I don't care how exactly"

kwiatekus:
Start with nodeJS support

aerfio:
Things to do:
- add new field to spec, e.g. "runtime", which value has to equal to the name of configmap with proper labels
  - this has to be somehow validated by webhook at runtime
- discuss whether we really need to enable users to bring their custom dockerfiles -> it brings A LOT of complexity, besides it goes against the idea of serverless and managed runtime, and we cannot guarantee that we'd somehow not brake anything during upgrade etc
- add more points here

aerfio:
After some internal discussions we've chosen to implement following runtimes for now:
- python 3.7.5
- nodejs 10
- nodejs 12

aerfio:
Update: we need to use 3.8.5 due to security issues...

PR #9117, opened 2020-07-27
Add function runtime dockerfiles

aerfio:
**Description**

- add node10 and python38 docker bases for functions (no implementation for now, just bases)

**Related issue(s)**

Part of https://github.com/kyma-project/kyma/issues/9055
Closes https://github.com/kyma-project/kyma/issues/8179


resources/serverless/values.yaml
```diff
@@ -73,6 +73,12 @@ images:
   functionRuntimeNodejs12:
     repository: "eu.gcr.io/kyma-project/function-runtime-nodejs12"
     tag: "1cc47db6"
+  functionRuntimeNodejs10:
```
dbadura, line 76: I wonder if such construction is more readable:
```
  functions:
    runtimes:
      nodejs12:
        repository: "eu.gcr.io/kyma-project/function-runtime-nodejs12"
        tag: "1cc47db6"
      python37:
        repository: "eu.gcr.io/kyma-project/function-runtime-python37"
        tag: "1cc47db6"
```

aerfio: probably it is :D gimme a second

components/function-runtimes/README.md
```diff
@@ -0,0 +1,5 @@
+# Function Runtimes
+
+## Overview
+
+This project includes Docker images, which are used as a base on which serverless Functions are build. See [this](../../docs/serverless) directory for more documentation for serverless Functions.
```
majakurcius, line 5:
```suggestion
This project includes Docker images which are used as a base for serverless Functions. See the [`docs/serverless`](../../docs/serverless) directory for more documentation on serverless Functions.
```
The comma after `which` must be removed, and `for` should be changed to `on`. Other changes are optional, but I would recommend them.

aerfio: Seems reasonable :D

components/function-runtimes/nodejs10/Dockerfile
```diff
@@ -0,0 +1,16 @@
+FROM kubeless/nodejs@sha256:aa819e9bee6c22db0b137595ac45765ef1cad64d0dc09502a008ce4446c607ef as source
```
m00g3n, line 1: The `kubeless/nodejs` image is base on `Debian GNU/Linux 9 (stretch)` and weights `610MB`.
The size of the `/kubeless_rt` directory that we copy is `5.5M`.

Can we somehow improve this?
The same question applies for python image.

aerfio: I'd rather be clear about what our sources are and from which image we've extracted those files

m00g3n: [kubeless/runtimes](https://github.com/kubeless/runtimes/tree/master/stable/nodejs) repository contains all the files (including `Dockerfiles`), lets investigate if we can improve this in a followup task;

PR #9126, opened 2020-07-28
Add ability to run python function in serverless

dbadura:
**Description**

- add ability to use python as a function.

**Related issue(s)**

### DIFF
--- a/components/function-controller/config/crd/bases/serverless.kyma-project.io_functions.yaml
+++ b/components/function-controller/config/crd/bases/serverless.kyma-project.io_functions.yaml
@@ -242,8 +242,6 @@ spec:
               type: string
             source:
               type: string
-          required:
-          - runtime
           type: object
       type: object
   version: v1alpha1
--- a/components/function-controller/config/crd/bases/serverless.kyma-project.io_functions.yaml
+++ b/components/function-controller/config/crd/bases/serverless.kyma-project.io_functions.yaml
@@ -18,6 +18,9 @@ spec:
   - JSONPath: .status.conditions[?(@.type=='Running')].status
     name: Running
     type: string
+  - JSONPath: .status.runtime
+    name: Runtime
+    type: string
   - JSONPath: .metadata.generation
     name: Version
     type: integer
@@ -190,12 +193,14 @@ spec:
               type: object
             runtime:
               enum:
-              - nodejs-12
+              - nodejs12
+              - nodejs10
+              - python37
               type: string
             source:
               description: Source defines the source code of a function
               type: string
-            type:
+            sourceType:
               type: string
           required:
           - source
@@ -231,10 +236,14 @@ spec:
               type: string
             runtime:
               enum:
-              - nodejs-12
+              - nodejs12
+              - nodejs10
+              - python37
               type: string
             source:
               type: string
+          required:
+          - runtime
           type: object
       type: object
   version: v1alpha1
--- a/components/function-controller/internal/controllers/serverless/runtime/runtime.go
+++ b/components/function-controller/internal/controllers/serverless/runtime/runtime.go
@@ -45,7 +45,7 @@ func GetRuntimeConfig(r serverlessv1alpha1.Runtime) Config {
 			DependencyFile:          "requirements.txt",
 			FunctionFile:            "handler.py",
 			DockerfileConfigMapName: "dockerfile-python-37",
-			RuntimeEnvs: []corev1.EnvVar{{Name: "PYTHONPATH", Value: "$(KUBELESS_INSTALL_VOLUME)/lib.python3.7/site-packages:$(KUBELESS_INSTALL_VOLUME)"},
+			RuntimeEnvs: []corev1.EnvVar{{Name: "PYTHONPATH", Value: "$(KUBELESS_INSTALL_VOLUME)/lib.python3.7/site-packages:$(KUBELESS_INSTALL_VOLUME)"}, // https://github.com/kubeless/runtimes/blob/master/stable/python/python.jsonnet#L45
 				{Name: "FUNC_RUNTIME", Value: "python37"}},
 		}
 	default:
--- a/components/function-controller/internal/controllers/serverless/runtime/runtime.go
+++ b/components/function-controller/internal/controllers/serverless/runtime/runtime.go
@@ -1,0 +1,73 @@
+package runtime
+
+import (
+	serverlessv1alpha1 "github.com/kyma-project/kyma/components/function-controller/pkg/apis/serverless/v1alpha1"
+	corev1 "k8s.io/api/core/v1"
+)
+
+type Runtime interface {
+	SanitizeDependencies(dependencies string) string
+}
+
+type Config struct {
+	Runtime                 serverlessv1alpha1.Runtime
+	DependencyFile          string
+	FunctionFile            string
+	DockerfileConfigMapName string
+	RuntimeEnvs             []corev1.EnvVar
+}
+
+func GetRuntimeConfig(r serverlessv1alpha1.Runtime) Config {
+	switch r {
+	case serverlessv1alpha1.Nodejs12:
+		return Config{
+			Runtime:                 serverlessv1alpha1.Nodejs12,
+			DependencyFile:          "package.json",
+			FunctionFile:            "handler.js",
+			DockerfileConfigMapName: "dockerfile-nodejs-12",
+			RuntimeEnvs: []corev1.EnvVar{{Name: "NODE_PATH", Value: "$(KUBELESS_INSTALL_VOLUME)/node_modules"},
+				{Name: "FUNC_RUNTIME", Value: "nodejs12"},
+			},
+		}
+	case serverlessv1alpha1.Nodejs10:
+		return Config{
+			Runtime:                 serverlessv1alpha1.Nodejs10,
+			DependencyFile:          "package.json",
+			FunctionFile:            "handler.js",
+			DockerfileConfigMapName: "dockerfile-nodejs-10",
+			RuntimeEnvs: []corev1.EnvVar{{Name: "NODE_PATH", Value: "$(KUBELESS_INSTALL_VOLUME)/node_modules"},
+				{Name: "FUNC_RUNTIME", Value: "nodejs10"},
+			},
+		}
+	case serverlessv1alpha1.Python37:
+		return Config{
+			Runtime:                 serverlessv1alpha1.Python37,
+			DependencyFile:          "requirements.txt",
+			FunctionFile:            "handler.py",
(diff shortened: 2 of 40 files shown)


aerfio:
To do:
- update webhook configuration (make sure that proper envs are set in place for particular runtime etc)
- write various tests
- make sure that proper envs are injected into function deployment

components/function-controller/config/samples/node10_test.yaml
```diff
@@ -0,0 +1,20 @@
+apiVersion: serverless.kyma-project.io/v1alpha1
```
aerfio, line 1: delete that file later, or add python hello world instead

components/function-controller/config/crd/bases/serverless.kyma-project.io_functions.yaml
```diff
@@ -4,7 +4,7 @@ apiVersion: apiextensions.k8s.io/v1beta1
 kind: CustomResourceDefinition
 metadata:
   annotations:
-    controller-gen.kubebuilder.io/version: v0.2.4
+    controller-gen.kubebuilder.io/version: v0.2.0-rc.0
```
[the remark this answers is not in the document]
aerfio: run `rm $(which controller-gen)` and then `make manifest`

components/function-controller/internal/controllers/serverless/runtime/runtime.go
```diff
@@ -0,0 +1,73 @@
...
+			},
+		}
+	case serverlessv1alpha1.Python37:
+		return Config{
+			Runtime:                 serverlessv1alpha1.Python37,
+			DependencyFile:          "requirements.txt",
+			FunctionFile:            "handler.py",
+			DockerfileConfigMapName: "dockerfile-python-37",
+			RuntimeEnvs: []corev1.EnvVar{{Name: "PYTHONPATH", Value: "$(KUBELESS_INSTALL_VOLUME)/lib.python3.7/site-packages:$(KUBELESS_INSTALL_VOLUME)"},
+				{Name: "FUNC_RUNTIME", Value: "python3"}},
```
aerfio, line 49: python37

components/function-controller/pkg/apis/serverless/v1alpha1/function_defaults_test.go
```diff
@@ -4,6 +4,8 @@ import (
 	"context"
 	"testing"
 
+	"github.com/stretchr/testify/require"
```
aerfio, line 7: please do not add one more testing library 😿

components/function-controller/pkg/apis/serverless/v1alpha1/function_types.go
```diff
@@ -23,6 +29,9 @@ type FunctionSpec struct {
 	// Deps defines the dependencies for a function
 	Deps string `json:"deps,omitempty"`
 
+	// +optional
+	Runtime Runtime `json:"runtime, omitempty"`
```
aerfio, line 33: i'm not sure whether that space after `runtime,` is correct, it could introduce some nasty bug

components/function-controller/pkg/apis/serverless/v1alpha1/runtime_validation.go
```diff
@@ -0,0 +1,25 @@
...
+
+func ValidateDependencies(runtime Runtime, dependencies string) error {
+	switch runtime {
+	case Nodejs10, Nodejs12:
+		return validateNodeJSDependecies(dependencies)
+	case Python37:
+		return nil
+	}
+	return fmt.Errorf("Cannot find runtime: %s", runtime)
+
```
aerfio, line 17: this part looks like it needs goimports, I may be wrong

aerfio:
image

What do you think about adding new printing column to function CR? so that the user sees the runtime immediately? Above you see just an example how it looks right now

dbadura:
> image
> 
> What do you think about adding new printing column to function CR? so that the user sees the runtime immediately? Above you see just an example how it looks right now

Good idea!

pPrecel:
Nice work! 👌  
I have tested cooperation of this functionality with the gitops functions and I observed a few issues. 
**First:** in `resources/serverless/values.yaml` we can read that tags nodejs10 and python37 runtimes are setted as `tba` is this correct? They didn't work for me and i replaced it with `dcc73092` but it didn't work with `nodejs10` with executor error `/bin/sh: /kubeless-npm-install.sh: Permission denied`
**Second:** When I have changed runtime from `nodejs10` to `nodejs12` controller updated the status of function but didn't trigger building a new image. 
**Third:** For git functionality we should update the `calculateGitImageTag` method in the function_controller.go because controller should trigger building new image when user change runtime. So we have to add runtime to the generated tag. My proposition:
```text
        data := strings.Join([]string{
		string(instance.GetUID()),
		instance.Status.Commit,
		instance.Status.Repository.BaseDir,
add ->          string(instance.Status.Runtime),
	}, "-")
```

components/function-controller/internal/controllers/serverless/runtime/runtime.go
```diff
@@ -0,0 +1,73 @@
...
+	case serverlessv1alpha1.Python37:
+		return Config{
+			Runtime:                 serverlessv1alpha1.Python37,
+			DependencyFile:          "requirements.txt",
+			FunctionFile:            "handler.py",
+			DockerfileConfigMapName: "dockerfile-python-37",
+			RuntimeEnvs: []corev1.EnvVar{{Name: "PYTHONPATH", Value: "$(KUBELESS_INSTALL_VOLUME)/lib.python3.7/site-packages:$(KUBELESS_INSTALL_VOLUME)"},
+				{Name: "FUNC_RUNTIME", Value: "python37"}},
+		}
+	default:
```
kfurgol, line 51: missing unit test for that case. I see that it's the same as `Nodejs12`, just wanted to point it out

components/function-controller/internal/controllers/serverless/runtime/runtime.go
```diff
@@ -0,0 +1,73 @@
...
+				{Name: "FUNC_RUNTIME", Value: "nodejs10"},
+			},
+		}
+	case serverlessv1alpha1.Python37:
+		return Config{
+			Runtime:                 serverlessv1alpha1.Python37,
+			DependencyFile:          "requirements.txt",
+			FunctionFile:            "handler.py",
+			DockerfileConfigMapName: "dockerfile-python-37",
+			RuntimeEnvs: []corev1.EnvVar{{Name: "PYTHONPATH", Value: "$(KUBELESS_INSTALL_VOLUME)/lib.python3.7/site-packages:$(KUBELESS_INSTALL_VOLUME)"},
```
kfurgol, line 48: I would change the name from `PYTHONPATH` to `PYTHON_PATH`

components/function-controller/internal/controllers/serverless/runtime/runtime.go
```diff
@@ -0,0 +1,73 @@
...
+
+func GetRuntimeConfig(r serverlessv1alpha1.Runtime) Config {
+	switch r {
+	case serverlessv1alpha1.Nodejs12:
+		return Config{
+			Runtime:                 serverlessv1alpha1.Nodejs12,
+			DependencyFile:          "package.json",
+			FunctionFile:            "handler.js",
+			DockerfileConfigMapName: "dockerfile-nodejs-12",
+			RuntimeEnvs: []corev1.EnvVar{{Name: "NODE_PATH", Value: "$(KUBELESS_INSTALL_VOLUME)/node_modules"},
```
kfurgol, line 28: Maybe it would be a good idea to set these values as a consts? (Speaking about `NODE_PATH`, `FUNC_RUNTIME` etc.)

components/function-controller/internal/controllers/serverless/suite_test.go
```diff
@@ -22,6 +27,10 @@ var config FunctionConfig
 var resourceClient resource.Client
 var testEnv *envtest.Environment
 
+const (
+	testNamespace = "tutaj"
```
kfurgol, line 31: not a fan of using non-english words in code ;) just a loose suggestion to change it to metasyntactic variable

components/function-controller/pkg/apis/serverless/v1alpha1/function_defaults.go
```diff
@@ -89,3 +93,9 @@ func (spec *FunctionSpec) defaultResources(ctx context.Context) {
 		spec.Resources.Limits[corev1.ResourceCPU] = newRerource
```
kfurgol, line 93: I know it is not a line you've changed, just found a typo:
`newRerource` -> `newResource`

components/function-controller/pkg/apis/serverless/v1alpha1/runtime_validation.go
```diff
@@ -0,0 +1,25 @@
...
+	case Nodejs10, Nodejs12:
+		return validateNodeJSDependecies(dependencies)
+	case Python37:
+		return nil
+	}
+
+	return fmt.Errorf("cannot find runtime: %s", runtime)
+}
+
+func validateNodeJSDependecies(dependencies string) error {
```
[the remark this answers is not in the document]
kfurgol: if you apply the suggestion, please do not forget to change the executions of the function too

kfurgol (commented):
Just a code review, overall it looks good to me. I put some suggestions below

components/function-controller/pkg/apis/serverless/v1alpha1/runtime_validation.go
```diff
@@ -0,0 +1,25 @@
...
+	"fmt"
+	"strings"
+)
+
+func ValidateDependencies(runtime Runtime, dependencies string) error {
+	switch runtime {
+	case Nodejs10, Nodejs12:
+		return validateNodeJSDependecies(dependencies)
+	case Python37:
+		return nil
```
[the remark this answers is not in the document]
aerfio: No validation here, as requirements.txt has it's own format, better to leave it be then fall into some kind of edge case

components/function-controller/internal/controllers/serverless/runtime/runtime.go
aerfio: https://github.com/kubeless/runtimes/blob/master/stable/python/python.jsonnet#L45 unfortunately it has to be set to that value. I'll add that link as comment

components/function-controller/internal/controllers/serverless/runtime/runtime.go
aerfio: That's correct :D

kfurgol (approved):

aerfio (approved):

parostatkiem (approved):

kazydek (approved):

piotrmsc (approved):

Issue #8875, opened 2020-06-22
Namespace creation failed during the second run of core-test-external-solution in stability checker


sayanh:
**Description**

It is found that namespace creation fails due to certain dangling resources. Hence, the second round of core-test-external-solution fails while creating namespace.

**Expected result**

Cleanup should be smooth.

**Actual result**

```
 Start of logs from container "test" in pod "oct-tp-testsuite-all-core-test-external-solution-0" in status "Failed"
  TESTING start
  time="2020-06-18T04:37:50Z" level=info msg="Cleanup mode: yes"
  time="2020-06-18T04:37:50Z" level=info msg="Step: 'Parallelized: Create namespace e2e-event-mesh, Create application e2e-event-mesh'"
  time="2020-06-18T04:38:11Z" level=error msg="Error in 'Parallelized: Create namespace e2e-event-mesh, Create application e2e-event-mesh': 1 error occurred:\n\t* object is being deleted: namespaces \"e2e-event-mesh\" already exists\n\n"
  time="2020-06-18T04:38:11Z" level=info msg="Cleanup: 'Parallelized: Create namespace e2e-event-mesh, Create application e2e-event-mesh'"
  time="2020-06-18T04:38:11Z" level=warning msg="Error during 'Parallelized: Create namespace e2e-event-mesh, Create application e2e-event-mesh' cleanup: 1 error occurred:\n\t* Operation cannot be fulfilled on namespaces \"e2e-event-mesh\": The system is ensuring all content is removed from this namespace.  Upon completion, this namespace will automatically be purged by the system.\n\n"
```
Reference:
https://storage.googleapis.com/kyma-prow-logs/logs/kyma-integration-gardener-azure/1268392110940229633/build-log.txt

Logs of stability checker

**Steps to reproduce**

Not easy to reproduce as it happens sometimes.

marcobebway:
**Description:**

After investigating this, we found that the namespace stayed in `Terminating` state because the service catalog controller could not delete the service instance and service binding. The service binding ready status remained `False` with the reason `ReferencesNonexistentServiceClass`.

```yaml
$ kubectl get servicebinding.servicecatalog.k8s.io -n e2e-event-mesh e2e-event-mesh -oyaml

apiVersion: servicecatalog.k8s.io/v1beta1
kind: ServiceBinding
metadata:
  creationTimestamp: "2020-07-30T15:16:47Z"
  deletionGracePeriodSeconds: 0
[...]

PR #9128, opened 2020-07-28
Force cleanup the created service instance and service binding in case of errors

marcobebway:
**Description**

The external solution E2E test sometimes fails to cleanup the created test namespace then exits with failure. And when the test suite executes it for the second time, it also fails because the test namespace still exists. The namespace is not deleted because the service catalog controller fails to cleanup the created service instance and service binding leaving its finalizer in their finalizers list. This PR fixes this behaviour by cleaning-up the finalizers list of the two resources so they can be garbage collected.

**Related issue(s)**

- Fixes https://github.com/kyma-project/kyma/issues/8875.

### DIFF
--- a/tests/end-to-end/external-solution-integration/pkg/helpers/k8s.go
+++ b/tests/end-to-end/external-solution-integration/pkg/helpers/k8s.go
@@ -39,10 +39,10 @@ func InClusterEndpoint(name, namespace string, port int) string {
 // AwaitResourceDeleted retries until the resources cannot be found any more
 func AwaitResourceDeleted(check func() (interface{}, error)) error {
 	return retry.Do(func() error {
-		_, err := check()
+		t, err := check()
 
 		if err == nil {
-			return errors.New("resource still exists")
+			return errors.New(fmt.Sprintf("resource still exists: %T", t))
 		}
 
 		if !k8serrors.IsNotFound(err) {
--- a/tests/end-to-end/external-solution-integration/pkg/step/step.go
+++ b/tests/end-to-end/external-solution-integration/pkg/step/step.go
@@ -19,8 +19,9 @@ type Step interface {
 
 // Runner executes steps in safe manner
 type Runner struct {
-	log     *logrus.Logger
-	cleanup CleanupMode
+	log             *logrus.Logger
+	cleanup         CleanupMode
+	cleanupBehavior CleanupBehavior
 }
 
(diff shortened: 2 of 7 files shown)


tests/end-to-end/external-solution-integration/pkg/helpers/k8s.go
```diff
@@ -39,10 +39,10 @@ func InClusterEndpoint(name, namespace string, port int) string {
 // AwaitResourceDeleted retries until the resources cannot be found any more
 func AwaitResourceDeleted(check func() (interface{}, error)) error {
 	return retry.Do(func() error {
-		_, err := check()
+		t, err := check()
 
 		if err == nil {
-			return errors.New("resource still exists")
+			return errors.New(fmt.Sprintf("resource still exists: %T", t))
```
nachtmaar, line 45: I think the fmt.Sprintf is unnecessary:

(nit)
```suggestion
			return fmt.Errorf("resource still exists: %T", t)
```

tests/end-to-end/external-solution-integration/pkg/step/step.go
```diff
@@ -100,3 +112,10 @@ func isNotFound(err error) bool {
 	})
 	return isNotFound
 }
+
+type CleanupBehavior string
+
+const (
+	CleanupBehaviorAllSteps         CleanupBehavior = "all"
```
nachtmaar, line 119: What is the reason behind introducing this? Why is a cleanup of all steps required even though the step did not get a change to create resources? 🤔

tests/end-to-end/external-solution-integration/pkg/testsuite/create_service_binding.go
```diff
@@ -80,3 +69,43 @@ func (s *CreateServiceBinding) isServiceBindingReady() error {
...
+		return err
+	}
+
+	// skip cleanup if the service binding is not marked for deletion or the finalizers list is empty
+	if sb.DeletionTimestamp == nil || len(sb.Finalizers) == 0 {
+		return nil
+	}
+
+	sb = sb.DeepCopy()
+	sb.Finalizers = []string{}
```
nachtmaar, line 105: I am curious. Is this a known problem? If not, we should report this to service-catalog so that they can fix the underlying issue.

marcobebway:
As discussed offline, we decided to close this PR and solve this issue as part of the service catalog.

PR #9129, opened 2020-07-28
Remove hardcoded Istio Gateway in Logging component

pkosiec:
**Description**

- Remove hardcoded Istio Gateway in Logging component

Previously, the gateway was hardcoded, which prevented using the component with different Istio Gateway, like `compass-gateway`. Now the gateway is configurable.


a-thaler:
Looks good, can you please apply the same for /resources/logging/templates/loki/virtualservice.yaml, that will be great to have it consistent in whole chart

pkosiec:
@a-thaler Yup, I applied it there while opening the PR. However I had an indentation issue, that's why all CI jobs fail. Now it should be fine.

PR #9134, opened 2020-07-29
Support the gitops reference in the function-controller

pPrecel:
**Description**

- Add possibility to use method "LastCommit" with every reference type (tag, branch, commit) 

**Related issue(s)**

[Issue #8959](https://github.com/kyma-project/kyma/issues/8959)


components/function-controller/internal/controllers/serverless/function_controller.go
```diff
@@ -94,9 +95,10 @@ func (r *FunctionReconciler) updateStatus(
```
tgorgol, line 95: I think we should also check if `instance.Status.Commit == commit` in line 92, otherwise if someone uses branch name as reference, the Repositories will be equal and the status won't be updated with the new `.Status.Commit`.

components/function-controller/internal/controllers/serverless/function_controller.go
```diff
@@ -64,11 +64,11 @@ func (r *FunctionReconciler) calculateImageTag(instance *serverlessv1alpha1.Func
 }
 
 func (r *FunctionReconciler) updateStatusWithoutRepository(ctx context.Context, result ctrl.Result, instance *serverlessv1alpha1.Function, condition serverlessv1alpha1.Condition) (ctrl.Result, error) {
-	return r.updateStatus(ctx, result, instance, condition, nil)
+	return r.updateStatus(ctx, result, instance, condition, nil, "")
 }
 
 func (r *FunctionReconciler) calculateGitImageTag(instance *serverlessv1alpha1.Function) string {
-	hash := sha256.Sum256([]byte(fmt.Sprintf("%s-%s", instance.GetUID(), instance.Status.Repository.Commit)))
+	hash := sha256.Sum256([]byte(fmt.Sprintf("%s-%s", instance.GetUID(), instance.Status.Repository.Reference)))
```
tgorgol, line 71: Shouldn't this be `instance.Status.Commit`?

tgorgol:
Tag support turned out to be a bit more tricky than we expected. For now we'll support only commit hash and branch as a reference.

PR #9145, opened 2020-07-30
Use newer moby package version in cbs

dariadomagala:
**Description**

- Use newer moby package version in cbs

**Related issue(s)**


dariadomagala:
this change is already on master

PR #9150, opened 2020-07-31
Rc3 event triggers fix

parostatkiem:
**Description**

- ...
- ...
- ...

**Related issue(s)**


adamwalach:
Since this is fix created based on old master version we can't just merge it.
Build job is also not able to proceed because of conflicts (we can't just resolve them cause that would mean rebase to the current master).
Because of that we generated docker image manually:
```
eu.gcr.io/kyma-project/console-backend-service:PR-9151
```

PR #9175, opened 2020-08-05
Show more informations about functions in the external-solution test

pPrecel:
**Description**

- Expand one more information when a function is deploying in external-solution test

### DIFF
--- a/tests/end-to-end/external-solution-integration/pkg/testsuite/deploy_function.go
+++ b/tests/end-to-end/external-solution-integration/pkg/testsuite/deploy_function.go
@@ -157,7 +157,7 @@ func (s *DeployFunction) isFunctionReady() error {
 	}
 
 	if !s.isReady(function) {
-		return fmt.Errorf("function \" %+v \"is not ready yet", function)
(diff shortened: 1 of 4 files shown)


tests/end-to-end/external-solution-integration/pkg/testsuite/deploy_function.go
```diff
@@ -157,7 +157,7 @@ func (s *DeployFunction) isFunctionReady() error {
 	}
 
 	if !s.isReady(function) {
-		return errors.New("function is not ready yet")
+		return fmt.Errorf("function \" %+v \"is not ready yet", function)
```
dbadura, line 160:
```suggestion
		return fmt.Errorf(`function "%+v" is not ready yet`, function)
```

magicmatatjahu, line 160: Is quotes necessary? Maybe use new lines.

pPrecel: Good point

PR #9176, opened 2020-08-05
Delete values schema json

sayanh:
**Description**

- Although kyma-operator takes each override from configmap/secret and inject to helm charts as string(which is again not correct as the data for a secret is a [byte map](https://github.com/kubernetes/kubernetes/blob/master/pkg/apis/core/types.go#L4896)) key:value pairs, the values `true` or `false` for `.Values.global.httpAdapter.tracingEnabled` were treated as boolean. Hence, helm lint/install failed as the schema strictly expects this value to be a string. Therefore, deleting the schema for the values is the way to go due to the unpredictable behavior of the Kyma-operator.

- Deleting all the values.schema.json from all charts due to [this](https://github.com/kyma-project/kyma/pull/9176#issuecomment-669068224) 

**Related issue(s)**

See https://github.com/kyma-project/control-plane/issues/104


a-thaler:
Could you please delete them for all charts? I think they are fully outdated and need to be re-introduced when the installer supports values.yaml

sayanh:
@a-thaler I removed all of them now.

Issue #5752, opened 2019-09-25
POC : GitOps support for serverless


devguyio:
**Description**

Propose a solution for GitOps support for serverless.

**Requirements**
* Function code could be versioned at GIT (possibility to provide a source code should be preserved)
* Function version can be fixed to specific  repo URL, branch and commit hash (optional)
* In case commit hash is not fixed, function is rebuild and restarted on every commit to the branch
* support for private repositories  - include a mechanism to reference a secret that is necessary to access private repos
* gitops configuration in Function CR only - exclude UI for the sake of MVP 
* support multiple functions in one repo

Abd4llA:
@akrausesap do you have any ideas in mind?

montaro:
@Abd4llA Do you have input from the customers and the stakeholders in terms of features to add here, please?

michal-hudy:
I think we should create GitOps not only for Functions, but for Kyma at all. Functions should support git repositories as the source, but this is separate functionality.

Description edited

michal-hudy:
> In case commit hash is not fixed, function is rebuild and restarted on every commit to the branch

Do we plan to monitor a repository? It will be pull or push? If pull then we are independent from the platform (GitHub, GitLab, GCP Source, etc.) but if it will be a push then we will need to have support for webhooks. Yes, we can use some ready solutions for that, but it needs to be investigated.

michal-hudy:
For POC scope private repositories are also supported, or only public?

michal-hudy:
Another thing 😄 Does the repository will contain `Dockerfile` and `Function` CR?

kwiatekus:
> Do we plan to monitor a repository? It will be pull or push? If pull then we are independent from the platform (GitHub, GitLab, GCP Source, etc.) but if it will be a push then we will need to have support for webhooks. Yes, we can use some ready solutions for that, but it needs to be investigated.

I'd start with easiest for the sake of the POC - pull.

kwiatekus:
> For POC scope private repositories are also supported, or only public? What about authentication?

Excelent point. I will add it to the the requirements

kwiatekus:
> Another thing 😄 Does the repository will contain `Dockerfile` and `Function` CR?
I imagine the code + dockerfile.
In the POC I'd leave the Function CR out

kwiatekus:
Investigate how addons are fetched form git. 
Maybe we could reuse the same mechanisms for git communication

michal-hudy:
[...]

PR #9213, opened 2020-08-10
Extend function controller code coverage

m00g3n:
**Description**

- Extend function controller code coverage
- Refactor function controller GitOps tests

**Related issue(s)**

See also: #5752
See also: #8959


components/function-controller/internal/controllers/serverless/function_reconcile_matchers_test.go
```diff
@@ -0,0 +1,91 @@
+package serverless
```
tgorgol, line 1: I noticed most test files use the `serverless` package instead of `serverless_test`, cluttering the scope. Maybe we could investigate how much effort would it require to move the unit tests to the `serverless_test` package? Not necessarily in this PR.

PR #9254, opened 2020-08-13
Introduce HTTP Output for Fluent-Bit configuration

NickyMateev:
**Description**

This change introduces HTTP Output plugin configuration as an option to set-up the Fluent-Bit deployment: see [Fluent Bit: Official Manual]() and [Fluent Bit Configuration parameters](https://github.com/helm/charts/tree/master/stable/fluent-bit#configuration)

The HTTP Output plugin allows to flush your records into a HTTP endpoint. For now the functionality is pretty basic and it issues a POST request with the data records in MessagePack (or JSON) format.

### DIFF
--- a/resources/logging/charts/fluentbit/templates/daemonset.yaml
+++ b/resources/logging/charts/fluentbit/templates/daemonset.yaml
@@ -77,10 +77,23 @@ spec:
           mountPath: /secure/forward-tls.key
           subPath: tls.key
 {{- end }}
-{{- if and (.Values.backend.forward.tls.verify) (.Values.backend.forward.tls.ca) }}
-        - name: forward-ca-secret
-          mountPath: /secure/forward-tls-ca.crt
-          subPath: forward-tls-ca.crt
+{{- if and (.Values.backend.http.tls_verify) (.Values.backend.http.tls_ca) }}
+        - name: http-ca-secret
+          mountPath: /secure/http-tls-ca.crt
+          subPath: http-tls-ca.crt
+{{- end }}
+{{- if and (.Values.backend.http.tls_cert) (.Values.backend.http.tls_key) }}
+        - name: http-tls-secret
+          mountPath: /secure/http-tls.crt
+          subPath: tls.crt
+        - name: http-tls-secret
+          mountPath: /secure/http-tls.key
+          subPath: tls.key
+{{- end }}
+{{- if and (.Values.backend.http.tls_verify) (.Values.backend.http.tls_ca) }}
+        - name: http-ca-secret
+          mountPath: /secure/http-tls-ca.crt
+          subPath: http-tls-ca.crt
 {{- end }}
 {{- with .Values.extraVolumeMounts }}
         {{- tpl . $ | nindent 8}}
--- a/resources/logging/charts/fluentbit/templates/daemonset.yaml
+++ b/resources/logging/charts/fluentbit/templates/daemonset.yaml
@@ -77,10 +77,10 @@ spec:
           mountPath: /secure/forward-tls.key
           subPath: tls.key
 {{- end }}
-{{- if and (.Values.backend.http.tls_verify) (.Values.backend.http.tls_ca) }}
-        - name: http-ca-secret
-          mountPath: /secure/http-tls-ca.crt
-          subPath: http-tls-ca.crt
+{{- if and (.Values.backend.forward.tls.verify) (.Values.backend.forward.tls.ca) }}
+        - name: forward-ca-secret
+          mountPath: /secure/forward-tls-ca.crt
+          subPath: forward-tls-ca.crt
 {{- end }}
 {{- if and (.Values.backend.http.tls_cert) (.Values.backend.http.tls_key) }}
         - name: http-tls-secret
--- a/resources/logging/charts/fluentbit/templates/daemonset.yaml
+++ b/resources/logging/charts/fluentbit/templates/daemonset.yaml
@@ -168,6 +168,16 @@ spec:
         secret:
           secretName: "{{ template "fluent-bit.fullname" . }}-forward-tls-secret"
 {{- end }}
+{{- if and (.Values.backend.http.tls_verify) (.Values.backend.http.tls_ca) }}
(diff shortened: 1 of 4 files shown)


NickyMateev:
A suggestion was made by @rakesh-garimella to use the Fluent-Bit `conf.extra` field and provide overrides for that in the form:
```
extra: #|
  #  [FILTER]
  #      Name              test
  #      Match             *
  #  [OUTPUT]
  #      Name              test
  #      Match             *
```

There is a problem with such an alternative approach - it couples all configuration and confidential information in one place:

```
Name  http
      Match *
      Host **<host>**
      Port 443
      HTTP_User **<user>**
      HTTP_Passwd **<password>**
      tls true
      tls.verify false
      tls.debug 1
      URI /
      Format json
```

Our deploy pipelines first apply any necessary Kyma override configurations and then fetch the all confidential data like usernames/passwords, certificates from a remote store in the form of K8S secrets and then they apply them.
Unlike the standard `forward` and `elasticsearch` configurations which allow for such a pipeline setup where first all the configuration overrides are applied for the fluent-bit output plugin (as ConfigMap overrides) and only afterwards the username, password or whatever authentication data is applied as a separate secret, in the `conf.extra` approach this separation does not seem possible.

a-thaler (changes requested):
Hi @NickyMateev , thanks for your PR. In general it looks good, only the TLS settings are not aligned with the TLS settings of the other configs and the options could be more complete, see my comments

resources/logging/charts/fluentbit/values.yaml
```diff
@@ -230,6 +233,20 @@ backend:
...
+    http_user: ""
+    http_passwd: ""
+    tls: "off"
+    tls_verify: on
+    tls_debug: 1
+    ## Specify the data format to be used in the HTTP request body
+    ## Can be either 'msgpack' or 'json'
+    format: msgpack
+    # json_date_format: double or iso8601
+    headers: []
```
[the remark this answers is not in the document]
NickyMateev: Sure thing, I added them as well.

resources/logging/charts/fluentbit/values.yaml
```diff
@@ -230,6 +233,20 @@ backend:
...
       debug: 1
+  http:
+    host: "127.0.0.1"
+    port: 80
+    uri: "/"
+    http_user: ""
+    http_passwd: ""
+    tls: "off"
+    tls_verify: on
+    tls_debug: 1
```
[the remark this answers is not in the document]
NickyMateev: Synced so it looks like the forward plugin 👍

NickyMateev:
@a-thaler - I've addressed your comments.

resources/logging/charts/fluentbit/templates/daemonset.yaml
```diff
@@ -82,6 +82,19 @@ spec:
...
 {{- end }}
+{{- if and (.Values.backend.http.tls_cert) (.Values.backend.http.tls_key) }}
+        - name: http-tls-secret
+          mountPath: /secure/http-tls.crt
+          subPath: tls.crt
+        - name: http-tls-secret
+          mountPath: /secure/http-tls.key
+          subPath: tls.key
+{{- end }}
+{{- if and (.Values.backend.http.tls_verify) (.Values.backend.http.tls_ca) }}
```
a-thaler, line 93: Here you make the condition dependent on the verify flag, in the configmap.yaml not. Seems inconsistent, please double-check.

resources/logging/charts/fluentbit/templates/daemonset.yaml
```diff
@@ -82,6 +82,19 @@ spec:
           mountPath: /secure/forward-tls-ca.crt
           subPath: forward-tls-ca.crt
 {{- end }}
+{{- if and (.Values.backend.http.tls_cert) (.Values.backend.http.tls_key) }}
+        - name: http-tls-secret
```
a-thaler, line 86: You need to add the two secrets to the secrets.yaml, at the moment that mount will fail

a-thaler (changes requested):
sorry for late review, vacation time..
Please have another look, the value of the crt file is not used anywhere as you forgot to intorduce the related secrets.

resources/logging/charts/fluentbit/templates/daemonset.yaml
NickyMateev: It's identical like it was already done for the forward plugin configuration - the same discrepancy can be observed there as well.

NickyMateev:
@a-thaler any idea why the `pre-master-kyma-gke-compass-integration` job is failing?

```
kyma installation error, description: install component logging
kyma installation error, description: install component logging
Installation errors until timeout:
----------
logging:
  Helm installation of release "logging" failed: unable to build kubernetes objects from release manifest: error validating "": error validating data: unknown object type "nil" in Secret.data.http-tls-ca.crt
```

I tested it multiple times yesterday locally and even if `tls_ca` is empty string value `""`, the corresponding  `Secret.data.http-tls-ca.crt` would also be `""` and Kyma installation would carry on and logs would start flushing to the HTTP endpoint that's specified.

a-thaler:
Thanks for fixing it. In general it looks good to me now. We will need some more time to test it manual again, however the integration job already found a problem, please see the logs of the prow job:

`Helm installation of release "logging" failed: unable to build kubernetes objects from release manifest: error validating "": error validating data: unknown object type "nil" in Secret.data.http-tls-ca.crt`

NickyMateev:
Indeed the same log error message I referenced in the previous comment - I would like to have your assistance on why this might be happening.

NickyMateev:
@a-thaler the problem seemed to be a mismatch in my local tiller version and the one in the prow cluster executing the jobs. Mine didn't exhibit some well-known Helm issue which I fixed with some additional validation: https://github.com/helm/helm/issues/5750

Everything seems to be OK now.

a-thaler (approved):
Thanks for fixing it

NickyMateev:
Thanks for the support @a-thaler ! 🍻

PR #9261, opened 2020-08-17
Improve ORY Hydra init container

Demonsthere:
**Description**

- Make the logic more resilient
**Related issue(s)**

### DIFF
--- a/resources/ory/charts/hydra/templates/deployment.yaml
+++ b/resources/ory/charts/hydra/templates/deployment.yaml
@@ -35,19 +35,7 @@ spec:
             - /bin/sh
             - -c
             - |
-              DB_URL_PORT=$(echo $DSN | cut -d '@' -f2 | cut -d '/' -f 1 )
-
-              if echo "$DB_URL_PORT" | grep -q ':'; then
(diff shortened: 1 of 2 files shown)


resources/ory/charts/hydra/templates/deployment.yaml
```diff
@@ -35,8 +35,15 @@ spec:
             - /bin/sh
             - -c
             - |
-              export DB_URL=$(echo $DSN | cut -d '@' -f2 | cut -d ':' -f1)
-              export DB_PORT=$(echo $DSN | cut -d '@' -f2 | cut -d ':' -f2 | cut -d '/' -f1)
+              DB_URL_PORT=$(echo $DSN | cut -d '@' -f2 | cut -d '/' -f 1 )
+
+              if echo "$DB_URL_PORT" | grep -q ':'; then
```
piotrmsc, line 40: could you add just simple comment here for the future, like in which flow how example DSN looks like and why we split. It will be easier for the future in case of debugging  :)

Tomasz-Smelcerz-SAP (approved):

PR #9278, opened 2020-08-19
Cypress POC Bump

Wawrzyn321:
**Description**

- Use Cypress instead of Testcafe
- Remove local cluster console tests override, as it wouldn't work anyway (and isn't needed when we use Cypress)

**Related issue(s)**
[Console](https://github.com/kyma-project/console/pull/1937)

### DIFF
--- a/resources/console/charts/web/values.yaml
+++ b/resources/console/charts/web/values.yaml
@@ -36,14 +36,14 @@ replicaCount: 1
 resources: {}
 test:
   acceptance:
-    enabled: true
+    enabled: false
     ui:
       # tests uses chromium that require more memory
       limits:
-        cpu: 900m
+        cpu: 400m
         memory: 1.5Gi
       requests:
-        cpu: 750m
+        cpu: 350m
         memory: 900Mi
       service_catalog:
         enabled: true
--- a/resources/console/charts/web/values.yaml
+++ b/resources/console/charts/web/values.yaml
@@ -40,10 +40,10 @@ test:
     ui:
       # tests uses chromium that require more memory
       limits:
-        cpu: 400m
+        cpu: 450m
         memory: 1.5Gi
       requests:
-        cpu: 350m
+        cpu: 380m
         memory: 900Mi
       service_catalog:
         enabled: true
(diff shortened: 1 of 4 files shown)


Wawrzyn321:
pre-master-kyma-gke-integration, pre-master-kyma-gke-upgrade passed, pre-master-kyma-integration timed out with console-web running.

Wawrzyn321:
even with higher cpu, `console-web` suffocated other tests.

Wawrzyn321:
now with reorganized tests structure.

Wawrzyn321:
all console-web passed. Excellent. Moving on with more tests.

Wawrzyn321:
magnificent.

Wawrzyn321:
even _compass-integration_ has passed. Let's go again.

Wawrzyn321:
it turns out bumping the memory was the culprit of timing out tests... I assumed they are failing because of timeout, but it they were failing because of cluster addons test (at least after removing memory bump changes). I'll try upping the Cypress timeout limit from 10s to 30s.

Wawrzyn321:
Even though, tests on kyma integration are failing, sometimes randomly. I think we'd better of disabling them anyway.

Wawrzyn321:
ohboyohboy

resources/console/charts/web/values.yaml
```diff
@@ -36,7 +36,7 @@ replicaCount: 1
 resources: {}
 test:
   acceptance:
-    enabled: false
+    enabled: true
```
parostatkiem, line 39: We could gain more stability with setting this to `false`...

parostatkiem (approved):

resources/console/charts/web/values.yaml
Wawrzyn321: It would be the same stability you gain by adding child's training wheels to a motorcycle...

magicmatatjahu (approved):

Wawrzyn321:
> THROUGH THE DARKNESS OF TESTCAFE'S PAST
> THE HASSELHOFF LONGS TO SEE
> ONE CHANTS OUT BETWEEN TWO COMMITS
> CYPRESS WALK WITH ME.

PR #9283, opened 2020-08-20
Make loki and jaeger datasource configurable

rakesh-garimella:
**Description**

- We should keep additional datasource in for grafana and loki configurable.
   So that we can control them via overrides if on of the component is not present

**Related issue(s)**

### DIFF
--- a/resources/monitoring/templates/grafana/configmaps-datasources.yaml
+++ b/resources/monitoring/templates/grafana/configmaps-datasources.yaml
@@ -32,7 +32,10 @@ data:
 {{- end }}
 {{- end }}
 {{- end }}
-{{- if .Values.grafana.additionalDataSources }}
-{{ toYaml .Values.grafana.additionalDataSources | indent 4}}
+{{- if .Values.grafana.additionalDataSources.LokiEnabled }}
+{{ toYaml .Values.grafana.additionalDataSources.LokiConfig | indent 4}}
+{{- end }}
+{{- if .Values.grafana.additionalDataSources.JaegerEnabled }}
+{{ toYaml .Values.grafana.additionalDataSources.JaegerConfig | indent 4}}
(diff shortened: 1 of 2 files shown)


resources/monitoring/templates/grafana/configmaps-datasources.yaml
```diff
@@ -32,7 +32,10 @@ data:
 {{- end }}
 {{- end }}
 {{- end }}
```
a-thaler, line 34: With that you destroy the idea of having additionalDatasources configurable in a flexible way. I suggest to just add full new entries and keep the additionalDataSources for the customer as it is:
```
{{- if .Values.grafana.dataSources.loki.enabled }}
{{ toYaml .Values.grafana.dataSources.loki.config | indent 4}}
{{- end }}
{{- if .Values.grafana.dataSources.jaeger.enabled }}
{{ toYaml .Values.grafana.dataSources.jaeger.config | indent 4}}
{{- end }}
{{- if .Values.grafana.additionalDataSources }}
{{ toYaml .Values.grafana.additionalDataSources | indent 4}}
{{- end }}

File: resources/monitoring/values.yaml [hunk 2]
```diff
@@ -516,18 +516,22 @@ grafana:
...
-      url: http://logging-loki.kyma-system:3100
-      editable: true
-      jsonData:
-        maxLines: 1000
-    - name: Jaeger
-      type: jaeger
-      access: proxy
-      url: http://tracing-jaeger-query.kyma-system:16686
-      editable: true
+    LokiEnabled: true
```
a-thaler, line 519: See comment above, should be a separate key outside of additionalDataSources.
However, you should use keys with first letter lowerCase and use nested structure like in:
loki:
  enabled: true
  config:
     ...

PR #9287, opened 2020-08-24
increase limits/requests.memory in */serverless/values.yaml

lauraanddola:
**Description**
serverless pod sometimes reported OOMKilled alert, want to fix it by increasing tis limits/requests.memory and cpu.


lauraanddola:
pls ignore this commit.

PR #9338, opened 2020-08-31
Configure liveness probe for dex

ebensom:
**Description**
Liveness probe is configured to mitigate the broken connectivity (through istio-proxy and through the Gardener VPN) issue of the HTTP client in dex [which is kept in open state](https://github.com/kyma-incubator/dex/blob/kyma-master/storage/kubernetes/client.go#L152-L157).

The default settings are tuned in accordance with the dex async health check params (15 seconds period, http.Client.Timeout = 15 seconds)

### DIFF
--- a/resources/dex/Chart.yaml
+++ b/resources/dex/Chart.yaml
@@ -1,5 +1,5 @@
 apiVersion: v1
-appVersion: "59c6ec8f"
+appVersion: "7d6c1cb1"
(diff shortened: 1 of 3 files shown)


Demonsthere (approved):
/meow

Tomasz-Smelcerz-SAP (approved):

PR #9357, opened 2020-09-03
Lower Prometheus query timeout

hisarbalik:
**Description**
Current query time out is too high, this PR reduce query time out to 30 seconds.

New value testet on weekly cluster to ensure prometheus has enough data collected and few complex queries fired on this data.

Queries executed in parallel and repeated every 5 seconds over 2 hours. In this time Prometheus query execution time measured max 2.5 seconds.

Following screenshoot show engine query duration in seconds, with following query ```prometheus_engine_query_duration_seconds{quantile="0.99"}```

query_timeout

### DIFF
--- a/resources/monitoring/values.yaml
+++ b/resources/monitoring/values.yaml
@@ -1552,7 +1552,7 @@ prometheus:
     ##
     query:
       maxConcurrency: 100
-      timeout: 1m
+      timeout: 30s
 
     ## Namespaces to be selected for PrometheusRules discovery.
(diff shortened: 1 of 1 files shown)


rakesh-garimella (approved):

rakesh-garimella:
Does it make sense to decrease it further than 30s ? Since max it took is 2.5 so 20s may be ?

hisarbalik:
> Does it make sense to decrease it further than 30s ? Since max it took is 2.5 so 20s may be ?

We have allowed 100 parallel queries same time, I think under load and high data conditions 30 sec is fine

PR #9362, opened 2020-09-03
Remove Knative Serving Charts

anishj0shi:
Remove knative-serving chart and knative-serving test

other references to knative-serving will be tackled with different PRs


resources/cluster-essentials/files/certificates.networking.internal.knative.dev.crd.yaml
```diff
@@ -1,31 +0,0 @@
-apiVersion: apiextensions.k8s.io/v1beta1
-kind: CustomResourceDefinition
-metadata:
-  annotations:
-    "helm.sh/resource-policy": keep
```
magicmatatjahu, line 5: `"helm.sh/resource-policy": keep` annotation tells Helm that resource should stay on cluster, but it's will unpinned from Helm release, so you should create upgrade job and remove whole crds related to `knative-serving`.

k15r: we will create documentation how to cleanup the cluster after knative serving was removed. This guarantees that customers who used knative serving for their own functionality will not lose any configuration.

alexandra-simeonova (approved):

Demonsthere (dismissed):
Please keep in mind that deleting helm resources this way introduces a problem: The SA you created for the job will persist the installation (as it does not have a succeed phase for the deletion hook), which someone could use to delete arbitrary helm releases. 
@pbochynski Your insight might be needed here

sayanh (approved):

resources/cluster-essentials/files/certificates.networking.internal.knative.dev.crd.yaml
nachtmaar: Migration guide can be found here: https://github.com/kyma-project/kyma/pull/9487#pullrequestreview-493484269

marcobebway (approved):

suleymanakbas91 (approved):

pbochynski (approved):

alexandra-simeonova (approved):

piotrmsc (approved):

PR #8898, opened 2020-06-25
Install Kyma with Istio 1.5

Tomasz-Smelcerz-SAP:
**Description**

Install Kyma with istio 1.5

Changes proposed in this pull request:

- Istio upgraded to v1.5.10

**Related issue(s)**
See also #8868 
Depends on: 
- https://github.com/kyma-project/kyma/issues/9094
- https://github.com/kyma-project/kyma/issues/9104
- https://github.com/kyma-project/kyma/issues/9105
- https://github.com/kyma-project/kyma/issues/9101

PR #9393, opened 2020-09-08
Fix mon test for istio 1.15

rakesh-garimella:
**Description**

- With istio-1.5 we are using pod monitors to fetch metrics from the istio-proxy pods. We should ignore the targets from the istio-proxy side car pods for the tests.

**Related issue(s)**


clebs (approved):

clebs (approved):

rakesh-garimella:
Related to istio upgrade: https://github.com/kyma-project/kyma/pull/8898

PR #9436, opened 2020-09-14
make matchExpression to support multiple lines

lauraanddola:
Need to make the `_match_re_` to match multiple conditions(e.g. if slack belongs to default1/2/3, then send to slack otherwise ignore it)....
```
  - receiver: "slack"
    continue: true # If set to `false`, it stops after the first matching.
    match_re:
        severity: critical
        namespace: ^(default1|default2|default3)*$
```


rakesh-garimella:
Hi @lauraanddola  can you please provide an example of override to how to pass this value ?

rakesh-garimella (approved):

PR #9467, opened 2020-09-16
Add few more items to the ignore list

rakesh-garimella:
**Description**

- We should ignore the namesapce with test and also pods having name test during the upgrade process. As these pods would be deleted eventually. 

**Related issue(s)**

### DIFF
--- a/tests/end-to-end/upgrade/pkg/tests/monitoring/targets_rules.go
+++ b/tests/end-to-end/upgrade/pkg/tests/monitoring/targets_rules.go
@@ -197,13 +197,13 @@ func (t TargetsAndRulesTest) testTargetsAreHealthy() error {
 }
 
 func shouldIgnoreTarget(target prom.Labels) bool {
-	var jobsToBeIgnored = []string{
+	jobsToBeIgnored := []string{
 		// Note: These targets will be tested here: https://github.com/kyma-project/kyma/issues/6457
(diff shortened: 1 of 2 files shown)


tests/end-to-end/upgrade/pkg/tests/monitoring/targets_rules.go
```diff
@@ -204,11 +203,33 @@ func shouldIgnoreTarget(target prom.Labels) bool {
...
+		}
+	}
+
 	for _, j := range jobsToBeIgnored {
 		if target["job"] == j {
 			return true
 		}
 	}
+
+	if strings.Contains(target["namespace"], namespacesToIgnore) {
```
clebs, line 229: Shouldn't this be either `strings.EqualFold` (for case insensitive) or just `target["namespace"] == namespacesToIgnore`?

The current code would return true in the following scenario, which probably is not correct:
```go
target["namespace"] = "testsFromCustomer"
namespacesToIgnore = "test"
```

clebs (dismissed):

clebs (approved):

k15r (approved):
could you kindly create a followup ticket so that we can tackle such exclusions in a more generic, less error-prone way? (e.g using labels)

rakesh-garimella:
Created a followup issue: https://github.com/kyma-project/kyma/issues/9469

Issue #9089, opened 2020-07-22
Adapt unhealthy target test to also consider 0/0 targets as unhealthy


rakesh-garimella:
**Description**

We currently have test to check the unhealthy targets. But due to some missing labels there can arise a situation where we have 0/0 targets. Which would mean we dont have unhealthy targets although it would mean the metrics are not being fetched by the service monitor. To avoid such scenarios we should extend the unhealthy target test to also check for 0/0 available targets.

![Screen Shot 2020-07-21 at 6 06 26 PM](https://user-images.githubusercontent.com/26568757/88522703-0c646800-cff7-11ea-9ab9-42e9376a37d2.png)

Hint: We should check for available service monitors and then check if they are up or not.

**Expected result**
We are able to identify 0/0 service monitor targets

**Actual result**

Currently 0/0 is assumed to be healthy targets which is wrong.

rakesh-garimella:
When the test is automated we should ignore the servicemonitors created by kiali operator and jaeger operator as these are created automatically and we cannot disable their creation.

PR #9481, opened 2020-09-18
Checking that all scrape pools have active targets

shorim:
**Description**

- build a list of scrape pools based on the available service monitors and pod monitors
- make sure that we don't have any scrape pool with 0 active targets in monitoring integration and upgrade test

**Related issue(s)**

#9089

### DIFF
--- a/tests/integration/monitoring/test-monitoring.go
+++ b/tests/integration/monitoring/test-monitoring.go
@@ -34,9 +34,6 @@ var k8sClient *kubernetes.Clientset
 var monitoringClient *monitoringv1.MonitoringV1Client
 var httpClient *http.Client
 
-// void struct is used to simulate a set data structure using a map
-type void struct{}
-
 func main() {
 	kubeConfig = loadKubeConfigOrDie()
 	k8sClient = kubernetes.NewForConfigOrDie(kubeConfig)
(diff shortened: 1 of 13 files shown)


shorim:
During the execution of the `pre-master-kyma-gke-compass-integration`prow job, `monitoring-prometheus-pushgateway` service monitor is created. However, there are no active targets for the scrape
 pool `kyma-system/monitoring-prometheus-pushgateway/0` which is causing the new monitoring test in this PR to fail.

targets_during_pre-master-kyma-gke-compass-integration

tests/integration/monitoring/test-monitoring.go
```diff
@@ -31,15 +31,21 @@ const expectedGrafanaInstance = 1
 
 var kubeConfig *rest.Config
 var k8sClient *kubernetes.Clientset
+var monitoringClient *monitoringv1.MonitoringV1Client
 var httpClient *http.Client
 
+// void struct is used to simulate a set data structure using a map
+type void struct{}
```
clebs, line 38: Is there a reason to have this type? There is no need for a type to represent/declare an empty struct.

clebs (changes requested):

tests/integration/monitoring/test-monitoring.go
shorim: I wanted to have a `set` to store scrape pools inside it, so I'm using a `map` to imitate the behavior of a `set` similar to the 2nd idea here https://yourbasic.org/golang/implement-set/

I'm using the `void` struct here https://github.com/kyma-project/kyma/pull/9481/files#diff-94952d952328e43b83c847ec008d2744R271 and I should have used it in the declaration here as well https://github.com/kyma-project/kyma/pull/9481/files#diff-94952d952328e43b83c847ec008d2744R260, but I missed that one out.

Do you think it would be better to remove the `void` type and replace `scrapePools[scrapePool] = void{}` with `scrapePools[scrapePool] = struct{}{}`?

clebs: Yes, it would be better to replace `void` with `struct{}{}`. The first one adds unnecessary complexity and indirection, whereas the second is standard and explicit.

clebs (approved):

piotrmiskiewicz (approved):
lgtm

clebs (approved):

PR #9504, opened 2020-09-22
Fix dockerfile for event-publisher-proxy

sayanh:
**Description**

- Fix dockerfile for event-publisher-proxy


nachtmaar (approved):
LGTM

PR #9524, opened 2020-09-24
Fix git ssh auth validation in gitrepository cr

### LINKED ISSUE #9383: Misleading error message for Git Functions when using ssh
**Description**

In new UI (not yet released) you can supply url to repo, which should be address to git repo with source code of function. Unfortunately in function-controller we do not explicitly say, that when we are using ssh (e.g. `git@github.com:kyma-project/kyma.git`) then we need to also provide a token, otherwise it bugs with following error 
image

**Expected result**
It doesn't allow me to create such a repository (preferably via webhook), or somehow informs me what is the cause of this error

**Actual result**
It gives me misleading info

**Steps to reproduce**
Provide any repo with function via ssh link with no credentials

**Troubleshooting**

dbadura:
**Description**

- add additional validation in case of using ssh in gitrepository.

**Related issue(s)**

fix #9383

### DIFF
--- a/components/function-controller/pkg/apis/serverless/v1alpha1/gitrepository_validation.go
+++ b/components/function-controller/pkg/apis/serverless/v1alpha1/gitrepository_validation.go
@@ -45,7 +45,8 @@ func (in *RepositoryAuth) validateSSHAuth(path string) *apis.FieldError {
 	}
 	var err *apis.FieldError
 	if in.Type != RepositoryAuthSSHKey {
-		err = apis.ErrInvalidValue(in.Type, fmt.Sprintf("%s.type", path))
+		err = apis.ErrGeneric(fmt.Sprintf("invalid value for git ssh, expected %s, current: %s",
+			RepositoryAuthSSHKey, in.Type), fmt.Sprintf("%s.type", path))
 	}
 
 	return err.Also(validateIfMissingFields(property{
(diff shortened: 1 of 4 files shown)


components/function-controller/pkg/apis/serverless/v1alpha1/gitrepository_validation.go
```diff
@@ -33,3 +38,27 @@ func (in *RepositoryAuth) validate(path string) *apis.FieldError {
...
+		err = apis.ErrGeneric(fmt.Sprintf("invalid value for git ssh, expected %s, current: %s",
+			RepositoryAuthSSHKey, in.Type), fmt.Sprintf("%s.type", path))
+	}
+
+	return err.Also(validateIfMissingFields(property{
+		name:  fmt.Sprintf("%s.secretName", path),
+		value: in.SecretName,
+	}))
+}
+func isRepoURLIsSSH(repoURL string) bool {
```
aerfio, line 57: Could you please write some more tests for that regexp? In first glance it looks kinda complicated

aerfio (changes requested):

pPrecel (approved):
We should add more tests for the magic regexp in the `isRepoURLIsSSH` function :D But it looks great! Thanks for your contribution. 

Also, for me adding this validation to the webhook is a good option, but maybe someone has something against or better ideas? For example @m00g3n

components/function-controller/pkg/apis/serverless/v1alpha1/gitrepository_validation.go
dbadura: Done

aerfio (approved):
LGTM now

PR #9525, opened 2020-09-24
Remove EMS prefix from env var in event-publisher-proxy

nachtmaar:
**Description**

- Remove EMS prefix from env variables

**Related issue(s)**


sayanh:
Another mention of EMS: https://github.com/kyma-project/kyma/blob/master/components/event-publisher-proxy/pkg/handler/handler.go#L64

PR #9619, opened 2020-10-02
Remove dummy NOTES files

pbochynski:
**Description**

Remove files that clutter the helm installation output but do not bring any values. Only few Kyma charts introduce NOTES.txt, so it doesn't make sense to keep them.

With this PR merged I can remove another 4 lines there:
https://github.com/kyma-incubator/local-kyma-k3d/blob/47073aab63775138ca77fadd2f6561291f1fc70c/kyma-k3d.sh#L44-L47


strekm (approved):
/meow

PK85 (approved):

PR #9620, opened 2020-10-05
Istio/upgrade rbac fix

Tomasz-Smelcerz-SAP:
**Description**

Extend Istio RBAC rules to provide support for old K8s version: 1.15.x
In this version deployments, replicasets etc. are by default considered to be in `extensions` api group.
This is not required in K8s versions >= 1.16.x, where the mentioned resources are by default served by `apps` api group.

**Related issue(s)**
See also: #9622


strekm (approved):
/meow

Demonsthere (approved):
/meow

PR #9625, opened 2020-10-06
Upgrade uaa-activator SI and SB creator

jasiu001:
During the execution of `uaa-activator` could happen following errors, one is a consequence of the other.

In general, `uaa-activator` consists of several steps. The third step is responsible for creating ServiceBinding (SB). ServiceBinding is provisioned by ServiceCatalog (SC),
Once the process of creating the SB is complete, a new secret related to SB should be showing up.
SC does not ensure the creation of a secret, even if SB is ready/true. SC can show in logs Secret is triggered to create:
```
Sep 16, 2020 @ 20:02:53.427	I0916 18:02:53.427360       1 controller_binding.go:540] ServiceBinding "kyma-system/uaa-issuer-secret" v12239: Creating/updating Secret "kyma-system/uaa-issuer-secret" with 14 keys
Sep 16, 2020 @ 20:02:53.434	I0916 18:02:53.434422       1 controller_binding.go:698] ServiceBinding "kyma-system/uaa-issuer-secret" v12239: Injected bind result
```
but there may be some delays in creating the secret. `uaa-activator` does not check if secret already exists and immediately goes to the next step where secret is required.
This results in an error:
```
while rendering uaa connector config: while fetching Secret kyma-system/uaa-issuer-secret: secrets "uaa-issuer-secret" not found
```

If the first run fails, the process is repeated. Then `uaa-activator` could fail in the second step (where ServiceInstance (SI) is created) with error:
```
Application with xsappname <name_depend_on_cluster_host> already exists. To create a new service instance, ensure that the xsappname specified in your application's xs-security.json file together with the selected service plan of the UAA service broker lead to a new appid. To update an existing service instance, use the update-service command instead.;
```
for some reason `ApiError` [IsAlreadyExists](https://github.com/kyma-project/kyma/blob/master/components/uaa-activator/internal/uaa/creator.go#L38) does not return true (or error is nil), new SI is created, which causes the above error to occur.

PR ensures a new Secret exists when new SB is created and removes failed SI before creating new one.

### DIFF
--- a/components/uaa-activator/internal/uaa/creator.go
+++ b/components/uaa-activator/internal/uaa/creator.go
@@ -7,6 +7,7 @@ import (
 
 	"github.com/kubernetes-sigs/service-catalog/pkg/apis/servicecatalog/v1beta1"
 	"github.com/pkg/errors"
+	v1 "k8s.io/api/core/v1"
 	apiErrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/util/retry"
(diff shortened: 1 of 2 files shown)


crabtree (approved):
Looks good but I have one overall concern. Are we certain that function that repeats until success will always eventually reach the success state?

jasiu001:
Under the hood, `UntilSuccess` method implements `PollImmediate` method from `k8s.io/apimachinery/pkg/util/wait` with timeout set in `GLOBAL_REPEAT_TIMEOUT` environment in `uaa-activator` job.

PR #9629, opened 2020-10-06
Change the tracing override yaml as per new structure is istio-1.5

rakesh-garimella:
**Description**

- Adjust the override yaml so that that value is applied properly

**Related issue(s)**


rakesh-garimella:
Closing this PR as it needs to be created against 1.16

PR #9638, opened 2020-10-06
update alpine-kubectl to have upgraded alpine packages

a-thaler:
**Description**

- update alpine-kubectl image to have upgraded alpine packes
- ...
- ...

**Related issue(s)**

https://github.com/kyma-project/test-infra/pull/2824


a-thaler:
The version of the image is causing problems with the memory footprint of the application-operator. It turned out that latest version is bundling prow-tools. We need to fix that properly first, will close PR

PR #9640, opened 2020-10-06
Fix comparing valueFrom in function-controller

aerfio:
**Description**

- as in title

Comparing string representation should be enough, see https://raw.githubusercontent.com/kubernetes/api/master/core/v1/generated.pb.go -> search for `func (this *EnvVarSource) String() string {`

**Related issue(s)**

### DIFF
--- a/components/function-controller/internal/controllers/serverless/function_controller.go
+++ b/components/function-controller/internal/controllers/serverless/function_controller.go
@@ -48,7 +48,9 @@ func (r *FunctionReconciler) envsEqual(existing, expected []corev1.EnvVar) bool 
 	for key, value := range existing {
 		expectedValue := expected[key]
 
-		if expectedValue.Name != value.Name || expectedValue.Value != value.Value || expectedValue.ValueFrom != value.ValueFrom { // valueFrom check is by reference
+
+
+		if expectedValue.Name != value.Name || expectedValue.Value != value.Value || expectedValue.ValueFrom.String() != value.ValueFrom.String() { // valueFrom check is by string representation
 			return false
 		}
 	}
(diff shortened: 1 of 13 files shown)


components/function-controller/internal/controllers/serverless/function_controller.go
```diff
@@ -48,7 +48,9 @@ func (r *FunctionReconciler) envsEqual(existing, expected []corev1.EnvVar) bool
 	for key, value := range existing {
 		expectedValue := expected[key]
 
-		if expectedValue.Name != value.Name || expectedValue.Value != value.Value || expectedValue.ValueFrom != value.ValueFrom { // valueFrom check is by reference
+
+
+		if expectedValue.Name != value.Name || expectedValue.Value != value.Value || expectedValue.ValueFrom.String() != value.ValueFrom.String() { // valueFrom check is by string representation
```
magicmatatjahu, line 53: If think that before string comparison, we should check existing of pointers.

magicmatatjahu (changes requested):
Please also add to `buildDeployment` method these lines:

```go
	envs := make([]corev1.EnvVar, 0)
	envs = append(envs, envVarsForDeployment...)
	envs = append(envs, rtmConfig.RuntimeEnvs...)
	envs = append(envs, instance.Spec.Env...)
```

Why? Fn-controller adds default envs at the end of array, and envs from user at the beginning, so user cannot redefine envs. See https://github.com/kyma-project/kyma/blob/master/components/function-controller/internal/controllers/serverless/build_resources.go#L331

components/function-controller/internal/controllers/serverless/function_controller.go
aerfio: it's already done by String() method

magicmatatjahu (dismissed):

magicmatatjahu (dismissed):

m00g3n (approved):

PR #9685, opened 2020-10-13
Use images from gcr repository in Istio components

kubadz:
**Description**

- use images from gcr repository in Istio components

**Related issue(s)**


Demonsthere (approved):
/meow

PR #9717, opened 2020-10-16
DO NOT MERGE!! - Kyma 1.16 Istio policy migration fixes - DO NOT MERGE!!

nachtmaar:
**Description**

This PR should not be merge into the release branch. Instead a new PR will be created which will contain these changes.

- Kyma 1.16 Istio policy migration fixes


resources/event-sources/post-upgrade.sh
```diff
@@ -24,4 +24,10 @@ function delete_orphaned_ksvcs() {
   done
 }
 
+
+function delete_orphaned_policies() {
+    kubectl delete policy -n kyma-integration -l application-name
```
sayanh, line 29: I think we'd need `--ignore-not-found` to make it robust

nachtmaar:
```suggestion
    kubectl delete policy --ignore-not-found -n kyma-integration -l application-name
```

PR #9733, opened 2020-10-20
downgrade etcd for helm broker

piotrmiskiewicz:
**Description**

- downgrade etc to 3.3.x due to security issues

**Related issue(s)**


polskikiel (approved):
don't we also want to downgrade the etcd client library in `go.mod`?

PR #9752, opened 2020-10-23
Changes in Docker registry docs

kazydek:
**Description**

- Updated the note with the recommendation to use an external registry for prod.
- Added info on the default PVC size.
- Added the point to the security doc clearly stating that admins have access to Secrets (used by build Jobs to pull and push Function images) in all non-system namespaces.


majakurcius (approved):

pPrecel (approved):
/meow

PR #9793, opened 2020-10-29
Implement mutation handler for bindings

polskikiel:
**Description**

- Implemented mutation handler for bindings

**Related issue(s)**

### DIFF
--- a/components/binding/cmd/manager/main.go
+++ b/components/binding/cmd/manager/main.go
@@ -3,6 +3,7 @@ package main
 import (
 	"github.com/kyma-project/kyma/components/binding/internal/controller"
 	"github.com/kyma-project/kyma/components/binding/internal/webhook"
+	"github.com/kyma-project/kyma/components/binding/internal/webhook/binding"
 	bindingsv1alpha1 "github.com/kyma-project/kyma/components/binding/pkg/apis/v1alpha1"
 
 	log "github.com/sirupsen/logrus"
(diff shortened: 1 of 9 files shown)


components/binding/cmd/manager/main.go
```diff
@@ -54,6 +55,9 @@ func main() {
 	mgr.GetWebhookServer().Register(
 		"/binding-mutating-pod",
 		&k8sWebhook.Admission{Handler: webhook.NewPodHandler(mgr.GetClient(), log.WithField("webhook", "Pod"))})
+	mgr.GetWebhookServer().Register(
+		"/binding-mutating",
+		&k8sWebhook.Admission{Handler: binding.NewMutationHandler(mgr.GetClient(), log.WithField("webhook", "binding-mutating"))})
```
jasiu001, line 60: let the names be consistent, maybe
`Pod-mutating`
`Binding-mutating`?

components/binding/internal/webhook/binding/mutate.go
```diff
@@ -0,0 +1,65 @@
...
+	return admission.PatchResponseFromRaw(req.Object.Raw, rawBinding)
+}
+
+func (h *MutationHandler) InjectDecoder(d *admission.Decoder) error {
+	h.decoder = d
+	return nil
+}
+
+func (h *MutationHandler) mutateBinding(ctx context.Context, binding *v1alpha1.Binding) error {
+	binding.Finalizers = []string{v1alpha1.BindingFinalizer}
```
jasiu001, line 63: I know it is our type but IMO good practice is to check if some finalizer already exist and add a new one instead override the old one

components/binding/internal/webhook/pod_webhook.go
```diff
@@ -4,6 +4,7 @@ import (
 	"context"
```
jasiu001, line 4: maybe be consist, let's move it to `/webhook/pod/mutate.go`? and then `matcher_gvc.go` will be in `/webhook/` instead in `common`, what do you think?

jasiu001 (approved):

jasiu001 (approved):

PR #9805, opened 2020-10-30
Bump Compass Runtime Agent and end to end tests

franpog859:
**Description**

- Bump Compass Runtime Agent and end to end tests

### DIFF
--- a/resources/compass-runtime-agent/values.yaml
+++ b/resources/compass-runtime-agent/values.yaml
@@ -3,9 +3,9 @@ global:
     containerRegistry:
       path: eu.gcr.io/kyma-project
     runtimeAgent:
-      version: "470796a1"
+      version: "c0341336"
     runtimeAgentTests:
-      version: "470796a1"
+      version: "cda9f380"
     compassExternalSolutionTests:
       version: "1a5b13b8"
   istio:
--- a/resources/compass-runtime-agent/values.yaml
+++ b/resources/compass-runtime-agent/values.yaml
@@ -3,9 +3,9 @@ global:
     containerRegistry:
       path: eu.gcr.io/kyma-project
     runtimeAgent:
-      version: "PR-9593"
+      version: "470796a1"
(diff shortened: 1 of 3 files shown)


Maladie (dismissed):

tests/end-to-end/external-solution-integration/chart/external-solution/values.yaml
```diff
@@ -3,7 +3,7 @@ containerRegistry:
 
 image:
   dir: ""
-  version: PR-9354
```
k15r, line 6: PR-9354 was merged in commit 415d68866
would it not make sense to use the image created by this commit?

resources/compass-runtime-agent/values.yaml
```diff
@@ -3,9 +3,9 @@ global:
     containerRegistry:
       path: eu.gcr.io/kyma-project
     runtimeAgent:
-      version: "PR-9593"
+      version: "470796a1"
     runtimeAgentTests:
-      version: "PR-9569"
```
k15r, line 8: PR-9569 was merged in commit cda9f3805

resources/compass-runtime-agent/values.yaml
```diff
@@ -3,9 +3,9 @@ global:
     containerRegistry:
       path: eu.gcr.io/kyma-project
     runtimeAgent:
-      version: "PR-9593"
```
k15r, line 6: PR-9593 was merged in commit c03413362

k15r (changes requested):

tests/end-to-end/external-solution-integration/chart/external-solution/values.yaml
franpog859: Done! It should make more sense right now ^^ Do you know maybe if we can check this information only from the Container Registry side?

k15r (approved):

Maladie (approved):

PR #9810, opened 2020-10-30
Add alert rule when one or more targets for job has scrape limit exce…

rakesh-garimella:
**Description**

- We have scrape sample limits in place for pod monitors. Whenever one of the target reaches the scrape sample limits then alert should be fired as this means the metrics from now on would be dropped.

**Related issue(s)**

### DIFF
--- a/resources/monitoring/templates/prometheus/kyma-rules/kyma-rules.yaml
+++ b/resources/monitoring/templates/prometheus/kyma-rules/kyma-rules.yaml
@@ -81,7 +81,7 @@ spec:
       annotations:
         description: One or more targets in the job monitoring-prometheus-istio-server have scrape limit exceeded
         summary: Scrape limit for one or more targets is exceeded.
-      expr: increase(prometheus_target_scrapes_exceeded_sample_limit_total{job="monitoring-prometheus-istio-server"}[1m]) > 1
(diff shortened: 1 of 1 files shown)


resources/monitoring/templates/prometheus/kyma-rules/kyma-rules.yaml
```diff
@@ -75,3 +75,13 @@ spec:
         severity: warning
       annotations:
         message: Pod {{`{{`}} $labels.namespace {{`}}`}}/{{`{{`}} $labels.pod {{`}}`}} ({{`{{`}} $labels.container {{`}}`}}) is OOMKilled for 5 minutes.
+  - name: prometheus-rules
+    rules:
+    - alert: ScrapeLimitForTargetExceeded
+      annotations:
+        description: One or more targets in the job monitoring-prometheus-istio-server have scrape limit exceeded
+        summary: Scrape limit for one or more targets is exceeded.
+      expr: increase(prometheus_target_scrapes_exceeded_sample_limit_total{job="monitoring-prometheus-istio-server"}[1m]) > 1
```
a-thaler, line 84: Why the value should be greater than 1? It should fire if it is greater than 0.

rakesh-garimella: fixed.

a-thaler (approved):

Issue #9689, opened 2020-10-13
Optimize observability setup for evaluation profile


a-thaler:
**Description**
in regards to https://github.com/kyma-project/kyma/issues/9673 the observability stack should be optimized in the evaluation profile to provide full feature-richness with lowest price.

AC: for all observability components:
- reduce retention times as much as possible
- remove any persistence
- lower memory and cpu so that it might crash with workloads more than 10 but also recovers
- Decide on a profile implementation mechanism with all the teams.
- Implement the chosen mechanism for the observability charts (either on the charts directly or as extra resource).

**Reasons**
EVakkuation must be cheap

**Attachments**

clebs:
### Findings for Monitoring minimal setup:
The minimal viable setup is as follows:
```yaml
  limits:
    cpu: 600m
    memory: 1500Mi
  requests:
    cpu: 150m
    memory: 1000Mi
...
retention: 2h
retentionSize: 1GB
grafana.persistence.enabled: "false"
```

The profile during 2h of stress test was:
![Screenshot 2020-10-19 at 17 29 54](https://user-images.githubusercontent.com/1059661/96476211-3663aa00-1235-11eb-9abb-3e27dba49a5e.png)

From that we can see that the current profile is the minimum viable since there is a spike at start that comes close to the limits. Should we make it less, prometheus would endlessly restart.

clebs:
### Findings for Tracing

The minimal viable setup is as follows:
```yaml
jaeger-operator:
resources:
  limits:
    cpu: 50m
    memory: 100Mi
  requests:
    cpu: 10m
    memory: 40Mi
(...)
jaeger:
  resources:
      limits:
        cpu: 50m
        memory: 60Mi
      requests:
        cpu: 10m
        memory: 40Mi
  storage:
      type: memory
      options:
        (...)
        memory:
          max-traces: 10000
kcproxy:
  resources:
    limits:
      cpu: 50m
      memory: 20Mi
    requests:
      cpu: 10m
      memory: 5Mi
```

With this setup, jaeger was stable during 4h running the stress tests generating traces and querying with a limit of 1000 traces. This should be more than enough for a trial.

clebs:
### Findings for Logging
The minimal viable setup is as follows:
```yaml
loki:
  resources:
    limits:
      cpu: 100m
      memory: 512Mi
    requests:
      cpu: 50m
      memory: 128Mi
  persistence:
    enabled: false

(...)

fluentbit:
  resources:
[...]

PR #9816, opened 2020-10-30
Add eval & prod profiles for observability

clebs:
**Description**

- Add evaluation and production profiles for observability components.

**Related issue(s)**
#9689

### DIFF
--- a/resources/monitoring/profile-production.yaml
+++ b/resources/monitoring/profile-production.yaml
@@ -1,0 +1,27 @@
+prometheus:
+  prometheusSpec:
+    resources:
+      limits:
+        cpu: 600m
+        memory: 1500Mi
+      requests:
+        cpu: 150m
+        memory: 1000Mi
+    retention: 30d
+    retentionSize: 15GB
+    storageSpec:
+      volumeClaimTemplate:
+        spec:
+          resources:
+            requests:
+              storage: 20Gi
+
+prometheus-istio:
+  server:
+    resources:
+      limits:
+        memory: 3Gi
+
+alertmanager:
+  alertmanagerSpec:
+    retention: 240h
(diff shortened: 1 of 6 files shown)


suleymanakbas91 (approved):

resources/monitoring/profile-production.yaml
```diff
@@ -0,0 +1,27 @@
+prometheus:
+  prometheusSpec:
+    resources:
+      limits:
+        cpu: 600m
+        memory: 1500Mi
```
a-thaler, line 6: Isn't there a mistake?

resources/monitoring/profile-production.yaml
```diff
@@ -0,0 +1,27 @@
...
+        cpu: 150m
+        memory: 1000Mi
+    retention: 30d
+    retentionSize: 15GB
+    storageSpec:
+      volumeClaimTemplate:
+        spec:
+          resources:
+            requests:
+              storage: 20Gi
```
a-thaler, line 17: Shouldn't we reduce the size heavily on evaluation profile?

clebs: that is the production profile. This was copied from:
https://github.com/kyma-project/kyma/blob/19a922844fe8b30d8f251f33dabdebed1bbd75c1/installation/resources/installer-config-production.yaml.tpl#L67

resources/monitoring/profile-production.yaml
clebs: True, should be 3Gi, will fix it.

resources/monitoring/profile-production.yaml
```diff
@@ -0,0 +1,27 @@
+prometheus:
+  prometheusSpec:
+    resources:
+      limits:
+        cpu: 600m
```
a-thaler, line 5: CPU requests and limits are also not fitting, see: https://github.com/kyma-project/kyma/blob/master/installation/resources/installer-config-production.yaml.tpl

clebs: Fixed thanks

Issue #9567, opened 2020-09-28
Support of istio 1.7 in observability components


a-thaler:
**Description**

In https://github.com/kyma-project/kyma/issues/9556 the instio setup will be updated to 1.7. We need to assure that this upgrade works smoothly with the observability stack (loki, prometheus, grafana, kiali, jaeger)

[x] adopt changes done in istio metrics (setup istio with prometheus and compare scrape config to kyma scrape config)
[x] validate that upgrade is not increasing overall memory/disk footprint
[x] validate that a fresh installation has no increased overall memory/disk footprint
[x] validate grafana dashboards, kiali dashboards, jaeger traces
[x] Loki exposure via JWT verification
[x] Sampling rate for the traces in istio needs to be verified if it needs adjustment.
[x] Update readme for tracing overrides
[x] Check connection with application mocks
**Reasons**

**Attachments**

https://github.com/kyma-project/kyma/issues/9556

rakesh-garimella:
[x] adopt changes done in istio metrics (setup istio with prometheus and compare scrape config to kyma scrape config)
  - With istio-1.7 the recommended changes is to use `prometheus.io` annotation into pods that could be used to scrape the pods. But this would not work in our case as we have to move away from `servicemonitors` and `podmonitors`
[ ] validate that upgrade is not increasing overall memory/disk footprint
[ ] validate that a fresh installation has no increased overall memory/disk footprint
[ ] validate grafana dashboards, kiali dashboards, jaeger traces
[ ] Loki exposure via JWT verification

a-thaler:
Don't forget to update the tracing troubleshooting section: https://kyma-project.io/docs/components/tracing/#troubleshooting-troubleshooting

rakesh-garimella:
Currently upgrade on istio-1.7 is broken as currently there is no check to prevent from upgrading from istio-1.7->istio-1.6. (Currently when we upgrade on istio-1.7 it fails as it tries to upgrade from istio-1.7->istio-1.6 and fails.)

rakesh-garimella:
[] validate that upgrade is not increasing overall memory/disk footprint
[] validate that a fresh installation has no increased overall memory/disk footprint
After running fresh istio-1.7 installation with stability checker for ~18h
```
Memory:
prometheus-istio-server: 1.45GB
Prometheus: 1.5GB
CPU:
prometheus-istio-server: 0.16
Prometheus:0.26
TSDB:
prometheus-istio-server: 145k
[...]

PR #9821, opened 2020-11-01
Remove unneeded relabelings to reduce load on prometheus memory

rakesh-garimella:
**Description**

- Remove unneeded relabelings to reduce load on prometheus memory
- Include job in the pod monitor since its a static value.
**Related issue(s)**

#9567

### DIFF
--- a/resources/monitoring/charts/prometheus-istio/values.yaml
+++ b/resources/monitoring/charts/prometheus-istio/values.yaml
@@ -1273,25 +1273,11 @@ serverFiles:
           regex: "15090"
           replacement: $1
           action: keep
-        - source_labels: [__meta_kubernetes_service_name]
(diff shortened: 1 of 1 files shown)


resources/monitoring/charts/prometheus-istio/values.yaml
```diff
@@ -1329,20 +1288,8 @@ serverFiles:
           regex: __meta_kubernetes_pod_label_(.+)
           replacement: $1
           action: labeldrop
-        - source_labels: [__meta_kubernetes_namespace]
```
a-thaler, line 1332: You still need to drop the source label like it is done on the block above, don't you? Have you verified that the source labels are not there?

a-thaler (approved):

Issue #9851, opened 2020-11-03
Install NATS Server


k15r:
As a kyma developer i want to be able to have a NATS server available in the cluster, so that i can send and receive events

Acceptance Criteria:
- [x] helm chart to install NATS using the (NATS operator)[https://github.com/nats-io/k8s/tree/master/helm/charts/nats-operator]
- [x] installation of the server is part of the optional eventing chart
- [x] installation of the server can be enabled/disabled using overrides on this chart

https://github.tools.sap/skydivingtunas/team/blob/master/roadmap.org#create-a-publishing-proxy-for-nats

PR #9864, opened 2020-11-04
Install NATS server

radufa:
**Description**

Install NATS server using Nats-operator

**Related issue(s)**

See also #9851


resources/eventing/charts/nats/values.yaml
```diff
@@ -0,0 +1,12 @@
+# Default values for nats.
+# This is a YAML-formatted file.
+# Declare variables to be passed into your templates.
+global:
```
nachtmaar, line 4: Why do you need a global value ?

> https://helm.sh/docs/chart_template_guide/subcharts_and_globals/
> Global values are values that can be accessed from any chart or subchart by exactly the same name. Globals require explicit declaration. You can't use an existing non-global as if it were a global.

resources/eventing/charts/nats/templates/10-deployment.yaml
```diff
@@ -0,0 +1,50 @@
+{{- if .Values.global.install.enabled }}
+
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: nats-operator
+  # Change to the name of the namespace where to install NATS Operator.
+  # Alternatively, change to "nats-io" to perform a cluster-scoped deployment in supported versions.
+  namespace: {{ .Values.global.namespace }}
```
nachtmaar, line 9: Namespace does not have to be defined. Helm will do this automatically.

resources/eventing/charts/nats/values.yaml
```diff
@@ -0,0 +1,12 @@
...
+  install:
+    enabled: false
+  namespace: nats
```
nachtmaar, line 7: We don't need this as kyma-installer is also passing the "--create-namespace <namespace-name>" when installing the chart.

resources/eventing/charts/nats/Chart.yaml
```diff
@@ -0,0 +1,6 @@
+apiVersion: v2
+name: nats
+description: Helm chart for nats
+type: application
+version: 0.1.0
+appVersion: 0.1.0
```
nachtmaar, line 6: (nit): f you would like to use appVersion, it should have the same version as nats and we should commit to keep it up2date :)

> appVersion: The version of the app that this contains (optional). This needn't be SemVer.
https://helm.sh/docs/topics/charts/

resources/eventing/charts/nats/templates/10-deployment.yaml
```diff
@@ -0,0 +1,50 @@
...
+    metadata:
+      annotations:
+        sidecar.istio.io/inject: "false"
+      labels:
+        name: nats-operator
+    spec:
+      serviceAccountName: nats-operator
+      containers:
+      - name: nats-operator
+        image: connecteverything/nats-operator:0.7.4
```
nachtmaar, line 25: Let's sync this image to the Kyma registry and use it: https://github.com/kyma-project/test-infra/blob/master/development/image-syncer/external-images.yaml

resources/eventing/charts/nats/Chart.yaml
radufa: removed

resources/eventing/charts/nats/README.md
```diff
@@ -0,0 +1,35 @@
+# NATS Chart
+
+This Helm chart deploy NATS: https://nats.io/
+Steps:
+- Create "nats" namespace:
+```bash
+$ kubectl create ns  install nats
```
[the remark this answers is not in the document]
radufa: done

resources/eventing/charts/nats/templates/10-deployment.yaml
radufa: removed

resources/eventing/charts/nats/values.yaml
radufa: It was prepared for the future structure, where "eventing" will have a hierarchy of charts.
Changed for now to be "local"

resources/eventing/charts/nats/values.yaml
radufa: It is still needed for "nats-operator-binding".
We will use "--create-namespace" for now, see also README.md.

resources/eventing/charts/nats/templates/10-deployment.yaml
radufa: Will be solved in https://github.com/kyma-project/kyma/issues/9875

resources/eventing/charts/nats/values.yaml
nachtmaar: You can use

 ```
{{ .Release.Namespace }}
```

to get the namespace from Helm dynamically.

File: resources/eventing/charts/nats/crds/customeresourcedefinition.yaml [hunk 7]
```diff
@@ -0,0 +1,35 @@
+apiVersion: apiextensions.k8s.io/v1beta1
```
nachtmaar, line 1: Let's also put the CRDs behind the feature flag ( I mean `install.enabled`).

resources/eventing/charts/nats/templates/00-prereqs.yaml
```diff
@@ -0,0 +1,130 @@
+{{- if .Values.install.enabled }}
+
+---
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: nats-operator
+  # Change to the name of the namespace where to install NATS Operator.
+  # Alternatively, change to "nats-io" to perform a cluster-scoped deployment in supported versions.
+  namespace: {{ .Values.namespace }}
```
nachtmaar, line 10: Helm labels are missing here. This results in an upgrade problem (from eventing without nats => eventing with nats):

```
Error: UPGRADE FAILED: rendered manifests contain a resource that already exists. Unable to continue with update: ServiceAccount "nats-operator" in namespace "nats" exists and cannot be imported into the current release: invalid ownership metadata; label validation error: missing key "app.kubernetes.io/managed-by": must be set to "Helm"; annotation validation error: missing key "meta.helm.sh/release-name": must be set to "eventing"; annotation validation error: missing key "meta.helm.sh/release-namespace": must be set to "eventing"
```

resources/eventing/charts/nats/templates/20-service.yaml
```diff
@@ -0,0 +1,17 @@
+{{- if .Values.install.enabled }}
+
+apiVersion: v1
+kind: Service
+metadata:
+  name: {{ .Values.nats.fullname }}
```
nachtmaar, line 6: (nit) We usually use the helpers functions in the eventing chart to build have a name which consists of the helm chart (`eventing`) and the name of the subchart. What is the intention behind not using it in this chart ? 🤔 

e.g. 
```
{{ include "nats.fullname" . }} # I know this helper does not exist yet
```

That way we would follow the same naming schema

resources/eventing/charts/nats/crds/customeresourcedefinition.yaml
radufa: Right, done

radufa: See: https://helm.sh/docs/chart_best_practices/custom_resource_definitions/
Templating is not supported in CRDs.

resources/eventing/charts/nats/templates/20-service.yaml
radufa: done

resources/eventing/charts/nats/templates/00-prereqs.yaml
radufa: This will be solved in the follower issue:
https://github.com/kyma-project/kyma/issues/9969

marcobebway (dismissed):
Kindly update this readme accordingly: https://github.com/kyma-project/kyma/blob/master/resources/eventing/README.md.

resources/eventing/charts/nats/templates/00-prereqs.yaml
```diff
@@ -0,0 +1,130 @@
+{{- if .Values.install.enabled }}
```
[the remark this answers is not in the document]
radufa: These are the original names from NATS repo. These files are only copied here. For now, it could be better to have the same name and structure as in the NATS repo. Later, we can decide if we want to modify them, which of course will generate some maintenance work for us...

resources/eventing/charts/nats/values.yaml
radufa: True, but I wanted to let it simple now, cause NATS operator has different behavior depending of the namespace.. :
```
  # Change to the name of the namespace where to install NATS Operator.
  # Alternatively, change to "nats-io" to perform a cluster-scoped deployment in supported versions.
```

nachtmaar (approved):

alexandra-simeonova (approved):

Issue #9402, opened 2020-09-09
POC Kyma Binding


PK85:
**Description**
POC about new credential binding mechanism. We'd like to decouple Kyma bindings from Service Catalog domain (currently ServiceBindingUsage) and make them lightweight and simple. 

More or less new CRD _from_ and _to_ sections must be configurable from type point of view. Section _to_ could be reused from SBU. The _from_ section in the first iteration must support secrets, and config maps could be added in the next iterations.

**Reasons**
Current credential binding into application is implemented with help of ServiceBindingUsage CRD. This is tightly coupled with service catalog domain and cannot be used without it. 

**AC**
- check if we have any alternatives to podpreset
- allow user to declare which secret will be injected into function and into deployment 
- allow to adjust environment variables keys (mappings). Design solution.
- create full-blown doc about that feature
- part of the kyma-project organization, kyma repository. Give a possibility to install "bindings" chart only. 
- check if we can removed podpreset CRD from kyma (if any component use it as well)

**Naming**
- api group name: bindings.kyma-project.io
- CRD: bindings.kyma-project.io/Bindings
- CRD: bindings.kyma-project.io/TargetKind

Binding proposal:
```
apiVersion: bindings.kyma-project.io/v1alpha1
kind: Binding
metadata:
 name: deploy-redis-client
spec:
 source:
   kind: secret
   name: redis-instance-credential
 target:
   kind: deployment
   name: redis-client
```

TargetKind proposal:
```
apiVersion: bindings.kyma-project.io/v1alpha1
kind: TargetKind
metadata:
   name: deployment
spec:
   displayName: Deployment
   resource:
     group: apps
     kind: deployment
     version: v1
   labelsPath: spec.template.metadata.labels
```

**Links**
- https://github.com/knative/pkg/tree/master/apis/duck (https://www.youtube.com/watch?v=kldVg63Utuw)

crabtree:
Maybe also worth checking: https://github.com/IBM/composable

jasiu001:
Existing solutions that combine Secret/ConfigMap with other resources:
**1) Knative duck [link](https://github.com/knative/pkg/tree/master/apis/duck)**
- the project is still being developed (last commit a few days ago) but still in the alpha stage
- presentation showed on "kubecon" presents only example with knative service
- very poor documentation, lack of examples. There is no proof it could be used with Secret and could be useful in binding case
[...]

PR #9876, opened 2020-11-05
Sync TargetKinds on startup

polskikiel:
**Description**
#9402 

- Implemented TargetKinds sync on startup

**Related issue(s)**

https://github.com/kyma-project/kyma/issues/9402


components/binding/cmd/manager/main.go
```diff
@@ -93,3 +105,28 @@ func fatalOnError(err error, msg string) {
 		log.Fatalf("%s: %s", msg, err.Error())
 	}
 }
+
+// syncTargetKinds assures that existing TargetKinds will be loaded before reconcilers are started
```
jasiu001, line 109: please move it to a separate file (inside `internal`), let's avoid a situation when the main file grows all the time

components/binding/cmd/manager/main.go
```diff
@@ -49,7 +55,11 @@ func main() {
 		logger.SetLevel(log.WarnLevel)
 	}
 
-	mgr, err := ctrl.NewManager(ctrl.GetConfigOrDie(), ctrl.Options{
+	config := ctrl.GetConfigOrDie()
+	cli, err := client.New(config, client.Options{Scheme: scheme})
```
jasiu001, line 59: do not create new client, manager gives you one:
```go
mgr.GetClient()
```
see e.g. method `controller.SetupTargetKindReconciler`

components/binding/internal/storage/error.go
```diff
@@ -0,0 +1,35 @@
+package storage
```
jasiu001, line 1: maybe should be in `internal/error/error.go`, then we can use it in other places without import package "storage", besides maybe all errors should be in one place, what do you think?

components/binding/internal/worker/targetkind.go
```diff
@@ -87,7 +71,29 @@ func (w *TargetKindWorker) RemoveProcess(targetKind *v1alpha1.TargetKind, log lo
...
+			return errors.Wrapf(err, "while set TargetKind phase to %s", status)
+		}
+	case v1alpha1.TargetKindFailed:
+		err = targetKind.Status.Failed()
+		if err != nil {
+			return errors.Wrapf(err, "while set TargetKind phase to %s", status)
+		}
+	}
+
+	log.Infof("TargetKind %q %s", targetKind.Name, status)
```
jasiu001, line 92: use passed log, there we have fields to grep logs

components/binding/internal/worker/targetkind.go
```diff
@@ -87,7 +71,29 @@ func (w *TargetKindWorker) RemoveProcess(targetKind *v1alpha1.TargetKind, log lo
...
 
-func (w *TargetKindWorker) isDifferentThanRegistered(targetKind *v1alpha1.TargetKind) (bool, error) {
+func (w *TargetKindWorker) register(targetKind *v1alpha1.TargetKind, status string) error {
+	err := w.storage.Register(*targetKind)
+	if err != nil {
+		return errors.Wrapf(err, "while registering TargetKind %q", targetKind.Name)
+	}
+	switch status {
+	case v1alpha1.TargetKindRegistered:
+		err = targetKind.Status.Registered()
```
jasiu001, line 81: we also have a message field in `.status`, it should inform why TK is failed

components/binding/cmd/manager/main.go
polskikiel: Unfortunately, you can't use the manager's client before it started

components/binding/cmd/manager/main.go
polskikiel: didn't find a better place than `/internal/syncer/syncer.go` as a separate struct, now execution looks like:
 ```
err = syncer.NewExecutor(cli, logger).TargetKinds(targetKindWorker)
```

File: components/binding/internal/storage/error.go (hunk 3 above)
Comment (polskikiel) [reply components/binding/internal/storage/error.go:1]: I think i can move that implementation to the error package why not 👍

Comment (polskikiel) [reply components/binding/internal/storage/error.go:1]: I'll also change the name of package to `errors` because error conflicts with the go type name

File: components/binding/internal/worker/targetkind.go (hunk 5 above)
Comment (polskikiel) [reply components/binding/internal/worker/targetkind.go:83]: added message on fail: 

```
Failed because a TargetKind with the same kind and different properties exists
```

jasiu001 (approved):

PR #9923, opened 2020-11-10
Add troubleshooting guide when production profile changes are not applied properly

rakesh-garimella:
**Description**

- If the production profile is applied to an existing running cluster then the change in PVC size is not applied properly. Added a troubleshoot guide/workaround to mitigate the issue.

**Related issue(s)**

### DIFF
--- a/docs/monitoring/05-04-production-profile.md
+++ b/docs/monitoring/05-04-production-profile.md
@@ -110,6 +110,30 @@ You can deploy a Kyma cluster with Monitoring configured to use the production p
     EOF
     ```
   2. Run the [cluster update process](/root/kyma/#installation-update-kyma).
+When the production overrides are applied to an already installed Kyma cluster, then the changes to the storage size of the PVC for Prometheus will not be applied. This is because the underlying Cloud infrastructure might not support dynamic resizing of the PVC. For a workaround, follow these steps:  
+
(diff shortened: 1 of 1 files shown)


suleymanakbas91 (approved):

docs/monitoring/05-04-production-profile.md
```diff
@@ -110,6 +110,29 @@ You can deploy a Kyma cluster with Monitoring configured to use the production p
     EOF
     ```
   2. Run the [cluster update process](/root/kyma/#installation-update-kyma).
+When the production overrides are applied to an already installed Kyma cluster, then the changes to the storage size of the PVC for Prometheus will not be applied. This is because the underlying Cloud infrastructure might not support dynamic resizing of the PVC. For a workaround, follow these steps:
```
alexandra-simeonova, line 113:
```suggestion

When the production overrides are applied to an already installed Kyma cluster, then the changes to the storage size of the PVC for Prometheus will not be applied. This is because the underlying Cloud infrastructure might not support dynamic resizing of the PVC. For a workaround, follow these steps:  
```

One more small thing, there should be an empty line here.

alexandra-simeonova (approved):
One more small thing, there should be an empty line here.

PR #9945, opened 2020-11-12
Implement random xsappname for uaa-activator component

jasiu001:
**Description**

- removes Secret from chart template, instead of this setting for SI to UAA are created in the code
- every run of the job generates a new `Xsappname` parameter value


piotrmiskiewicz (approved):
lgtm

PR #9989, opened 2020-11-19
Allow privileged containers

montaro:
See also #9546


installation/resources/installer-cr-cluster.yaml.tpl
```diff
@@ -9,6 +9,7 @@ metadata:
 spec:
```
montaro, line 9: This file is copied from the master branch, I don't know why it's included here!

marcobebway (approved):

PR #9991, opened 2020-11-20
Run serverless tests as normal application

dbadura:
**Description**

- run test as normal binary, first parameter is the name of test suite.

**Related issue(s)**


resources/serverless/templates/_helper_tests.txt
```diff
@@ -1,6 +1,6 @@
 echo 'TESTING has started'
 sleep 20
-/app/main.test -test.v
+/app/main.test kyma-integration
```
aerfio, line 3: image
Removing that flag will disable logging in time of running a test, why do it? If you want to pass some kind of option to the binary do it as an additional flag or by env

tests/function-controller/Dockerfile
```diff
@@ -15,14 +15,14 @@ RUN GO111MODULE=on go mod download
...
-COPY ./main_test.go ${BASE_APP_DIR}
+COPY ./cmd/ ${BASE_APP_DIR}/cmd/
 COPY ./licenses/ ${BASE_APP_DIR}/licenses/
 
 #
 # Build tests
 #
 
-RUN CGO_ENABLED=0 go test main_test.go -c
+RUN CGO_ENABLED=0 go build -o main.test ./cmd/main.go
```
aerfio, line 25: Go back to the proper way -> right now we're running test suite in that binary, so please compile it as test binary

tests/function-controller/cmd/main.go
```diff
@@ -2,57 +2,70 @@ package main
...
 }
 
-type testSuite func(*rest.Config, testsuite.Config, *logrus.Entry) (step.Step, error)
+func main() {
+	logf := logrus.New()
+	logf.SetFormatter(&logrus.TextFormatter{})
+	logf.SetReportCaller(false)
 
-func runTests(t *testing.T, testFunc testSuite) {
-	rand.Seed(time.Now().UnixNano())
```
aerfio, line 38: Are you sure that there are no things in the whole tests that would need this? I remember that we generated namespace name, right now we basing that on current test time, but maybe we have similar logic somewhere else?

tests/function-controller/cmd/main.go
```diff
@@ -2,57 +2,70 @@ package main
...
+
+	for _, scenario := range pickedScenarios {
+		runScenario(scenario, scenarioName, logf, cfg, restConfig)
+	}
+}
+
+type testSuite func(*rest.Config, testsuite.Config, *logrus.Entry) (step.Step, error)
+
+func runScenario(testFunc testSuite, name string, logf *logrus.Logger, cfg config, restConfig *rest.Config) {
+	rand.Seed(time.Now().UnixNano())
```
aerfio, line 61: ah, you've moved it here - but why?

tests/function-controller/cmd/main.go
```diff
@@ -61,6 +74,8 @@ func loadConfig(prefix string) (config, error) {
 	return cfg, err
 }
 
-func failOnError(g *gomega.GomegaWithT, err error) {
-	g.Expect(err).NotTo(gomega.HaveOccurred())
+func failOnError(err error) {
+	if err != nil {
+		panic(err)
+	}
```
aerfio, line 80: why change it?

tests/function-controller/cmd/main.go
```diff
@@ -2,57 +2,70 @@ package main
...
+	logf.SetFormatter(&logrus.TextFormatter{})
+	logf.SetReportCaller(false)
 
-func runTests(t *testing.T, testFunc testSuite) {
-	rand.Seed(time.Now().UnixNano())
-	g := gomega.NewGomegaWithT(t)
+	if len(os.Args) < 2 {
+		logf.Errorf("Scenario not specified. Specify it as the first argument")
+		os.Exit(2)
+	}
```
aerfio, line 36: it looks kinda clunky, especially given that flags are listed in os.Args, go back to flags or envs
see:
```go
package main

import (
	"flag"
	"fmt"
	"os"
)

func main() {
	data := flag.Bool("test", false, "sample usage")
	flag.Parse()

	for i, item := range os.Args {
		fmt.Printf("Os arg #%d: %s\n", i+1, item)
	}

	fmt.Println(*data)
}
![image](https://user-images.githubusercontent.com/17271979/100081965-ed140500-2e47-11eb-83a9-b17a267c0760.png)

```

aerfio (changes requested):

tests/function-controller/cmd/main.go
dbadura: I can log the error and then use `os.exit(1)`

tests/function-controller/cmd/main.go
aerfio: discussed offline - it stays

aerfio (approved):

aerfio:
Update PR description please

Issue #9673, opened 2020-10-12
Evaluation profile


pbochynski:
**Description**

Provide a way to install Kyma with the evaluation profile, which contains the features but has low resource requirements (CPU & memory). The profile should work with different target platforms (GKE, AKS, Gardener, minikube, k3s or kind).
To achieve low resource consumption some trade-offs are required (e.g. short retention period of logs and metrics). Take istio [demo profile](https://istio.io/latest/docs/setup/additional-setup/config-profiles) as an example.

The profile is specified in the installation CR in the spec:
```
profile: evaluation|production
```
For kyma CLI you can select profile with the flag for install command:
```
kyma install --profile evaluation
```
We can define profile also for the component (future)

Installer can read the profile value and merge overrides with the profile-xxxx.yaml that exists in the component chart root folder (the location can be changed to installation folder if keeping the file in the chart is not recommended - @Tomasz-Smelcerz-SAP to verify)

The profile should address the request CPU, request memory, and autoscaling. 

Subtasks:
- [x] low resource settings for istio (demo with some disabled modules that are installed with monitoring and tracing) #9759 
- [x] low resource settings for ory #9757 
- [x] low resource settings for monitoring (limited metrics retention) https://github.com/kyma-project/kyma/issues/9689
- [x] evaluation profile for serverless - #9738 
- [x] evaluation profile for rafter - #9860
- [x] evaluation profile for console - #9898
- [x] evaluation profile for Application Connector #9930
- [x] evaluation profile is described in the documentation with all the limitations
- [x] provisioned should handle profile parameter https://github.com/kyma-project/control-plane/issues/322
- [x] kyma installer takes profile from installation CR
- [x] Support in CLI https://github.com/kyma-project/cli/issues/640  
- [x] Pipeline installing evaluation profile is green

Stretched:
- [ ] separate resource profile from target platform - #9442
- [ ] cleanup deprecated installation CR files - #8130 

**Reasons**

People want to try Kyma features on different platforms and do not want to pay 300$ per month for hello world examples. The default production profile is too big (expensive). The local profile is bound to minikube only and doesn't contain monitoring and tracing.

kazydek:
[...]

PR #9994, opened 2020-11-20
Evaluation profile for event mesh

berkanercan:
**Description**

- Added evaluation profile resource values to knative-eventing helm chart
- Added evaluation profile resource values to the helm chart of application operation under components/application-operator

These changes can handle two events per second.

*Testing*
A kyma cluster with an external application binded is used to test the resource values. 

**Related issue(s)**
[`See also #9673 `](https://github.com/kyma-project/kyma/issues/9673)


resources/nats-streaming/templates/destination-rule.yaml
```diff
@@ -1,4 +1,4 @@
-apiVersion: networking.istio.io/v1beta1
+apiVersion: networking.istio.io/v1alpha3
```
k15r, line 1: ????

resources/nats-streaming/values.yaml
```diff
@@ -28,7 +28,7 @@ global:
       # If set to "-", storageClassName: "", which disables dynamic provisioning
       # If defined, storageClassName: <storageClass>
       # storageClass: "-"
-      enabled: true
+      enabled: false
```
k15r, line 31: why do we disable this?

resources/nats-streaming/templates/destination-rule.yaml
k15r, line 1: why do we downgrade the apiversion?

berkanercan:
This PR is obsolete.

Issue #10001, opened 2020-11-23
Serverless integration test suite on lightweight cluster (k3s) 


kwiatekus:
**Description**

Implement a minimalistic Serverless TestSuite running on k3s 
 - build function ( various runtimes : https://github.com/kyma-project/kyma/issues/9678 )
 - run function( triggered internally )
 - consumption of injected secrets
 - consumption of config map

Minimalistic test suite means:
 - no istio
 - no eventing
 - serverless in isolation

**Reasons**

The goal is to have a quick  test that proves that basic functionalities work fine  
https://github.com/kyma-project/kyma/issues/10050

kwiatekus:
Testing various runtime images : https://github.com/kyma-project/kyma/issues/9678

PR #10026, opened 2020-11-24
Fix function reconcile datarace with configmaps controller

aerfio:
**Description**

- added needed fix and proper comment for posterity
- datarace is explained in that comment

**Related issue(s)**

https://github.com/kyma-project/kyma/issues/10001


dbadura (dismissed):

aerfio:
After waiting a couple of seconds after starting controller I've not observed this behaviour, so I'll close this for now without merging

Issue #9861, opened 2020-11-04
Connect Publishing Proxy for BEB to application


k15r:
As a kyma user i want to be able to send events from my connected application to the eventing backend, so that i can later consume them in my workload

Acceptance Criteria:
- [x] proxy is installed as an optional component using kyma installer (SKR)
- [x] cloudevents 1.0 are accepted and forwarded to BEB
- [x] events sent from a connected application get forwarded to BEB
- [x] legacy events sent from an application shall be converted to cloudevents 1.0 automatically
- [x] OSS still works using event mesh

berkanercan:
In the application operator, we need a switch to decide which charts to deploy. Option 1 is event mesh: knative eventing charts. Option 2: Event publish proxy; which should be configured with the application validator.

PR #10051, opened 2020-11-26
Connect event publisher proxy to application operator

sayanh:
**Description**

- Based on a global flag, `global.isBEBEnabled`=true/false, validator should deploy pods in `kyma-integration` ns for eventing.
- If BEB is enabled in a cluster, event publisher proxy should be connected to connectivity validator and the necessary wiring should be done by application operator

**Related issue(s)**

See https://github.com/kyma-project/kyma/issues/9861

### DIFF
--- a/components/application-connectivity-validator/internal/validationproxy/handler.go
+++ b/components/application-connectivity-validator/internal/validationproxy/handler.go
@@ -22,6 +22,8 @@ import (
 
 const (
 	CertificateInfoHeader = "X-Forwarded-Client-Cert"
+	// In a BEB enabled cluster, validator should forward the event coming to /{application}/v2/events and /{application}/events to /publish endpoint of EventPublisherProxy(https://github.com/kyma-project/kyma/tree/master/components/event-publisher-proxy#send-events)
+	BEBEnabledPublishEndpoint = "/publish"
 )
 
 type ProxyHandler interface {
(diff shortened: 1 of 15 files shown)


alexandra-simeonova (approved):

marcobebway (approved):

components/application-connectivity-validator/internal/validationproxy/handler.go
```diff
@@ -164,9 +172,21 @@ func (ph *proxyHandler) getClientIDsFromResource(applicationName string) ([]stri
...
+	case strings.HasPrefix(path, ph.eventServicePathPrefixV1):
+		return ph.eventsProxy, nil
+
+	// For a cluster which is BEB enabled, events reaching /{application}/v2/events will be routed to /publish endpoint of event-publisher-proxy
+	// For a cluster which is not BEB enabled, events reaching /{application}/v2/events will be routed to /{application}/v2/events endpoint of event-service
+	case strings.HasPrefix(path, ph.eventServicePathPrefixV2):
+		if ph.isBEBEnabled {
+			return ph.eventMeshProxy, nil
+		}
 		return ph.eventsProxy, nil
```
rafalpotempa, line 186: Isn't that equivalent to `/{application}/v1/events` which contradicts the comment or does the endpoint depend on eventService version for an application?

rafalpotempa (approved):

components/application-connectivity-validator/internal/validationproxy/handler.go
marcobebway: In case of `/v2/events` (which is deprecated), we handle it differently than `/v1/events`. Both types are handled by the same service, but two different endpoints.

Issue #9898, opened 2020-11-09
Define evaluation and production profile for console chart


kwiatekus:
**Description**

Define `production` and `evaluation` profiles for console chart.
 - For production, create a `profile-production.yaml` file. We can start with the current default values ( for cbs and console-web). Think which resources require higher values
 - For evaluation, create `profile-evaluation.yaml` file - request less cpu time and memory.

The defaults ( default values from values.yaml  - means "no profile" or default profile ) should be matching values in the middle between evaluation and production.

Current resources utilisation ( on  cluster with little trafic )
```
Name                                     CPU Requests  CPU Limits  Memory Requests  Memory Limits  
----                                     ------------  ----------  ---------------  ------------- 
console-web-77d55fd8db-5b52v                   0 (0%)      0 (0%)        64Mi (0%)     192Mi (1%)     
console-backend-865944bd45-cmnjx             75m (1%)   250m (6%)       128Mi (1%)     556Mi (4%)     
```
**Reasons**

There is no golden setup that fulfils all possible installation cases. Evaluation of kyma should be cheap and possible to realise on one node cluster. Production grade installation requires significant amount of processing time and memory ( especially on CBS)

**Attachments**

pbochynski:
Resolved by https://github.com/kyma-project/kyma/pull/10133

PR #10056, opened 2020-11-26
Create evaluation profile for console and console backend service

dariadomagala:
**Description**

- Create evaluation profile for console and console backend service

**Related issue(s)**


Sawthis:
It resolves the #9898 issue.

parostatkiem (commented):
What about the `cpu` limits/requests?

resources/console/charts/web/templates/deployment.yaml
```diff
@@ -35,6 +35,8 @@ spec:
         - name: console
           image: "{{ .Values.global.containerRegistry.path }}/console:{{ .Values.console.image.tag }}"
           imagePullPolicy: {{ .Values.console.image.pullPolicy }}
+          resources: 
+            {{- toYaml .Values.deployment.resources | nindent 12 }}
```
parostatkiem, line 39: what about the `core-ui` container? And other "frontend" deployments like `log-ui` or `service-catalog-io`?

resources/console/profile-evaluation.yaml
```diff
@@ -0,0 +1,14 @@
+web:
+  deployment:
+    resources:
+      requests:
+        memory: 16Mi
```
parostatkiem, line 5: On two different clusters I checked, the pod from this deployment consumes exactly 19Mi (2Mi Console; 17Mi Core-ui) while running. I think there's no case when it'd take less, so I suggest to use this value as the `request`.

resources/console/profile-evaluation.yaml
```diff
@@ -0,0 +1,14 @@
...
+      limits:
+        memory: 32Mi
+backend: 
+  deployment:
+    resources:
+      requests:
+        memory: 32Mi
```
parostatkiem, line 12: On two different clusters I checked, the pod from this deployment consumes 104Mi and 116Mi at a `standby` state. I think we should request a value closer to these numbers.

resources/console/profile-production.yaml
```diff
@@ -0,0 +1,14 @@
...
+    resources:
+      requests:
+        memory: 64Mi
+      limits:
+        memory: 128Mi
+backend: 
+  deployment:
+    resources:
+      requests:
+        memory: 64Mi
```
parostatkiem, line 12: ^ As in the profile-evaluation case - I think it'll never consume less than ~100Mi.

resources/console/profile-evaluation.yaml
parostatkiem: Here's the 12-hours `weekly` cluster summary from Grafana:
image

resources/console/profile-evaluation.yaml
parostatkiem: Also, here's the `weekly` cluster 12h summary. A bit different figures but still, I think it should be much less for the `console` container (as you currently apply) or much more for the whole pod.
image

parostatkiem (changes requested):

resources/console/profile-evaluation.yaml
dariadomagala: This is the evaluation profile, so weekly is not a good source of information. This cluster runs lots of tests all the time, and we won't have those tests in the evaluation profile.

resources/logging/profile-evaluation.yaml
```diff
@@ -17,3 +17,12 @@ fluent-bit:
...
       cpu: 20m
       memory: 40Mi
+
+logui:
+  resources: 
+    requests:
+      cpu: 1m
+      memory: 4Mi
+    limits:
+      cpu: 1m
```
parostatkiem, line 27: Look what few page reloads have done. I think `3m` value would be better.
image

parostatkiem (changes requested):

parostatkiem (approved):
🌵

resources/console/profile-evaluation.yaml
```diff
@@ -0,0 +1,25 @@
...
+      limits:
+        cpu: 10m
+        memory: 32Mi
+backend:
+  resources:
+    requests:
+      cpu: 2m
+      memory: 40Mi
+    limits:
+      cpu: 10m
```
adamwalach, line 24: I think limit values for CPU are a bit to low, take a look on values of other components (search for "Describe nodes (after installation)"):
https://storage.googleapis.com/kyma-prow-logs/logs/adamwalach_test_of_prowjob_kyma-integration-evaluation-gardener/1334523088292810752/build-log.txt
Maybe we could set it to a value closer to `50m`?
That may improve startup time as well
That comment is about all cpu limits -  for other components as well.

dariadomagala:
PR recreated here - https://github.com/kyma-project/kyma/pull/10133

PR #10063, opened 2020-11-26
Update eventing-controller k8s libs version to v0.18.6

montaro:
Update eventing-controller k8s libs version to v0.18.6

Resolves https://github.tools.sap/kyma/security-scans/issues/4149


montaro:
Will be updated in another PR to v0.18.12

PR #10085, opened 2020-11-30
Upgrade grafana chart

skhalash:
**Description**

- Upgrade Grafana chart to v6.1.9 of the upstream

**Related issue(s)**

### DIFF
--- a/resources/monitoring/charts/grafana/Chart.yaml
+++ b/resources/monitoring/charts/grafana/Chart.yaml
@@ -1,6 +1,7 @@
 apiVersion: v1
 name: grafana
-version: 5.2.1
+version: 6.1.9
+appVersion: 7.3.3
 kubeVersion: "^1.8.0-0"
 description: The leading tool for querying and visualizing time series and metrics.
 home: https://grafana.net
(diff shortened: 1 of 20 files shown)


resources/monitoring/charts/grafana/Chart.yaml
```diff
@@ -1,6 +1,7 @@
 apiVersion: v1
 name: grafana
-version: 5.2.1
+version: 6.1.9
+appVersion: 7.3.3
```
suleymanakbas91, line 4: We decided to delete `appVersion` field as it is mostly forgotten to be maintained: https://github.com/kyma-project/kyma/commit/a83c6821641471141dfd547cebcec0abfbb23530. Could you please delete it from all the subcharts?

resources/monitoring/charts/grafana/values.yaml
```diff
@@ -63,6 +67,13 @@ image:
   # pullSecrets:
   #   - myRegistrKeySecretName
 
+testFramework:
```
suleymanakbas91, line 70: You can delete this part as we are not using the Grafana test framework.

suleymanakbas91 (changes requested):

resources/monitoring/charts/grafana/Chart.yaml
skhalash: Done

resources/monitoring/charts/grafana/values.yaml
skhalash: Done

suleymanakbas91 (approved):

PR #10117, opened 2020-12-02
Remove duplicated function tests from e2e pipeline

aerfio:
**Description**

- I'm removing those tests as we've migrated those into our k3s pipeline
- add using secret as env source in k3s pipeline test

**Related issue(s)**

#10001

### DIFF
--- a/tests/function-controller/testsuite/teststep/secret.go
+++ b/tests/function-controller/testsuite/teststep/secret.go
@@ -1,0 +1,43 @@
+package teststep
+
+import (
+	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
+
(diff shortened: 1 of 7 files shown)


tests/function-controller/testsuite/teststep/secret.go
```diff
@@ -0,0 +1,43 @@
+package teststep
```
kfurgol, line 1: What do you think about moving this file to `pkg`?

kfurgol (approved):
lgtm

tests/function-controller/testsuite/teststep/secret.go
aerfio: There's already secret.go in pkg, but this file and that pkg/secret.go does similar things

PR #10127, opened 2020-12-03
Mark SKR compatible test definitions

tobiscr:
**Description**
Adding a label to test definitions which are SKR compatible can be executed as part of the SKR e2e test.

Only test definitions which could be successfully executed on an SKR cluster were enabled.

**Related issue(s)**


Demonsthere (approved):
/meow

piotrmsc (approved):

franpog859 (approved):

aerfio (approved):

PR #10131, opened 2020-12-03
Bump version of postgresql in ory charts

kubadz:
The bump is necessary because the currently used version (11.9) contains vulnerabilities that are fixed in 11.10.

**Description**

- bump version of postgresql in ory charts

**Related issue(s)**


Demonsthere (approved):
/meow

PR #10143, opened 2020-12-05
reduce dropped labels for istio, make istio scrape config configurable

a-thaler:
**Description**
The scrape envoyStats scrape config for the prometheus-istio instance is not configurable yet via overrides. This PR moves the config to the "additionalScrapeConfigs" section so that a multi-line string can be provided. Furthermore, it allows to have the sample-limit and metricsRelabelling regexp configurable.

- move envoyStats scrape config to additional section
- have sample-limit configurable
- have regexp for relabelling configurable
- reduce the amount of labels dropped for istio to have kiali working again

**Related issue(s)**

### DIFF
--- a/resources/monitoring/charts/prometheus-istio/values.yaml
+++ b/resources/monitoring/charts/prometheus-istio/values.yaml
@@ -1255,53 +1255,43 @@ serverFiles:
         static_configs:
           - targets:
             - localhost:9090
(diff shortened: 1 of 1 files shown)


rakesh-garimella (approved):
We would need a troubleshooting guide for users if they want to reduce the memory usage of prometheus-istio server by re-introducing the drop labels (at the cost of no kiali graphs)

PR #10160, opened 2020-12-08
Add k8s test to fast-integration

aerfio:
**Description**

- as in title, add scaffold for future tests


tests/fast-integration/test/fixtures.js
```diff
@@ -0,0 +1,29 @@
+const deploymentYamlString = `apiVersion: apps/v1
```
crabtree, line 1: For me it's a bit too generic, what about renaming it to `httpbinDeployment`?

aerfio: It will be gone in next PR where we rewrite that test to https://github.com/kyma-incubator/local-kyma/blob/main/app-connector-example.sh

crabtree (approved):

tests/fast-integration/test/k8s-test.js
```diff
@@ -0,0 +1,55 @@
...
+
+  before(async function () {
+    try {
+      await k8sDynamicApi.delete(deployObj);
+    } catch (error) {
+      console.warn(error.body.message);
+    }
+  });
+
+  it("create deployment", async function () {
```
pbochynski, line 22: I think it should go to `before` section. The idea is that deployments, functions are created before tests not during the test. In the test we should check if deployment has status ready.

PR #10170, opened 2020-12-09
Fast integration tests README

pbochynski:


tests/fast-integration/README.md
```diff
@@ -0,0 +1,59 @@
...
+
+Install dependencies:
+```
+cd kyma/tests/fast-integration
+npm install
+```
+
+Install mocha:
+```
+npm install -g mocha
```
crabtree, line 40: You don't need this if you run our tests using `npm test`

pbochynski: Fixed.

tests/fast-integration/README.md
```diff
@@ -0,0 +1,54 @@
...
+npm test
+```
+
+
+# FAQ
+
+## Why don't you use Octopus?
+Octopus is a great tool for running tests inside Kubernetes cluster in a declarative way. But it is not the right tool for fast integration testing. The goal is to execute the tests in 4 minutes. With Octopus you need 4 minutes or more before test even start (2 minutes to build test image and push it to docker registry, 1 minute to deploy Octopus, 1 minute to deploy the test pod). 
+
+## Why tests are written in node.js not in golang?
```
aerfio, line 49: Kinda nitpick, but: 
```suggestion
## Why tests are written in node.js not in Go?
```

tests/fast-integration/README.md
```diff
@@ -0,0 +1,54 @@
...
+
+Several reasons:
```
aerfio, line 51: I'd add some note about more mature testing frameworks and overall lighter dependencies - I know all of those jokes about node_modules, but in this case we have only 1 big dependency - @kubernetes/client-node. If we used Go we'd have much more dependencies, like k8s.io/*, go-client, controller-runtime, etc etc

aerfio (commented):
LGTM, see two small nitpicks. That part about lighter dependencies may not fit nicely there, you might drop it if it feels off the point

tests/fast-integration/README.md
pbochynski: Done.

crabtree (approved):

aerfio (approved):

klaudiagrz (approved):

klaudiagrz (approved):
TWs will do a post-review later on. I'm approving the PR just to unblock it :)

PR #10173, opened 2020-12-09
Fix problems in istio-proxy-reset job

Tomasz-Smelcerz-SAP:
**Description**
Improve istio-proxy-reset-job.
This PR addressed two problems:
- "null" parent for plain old Pods without backing deployment/replicaset etc.
- Poor Pod selection: Hard-coded old version of Istio Proxy to select the Pods that should be migrated. Logic is reversed now: the user provides the expected Istio Proxy version, and the logic selects all Pods that don't match this version.

- Improved Pod matching logic
- Fixed problem with null parent
- Put the script into a separate file to support manual testing/maintenance
- Add dry-run option to support manual testing/maintenance
- Expose job's parameters as entries in values.yaml to simplify future maintenance

### DIFF
--- a/resources/istio/templates/istio-proxy-reset.yaml
+++ b/resources/istio/templates/istio-proxy-reset.yaml
@@ -56,61 +56,14 @@ spec:
       containers:
         - name: proxy-reset
           image: eu.gcr.io/kyma-project/incubator/develop/k8s-tools:20201023-5de446cf
+          env:
+          - name: EXPECTED_ISTIO_PROXY_IMAGE
+            value: {{ .Values.kyma.proxyResetJob.expectedIstioProxyImage }}
+          - name: COMMON_ISTIO_PROXY_IMAGE_PREFIX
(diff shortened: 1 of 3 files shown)


resources/istio/templates/istio-proxy-reset.yaml
```diff
@@ -56,61 +56,14 @@ spec:
       containers:
         - name: proxy-reset
           image: eu.gcr.io/kyma-project/incubator/develop/k8s-tools:20201023-5de446cf
+          env:
+          - name: EXPECTED_ISTIO_PROXY_IMAGE
+            value: {{ .Values.kyma.proxyResetJob.expectedIstioProxyImage }}
```
Demonsthere, line 61:
```suggestion
            value: {{ .Values.kyma.proxyResetJob.commonIstioProxyImagePrefix }}:{{ .Chart.Version }}
```

Demonsthere (dismissed):
/meow

strekm (approved):
/meow

PR #10192, opened 2020-12-10
Fix monitoring test

tobiscr:
**Description**

Fix instable monitoring test caused by exceeded scrape limit.

**Related issue(s)**


rakesh-garimella (approved):

rakesh-garimella:
We have to ignore this alert as we had to introduce the sample_limits to protect prometheus OOM on long running cluster with dynamic workload creation. This can be reverted when this [issue](https://github.com/istio/istio/issues/21377) is fixed.

PR #10268, opened 2020-12-30
Remove duplicate resource element

berkanercan:
**Description**

Both myself and another team has added resources to this file. Since mine has if statement in it. I have removed mine.


sayanh:
Please add a description

components/application-operator/charts/application/templates/deployment.yaml
```diff
@@ -136,8 +136,4 @@ spec:
         ports:
         - containerPort: {{ .Values.eventService.deployment.args.externalAPIPort }}
           name: http-api-port
-        {{- if .Values.eventService.deployment.resources }}
-        resources:
-        {{- toYaml .Values.eventService.deployment.resources | nindent 10 }}
-        {{- end }}
 {{- end }}
```
sayanh, line 139: Following are my concerns:
- This change comes with an image change as well. I’d include the change in image as well in the same PR so that the context is preserved.
- Values still exist [here](https://github.com/kyma-project/kyma/blob/master/components/application-operator/charts/application/values.yaml#L81) which need to be removed as well.

berkanercan:
\test all

janmedrek (approved):

sayanh (approved):

PR #10275, opened 2021-01-04
Pamela resources profiling

parostatkiem:
**Description**

- add evaluation/production profiles handling to Pamela chart
- adjust evaluation and production profile values basing on observations
- increase Console-backend memory limit to `200Mi` in production profile

**Related issue(s)**

### DIFF
--- a/resources/console/profile-evaluation.yaml
+++ b/resources/console/profile-evaluation.yaml
@@ -26,8 +26,8 @@ backend:
 pamela:
   resources:
     limits:
-      cpu: 120m
-      memory: 300Mi
+      cpu: 40m
+      memory: 150Mi
     requests:
-      cpu: 50m
-      memory: 190Mi
+      cpu: 15m
+      memory: 55Mi
--- a/resources/console/profile-evaluation.yaml
+++ b/resources/console/profile-evaluation.yaml
@@ -26,7 +26,7 @@ backend:
 pamela:
   resources:
     requests:
-      cpu: 15m
+      cpu: 8m
       memory: 55Mi
     limits:
       cpu: 40m
--- a/resources/console/profile-evaluation.yaml
+++ b/resources/console/profile-evaluation.yaml
@@ -23,3 +23,11 @@ backend:
     limits:
       cpu: 50m
       memory: 80Mi
+pamela:
+  resources:
+    limits:
+      cpu: 120m
+      memory: 300Mi
+    requests:
+      cpu: 50m
+      memory: 190Mi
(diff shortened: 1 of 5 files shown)


resources/console/profile-evaluation.yaml
```diff
@@ -23,3 +23,11 @@ backend:
...
       cpu: 50m
       memory: 80Mi
+pamela:
+  resources:
+    requests:
+      cpu: 8m
+      memory: 55Mi
+    limits:
+      cpu: 40m
+      memory: 150Mi
```
dariadomagala, line 33: Weekly and nightly clusters use 180Mi all the time, 150 might be not enough

dariadomagala (commented):
Also, Krzysiu suggested to also fix this issue:
https://kyma-community.slack.com/archives/CD11JLTTN/p1608465633047000

resources/console/profile-evaluation.yaml
parostatkiem: No, weekly uses 53MB normally, 63MB on peak...
image

resources/console/profile-production.yaml
```diff
@@ -22,4 +22,12 @@ backend:
       memory: 40Mi
     limits:
       cpu: 50m
-      memory: 150Mi
+      memory: 200Mi
```
parostatkiem, line 25: @dariadomagala I raised this value. I assumed the man with the problem was using a production cluster.

resources/console/profile-evaluation.yaml
dariadomagala: You're right, I selected the wrong color on the graph, my bad. If it's 63, then 150 might be a bit too much. same with production profile. I think 200 would be enough there

parostatkiem: 😕 so.. 200 or 150? I think `150Mi` would be perfect for the production profile and `100Mi` for the evaluation one. Do you agree?

dariadomagala: "150 might be a bit too much" was about the evaluation profile, "200 would be enough there" was about the production profile 🙂 sorry for the confusion.
Those values that you wrote should be fine 🙂

resources/console/profile-production.yaml
dariadomagala: as we discussed, this issue should be moved to a separate task after all, please remove this change :)

kazydek (approved):

akucharska (approved):

Issue #9665, opened 2020-10-08
Transform projects to go modules


dbadura:
**Description**
In current repository we have:
- 24 projects which use `go dep`
- 22 projects which use `go modules`
The main problem is that the `dep` procjet is abandoned and will not receive further updates.

**PoC**
Under this PoC I want to:
- migrate as many project as I can within few days
- evaluate the complexity of migration of rest of components
- adjust generic makefile to use `go mod`

Expected problems:
- big problems with dependencies
- version conflicts

**TODO**:
- [ ] Adjust Generic Makefile to use `go.mod`
- [ ] Migrate projects 
- [ ] Migrate tests
- [ ] Remember about security scans
**Reasons**
The `dep` tool is not supported.

dbadura:
I have transfered following components:
Components
-- |
./components/event sources
./components/app-broker
./components/event service
./components/kubeconfig iam service
./components/apiserver-proxy
 

Tests
-- |
./tests/application-connector-test
./tests/application gateway legacy tests
./tests/application registry test
./tests/integration/monitoring
./tests/integration/dex
./tests/integration/event-service
./tests/rafter

The following components won't be transfered because they are going to be removed:

Test | reason
-- | -- 
./tests/compass-runtime-agent | team refactoring this
./tests/console-backend-service | to be removed
./tests/contract/knative-channel-kafka | to be removed after beb implementation
./tests/end-to-end/external-solution-integration | blocked by k8s version mismatch
./tests/end-to-end/upgrade | blocked by k8s version mismatch
./tests/integration/apiserver-proxy | to be removed
./tests/integration/logging | blocked compass rtm agent-test
./tests/service-catalog | blocked by app broker (to old k8s version)
./tools/failery | cbs uses

More description:
external-solution-integration: service binding usage has k8s in 0.18 but app broker and event-sources has k8s in 0.17
upgrade: service binding usage has k8s in 0.18 but app broker and event-sources has k8s in 0.17
service-catalog: service binding usage has k8s in 0.18 but app broker has k8s in 0.17

PR #10277, opened 2021-01-04
Transform Event-Service integration test to use go modules isntead of dep 

dbadura:
**Description**

- transform project to use go modules

**Related issue(s)**

#9665


berkanercan (approved):
looking good

k15r (approved):

PR #10278, opened 2021-01-04
Transform dex integration test to use go modules isntead of dep 

dbadura:
**Description**

- transform project to use go modules

**Related issue(s)**

#9665


tests/integration/dex/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM eu.gcr.io/kyma-project/test-infra/buildpack-golang-toolbox:v20190913-65b55d1 as builder
+FROM golang:1.14-alpine3.12 as builder
```
piotrmsc, line 1: @dbadura hmm, I think kyma buildpack should be still used in place?

dbadura: Generally it can stay but we don't use tools from buildpack. Only golang is required. 
I decided to remove the heavy buildpack (1,6GB) and use lighter alpine golang (370mb).

piotrmsc: hmmm, so maye perhaps such buildpack should be also moved to kyma gcr? 
Difference in size is totally a ➕ but won't this make triage worse?

piotrmsc (approved):

Issue #10037, opened 2020-11-25
Function image build scheduler 


kwiatekus:
**Description**

Improve function build mechanism so that it builds 100 functions on small cluster in a way that cluster remains operable.

Build service should schedule parallel builds in a way that it doesn't exceed dedicated resources.
During the build process ( of all the lambdas ) the build scheduler should not exhaust all available resources 

Define default configuration per profile, ie:
 - single build allowed at a time in evaluation profile. Additional build requests would need to be queued.
 - 5 simultaneous builds allowed in production. Anything beyond the limit goes into queue.

Perform an empirical stress test that proves the solution.

PR #10283, opened 2021-01-05
Function images build scheduling

kfurgol:
**Description**

- Added mechanism that will requeue reconciler if there are more function-controller related jobs than given amount
- Added unit tests
- Added profiles settings
- Removed configuration of maximum concurrent reconciles

**Related issue(s)**
#10037


components/function-controller/internal/controllers/serverless/function_reconcile.go
```diff
@@ -58,7 +58,7 @@ func (r *FunctionReconciler) SetupWithManager(mgr ctrl.Manager) error {
 		Owns(&corev1.Service{}).
 		Owns(&autoscalingv1.HorizontalPodAutoscaler{}).
 		WithOptions(controller.Options{
-			MaxConcurrentReconciles: r.config.MaxConcurrentReconciles,
+			MaxConcurrentReconciles: 1, // Job build scheduling mechanism requires this parameter to be set to 1
```
dbadura, line 61: I would add more description about why the value is 1.

dbadura: Link to the issue

dbadura:
Did you take into consideration that our test are run in parallel (3 functions at one time) and this change can introduce to instability of the tests, because the third function to build could timeout.
Did you run serverless tests several times to check this?
Maybe it's not a problem, because the function build so fast that reaching the timeout is not possible.

resources/serverless/profile-production.yaml
```diff
@@ -16,4 +16,9 @@ webhook:
...
   resources:
     requests:
-      cpu: 50m
\ No newline at end of file
+      cpu: 50m
+
+containers:
+  manager:
+    envs:
+      functionBuildMaxSimultaneousJobs: "5"
```
tgorgol, line 24: I wonder if we wouldn't be able to handle more than 5 simultaneous builds in production configuration.

tgorgol (commented):
Good job, few comments/suggestions from me

resources/serverless/profile-production.yaml
kfurgol: Indeed maybe we should revisit the acceptance criteria of the task, let's discuss

kfurgol: As discussed, for now, we'll go with 5

dbadura (changes requested):
Updated the comment

components/function-controller/internal/controllers/serverless/function_reconcile.go
```diff
@@ -58,7 +58,7 @@ func (r *FunctionReconciler) SetupWithManager(mgr ctrl.Manager) error {
 		Owns(&corev1.Service{}).
 		Owns(&autoscalingv1.HorizontalPodAutoscaler{}).
 		WithOptions(controller.Options{
-			MaxConcurrentReconciles: r.config.MaxConcurrentReconciles,
+			MaxConcurrentReconciles: 1, // Build job scheduling mechanism requires this parameter to be set to 1 https://github.com/kyma-project/kyma/issues/10037
```
dbadura, line 61: After some time, adding the link is not good thing, because the issue itself doesn't say why the value is 1.
It only describe that we want to implement concurrent build jobs limit.

I would add description why the value is 1. How it affects the scheduling mechanism. Why will happen when the value is set to something different bigger than 1.

dbadura (changes requested):
Added the comment :)

aerfio:
pre-master-serverless-integration-k3s

tgorgol (approved):
LGTM

components/function-controller/internal/controllers/serverless/function_reconcile.go
```diff
@@ -58,7 +58,7 @@ func (r *FunctionReconciler) SetupWithManager(mgr ctrl.Manager) error {
 		Owns(&corev1.Service{}).
 		Owns(&autoscalingv1.HorizontalPodAutoscaler{}).
 		WithOptions(controller.Options{
-			MaxConcurrentReconciles: r.config.MaxConcurrentReconciles,
+			MaxConcurrentReconciles: 1, // Build job scheduling mechanism requires this parameter to be set to 1. The mechanism is based on getting active and stateless jobs, concurrent reconciles makes it non deterministic. https://github.com/kyma-project/kyma/issues/10037
```
dbadura, line 61:
```suggestion
			MaxConcurrentReconciles: 1, // Build job scheduling mechanism requires this parameter to be set to 1. The mechanism is based on getting active and stateless jobs, concurrent reconciles makes it non deterministic . Value 1 removes data races while fetching list of jobs. https://github.com/kyma-project/kyma/issues/10037
```

dbadura (approved):

components/function-controller/README.md
```diff
@@ -38,6 +38,7 @@ To develop the component, use the formulae declared in the [generic](/common/mak
...
 | **APP_FUNCTION_BUILD_REPOFETCHER_IMAGE**                  | Full name of the Repo-Fetcher init container used for cloning repository for the Kaniko executor                                                                                                                                                                                                             | `eu.gcr.io/kyma-project/function-build-init:305bee60`                                                                                                    |
+| **APP_FUNCTION_BUILD_MAX_SIMULTANEOUS_JOBS**              | Maximum number of running simultaneous build jobs                                                                                                                                                                                                                                                            | `5`                                                                                                                                                      |
```
kazydek, line 41:
```suggestion
| **APP_FUNCTION_BUILD_MAX_SIMULTANEOUS_JOBS**              | Maximum number of build jobs running simultaneously                                                                                                                                                                                                                                                            | `5`                                                                                                                                                      |
```

tgorgol (approved):

dbadura (approved):

kazydek (approved):
🎉

Issue #10089, opened 2020-11-30
Redesign of fluent-bit configuration improving usabilities of shipped logs


a-thaler:
**Description**

Major goal of the logging component is to ship logs of system and workload components to different logging backends in a way that they are most queryable in that backend.
A challenge with kyma is the heterogenousy of log formats. Every workload (or system component) is logging in a different format. Using the kubernetes filter, it will parse JSON documents only on-demand (so you can mix plain text and JSON). However, shipping the differnet log types to a stack like ELK, different JSON formats are even a bigger problem as the JSON attribute types must be of the same type for an successful indexing.
With that, by default JSON documents should not be parsed and shipped as plain text log to the different backend systems as we simply cannot give any guarantees on the formats.

That situation we can improve by introducing a well-defined best practice log format and giving a chance to workloads to be conform to that format. Having then an additional filter applied in fluent-bit will enable us to send well-formed logs for known applications to the backend, while everything else gets shipped as plain text.
That should be the case for system components. A custom workload should have the chance to participate in that as well but also to define easy it's own format and enable shipping for it.

In order to achieve it the current fluent-bit chart needs a redesign as it is not flexible enough. Further points with current config approach is:
- Not all features of the standard plugins are configurable via values.yaml
- You can mix-match the output plugins via the tags to different inputs, but there is no flexibility on having various inputs nor filter combinations (only way to introduce custom inputs or filters is via one extra.conf, sometimes you want to have multiple even)
- The kubernetes filter plugin should be configured differently dependent on the output and/or input

Goal:
- come up with a new config format which is flexible to configure (fixing the points above) but brings a very good default out-of-the-box
- The default should cover already the discussed option to process and ship logs coming from workloads with a well-defined format
- The config should not be dependent on the "overrides" limitation, assume full values.yaml expressiveness as planned to be introduced soon

**Reasons**

**Attachments**

suleymanakbas91:
With https://github.com/kyma-project/kyma/pull/10284, I've refactored the chart and made it very similar to the upstream chart. I've also added new configuration options (config.parsers.additional, config.filters.additional, etc.) to enable passing extra configuration options to each part of the config file. 

As an example, I pass this overrides file to a Kyma installation: 

```yaml
---
apiVersion: v1
kind: ConfigMap
metadata:
  labels:
    installer: overrides
    component: logging
    kyma-project.io/installation: ""
  name: logging-overrides
  namespace: kyma-installer
data:
[...]

PR #10284, opened 2021-01-05
Refactor Fluent Bit Chart

suleymanakbas91:
**Description**

- Refactor Fluent Bit Chart

**Related issue(s)**

See https://github.com/kyma-project/kyma/issues/10089


resources/logging/charts/fluent-bit/templates/configmap.yaml
```diff
@@ -0,0 +1,227 @@
...
+        Name                kubernetes
+        Match               {{ .Values.config.filters.kubernetes.match }}
+        Kube_URL            https://kubernetes.default.svc:443
+        Merge_Log           {{ .Values.config.filters.kubernetes.mergeLog }}
+        Keep_Log            {{ .Values.config.filters.kubernetes.keepLog }}
+        {{- if .Values.config.filters.kubernetes.mergeLogKey }}
+        Merge_Log_Key       {{ .Values.config.filters.kubernetes.mergeLogKey }}
+        {{- end }}
+        K8S-Logging.Parser  {{ .Values.config.filters.kubernetes.parser }}
+        K8S-Logging.Exclude {{ .Values.config.filters.kubernetes.exclude }}
```
skhalash, line 83: I rendered the template before and after this change and found out that there was a small difference:

```
    [FILTER]                                                         [FILTER]
        ...                                                              ...
        K8S-Logging.Parser  On                                           K8S-Logging.Parser  true
        K8S-Logging.Exclude On                                           K8S-Logging.Exclude true
```
Not sure if it's important though (even though in the docs they use On instead of true for booleans)

resources/logging/charts/fluent-bit/values.yaml
```diff
@@ -0,0 +1,308 @@
...
+
+# Make use of a pre-defined configmap instead of the one templated here
+existingConfigMap: ""
+
+# Fluentbit configuration section
+# https://docs.fluentbit.io/manual/administration/configuring-fluent-bit/configuration-file
+config:
+  service:
+    flush: 1
+    logLevel: error
```
skhalash, line 127: The log level was changed from `info` to `error`. Is it necessary? Fluentbit doesn't seem to be too chatty, `info` could be useful for debugging

resources/logging/charts/fluentbit/templates/secrets.yaml
```diff
@@ -1,139 +0,0 @@
...
-  namespace: {{ .Release.Namespace }}
-  labels:
-    {{- include "fluent-bit.metaLabels" . | trim | nindent 4 }}
-  annotations:
-    {{- if .Values.globalAnnotations }}
-    {{- toYaml .Values.globalAnnotations | trim | nindent 4 }}
-    {{- end }}
-type: kubernetes.io/tls
-data:
-  tls.crt: {{ required ".Values.prometheusPushGateway.tls.cert is required when TLS Authentication is enabled!" .Values.prometheusPushGateway.tls.cert }}
```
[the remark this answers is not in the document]
suleymanakbas91: I see that it does not exist on the upstream chart, and it is probably not used by anyone. To simplify the chart and make it look similar to the upstream, I deleted it.

resources/logging/charts/fluent-bit/templates/configmap.yaml
suleymanakbas91: Fixed.

resources/logging/charts/fluent-bit/values.yaml
suleymanakbas91: I see that Fluent Bit Pods are the most log generating ones on the SKR, that's why I wanted to lower it. But, I'm changing it to `warn` now not to be too harsh at once. This will be changed anyway in the coming weeks and will be adaptable in global level, once we introduce Kyma log level option.

skhalash: Ah, OK, good to know

resources/logging/charts/fluentbit/templates/secrets.yaml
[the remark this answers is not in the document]
skhalash: Makes sense!

skhalash (dismissed):
Looks good to me!

erichorwath:
Is it possible to share one logging-fluent-bit-configmap example which has been generated by this new Helm chart along with production-like values (and preferably auditlog and kibana enabled). This makes it easier for me to compare it with the (hardcoded) configmap, we use: https://github.wdf.sap.corp/SCMOptimizer/OptimizationServices/blob/master/helm/optimization-service-kyma-overrides/templates/logging-fluent-bit-config.yaml

resources/logging/charts/fluent-bit/templates/_helpers.tpl
```diff
@@ -0,0 +1,80 @@
...
+{{- end -}}
+{{- end -}}
+
+{{- define "helm-toolkit.utils.joinListWithComma" -}}
+{{- $local := dict "first" true -}}
+{{- range $k, $v := . -}}{{- if not $local.first -}},{{- end -}}{{- $v -}}{{- $_ := set $local "first" false -}}{{- end -}}
+{{- end -}}
+
+{{- define "loki.namespace.filter" -}}
+{{- if .Values.conf.Input.Kubernetes_loki.exclude.namespaces -}}
```
rakesh-garimella, line 71: Needs to be fixed as the plugin has been renamed to `tail` and `Kuberenetes_loki` does not exist

suleymanakbas91: Nice catch! Fixed.

rakesh-garimella (approved):

Issue #9068, opened 2020-07-20
Application Gateway - requestTimeout and proxyTimeout parameters


VishnAndr:
**Description**
Request to expose **requestTimeout** and **proxyTimeout** parameters in application operator for application gateway component.
Or even better make it configurable in console UI on application level / namespace level.

**Reasons**

By default, the application gateway component is created with **requestTimeout** and **proxyTimeout** set as 10 seconds.

Sometimes an external (connected via application connector) system is not fast enough and can not process the request within 10 seconds interval.

An example is SAP Cloud for Customer system, where POST or PATCH operations may include synchronous execution of workflows, custom enhancement logic or external system calls.

**Attachments**

[Slack thread about the issue](https://kyma-community.slack.com/archives/CD0K2NSQZ/p1594886966112900)

**Workaround**
Directly edit `<appName>-application-gateway` deployment with
`$ kubectl edit deployment <appName>-application-gateway`
then change
.spec.template.spec.containers.args : 
        - --requestTimeout=10
        - --proxyTimeout=10
to the required values.

Or 
`$ kubectl get deployment <appName>-application-gateway -o yaml >> <appName>-application-gateway.yaml`
then change the same requestTimeout and proxyTimeout to the required values in the `<appName>-application-gateway.yaml` file
then apply the changes with
`$ kubectl apply -f  <appName>-application-gateway.yaml`

VishnAndr:
Or maybe combine them into one parameter? I personally can't imagine the case where two separate parameters with different values would be required.

ralikio:
We have just finished implementation of mechanism that allows to override helm configuration when creating application CRD.  The feature uses labels to provide new values for selected parameters. The details can be seen in the video attached to the description of https://github.com/kyma-project/kyma/pull/10289.

VishnAndr:
@ralikio great!! Thank you! Is it going to be in 1.19 or 1.20?

Issue #5509, opened 2019-08-30
Provision application-gateway with --skipVerify=true


johnkbarrow:
**Description**
You can override the --skipVerify flag for an application gateway instance once the Application has been created with a kubectl patch. It would be nice if this flag could be set to true when you create the Application.

**Reasons**
You need to set --skipVerify=true if you want to connect a local SAP Commerce Cloud instance with a local Kyma instance.
See https://www.sap.com/cxworks/article/468901527/Working_with_Local_Instances_of_SAP_Commerce_Cloud_and_Project_Kyma

**Attachments**

ralikio:
Feature implemented and merged in https://github.com/kyma-project/kyma/pull/10289.

PR #10289, opened 2021-01-07
Label Overrides Extension

ralikio:
**Description**

The PR contains changes that allow customisations of application-gateway deployments without need to use `kubectl edit` command. The mechanism is based on values provided in labels of Application CR in the form of `override.<HELM_PROP>=<VALUE>`. After resource creation the values are injected into helm overrides context so that they can be used in Helm Charts as any other values.

https://user-images.githubusercontent.com/74771103/103879482-38246780-50d8-11eb-8a1c-26a05bc9e69c.mov

- Extending override mechanism to include values provided in labels.
- Unit tests.

**Related issue(s)**
#5509 #9068


components/application-operator/pkg/kymahelm/application/overrides.go
```diff
@@ -13,3 +23,52 @@ type OverridesData struct {
 	StrictMode                            string `json:"strictMode,omitempty"`
 	IsBEBEnabled                          bool   `json:"isBEBEnabled,omitempty"`
 }
+
+type OverridesMap map[string]interface{}
+
+type StringMap map[string]string
+
+func MergeLabelOverrides(labels StringMap, target map[string]interface{}) {
```
akgalwas, line 31: The code could be a bit shorter:
```
func MergeLabelOverrides(labels StringMap, target map[string]interface{}) {
	for key, value := range labels {
		if strings.HasPrefix(key, overridePrefix) {
			preParsedKey := strings.Trim(strings.TrimPrefix(key, overridePrefix),".")

			if len(preParsedKey) == 0 {
				continue
			}

			utils.MergeMaps(target, unwind(preParsedKey, value))
		}
	}
}

func unwind(key string, value string) map[string]interface{} {
	index := strings.Index(key, ".")
	if index < 0 {
		return map[string]interface{}{key: value}
	}

	return map[string]interface{}{key[:index]: unwind(key[index + 1:], value)}
}
```

Moreover it more nicely communicates the recursive nature of the unwind function ( stop conditions are placed first, and recursive call is the last thing in the function body)

ralikio: Corrected.

components/application-operator/pkg/application-controller/reconciler_test.go
```diff
@@ -289,6 +291,7 @@ func TestApplicationReconciler_Reconcile(t *testing.T) {
 		assert.NotNil(t, result)
 		managerClient.AssertExpectations(t)
 		ApplicationReleaseManager.AssertExpectations(t)
+		ApplicationReleaseManager.AssertCalled(t, "UpgradeApplicationRelease", ofType)
```
koala7659, line 294: This is not needed AssertExpectations already asserts that everything specified with On and Return was
in fact called as expected.

components/application-operator/pkg/application-controller/reconciler_test.go
```diff
@@ -274,6 +274,8 @@ func TestApplicationReconciler_Reconcile(t *testing.T) {
 		ApplicationReleaseManager := &helmmocks.ApplicationReleaseManager{}
 		ApplicationReleaseManager.On("CheckReleaseExistence", applicationName).Return(true, nil)
 		ApplicationReleaseManager.On("CheckReleaseStatus", applicationName).Return(releaseStatus, statusDescription, nil)
+		ofType := mock.AnythingOfType("*v1alpha1.Application")
```
koala7659, line 277: As we don't need to call AssertCalled this can be defied in place as 

`ApplicationReleaseManager.On("UpgradeApplicationRelease", mock.AnythingOfType("*v1alpha1.Application")) `

components/application-operator/pkg/application-controller/reconciler_test.go
```diff
@@ -444,6 +449,43 @@ func TestApplicationReconciler_Reconcile(t *testing.T) {
...
+
+		// when
+		result, err := reReconciler.Reconcile(request)
+
+		// then
+		assert.NoError(t, err)
+		assert.NotNil(t, result)
+		managerClient.AssertExpectations(t)
+		ApplicationReleaseManager.AssertExpectations(t)
+		ApplicationReleaseManager.AssertCalled(t, "UpgradeApplicationRelease", ofType)
```
koala7659, line 487: This is not needed AssertExpectations already asserts that everything specified with On and Return was
in fact called as expected.

components/application-operator/pkg/application-controller/reconciler_test.go
```diff
@@ -444,6 +449,43 @@ func TestApplicationReconciler_Reconcile(t *testing.T) {
...
+			"Get", context.Background(), namespacedName, mock.AnythingOfType("*v1alpha1.Application")).
+			Run(setupAppInstance).Return(nil)
+		managerClient.On("Update", context.Background(), mock.AnythingOfType("*v1alpha1.Application")).
+			Run(statusChecker.checkStatus).Return(nil)
+
+		ApplicationReleaseManager := &helmmocks.ApplicationReleaseManager{}
+		ApplicationReleaseManager.On("CheckReleaseExistence", applicationName).Return(true, nil)
+		ApplicationReleaseManager.On("CheckReleaseStatus", applicationName).Return(releaseStatus, statusDescription, nil)
+
+		ofType := mock.AnythingOfType("*v1alpha1.Application")
```
koala7659, line 470: This is not needed now and can be replaced by  `ApplicationReleaseManager.On("UpgradeApplicationRelease", mock.AnythingOfType("*v1alpha1.Application"))`

components/application-operator/pkg/kymahelm/application/applicationreleasemanager.go
```diff
@@ -30,6 +30,7 @@ type ApplicationReleaseManager interface {
 	CheckReleaseExistence(name string) (bool, error)
 	CheckReleaseStatus(name string) (hapi_4.Status, string, error)
 	UpgradeApplicationReleases() error
+	UpgradeApplicationRelease(application *v1alpha1.Application)
```
koala7659, line 33: Suggestion - please define this function as a returning error result

components/application-operator/pkg/kymahelm/application/applicationreleasemanager.go
```diff
@@ -74,20 +75,24 @@ func (r *releaseManager) UpgradeApplicationReleases() error {
...
-		status, description, err := r.upgradeChart(&app)
-		if err != nil {
-			log.Errorf("Failed to upgrade release %s: %s", app.Name, err.Error())
+		r.UpgradeApplicationRelease(&app)
+	}
+
+	return nil
+}
+
+func (r *releaseManager) UpgradeApplicationRelease(app *v1alpha1.Application) {
```
koala7659, line 84: I understand that ignoring errors is required here however this is not a good practice to do it outside caller method and unusual with goLang practices. This function may return an error which can be just ignored. I suggest following code:
```
func (r *releaseManager) UpgradeApplicationRelease(app *v1alpha1.Application) error {

	if err != nil {
		log.Errorf("Failed to upgrade release %s: %s", app.Name, err.Error())

			setCurrentStatus(&app, status.String(), description)
		setCurrentStatus(app, status.String(), description)

			err = r.updateApplication(&app)
			if err != nil {
				log.Errorf("Failed to upgrade %s CR: %s", app.Name, err.Error())
			}
		err = r.updateApplication(app)
		if err != nil {
			log.Errorf("Failed to upgrade %s CR: %s", app.Name, err.Error())
		}
	}
 return err
}
```

components/application-operator/pkg/application-controller/reconciler_test.go
```diff
@@ -444,6 +449,43 @@ func TestApplicationReconciler_Reconcile(t *testing.T) {
 		managerClient.AssertExpectations(t)
 		ApplicationReleaseManager.AssertExpectations(t)
 	})
+
+	t.Run("should invoke release manager upgrade if application override labels has changed", func(t *testing.T) {
```
koala7659, line 453: Please reconsider this test

koala7659 (commented):
Please see my comments

tests/application-operator-tests/test/testkit/k8sresources_client.go
```diff
@@ -16,7 +18,7 @@ import (
 )
 
 type K8sResourcesClient interface {
-	GetDeployment(ctx context.Context, name string, options metav1.GetOptions) (interface{}, error)
```
koala7659, line 19: The client may use previous interface. Please see my comment to suite.go file

tests/application-operator-tests/test/testkit/k8sresources_client.go
```diff
@@ -83,7 +87,7 @@ func initClient(k8sConfig *restclient.Config, namespace string) (K8sResourcesCli
 	}, nil
 }
 
-func (c *k8sResourcesClient) GetDeployment(ctx context.Context, name string, options metav1.GetOptions) (interface{}, error) {
```
koala7659, line 86: The client may use previous interface. Please see my comment to suite.go file

tests/application-operator-tests/test/testkit/k8sresources_check.go
```diff
@@ -138,3 +140,9 @@ func (c *K8sResourceChecker) checkResourceRemoved(_ interface{}, err error) bool
 
 	return false
 }
+
```
koala7659, line 143: This method is not needed if we leave previous implementation

tests/application-operator-tests/test/applicationcontroller/suite.go
```diff
@@ -64,6 +72,59 @@ func (ts *TestSuite) CreateApplication(t *testing.T, accessLabel string, skipIns
...
+
+func (ts *TestSuite) WaitForRunArg(t *testing.T, name string, expectedArg string) {
+	err := testkit.WaitForFunction(defaultCheckInterval, assessLabelWaitTime, func() bool {
+		return ts.containsArg(t, name, expectedArg)
+	})
+
+	require.NoError(t, err)
+}
+
+func (ts *TestSuite) containsArg(t *testing.T, name string, expectedArg string) bool {
```
koala7659, line 113: With k8sClient using previous interface for getting Deployments we can do in place casting:

```
func (ts *TestSuite) containsArg(t *testing.T, name string, expectedArg string) bool {
	deployment, err := ts.k8sClient.GetDeployment(context.Background(), name, metav1.GetOptions{})
	require.NoError(t, err)

	for _, container := range deployment.(*v1.Deployment).Spec.Template.Spec.Containers {
		for _, arg := range container.Args {
			if arg == expectedArg {
				return true
			}
		}
	}

	return false
}
```

tests/application-operator-tests/test/applicationcontroller/tests/applicationcontroller_test.go
```diff
@@ -54,3 +56,45 @@ func TestApplicationOperator_SkipProvisioning(t *testing.T) {
 
 	testSuite.CleanUp()
 }
+
+func TestApplicationOperator_LabelOverrides(t *testing.T) {
```
koala7659, line 60: Maybe it would be good idea to test negative case. For example if application is labeled with any other value (nochange.deployment.args.skipVerify) . This should not cause application upgrade and does not affect the  gateway deployment

koala7659 (commented):
Please see my comments

components/application-operator/pkg/application-controller/reconciler.go
```diff
@@ -129,6 +129,8 @@ func (r *applicationReconciler) manageInstallation(application *v1alpha1.Applica
 		}
 
 		return r.installApplication(application)
+	} else if !application.ShouldSkipInstallation() {
+		r.releaseManager.UpgradeApplicationRelease(application)
```
koala7659, line 133: This reveals problem. We probably reinstall application whenever any update occurs (reconcile method is called). We should do it only if the update concerns change of override labels for the application.

koala7659 (changes requested):
Please see my comments

tests/application-operator-tests/test/applicationcontroller/tests/applicationcontroller_test.go
ralikio: I do agree that such cases must be tested but I have put them in smaller unit tests not to make whatever runs against cluster even heavier. For example, the mentioned labels without overrides are tested by kyma/components/application-operator/pkg/kymahelm/application/overrides_test.go:11. I admit that the test case was not obvious. Let me just add another one there to assure that labels without override prefix does not cause any upgrades.

ralikio:
\retest

koala7659 (dismissed):

components/application-operator/pkg/utils/string_map.go
```diff
@@ -0,0 +1,45 @@
...
+	for key, applicationValue := range labels {
+		if releaseVal, ok := m[key]; !ok || applicationValue != releaseVal {
+			return false
+		}
+	}
+
+	return true
+}
+
+func (m StringMap) flatten(parentKey string, source map[string]interface{}) {
```
akgalwas, line 31: It could be a bit shorter:
```
func flatten(prefix string, source InterfaceMap) {	
	for key, value := range source {
		newKey := prefix + key
		
		if casted, ok := value.(InterfaceMap); ok {
			flatten(newKey + Separator, casted)
		} else {
			m[newKey] = value.(string)
		}
	}
}
```

Flatten method could be also implemented as a regular function (not a method receiver):

```
func flatten(prefix string, source InterfaceMap, target StringMap) {	
	for key, value := range source {
		newKey := prefix + key
		
		if casted, ok := value.(InterfaceMap); ok {
			flatten(newKey + Separator, casted, target)
		} else {
			target[newKey] = value.(string)
		}
	}
}
```
NewStringMap could be implemented as follows:
```
func NewStringMap(source InterfaceMap) StringMap {
	return flatten(source, make(StringMap))
}
```

ralikio: Corrected according to the first snippet. Do you have in mind any performance benefits in terms of using functions over methods?

koala7659 (approved):
I think everything is fine

janmedrek (approved):

Issue #10036, opened 2020-11-25
Use non-privileged build tool


kwiatekus:
**Description**

We are looking for the docker build tool that allows for as little permissions as possible:
- can run in non-privileged container (MUST)
- can run as non-root ( STRETCH)

The most promising are kaniko and makitu.
We need to setup those tools in clusters with non-privileged policy enabled and compare them

**Reason**

We need to enable non-privileged container policy
It is a follow up for https://github.com/kyma-project/kyma/issues/9976

PR #10292, opened 2021-01-07
Non-privileged serverless container builder

pPrecel:
**Description**

- Add `Pod Security Policy` for running building jobs with smaller privileges
- Add mechanism of injecting `ServiceAccounts`, `Roles` and `RoleBindings` to every existing namespace

**Related issue(s)**

https://github.com/kyma-project/kyma/issues/10036

### DIFF
--- a/components/function-controller/internal/controllers/kubernetes/shared.go
+++ b/components/function-controller/internal/controllers/kubernetes/shared.go
@@ -11,6 +11,9 @@ import (
 const (
 	ConfigLabel              = "serverless.kyma-project.io/config"
 	RuntimeLabel             = "serverless.kyma-project.io/runtime"
+	RbacLabel                = "serverless.kyma-project.io/rbac"
+	RoleLabelValue           = "role"
+	RoleBindingLabelValue    = "rolebinding"
 	CredentialsLabelValue    = "credentials"
 	ServiceAccountLabelValue = "service-account"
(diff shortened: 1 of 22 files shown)


components/function-controller/internal/controllers/kubernetes/shared.go
```diff
@@ -19,6 +22,8 @@ const (
 type Config struct {
 	BaseNamespace                 string        `envconfig:"default=kyma-system"`
 	ExcludedNamespaces            []string      `envconfig:"default=istio-system;knative-eventing;kube-node-lease;kube-public;kube-system;kyma-installer;kyma-integration;kyma-system;natss;compass-system"`
+	RoleRequeueDuration           time.Duration `envconfig:"default=1m"`
+	RoleBindingRequeueDuration    time.Duration `envconfig:"default=1m"`
```
aerfio, line 26: add those values to helm chart as envs, so that we can override them during installation if that's needed

aerfio (changes requested):
Overall LGTM, just 1 small comment

components/function-controller/internal/controllers/kubernetes/shared.go
pPrecel: done

aerfio (approved):

PR #10295, opened 2021-01-07
Increase ory-hydra probes timeout to 5 secs

ebensom:
**Description**

It seems that sporadically it takes more than 1 second for ory-hydra to respond to the health and live requests.
When such thing happens, istio-proxy times out after one second (since that is the default value for `timeoutSeconds` which is propagated to Istio during probe-rewrite) and responds with 500 instead.

The issue causes SLA availability budget burn for ory-hydra.


strekm (approved):
/meow

PR #10315, opened 2021-01-12
Migrate App Conn Validator logging to Zap

franpog859:
**Description**

- Migrate Application Connectivity Validator logging to the [unified approach proposed here](https://github.com/kyma-project/community/tree/master/internal/proposals/logs)
- Prepare a common library so every other component can easily be migrated to the unified approach


components/application-connectivity-validator/cmd/applicationconnectivityvalidator/options.go
```diff
@@ -46,6 +48,8 @@ func parseArgs() *options {
 	kubeConfig := flag.String("kubeConfig", "", "Path to a kubeconfig. Only required if out-of-cluster.")
 	masterURL := flag.String("masterURL", "", "The address of the Kubernetes API server. Overrides any value in kubeconfig. Only required if out-of-cluster.")
 	syncPeriod := flag.Duration("syncPeriod", 120*time.Second, "Sync period in seconds how often controller should periodically reconcile Application resource.")
+	logFormat := flag.String("logFormat", "text", "Logs format.")
```
janmedrek, line 51: Is that in line with the proposal? I believe the parameters should be passed as environment variables.

janmedrek (changes requested):

components/application-connectivity-validator/cmd/applicationconnectivityvalidator/options.go
dbadura: Shall we change the proposal?
In my opinion if we accept command line arguments, we shouldn't introduce the approach with envs.

dbadura:
Do we need to double the folder `logger`?

components/application-connectivity-validator/cmd/applicationconnectivityvalidator/options.go
janmedrek: I don't like that to be honest. This way the Unified Logging is not unified at all, the goal was to set up a *single* way in which the level of logging can be changed - introducing multiple other possibilities breaks the idea of the whole proposal. Is introducing envs really that much work to consider changing the proposal?

Have you done the research on how many components use the parameters instead of the envs? If there is a vast majority doing so then we can consider that but still, we should communicate that clearly.

franpog859: Some of the components use command line arguments instead of env variables. I think it's much cleaner to use it this way and not to introduce a separate mechanism just for logging purposes. Another option is to migrate all the options to the env variables but I'm afraid it could end up with a scope creep. I think this solution is ok because the logging information is propagated via Helm charts and is easily changeable as the proposal states. What do you think?

janmedrek: With this solution, we will have one way to set up the value at the Helm chart level (which is okay), but changing that in the deployment spec would be one of two alternative ways (envs vs parameters).

I just want this to be established in a unified manner. I do not see a problem with introducing a small part of logic that would fetch the env variable in a component utilizing parameters (or vice versa in other cases, if we decide to go that way). We do not need to migrate all the existing options to env variables only for that purpose.

franpog859:
@dbadura I changed it to:
```
logging
├── logger
└── tracing
```
I didn't want to create two totally separate libraries as the tracing one is useful while used with logger one

components/application-connectivity-validator/cmd/applicationconnectivityvalidator/options.go
franpog859: 👍 I decided to go with the env variable as the majority of our components use them and it's fine with the proposal

janmedrek (dismissed):

dbadura:
> @dbadura I changed it to:
> 
> ```
> logging
> ├── logger
> └── tracing
> ```
> 
> I didn't want to create two totally separate libraries as the tracing one is useful while used with logger one

What do you think about such design:
> ```
> logging
> ├──tracing
>└── all logging go files
> ```

dbadura (dismissed):

franpog859:
I created a follow-up issue for the short documentation on how to use the unified logging library - https://github.com/kyma-project/kyma/issues/10458

Maladie (approved):

common/logging/logger/format.go
```diff
@@ -0,0 +1,44 @@
...
+	var format = Format(input)
+	switch format {
+	case JSON, TEXT:
+		return format, nil
+	default:
+		return format, errors.New(fmt.Sprintf("Given log format: %s, doesn't match with any of %v", format, all_formats))
+	}
+}
+
+func (f Format) ToZapEncoder() zapcore.Encoder {
```
m00g3n, line 30: Is the scope of this method must me public?

m00g3n (commented):
LGTM just couple of minor comments

m00g3n:
LGTM

m00g3n (dismissed):

Maladie (approved):

janmedrek (approved):

PR #10327, opened 2021-01-13
Update the curl version to latest in the end to end external solution test

montaro:
Fix for the security issue: CVE ID: CVE-2020-8169

Installing the latest version of cURL in alpine 3.10 is curl-7.66.0-r3 which is not matching the recommended version of cURL 7.71.0, so I force installing the latest version of cURL from the edge branch.

### DIFF
--- a/tests/end-to-end/external-solution-integration/Dockerfile
+++ b/tests/end-to-end/external-solution-integration/Dockerfile
@@ -12,6 +12,7 @@ LABEL source=git@github.com:kyma-project/kyma.git
 
 COPY --from=builder /e2e /
 
+# temporarily install latest cURL to mitigate the security issues
 RUN apk update && apk add ca-certificates && apk add --no-cache --update-cache --repository http://nl.alpinelinux.org/alpine/edge/main/ curl && rm -rf /var/cache/apk/*
 
 ENTRYPOINT ["/e2e"]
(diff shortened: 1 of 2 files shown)


tests/end-to-end/external-solution-integration/Dockerfile
```diff
@@ -12,6 +12,7 @@ LABEL source=git@github.com:kyma-project/kyma.git
 
 COPY --from=builder /e2e /
 
-RUN apk update && apk add ca-certificates && apk add curl && rm -rf /var/cache/apk/*
+# temporarily install latest cURL to mitigate the security issues
+RUN apk update && apk add ca-certificates && apk add --no-cache --update-cache --repository http://nl.alpinelinux.org/alpine/edge/main/ curl && rm -rf /var/cache/apk/*
```
sayanh, line 16: As we updating the alpine cache to `edge` I think 
```
FROM alpine:3.10
```
this is a smaller change that will do that job.

At the same time, let's create an issue on us to change `edge` later to a released version when it is possible.

sayanh:
Can you please change the image(to `eu.gcr.io/kyma-project/external-solution-end-to-end-tests:PR-10327`) for the test in this PR itself so that it proves that nothing is breaking?

tests/end-to-end/external-solution-integration/Dockerfile
montaro: @sayanh I can upgrade the `apline` version to the latest, however, I tested that locally and the `curl` version available is still not the latest `curl-7.69`

sayanh (approved):

montaro:
Upgrading to `alpine:edge` to get the `curl 7.74.0`

PR #10350, opened 2021-01-15
Add Istio production profile

kubadz:
**Description**

I intentionally omitted adding a production file for minikube, as it doesn't make sense

- add production profile for Istio

**Related issue(s)**


Demonsthere (approved):
/meow

PR #10370, opened 2021-01-18
Builds for debug images

ralikio:
**Description**

The pull request contains changes required to enhance app debugging process by automatically building containers with dlv debugger. Currently, it is mandatory for developers to build debug image manually. Debugging is a common process in development and so should be made much easier than it is now, not requiring creation of own Dockerfile that does not differ across multiple development setups.

- Dockerfile with `.debug` suffix containing all necessary debug utilities.
- Common makefile update for debug image builds during release process.
- New Entrypoint script in components Dockerfiles to allow components' processes to be run with PID of value 1.

If changes are accepted they need to be propagated to all Kyma components.

**Related issue(s)**


common/makefiles/generic-make-go.mk
```diff
@@ -87,9 +88,13 @@ release-dep: resolve dep-status verify build-image push-image
 .PHONY: build-image push-image
 build-image: pull-licenses
 	docker build -t $(IMG_NAME) .
+	docker build --build-arg BASE_IMAGE_NAME=$(DEBUG_IMG_NAME) -t $(DEBUG_IMG_NAME).debug - < $(SCRIPTS_DIR)/debug/debug.Dockerfile
```
ralikio, line 91: Typo in BASE_IMAGE_NAME value, should be `BASE_IMAGE_NAME=$(IMG_NAME)`

Issue #10110, opened 2020-12-01
Improved operations for fluent-bit with dashboard, alert rules and aliases


a-thaler:
**Description**

Fluent-bit documents where to find recommended dashboards and alert rules. We should use them. Furthermore, there is an alias feature available so that the outputs are well-named in the related fluent-bit metrics.

- [ ] Bundle proper dashboards and alerts, see https://docs.fluentbit.io/manual/administration/monitoring#grafana-dashboard
- [x] Introduce aliases for all pre-configured outputs, see https://docs.fluentbit.io/manual/v/1.5/administration/monitoring#configuring-aliases (implemented with https://github.com/kyma-project/kyma/pull/10284)

Requirements:
- There is a nice way to see in grafana the throughput and health infos for fluent-bit pipelines
- Alerts are getting raised if an error rate for a pipeline is getting to high over some time range

**Reasons**

**Attachments**

suleymanakbas91:
Aliases are already added with this PR https://github.com/kyma-project/kyma/pull/10284.

PR #10463, opened 2021-01-27
Add alert and improve the dashboard for Fluent Bit

suleymanakbas91:
**Description**

- Add alert and improve the dashboard for Fluent Bit

**Related issue(s)**

See https://github.com/kyma-project/kyma/issues/10110


tobiscr (approved):

skhalash (approved):
While testing we realized that the Loki plugin has a bug, which takes down the metrics endpoint if Loki is not reachable. Simply speaking, if FluentBit cannot push to Loki -> the whole monitoring is broken. Unfortunately, nothing to do on our side. Hopefully, it will be fixed in the last version we are going to upgrade to.

PR #10464, opened 2021-01-27
Work around to use wildcard in path for apirules

sayanh:
**Description**

- Working around to use wildcard in path for apirules

**Related issue(s)**


sayanh:
Not needed anymore.

Issue #10201, opened 2020-12-11
Function should not create HPA when min and max replicas are set to 1


rakesh-garimella:
**Description**

If we create a function and set min and max replicas to 1 then a HPA is created with min and max replicas set to 1. As the max replicas is already reached it triggers prometheus alerts (Hpa Maxed Out). Which is not nice since by setting min and max replicas to same value the user does not want to scale out. 

**Expected result**

We should not create HPA when min and max replicas are set to same value. It should be created when min replicas is less than max replicas.

aerfio:
@kwiatekus @tgorgol The fix is already merged so I close this

PR #10482, opened 2021-01-28
Do not create hpa when minReplicas==maxReplicas in function-controller

aerfio:
**Description**

- If function uses fixed number of replicas no HPA will be used

**Related issue(s)**

#10201


components/function-controller/internal/controllers/serverless/horizontalpodautoscaler.go
```diff
@@ -29,9 +29,9 @@ func (r *FunctionReconciler) onHorizontalPodAutoscalerChange(ctx context.Context
 	newHpa := r.buildHorizontalPodAutoscaler(instance, deploymentName)
 
 	switch {
-	case len(hpas) == 0:
+	case len(hpas) == 0 && instance.Spec.MinReplicas != instance.Spec.MaxReplicas:
 		return r.createHorizontalPodAutoscaler(ctx, log, instance, newHpa)
-	case len(hpas) > 1: // this step is needed, as sometimes informers lag behind reality, and then we create 2 (or more) hpas by accident
+	case len(hpas) > 1 || instance.Spec.MinReplicas == instance.Spec.MaxReplicas: // this step is needed, as sometimes informers lag behind reality, and then we create 2 (or more) hpas by accident
```
aerfio, line 34: case len(hpas) > 1 || (len(hpas)==1 && instance.Spec.MinReplicas == instance.Spec.MaxReplicas)

components/function-controller/internal/controllers/serverless/horizontalpodautoscaler.go
```diff
@@ -21,22 +21,39 @@ func (r *FunctionReconciler) isOnHorizontalPodAutoscalerChange(instance *serverl
 	}
 
 	newHpa := r.buildHorizontalPodAutoscaler(instance, deployments[0].GetName())
-	return !(len(hpas) == 1 &&
-		r.equalHorizontalPodAutoscalers(hpas[0], newHpa))
+	scalingEnabled := isScalingEnabled(instance)
+	numHpa := len(hpas)
+
+	return (scalingEnabled && numHpa != 1) ||
```
aerfio, line 27: Could you add test for that function? It's now grown big enough to have those I think

aerfio (commented):
Look great @tgorgol 👍🏻

components/function-controller/internal/controllers/serverless/function_reconcile_test.go
```diff
@@ -870,6 +870,134 @@ var _ = ginkgo.Describe("Function", func() {
...
+		gomega.Expect(err).To(gomega.BeNil())
+		gomega.Expect(result.Requeue).To(gomega.BeFalse())
+		gomega.Expect(result.RequeueAfter).To(gomega.Equal(time.Minute * 5))
+
+		gomega.Expect(resourceClient.Get(context.TODO(), request.NamespacedName, function)).To(gomega.Succeed())
+		gomega.Expect(function.Status.Conditions).To(gomega.HaveLen(conditionLen))
+		gomega.Expect(reconciler.getConditionStatus(function.Status.Conditions, serverlessv1alpha1.ConditionConfigurationReady)).To(gomega.Equal(corev1.ConditionTrue))
+		gomega.Expect(reconciler.getConditionStatus(function.Status.Conditions, serverlessv1alpha1.ConditionBuildReady)).To(gomega.Equal(corev1.ConditionTrue))
+		gomega.Expect(reconciler.getConditionStatus(function.Status.Conditions, serverlessv1alpha1.ConditionRunning)).To(gomega.Equal(corev1.ConditionTrue))
+		gomega.Expect(reconciler.getConditionReason(function.Status.Conditions, serverlessv1alpha1.ConditionRunning)).To(gomega.Equal(serverlessv1alpha1.ConditionReasonDeploymentReady))
```
rJankowski93, line 999: Should we check if change MaxReplicas to 2 will remove the HPA?

tgorgol: We're checking the transition from a fixed number of replicas to a range earlier (starting at line 904), do you think repeating this check here would allow us to check different logic?

rJankowski93: You're right. I thought that on the start function don't have min and max, but now I see that after line 882 min=1 and max=2, so scenario "deleting hpa" is checked.

tgorgol (approved):

Issue #10336, opened 2021-01-14
TLS Certificate management


strekm:
**Description**

Implement a new way of certificate management. Integrate [cert-manager](https://cert-manager.io/docs/) as an optional Kyma component, this component should be considered as a pre-requisite for Kyma. Certificates shouldn't be managed by the helm, this solution proved to be hard in maintenance & buggy. 

Introduce Kyma pre-requisite component to manage certificate installation. Following options should be possible:
- for local development cert-manager shouldn't be used, user provides certs
- usage of gardener
- cert-manager + domain name & issuer provided 

AC:
- no support for xip
- by default certificate manager is not installed on local cluster (ks3)
- the user is able to provide domain & certificates
- certificate manager is an optional Kyma component
- documentation updated

Consider removal of apiserver-proxy & iam-kubeconfig-svc components

**Reasons**

Improved & unified certificates management will improve maintaining certificates, refreshing certificates and it will make Kyma upgrades less buggy in certificates POV. 

**Attachments**

PR #10489, opened 2021-01-29
Add cert-manager helm chart, add cert-manager CRDs to cluster-essentials

mjakobczyk:
**Description**

- add `cert-manager` Helm chart to resources
- split `cert-manager` CRDs to multiple files and add them to `files` CRDs dir
- update Istio label job to include `cert-manager`

**Related issue(s)**

Implements #10336


resources/istio/templates/job-namespace-label.yaml
```diff
@@ -25,10 +25,20 @@ spec:
           - /bin/bash
           - -c
           - |
-            set -e
+            set +e
             {{- range .Values.kyma.namespaces2Label }}
-            echo "---> Setting label to {{ . }}"
-            kubectl label namespace {{ . | quote }} "istio-injection=disabled" --overwrite
+            if RESULT=$(kubectl get namespace {{ . }} 2>&1) ; then
```
Demonsthere, line 30:
```suggestion
            if RESULT=$(kubectl get namespace --ignore-not-found=true {{ . }} 2>&1) ; then
```

resources/istio/templates/job-namespace-label.yaml
```diff
@@ -25,10 +25,20 @@ spec:
...
+            else
+              NS_NOT_FOUND_ERROR="namespaces {{ . }} not found"
+              if [[ $RESULT =~ .*"$NS_NOT_FOUND_ERROR".* ]] ; then
+                echo "---> Namespace {{ . }} could not be labelled, as it was not found. Skipping."
+              else
+                echo $RESULT
+                exit 1
+              fi
+            fi
             {{- end }}
```
Demonsthere, line 42:
```suggestion
            {{- end }}
       env:
         - name: HOME
           value: /tmp
```

resources/istio/templates/job-namespace-label.yaml
```diff
@@ -25,10 +25,23 @@ spec:
           - /bin/bash
           - -c
           - |
-            set -e
+            set +e
```
Tomasz-Smelcerz-SAP, line 28: I'd say it's against good bash practice. Every script should have "-e" at the top.
You can enclose individual statements, that are expected to cause errors, in the +e ... -e "guard", although I'd use that as a last resort technique. Try to use the command so that it doesn't fail or handle the failures at the place they occur.

Tomasz-Smelcerz-SAP (changes requested):

resources/istio/templates/job-namespace-label.yaml
```diff
@@ -25,10 +25,23 @@ spec:
...
             {{- range .Values.kyma.namespaces2Label }}
-            echo "---> Setting label to {{ . }}"
-            kubectl label namespace {{ . | quote }} "istio-injection=disabled" --overwrite
+            if RESULT=$(kubectl get namespace {{ . }} 2>&1) ; then
+              echo "---> Setting label to {{ . }}"
+              kubectl label namespace {{ . | quote }} "istio-injection=disabled" --overwrite
```
Tomasz-Smelcerz-SAP, line 32: Can we improve this a little bit? I personally find mixing bash with Helm templates a terrible idea. In this way you can't actually run/test/lint the script locally. I prefer the "old-school" approach: Have a normal bash script in file, and invoke it with paramers. Then you can do both: run/test it locally and invoke it in the job, rendering it's parameters using templates, instead of "injecting" them directly into script's statements.

resources/istio/templates/job-namespace-label.yaml
mjakobczyk: That was for debugging purposes during script development. Changed back to `-e`.

Tomasz-Smelcerz-SAP (dismissed):

Tomasz-Smelcerz-SAP (dismissed):

majakurcius (commented):
Please read through my comments and apply :)

resources/cert-manager/README.md
```diff
@@ -0,0 +1,61 @@
...
+The `crds.yaml` file has been split into multiple files. Each of these contains exactly one CRD with its name indicating the resource type. They are available in the `files/` directory.
+
+### Installation
+
+To install `cert-manager`, use the Kyma Installer and follow these steps:
+
+1. Manually apply CRDs from the `files/` directory:
+
+   ```bash
+   # Kubernetes 1.15+
```
majakurcius, line 35: I just noticed that. In prerequisites you have Kubernetes 1.11+, but in the instructions for the **only** way to install it, you have Kubernetes 1.15+? 
I'm not sure about this...

resources/cert-manager/README.md
```diff
@@ -0,0 +1,61 @@
...
+
+Before you configure `cert-manager` to issue certificates, you must first create Issuer resources. To learn more, read about [Issuers configuration](https://cert-manager.io/docs/configuration/).
+
+One of the common use cases for `cert-manager` is securing ingress resources. Read the [Securing Ingresses Resources documentation](https://cert-manager.io/docs/usage/ingress/) for more details.
+
+For more information, check the [official `cert-manager` documentation](https://cert-manager.io/docs/).
+
+## Troubleshooting
+
+One potential issue may occur while creating an Issuer or ClusterIssuer resources. It might result in failed calling webhook "webhook.cert-manager.io". A workaround that can be applied in this case is to delete both cert-manager's webhooks:
```
majakurcius, line 56:
```suggestion
One potential issue may occur while creating Issuer or ClusterIssuer resources. It might result in a failed call to the `webhook.cert-manager.io` webhook. A workaround for this is to delete these two `cert-manager`'s webhooks:
```
- The `an` would refer to the main noun, but the main noun (`resources`) is in plural, so it should be removed :) 
Unless you wrote it as `an Issuer's resources`, then `an` would refer to the Issuer.

- `that can be applied in this case` -> `for this`, nothing wrong with the way you wrote it, but this way it's shorter and conveys the same message :)

- `both` -> `these two` or `these`, unless the `cert-manager` has only two webhooks, then `both` is also fine.

resources/cert-manager/README.md
```diff
@@ -0,0 +1,61 @@
...
+>to renew certificates at an appropriate time before expiry.
+
+
+For more information check [links](#links) section.
+
+## Prerequisites
+
+- Kubernetes 1.11+
+
+## Details
```
majakurcius, line 22: Please revert it to the way it was - remove the `## Details` section and change its subsections back to sections.

resources/cert-manager/README.md
```diff
@@ -0,0 +1,55 @@
...
+For more information, check the [official `cert-manager` documentation](https://cert-manager.io/docs/).
+
+## Troubleshooting
+
+One potential issue may occur while creating Issuer or ClusterIssuer resources. It might result in a failed call to the `webhook.cert-manager.io` webhook. A workaround for this is to delete these two `cert-manager`'s webhooks:
+
+   ```bash
+   $ kubectl delete mutatingwebhookconfiguration.admissionregistration.k8s.io cert-manager-webhook
+   $ kubectl delete validatingwebhookconfigurations.admissionregistration.k8s.io cert-manager-webhook
+   ```
```
majakurcius, line 55:
```suggestion
```bash
$ kubectl delete mutatingwebhookconfiguration.admissionregistration.k8s.io cert-manager-webhook
$ kubectl delete validatingwebhookconfigurations.admissionregistration.k8s.io cert-manager-webhook
```
```

**Review by majakurcius — APPROVED** [review]

**Review by majakurcius — APPROVED** [review]

File: resources/cert-manager/values.yaml [hunk 9]
```diff
@@ -0,0 +1,354 @@
...
+  imagePullSecrets: []
+  # - name: "image-pull-secret"
+
+  # Optional priority class to be used for the cert-manager pods
+  priorityClassName: ""
+  rbac:
+    create: true
+
+  podSecurityPolicy:
+    enabled: false
```
Demonsthere, line 17: Please enable this, as PSP is already enabled on master :)

resources/cert-manager/README.md
mjakobczyk: We have agreed to remove `Prerequisites` section entirely.

Demonsthere (dismissed):
/meow

majakurcius (approved):

Tomasz-Smelcerz-SAP (approved):

resources/istio/values.yaml
```diff
@@ -9,6 +9,7 @@ kyma:
   namespaces2Label:
     - istio-system
     - kube-system
+    - cert-manager
```
pbochynski, line 12: Does it make sense? Job in istio will be executed before cert-manager is deployed. Namespace won't be labeled.

mjakobczyk: It could be labelled as long as cert-manager component is installed before Istio. The job itself belongs to Istio, so the order in this file does not matter.

pbochynski (changes requested):
Separate CRDs from helm chart. Put them in separate folder and apply in the new installation library.
Remove the script that labels cert-manager namespace. Just create namespace for cert-manager with istio-injection disabled annotation

resources/istio/values.yaml
pbochynski: I cannot agree. Introducing dependencies between components based on the order of installation is against any good practices. I spent last year fighting with dependencies and making parallel installation possible.

mjakobczyk: I mean in this case I have not found any better way to deal with this to support old installer approach. Of course it is not a good practice to introduce any kind of dependencies. Here I simply reused what was in-place.

pbochynski (approved):

resources/cert-manager/README.md
```diff
@@ -0,0 +1,57 @@
...
+
+To install `cert-manager`, use the Kyma Installer and follow these steps:
+
+1. Manually apply CRDs from the project root directory:
+
+   ```bash
+   $ kubectl apply -f ./installation/resources/crds/cert-manager
+   ```
+
+2. Manually apply namespace resource from the project root directory:
```
majakurcius, line 29:
```suggestion
2. Manually apply the Namespace resource from the project root directory:
```
or `Namespace resources`

majakurcius (dismissed):

majakurcius (approved):
Approving conditionally, changes from one commit that got removed by the rebase need to be re-applied in a follow-up.

pbochynski (approved):

Demonsthere (approved):
/meow

PR #10490, opened 2021-01-29
Mute istio discovery, ingressgateway on evaluation and production profile

dbadura:
**Description**

- mute istio discovery and other istio components except istio-proxy

**Related issue(s)**


Demonsthere (approved):
/meow

PR #10498, opened 2021-02-01
modified components list for parallel-installation

strekm:
**Description**

- modified components list for parallel-installation
- introducing k3s components list

**Related issue(s)**


tobiscr:
Yes, agree that these components currently fail on K3s.
But this change would imply that either the user or the CLI have to be aware that K3s needs a special configuration.
Can't we fix that directly on the component level - means the component won't install itself if it's not able to run inside of the target cluster?

tobiscr:
We will take care on CLI level to remove these incompatible components from the components list.

PR #10513, opened 2021-02-03
Update cluster-essentials to support SecurityContext

Demonsthere:
**Description**

- Add SecurityContext to cluster-essentials
- Add global PSP to kyma

**Related issue(s)**
- Implements https://github.com/kyma-project/kyma/issues/9538


resources/cluster-essentials/templates/psp-unprivileged.yaml
```diff
@@ -0,0 +1,26 @@
...
+  name: kyma.unprivileged
+  annotations:
+    helm.sh/hook: "pre-upgrade, pre-install"
+    helm.sh/hook-weight: "0"
+    helm.sh/hook-delete-policy: "before-hook-creation"
+spec:
+  allowPrivilegeEscalation: false
+  allowedCapabilities:
+    # Istio-init required
+    - NET_ADMIN
```
mr-ma, line 14: I'm not sure if we should limit allowed capabilities at this stage. This was not in the scope of the privileged containers issue, so I'm not sure which components might get affected. Not listing allowed capabilities here seems more safe for this iteration.

resources/cluster-essentials/templates/psp-privileged.yaml
```diff
@@ -0,0 +1,24 @@
...
+apiVersion: policy/v1beta1
+kind: PodSecurityPolicy
+metadata:
+  name: kyma.privileged
+  annotations:
+    helm.sh/hook: "pre-upgrade, pre-install"
+    helm.sh/hook-weight: "0"
+    helm.sh/hook-delete-policy: "before-hook-creation"
+spec:
+  allowPrivilegeEscalation: true
```
mr-ma, line 11: I'm also not sure why do we need to have a privileged psp. I know Gardener does this, but I don't see its usage in the case of Kyma right now.

resources/cluster-essentials/templates/rbac-clusterrole.yaml
```diff
@@ -0,0 +1,28 @@
...
+  name: kyma:psp:privileged
+  annotations:
+    helm.sh/hook: "pre-upgrade, pre-install"
+    helm.sh/hook-weight: "0"
+    helm.sh/hook-delete-policy: "before-hook-creation"
+rules:
+- apiGroups: ["policy"] # "" indicates the core API group
+  resources: ["podsecuritypolicies"]
+  verbs: ["use"]
+  resourceNames: ["kyma.privileged"]
```
mr-ma, line 14: same as above, I'm not sure what privileged psp brings to us.

resources/cluster-essentials/templates/psp-privileged.yaml
Demonsthere: I added it as it may be prefered/wanted by some OpenSource users, or some use cases (evaluation profile?). This approach simplifies this.

mr-ma: Ok makes sense!

strekm (approved):
/meow

suleymanakbas91 (approved):

Issue #10218, opened 2020-12-14
application name needs to be cleaned from non alphanumeric characters before sending to BEB


k15r:
```filter:
dialect: beb
    filters:
    - eventSource:
        property: source
        type: exact
        value: /default/sap.kyma/kh
      eventType:
        property: type
        type: exact
        value: sap.kyma.custom.commerce-mock.order.cancelled.v1
```

the application name needs to be cleaned to match `/\w+/`, otherwise this error happens on webhook creation:
DomainNamespace segment of Type attribute does not comply with SAP Event Specification

k15r:
in the subscription this means that the eventType value needs to adhere to the same rule

berkanercan:
Summary of the issue: We include application names in the `eventType` in order to differentiate same events that are sent from different applications. As this bug points out, having dash( ex. `commerce-mock`) in `eventType.value` is not compliant with SAP Event Specification. 

If we replace the dash in the application name with a dot, it will be compliant but there is an edge case:

Application names are following [RFC 1123](https://kubernetes.io/docs/concepts/overview/working-with-objects/names/), which allow them to have multiple dashes. This means we can have `commerce-mock` and `commerce--mock` as separate applications. If we replace dashes with a dot, we will be casting both `commerce-mock` and `commerce--mock` into `commerce.mock`.

In the future, applications will be sending events directly to BEB and not over Kyma. This means we won't be including the application name in the event type.

Considering all these information, what is the best way to follow?

sayanh:
On another thought, `\w+`  matches _ (underscore) but RFC 1123 doesn’t allow `_`.
So, `test-1` as application can be `test_1` in event_type.
and `test--1` as application can be `test__1` in event_type.
And BEB does not like it either. So `\w+` is not good enough.

k15r:
Requirements:
- [ ] no event collisions for different applications
- [ ] it shall be possible to optionally omit the application name from the event type. This would make subscriptions portable between clusters.

Possible solution:
- [ ] application name cleaned from all forbidden characters will be used by default in event-types
       `sap.kyma.custom.<clean-application-name>.<event-type>`. As this can cause collisions if application exist with same cleaned name, the documentation needs to be updated to warn about this behaviour.
- [ ] a new label shall be introduced on the application resource. This label `application-type: <sometype eg. commerce>` will be used by the event-publish-proxy to replace the `clean-application-name` part in the new event-type by the value of the label.
[...]

PR #10524, opened 2021-02-04
Update the images for the Event publishers and controllers

marcobebway:
**Description**

Update the images for the Event publishers and controllers:

- `event-publisher-nats`
- `event-publisher-proxy`
- `eventing-controller`
- `eventing-controller-nats`

**Related issue(s)**
#10218

### DIFF
--- a/resources/eventing/values.yaml
+++ b/resources/eventing/values.yaml
@@ -14,6 +14,11 @@ global:
   # if true BEB is enabled else NATS is enabled
   isBEBEnabled: false
 
+  # eventTypePrefix is the prefix of the eventType
+  # note that the eventType format is: eventTypePrefix.applicationName.eventName.eventVersion
+  # for example: sap.kyma.custom.myapp.order.created.v1 (where the eventName is order.created)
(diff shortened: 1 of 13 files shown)


resources/eventing/values.yaml
```diff
@@ -14,6 +14,11 @@ global:
   # if true BEB is enabled else NATS is enabled
   isBEBEnabled: false
 
+  # eventTypePrefix is the prefix of the eventType
+  # note that the eventType format is: eventTypePrefix.applicationName.eventName.eventVersion
+  # for example: sap.kyma.custom.myapp.order.created.v1 (where the eventName is order.created)
+  eventTypePrefix: "sap.kyma.custom"
```
radufa, line 20: For the open source offering, an empty "eventTypePrefix" is preferable as the current default one: "sap.kyma.custom". Of course this should be configurable as a global value and used by the calling applications, if necessary

radufa: A following issue was created: https://github.com/kyma-project/kyma/issues/10579

radufa (approved):

PR #10319, opened 2021-01-13
Fix problem with certificates being removed from the secret during upgrade

Maladie:
**Description**

Changes proposed in this pull request:

- Add watcher.sh ability to create necessary secret on it's own

- Remove `app-connector-certs` secret from charts. From now on it will be created by watcher.sh, Runtime Agent (in Compass) or Certs Setup Job (on standalone Kyma)

**Related issue(s)**

https://github.com/kyma-project/kyma/issues/10120

PR #10128, opened 2020-12-03
Extending Compass Runtime Agent with a job to perform runtime synchronisation with compass after release upgrade

koala7659:

PR #10527, opened 2021-02-04
Delete not needed anymore job to resync SKR after Kyma upgrade

koala7659:
The job has been initially added in [this PR](https://github.com/kyma-project/kyma/pull/10128) as a workaround for problem with deleted SKR certificate during upgrade.

It is not needed after changes done in [this PR](https://github.com/kyma-project/kyma/pull/10319/files)


franpog859 (approved):
Cool!

PR #10533, opened 2021-02-05
Mute proxy

dbadura:
**Description**

- Mute istio-proxy
https://istio.io/latest/docs/tasks/observability/logs/access-log/

**Related issue(s)**

### DIFF
--- a/resources/istio/files/istio-operator-cluster-evaluation.yaml
+++ b/resources/istio/files/istio-operator-cluster-evaluation.yaml
@@ -172,6 +172,7 @@ spec:
             maxSurge: 100%
             maxUnavailable: 25%
   meshConfig:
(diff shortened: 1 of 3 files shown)


resources/istio/files/istio-operator-cluster-evaluation.yaml
```diff
@@ -172,7 +172,7 @@ spec:
             maxSurge: 100%
             maxUnavailable: 25%
   meshConfig:
-    accessLogFile: /dev/stdout
+    accessLogFile: ""
```
dbadura, line 175: I can add docs about it in comment, what do u think?

colunira: That's a good idea. Please add it in every file, so nobody forgets about it :)

strekm (approved):
/meow

PR #10585, opened 2021-02-12
Bump serverless images

aerfio:
**Description**

- as in title

### DIFF
--- a/resources/serverless/values.yaml
+++ b/resources/serverless/values.yaml
@@ -77,8 +77,8 @@ global:
 images:
   manager:
     repository: "eu.gcr.io/kyma-project/function-controller"
-    tag: "PR-10545"
-    pullPolicy: Always
+    tag: "8a8658b2"
+    pullPolicy: IfNotPresent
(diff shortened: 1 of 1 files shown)


resources/serverless/values.yaml
```diff
@@ -156,7 +156,7 @@ containers:
       functionBuildExecutorImage:
         value: gcr.io/kaniko-project/executor:v1.3.0
       functionBuildRepoFetcherImage:
-        value: eu.gcr.io/kyma-project/function-build-init:PR-10558
+        value: eu.gcr.io/kyma-project/function-build-init:f27d4182
```
dbadura, line 159: We could use also `8a8658b2`, because we build all images together.

dbadura (approved):

resources/serverless/values.yaml
aerfio: But there were no changes to that image in that PR, so if I wanted to see the changes based on that commit id then I'd find nothing

PR #10612, opened 2021-02-16
Certificates management for new Kyma

colunira:
**Description**

With the new Kyma we get rid of some components. This allows us to simplify the way we manage certificates and we are able to introduce some new solutions. In this PR I introduce a new chart - certificates. It will only work with the new Kyma - without console, apiserver-proxy and iam-kubeconfig-service.

- Introduce a new chart - certificates
- Add the new chart to components list and remove apiserver-proxy, iam-kubeconfig-service and console
- Set global.environment.gardener to false as default in the core component

**Related issue(s)**
https://github.com/kyma-project/kyma/issues/10336


Demonsthere (dismissed):
/meow

resources/certificates/README.md
```diff
@@ -0,0 +1,19 @@
...
+
+- `kyma-gateway-certs`
+- `apiserver-proxy-tls-cert`
+- optional `ingress-tls-cert`, which contains a copy of the TLS certificate from the `kyma-gateway-certs` Secret. It is only created if the certificate is self-signed.
+
+Each job in this chart handles a different scenario:
+
+- job-user-provided - Requires cert-manager installed on the cluster. User provides a ClusterIssuer CR. Then, Certificate CRs are created, which generate Secrets with TLS Certificate and private key based on the provided ClusterIssuer.
+- job-legacy - Support for the old way of managing certificates. User provides TLS Key and Certificate by overrides `global.tlsKey` and `global.tlsCrt`. Job puts them into secrets.
+- job-gardener - This scenario is used when working on Gardener. Certificate CRs are created, which generate Secrets with TLS Certificate and private key.
```
superojla, line 19:
```suggestion
- `job-user-provided` - Requires cert-manager installed on the cluster. User provides a ClusterIssuer CR. Then, Certificate CRs are created, which generate Secrets with TLS Certificate and private key based on the provided ClusterIssuer.
- `job-legacy` - Support for the old way of managing certificates. User provides TLS Key and Certificate by overrides `global.tlsKey` and `global.tlsCrt`. Job puts them into Secrets.
- `job-gardener` - This scenario is used when working on Gardener. Certificate CRs are created, which generate Secrets with TLS Certificate and private key.
```

superojla (dismissed):
lgtm 😸

resources/core/charts/gateway/templates/kyma-gateway-certs.yaml
```diff
@@ -1,4 +1,5 @@
-{{ if not .Values.global.environment.gardener }}
+{{ if not .Values.global.newInstallation }}
```
pbochynski, line 1: I don't like it. We were talking about that - we should not manage certs with helm. But introducing a global flag for that is ugly.

colunira: This flag was introduced to support new and old way of managing the certificates in parallel. In the new way this secret is not created in this place, but in the certificates chart. Once we get rid of kyma-installer this secret can be removed, as well as xip-patch chart.

I set this flag to false as a default, so we don't have to modify the old way of installation in any way. Temporarily we need to set it to true when installing Kyma in the new way.

pbochynski: Still I don't see any reason to introduce global flag. New installation should not install core, because you moved the secret and there is nothing more there. Old installation doesn't install certificates. Why do you need the flag?

pbochynski (changes requested):
Remove new installation global flag

resources/core/charts/gateway/templates/kyma-gateway-certs.yaml
```diff
@@ -1,4 +1,4 @@
-{{ if not .Values.global.environment.gardener }}
+ {{ if not .Values.global.environment.gardener }}
```
pbochynski, line 1: Why do you update the chart if core is removed from the component list? Shouldn't it be removed?

pbochynski (changes requested):

resources/core/charts/gateway/templates/kyma-gateway-certs.yaml
colunira: I'm removing it from components list for the new installation, but it is still being used by the old one.

pbochynski (dismissed):

pbochynski (approved):

superojla (approved):

Tomasz-Smelcerz-SAP (approved):

PR #10635, opened 2021-02-17
Make kyma-admin user name settable

jasiu001:

- Make kyma-admin user name settable, it is necessary to inject by overrides another name of the admin (owner of the cluster/SKR)


strekm (approved):
/meow

PR #10642, opened 2021-02-17
Fluent bit increase mem limits

dekiel:
Increase memory limit for fluent-bit.
Fixing OOMs in evaluation profile.
See https://github.com/kyma-project/test-infra/issues/3300


a-thaler:
There is a "test_file" committed with the PR, please remove

a-thaler:
The default limit in the values.yaml is as well at 128Mi. I guess we should increase that and remove a special setting from eval profile

rakesh-garimella (approved):

PR #10648, opened 2021-02-17
Istio increase mem limits

dekiel:
Increasing evaluation profile memory limits for pilot-discovery process.

See https://github.com/kyma-project/test-infra/issues/3300


strekm (approved):
/meow

strekm (approved):
/meow

Issue #10669, opened 2021-02-18
Enable in-cluster Eventing without creating Kyma Applications


marcobebway:
**Description**

The Eventing controllers (**eventing-controller-nats** and **eventing-controller**), should create NATS **Subjects** and  **Webhooks** (respectively) event if the given Kyma subscription contains an event-type for non-existing Kyma Application. 

In addition, in-cluster legacy events should be supported.

This decouples in-cluster enabled Eventing from creating Kyma Applications.

PR #10670, opened 2021-02-18
Enable in-cluster Eventing without creating Kyma Applications

marcobebway:
**Description**

Enable in-cluster Eventing without creating Kyma Applications:
- Update the Event publishers to support publishing in-cluster legacy events.
- Update the Eventing controllers (eventing-controller-nats and eventing-controller), to be able to create NATS Subjects and Webhooks (respectively) event if the given Kyma subscription contains an event-type for non-existing Kyma Application.

**Related issue(s)**
#10669


components/eventing-controller/pkg/handlers/eventtype/clean_test.go
```diff
@@ -25,60 +26,65 @@ func TestCleaner(t *testing.T) {
...
+			givenApplicationName: "testapp",
+			givenEventType:       "sap.kyma.testapp.order.created", // missing the version
 			wantError:            true,
 		},
 		{
 			givenEventTypePrefix: "sap.kyma",
 			givenApplicationName: "testapp",
-			givenEventType:       "sap.kyma.test-app.order.created.v1",
-			wantEventType:        "sap.kyma.testapp.order.created.v1",
+			givenEventType:       "sap.testapp.order.created.v1", // prefix not matched
```
sayanh, line 79: How about using `name` for the testCase struct which will explain the context?

components/eventing-controller/pkg/handlers/eventtype/clean_test.go
```diff
@@ -25,60 +26,65 @@ func TestCleaner(t *testing.T) {
 			givenApplicationName: "testapp",
 			givenEventType:       "sap.kyma.testapp.order.created.v1",
 			wantEventType:        "sap.kyma.testapp.order.created.v1",
+			wantError:            false,
```
sayanh, line 29: What about adding tests for Clean() with empty application list?

components/eventing-controller/pkg/handlers/eventtype/clean_test.go
sayanh, line 79: How about adding `name` for the testCase struct to provide some context and later use `t.Run("...context...", func(){})`?

components/eventing-controller/pkg/handlers/eventtype/clean_test.go
marcobebway: Done

components/eventing-controller/pkg/handlers/eventtype/clean_test.go
marcobebway: Done

sayanh (approved):

pbochynski:
Is there any documentation how to use in-cluster messaging? What is the recommendation about event types? Can we publish directly to nats?

marcobebway:
> Is there any documentation how to use in-cluster messaging? What is the recommendation about event types? Can we publish directly to nats?

This will be done in this PR: https://github.com/kyma-project/kyma/pull/10820.

sayanh (approved):

PR #10716, opened 2021-02-23
Remove production config

adamwalach:
**Description**

Since we introduced profiles support for all our components this file is no longer used.
All refeneces in test-infra are also gone.

- removed `installer-config-production.yaml.tpl`

**Related issue(s)**
https://github.com/kyma-project/test-infra/pull/3142


strekm (approved):
/meow

Issue #10521, opened 2021-02-04
Event-publishing-proxy fails forwarding proper cloudevent


k15r:
this request 

```curl -v -k --data @<(cat <<EOF
        {
            "specversion": "1.0",
            "source": "varkes",
            "type": "${3:-stage.com.external.solution.order.created}",
            "eventtypeversion": "v1",
            "id": "A234-1234-1234",
            "data" : "{\"foo\":\"bar\"}"
        }
EOF
) -H "Content-Type: application/cloudevents+json" localhost:8080/publish
```

to the event-publishing-proxy should result in a 204 HTTP_RESPONSE, but it fails with  
```
time="2021-02-04T09:42:46Z" level=info msg="Event dispatched id:[A234-1234-1234] statusCode:[400] duration:[927.595928ms] responseBody:[{\"error\":{\"code\":\"INVALID_HTTP_HEADER\",\"message\":\"Required header 'Content-Type' is absent\"}}]"
```

Legacy events received on the legacy endpoint work as expected:
```
curl -v -k --data @<(cat <<EOF
 {
            "event-type": "stage.com.external.solution.order.created",
            "event-type-version": "v1",
            "event-time": "2020-04-02T21:37:00Z",
             "data" : "{\"foo\":\"bar\"}"
        }
EOF
        ) -H "Content-Type: application/json" localhost:8080/foo/v1/events
*   Trying ::1...
* TCP_NODELAY set
* Connected to localhost (::1) port 8080 (#0)
> POST /foo/v1/events HTTP/1.1
> Host: localhost:8080
> User-Agent: curl/7.64.1
> Accept: */*
> Content-Type: application/json
> Content-Length: 210
>
* upload completely sent off: 210 out of 210 bytes
< HTTP/1.1 200 OK
< Content-Type: application/json
< Date: Thu, 04 Feb 2021 09:57:29 GMT
< Content-Length: 76
<
{"event-id":"b69ff2f8-f70d-4ff0-b695-abb75c73b44d","status":"","reason":""}
* Connection #0 to host localhost left intact
* Closing connection 0
```

k15r:
upon further investigation i found that the issue is not the "Content-Type" that is used to indicate that the request itself is a structured cloudevent, but the datacontenttype in the structured cloudevent.

the following request works as expected:

```
curl -v -k -H "Content-Type: application/cloudevents+json" --data @<(cat <<EOF
        {
            "specversion": "1.0",
[...]

PR #10721, opened 2021-02-23
Preserve CE event's mode (structured or binary) 

radufa:
**Description**

- A structured ce will be not converted to binary mode anymore.

**Related issue(s)**

See also: https://github.com/kyma-project/kyma/issues/10521

### DIFF
--- a/components/event-publisher-proxy/pkg/cloudevents/utils.go
+++ b/components/event-publisher-proxy/pkg/cloudevents/utils.go
@@ -10,10 +10,6 @@ import (
 	cehttp "github.com/cloudevents/sdk-go/v2/protocol/http"
 )
 
-const (
-	CEStructuredMode = "application/cloudevents+json"
-)
-
 // WriteRequestWithHeaders writes a CloudEvent HTTP request with the given message and adds the given headers to it.
 func WriteRequestWithHeaders(ctx context.Context, message binding.Message, req *http.Request, headers http.Header, transformers ...binding.Transformer) error {
 	err := cehttp.WriteRequest(ctx, message, req, transformers...)
--- a/components/event-publisher-proxy/pkg/cloudevents/utils.go
+++ b/components/event-publisher-proxy/pkg/cloudevents/utils.go
@@ -10,6 +10,10 @@ import (
 	cehttp "github.com/cloudevents/sdk-go/v2/protocol/http"
 )
 
+const (
+	CEStructuredMode = "application/cloudevents+json"
+)
+
 // WriteRequestWithHeaders writes a CloudEvent HTTP request with the given message and adds the given headers to it.
 func WriteRequestWithHeaders(ctx context.Context, message binding.Message, req *http.Request, headers http.Header, transformers ...binding.Transformer) error {
 	err := cehttp.WriteRequest(ctx, message, req, transformers...)
--- a/components/event-publisher-proxy/pkg/handler/http/handler.go
+++ b/components/event-publisher-proxy/pkg/handler/http/handler.go
@@ -152,7 +152,7 @@ func (h *Handler) publishCloudEvents(writer http.ResponseWriter, request *http.R
 		return
 	}
 
-	if request.Header.Get(cev2http.ContentType) == cloudevents.CEStructuredMode {
+	if request.Header.Get(cev2http.ContentType) == cev2event.ApplicationCloudEventsJSON {
 		ctx = binding.WithForceStructured(ctx)
 	} else {
 		ctx = binding.WithForceBinary(ctx)
--- a/components/event-publisher-proxy/pkg/handler/http/handler.go
+++ b/components/event-publisher-proxy/pkg/handler/http/handler.go
@@ -154,6 +154,8 @@ func (h *Handler) publishCloudEvents(writer http.ResponseWriter, request *http.R
 
 	if request.Header.Get(cev2http.ContentType) == cloudevents.CEStructuredMode {
 		ctx = binding.WithForceStructured(ctx)
+	} else {
+		ctx = binding.WithForceBinary(ctx)
 	}
 
 	h.receive(ctx, event)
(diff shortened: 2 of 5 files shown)


components/event-publisher-proxy/pkg/cloudevents/utils.go
```diff
@@ -10,6 +10,10 @@ import (
 	cehttp "github.com/cloudevents/sdk-go/v2/protocol/http"
 )
 
+const (
+	CEStructuredMode = "application/cloudevents+json"
```
sayanh, line 14: Instead of redefining, can we `sdkgocloudevents.ApplicationCloudEventsJSON` where the import is: `sdkgocloudevents "github.com/cloudevents/sdk-go/pkg/cloudevents"`

components/event-publisher-proxy/pkg/handler/http/handler_test.go
```diff
@@ -460,14 +464,27 @@ func setupTestResources(t *testing.T, port, maxRequestSize int, applicationName,
...
+		if r.Header.Get(cev2http.ContentType) == cloudevents.CEStructuredMode {
+			// CE structured mode
+			message := cev2http.NewMessageFromHttpRequest(r)
+			defer func() { _ = message.Finish(nil) }()
+			if event, err := binding.ToEvent(context.Background(), message); err != nil {
+				return err
+			} else if strings.Contains(event.Type(), appName) {
+				return nil
+			} else {
+				return errors.New(fmt.Sprintf("expected the event-type [%s] to contain the application name [%s] for structured mode", event.Type(), appName))
```
sayanh, line 476: `errors.Newf("%s", str)` where the import is `"gopkg.in/errgo.v2/fmt/errors"` can be handy.

components/event-publisher-proxy/pkg/handler/http/handler_test.go
```diff
@@ -460,14 +464,27 @@ func setupTestResources(t *testing.T, port, maxRequestSize int, applicationName,
 
 func validateApplicationName(appName string) testingutils.Validator {
 	return func(r *http.Request) error {
+		if r.Header.Get(cev2http.ContentType) == cloudevents.CEStructuredMode {
+			// CE structured mode
+			message := cev2http.NewMessageFromHttpRequest(r)
+			defer func() { _ = message.Finish(nil) }()
+			if event, err := binding.ToEvent(context.Background(), message); err != nil {
+				return err
+			} else if strings.Contains(event.Type(), appName) {
```
sayanh, line 473: Why is this check needed?

Can we somehow remove if...elseif...else block?

components/event-publisher-proxy/pkg/sender/nats.go
```diff
@@ -41,16 +41,13 @@ func (h *NatsMessageSender) Send(ctx context.Context, event *cev2event.Event) (i
...
-	defer cancel()
-
 	c, err := cev2.NewClient(p)
 	if err != nil {
 		h.Logger.Errorf("Failed to create client, %s", err.Error())
 		return http.StatusInternalServerError, err
 	}
 
-	err = c.Send(ctxWithCancel, *event)
+	err = c.Send(ctx, *event)
```
sayanh, line 50: Why are we removing the context with cancel?

components/event-publisher-proxy/pkg/handler/http/handler.go
```diff
@@ -136,7 +136,7 @@ func (h *Handler) publishCloudEvents(writer http.ResponseWriter, request *http.R
 	ctx, cancel := context.WithTimeout(request.Context(), h.RequestTimeout)
 	defer cancel()
 
-	message := cehttp.NewMessageFromHttpRequest(request)
+	message := cev2http.NewMessageFromHttpRequest(request)
```
k15r, line 139: why do we:
transform from request to cloudevent
validate cloudevent
transform back to request

when we could just: proxy request, and send response back to sender?

k15r (changes requested):

components/event-publisher-proxy/pkg/handler/http/handler.go
k15r: the location for this comments is probably not correct. it applies to the entire PR

now that we do not need to use a specific encoding for the cloudevent, we could just proxy the entire thing and rely on that the backend does also some validation before accepting the CE

components/event-publisher-proxy/pkg/cloudevents/utils.go
radufa: Done

components/event-publisher-proxy/pkg/handler/http/handler_test.go
radufa: The "appName" location is found in different places dependent of CE event (structured or binary). The test check both structured and binary events

radufa: Refactored

components/event-publisher-proxy/pkg/sender/nats.go
radufa: ctx is already a "cancelable" context.

components/event-publisher-proxy/pkg/handler/http/handler.go
radufa: Cause is a major change, it will be solved in #10780

components/event-publisher-proxy/pkg/handler/http/handler_test.go
radufa: Avoid to change the go.mod for now.

components/event-publisher-proxy/pkg/handler/http/handler.go
k15r: i agree, and it is not our main priority now.

k15r (approved):

sayanh (approved):

Issue #10728, opened 2021-02-23
Enhance NATS dispatcher performance by reusing the cloudevents client


marcobebway:
**Description**

Enhance NATS dispatcher performance by reusing the cloudevents client while dispatching.

PR #10729, opened 2021-02-23
Reuse the same cloudevents client while dispatching from NATS

marcobebway:
**Description**

Reuse the same cloudevents client while dispatching from NATS.

**Related issue(s)**
https://github.com/kyma-project/kyma/issues/10728


radufa:
http.StatusOK could be used here: https://github.com/kyma-project/kyma/blob/56b792e39ce009052267937f2a5582b2e16560e5/components/eventing-controller/pkg/handlers/nats.go#L186

radufa (approved):

radufa (approved):

Issue #10317, opened 2021-01-13
Cover SKR upgrade Azure -> BEB scenario


k15r:
As SAP i want all SKR instances to be switched from azure to BEB.

Acceptance Criteria:
* KEB needs to provision BEB on upgrade
* Azure eventhubs needs to be removed from the cluster

radufa:
The associated PR for the management plane config:
 https://github.tools.sap/kyma/management-plane-config/pull/839

radufa:
Another associated PR for the management plane config, dev environment: 
https://github.tools.sap/kyma/management-plane-config/pull/853

radufa:
Associated issue: **"Add a configurable suffix to "EMS_PUBLISH_URL""**  #10563

radufa:
The update of KEB documentation: https://github.com/kyma-project/control-plane/pull/509

radufa:
Related issue: https://github.com/kyma-project/kyma/issues/10593

PR #10733, opened 2021-02-24
Post upgrade hook for eventing with subscriptions.kyma-project.io

sayanh:
**Description**

- Post upgrade hook for eventingv2

**Related issue(s)**

See https://github.com/kyma-project/kyma/issues/10317


installation/resources/installer-cr-cluster.yaml.tpl
```diff
@@ -37,8 +37,8 @@ spec:
       namespace: "kyma-system"
     - name: "helm-broker"
       namespace: "kyma-system"
-    # - name: "eventing"
-    #   namespace: kyma-system
+    - name: "eventing"
```
sayanh, line 40: Will take it out before merging!

components/event-sources/upgrade-job/process/unlabel_namespaces.go
```diff
@@ -0,0 +1,64 @@
...
+				},
+			},
+		}
+
+		patchLabelsData, err := json.Marshal(patchLabels)
+		if err != nil {
+			return errors.Wrapf(err, "failed to marshal labels from ns: %s", ns.Name)
+		}
+
+		_, err = s.process.Clients.Namespace.Patch(ns.Name, patchLabelsData)
```
marcobebway, line 45: Please make sure we preserve already existing labels. Probably having a test for that would make sense.

components/event-sources/upgrade-job/process/convert_triggers_to_kyma_subs.go
```diff
@@ -0,0 +1,125 @@
...
+	}
+}
+
+func (s ConvertTriggersToKymaSubscriptions) Do() error {
+	for _, trigger := range s.process.State.Triggers.Items {
+		// Generate equivalent subscription for this trigger
+		sub := s.NewSubscription(&trigger)
+
+		_, err := s.process.Clients.KymaSubscription.Create(sub)
+		if err != nil {
```
marcobebway, line 39: Does it make sense to not return an error if `k8serrors.IsAlreadyExists(err)` is true? 

Because this means that we already migrated this trigger before and we don't need to return an error. Or maybe delete and recreate again instead of skipping.

components/event-sources/upgrade-job/process/patch_connectivity_validators.go
```diff
@@ -0,0 +1,94 @@
...
+		"--cacheCleanupMinutes=2",
+	}
+	return args
+}
+
+func getDeploymentPatchData(oldDeployment appsv1.Deployment) ([]byte, error) {
+	appName := oldDeployment.Annotations[appReleaseNameKey]
+	var oldContainer corev1.Container
+	for _, c := range oldDeployment.Spec.Template.Spec.Containers {
+		oldContainer = c
```
marcobebway, line 73: This will always evaluate to the last item in the `oldDeployment.Spec.Template.Spec.Containers` list, Could you please explain why is it like that?

components/event-sources/upgrade-job/process/add_steps.go
```diff
@@ -0,0 +1,16 @@
...
+		NewSaveCurrentState(p),
+		NewDeleteEventServiceDeployments(p),
+		NewDeleteEventSources(p),
+		NewUnLabelNamespace(p),
+		NewDeleteBrokers(p),
+		NewConvertTriggersToKymaSubscriptions(p),
+		NewDeleteTriggers(p),
+		NewDeleteAppSubscriptions(p),
+		NewPatchConnectivityValidators(p),
+		NewCleanUp(p),
```
marcobebway, line 14: Should `channels` be deleted as well?

components/event-sources/upgrade-job/process/clean_up.go
```diff
@@ -0,0 +1,32 @@
...
+func NewCleanUp(p *Process) CleanUpStep {
+	return CleanUpStep{
+		name:    "Cleanup of the backed up Config Map",
+		process: p,
+	}
+}
+
+func (s CleanUpStep) Do() error {
+	err := s.process.Clients.ConfigMap.Delete(BackedUpConfigMapNamespace, BackedUpConfigMapName)
+	if err != nil {
```
marcobebway, line 23: In case of error, does it make sense to al least label the ConfigMap with something indicates that migration actually was done? Because only the cleanup in the end failed, but actually the migration was done. This is purely for troubleshooting.

resources/eventing/templates/pre-upgrade-job.yaml
```diff
@@ -0,0 +1,78 @@
...
+  labels: {{- include "eventing.preUpgrade.labels" . | nindent 4 }}
+  name: {{ template "eventing.fullname" . }}-remove-backendmodule
+  annotations:
+    helm.sh/hook-weight: "0"
+    helm.sh/hook: "pre-install,pre-upgrade"
+    helm.sh/hook-delete-policy: "before-hook-creation,hook-succeeded"
+rules:
+  - apiGroups: ["ui.kyma-project.io"]
+    resources: ["backendmodules"]
+    verbs: ["delete"]
```
marcobebway, line 23: Why do we need this?

components/event-sources/upgrade-job/process/delete_app_subscriptions_test.go
```diff
@@ -0,0 +1,40 @@
...
+		}
+		err := p.Execute()
+		g.Expect(err).Should(gomega.BeNil())
+		g.Expect(len(e2eSetup.appSubscriptions.Items)).To(gomega.Equal(1))
+
+		// Check for deleted subscriptions for application
+		for _, oldAppSub := range e2eSetup.appSubscriptions.Items {
+			if oldAppSub.Labels[applicationNameKey] != "" {
+				_, err := p.Clients.Subscription.Get(KymaIntegrationNamespace, oldAppSub.Name)
+				g.Expect(k8serrors.IsNotFound(err)).To(gomega.BeTrue())
```
marcobebway, line 36: General comment regarding testing the deletion steps, we only check if the objects were deleted or not, but we also need to check that those objects were created in the first place.

Something like:
- Test setup.
- Check resources created.
- Run step.
- Check resources deleted.

components/event-sources/upgrade-job/process/unlabel_namespaces.go
sayanh: It does preserve. Test is adapted to make sure.

components/event-sources/upgrade-job/process/convert_triggers_to_kyma_subs.go
```diff
@@ -0,0 +1,125 @@
...
+	var triggerEventTypeVersion, triggerEventSource, triggerEventType string
+	if trigger.Spec.Filter != nil {
+		attributes = *trigger.Spec.Filter.Attributes
+		triggerEventTypeVersion = attributes[eventVersionKey]
+		triggerEventSource = attributes[eventSourceKey]
+		triggerEventType = attributes[eventTypeKey]
+	}
+	for _, app := range s.process.State.Applications.Items {
+		if triggerEventSource == app.Name {
+			cleanAppName = kymaeventingapp.CleanName(&app)
```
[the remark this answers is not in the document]
sayanh: Fixed

components/event-sources/upgrade-job/process/convert_triggers_to_kyma_subs.go
```diff
@@ -0,0 +1,125 @@
...
+		bebFilters = &kymaeventingv1alpha1.BebFilters{
+			Dialect: "",
+			Filters: []*kymaeventingv1alpha1.BebFilter{
+				bebFilter,
+			},
+		}
+	}
+
+	if trigger.Spec.Subscriber.URI != nil {
+		sink = trigger.Spec.Subscriber.URI.String()
```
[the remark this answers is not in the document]
sayanh: Added logic for better sink URL transformation. IMO there should not be any transformation:
- if the ns of the trigger does not match ns in the cluster-local URL of the sink
- if the sink URL is not a cluster-local URL
In all these cases, the subscription will fail to work(will not be `ready`) and the migration guide should be referred for best-practices.

components/event-sources/upgrade-job/process/delete_event_service_deploys.go
```diff
@@ -0,0 +1,35 @@
...
+var _ Step = &DeleteEventServiceDeployments{}
+
+type DeleteEventServiceDeployments struct {
+	name    string
+	process *Process
+}
+
+func NewDeleteEventServiceDeployments(p *Process) DeleteEventServiceDeployments {
+	return DeleteEventServiceDeployments{
+		name:    "Delete Event ServiceDeployments",
```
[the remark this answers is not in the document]
sayanh: Fixed

components/event-sources/upgrade-job/process/convert_triggers_to_kyma_subs.go
sayanh: Fixed

components/event-sources/upgrade-job/process/patch_connectivity_validators.go
sayanh: Right, and it is also true that there is only one container in this connectivity validator [deployment](https://github.com/kyma-project/kyma/blob/master/components/application-operator/charts/application/templates/validator.yaml#L34). Hence it works reliably.

components/event-sources/upgrade-job/process/patch_connectivity_validators.go
```diff
@@ -0,0 +1,94 @@
...
+	}
+
+	desiredContainer := oldContainer.DeepCopy()
+	desiredContainer.Args = getNewContainerArgs(appName)
```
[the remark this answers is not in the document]
sayanh: We could use that. There is a chance of having problems as these deployments are based on Helm. So if there is some change in the args for this deployment then we may not get this working. This approach adds the args which works and should be there no matter what the old contents were which IMO is safer.

components/event-sources/upgrade-job/process/add_steps.go
sayanh: Channels are deleted through 2 actions(which I already took care):
- deletion of HTTPSources
- deletion of Brokers
Are there any specific channels that will be left over?

resources/eventing/templates/pre-upgrade-job.yaml
sayanh: The old `Backendmodule` had namespace annotation with `knative-eventing` which should be replaced with the new one with `kyma-system`. Hence it is needed.

components/event-sources/upgrade-job/process/clean_up.go
sayanh: In such a case, the job fails anyway hence a human intervention is needed to fix it. The label won't add any value IMO.

marcobebway: The label will help whoever troubleshooting what happens that actual migration is done, but only cleaning up the ConfigMap failed.

installation/resources/installer-cr-cluster.yaml.tpl
sayanh: It is out.

components/event-sources/upgrade-job/process/convert_triggers_to_kyma_subs.go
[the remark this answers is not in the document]
marcobebway: I agree.

marcobebway (approved):

PR #10698, opened 2021-02-22
CLI with install, uninstall and download version

pbochynski:
**Description**

Changes proposed in this pull request:

- install given version
- download charts from version
- uninstall Kyma
- provision/deprovision k3d cluster
- CLI with commands to install/upgrade Kyma, provision/deprovision k3d cluster
- flag to install new eventing (instead of knative), discovers: https://github.com/kyma-project/kyma/issues/10720
- upgrade test testing if connected application can survive the upgrade, discovers: https://github.com/kyma-project/kyma/issues/10759

PR #10749, opened 2021-02-25
Fix event type and source in fast-integration test

montaro:


pbochynski (changes requested):
The change will fix the test for new-eventing and will break it for knative implementation. The fix is in other PR: https://github.com/kyma-project/kyma/pull/10698/files#diff-f05d98841f4a63b71b83a91fb79bc0b79651616cde2938a897a6bdb46b931aa3

PR #10776, opened 2021-03-02
Use new eventing in the default set of components for new CLI installer

kwiatekus:
**Description**

- remove knative based eventing charts in favour of new eventing

This PR should only affect default set of kyma components when installed via new `kyma alpha deploy` command

**Related issue(s)**


strekm (approved):
/meow

PR #10797, opened 2021-03-04
dynamic configuration for grafana datasources

a-thaler:
**Description**
Part of the Grafana helm chart is a mechanism to load the datasources configurations dynamically via a sidecar. Kyma leverages the feature already for dashboards. As kyma brings already 3 datasources we should leverage the feature here as well, locating the datasource configuration also in the component related to that (like the datasource for loki in the logging component)

- Enabled the dynamic loading for all namespaces based on configmaps
- Migrated the three datasources into configmaps located in the related component
- Renamed existing dashboards configmaps to have "grafana" in the name as well
- Added a new derivedField to the loki datasource definition to support traceId integration to logs

**Related issue(s)**

### DIFF
--- a/resources/logging/charts/loki/templates/kyma-additions/grafana-datasource.yaml
+++ b/resources/logging/charts/loki/templates/kyma-additions/grafana-datasource.yaml
@@ -1,0 +1,28 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: {{ template "loki.fullname" . }}-grafana-datasource
+  labels:
+    grafana_datasource: "1"
+    app: {{ template "loki.name" . }}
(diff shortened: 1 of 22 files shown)


resources/logging/charts/loki/templates/kyma-additions/grafana-datasource.yaml
```diff
@@ -0,0 +1,24 @@
+apiVersion: v1
```
suleymanakbas91, line 1: We also make this and the one in tracing configurable just like in prometheus-istio.

suleymanakbas91 (approved):

majakurcius (approved):

PR #10798, opened 2021-03-04
Add batteries-included mode to certificates

mjakobczyk:
**Description**

For now `certificates` component supports legacy certificates management, user-provided scenario and Gardener. This improvement defines a default behaviour when user does not want to make use of any of the mentioned modes. As a result, a ready-to-go component kicks off without any additional configuration required (so called `batteries-included` later on).

- define fallback mode called `batteries-included` to `certificates` component
- generate default `ClusterIssuer` and `Certificate`
- handle fallback mode for GKE clusters

**Related issue(s)**

Implements #10336


mjakobczyk:
As our approach has changed, batteries-included mode was implemented in installation lib - https://github.com/kyma-project/hydroform/pull/263. Closing.

PR #10808, opened 2021-03-05
Fix fast-integration tests for new nats

berkanercan:
**Description**

This PR applies the necessary changes to make fast-integration work with new nats backend. 

- Install.js is receiving an options object to skip the installation of knative related components

This PR should be merged when we switch to new nats solution from knative.


k15r:
outdated, closing

PR #10813, opened 2021-03-05
Kcproxy optional

### LINKED ISSUE #10811: Configurable authentication for kiali and jaeger UI
**Description**
At the moment the kiali and jaeger UI require authentication against the in-cluster dex by default. There is no way to disable it fully nor you can configure a different OIDC provider.

Goal: Make the authentication for the UIs optional (enabled by default) and have the OIDC provider configurable

Acceptance:
- A new override exists to disable authentication fully for the UIs, resulting in not deploying any unneeded resources (like the kxproxy component)
- Overrides exists to configure a custom OIDC provider with related clientId/secret

**Reasons**
Probably dex will be removed from the kyma stack by default and users should provide OIDC providers as optional config input.

**Attachments**

JeremyHarisch:
**Description**

This PR changes authentication for the UIs of Kiali and Jaegger to optional. Default is set to enabled. Furthermore, 'discoveryurl' was moved to values.yaml to make the configuration more simple.

**Related issue(s)**

Resolves #10811

### DIFF
--- a/resources/kiali/values.yaml
+++ b/resources/kiali/values.yaml
@@ -27,6 +27,7 @@ global:
     enabled: true
 
 kcproxy:
+  enabled: false
   replicaCount: 1
   port: 3000
   config:
--- a/resources/kiali/values.yaml
+++ b/resources/kiali/values.yaml
@@ -33,6 +33,7 @@ kcproxy:
   config:
     clientId: "kiali"
     clientSecret: "hiFWLWqIxw5d3gl"
+    discoveryUrl: https://dex.{{ .Values.global.ingress.domainName }}
     tlsVerify: false
     resources:
       uri: "/*"
--- a/resources/kiali/values.yaml
+++ b/resources/kiali/values.yaml
@@ -27,7 +27,7 @@ global:
     enabled: true
 
 kcproxy:
-  enabled: false
+  enabled: true
   replicaCount: 1
   port: 3000
   config:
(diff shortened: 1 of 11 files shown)


resources/kiali/values.yaml
```diff
@@ -27,11 +27,13 @@ global:
     enabled: true
 
 kcproxy:
+  enabled: true
   replicaCount: 1
   port: 3000
   config:
     clientId: "kiali"
     clientSecret: "hiFWLWqIxw5d3gl"
+    discoveryUrl: https://dex.{{ .Values.global.ingress.domainName }}
```
skhalash, line 36: Hmm, what is the reason for introducing this property?

skhalash (changes requested):
Consider renaming `kcproxy-virtualservice.yaml` to just `virtualservice.yaml` in the `jaeger` chart since it will route the traffic to one of two services (and also to be consistent with the `kiali` chart)

resources/kiali/values.yaml
JeremyHarisch: Icke asked me to introduce it so that if this URL needs to be adapted, everything can be done in the 'values.yaml'. 
The property gets used in here: https://github.com/kyma-project/kyma/pull/10813/files#diff-3970e9005cb9dd2ffec95600b78d8362922633355d029c151838d6620858e6afR36

The same goes for the tracing charts.

skhalash: Thanks, I see!

@a-thaler Is it something we want to make overridable? Because if it's just because of code reuse, we could just put it into a template (a *.tpl file)

a-thaler: It was about having the OIDC issuer related settings configurable. In future there might be no dex and you can enable it with your custom settings

skhalash (approved):
LGTM!

clebs (approved):

PR #10832, opened 2021-03-08
Fix pods of knative eventing to use correct psp

berkanercan:
**Description**

The following components were choosing garneder-privileged psp over 001-kyma-unprivileged, since their pods didn't have the correct securityContext in the container level. 
- eventing-controller
- sources-controller
- eventing-webhook
- knative-eventing-kafka-channel-controller
- event-soruces-controller-manager


sayanh (commented):
The thing is these charts won't be installed or upgraded in the next version. Are these changes needed?

Issue #10730, opened 2021-02-24
istio-proxy-reset job doesn't restart function and eventing controller deployments


ebensom:
**Description**

The `istio-proxy-reset` job doesn't seem to have any effect on specific workloads created by either the eventing controller or the function controller .

The `kubectl rollout restart` command used by the `istio-proxy-reset` script works by changing an annotation in the deployment pod spec, however the annotations are controlled by the controller, and immediately rolled back.
For example: https://github.com/kyma-project/kyma/blob/274bac2906a69feed1e516f879ba34fa8de7d144/components/function-controller/internal/controllers/serverless/deployment.go#L100

Solution could be to change the behavior of the function and eventing controllers to omit external annotations, or adapt the `istio-proxy-reset` script to forcefully delete these pods instead of restarting them.

**Steps to reproduce**

Issue `kubectl rollout restart` on any deployment created by either the function controller or eventing controller.

a-thaler:
We should solve it properly by adjusting the controllers. Everyone will expect to have the rollout command working properly I assume.

sayanh:
@ebensom Why do you think it is done in the eventing-controller?

kwiatekus:
@aerfio fixed the annotation handling on serverless deployments

sayanh:
This does not impact eventing controllers.

strekm:
@k15r is it done? is there something that needs to be implemented in istio-proxy-reset job?

k15r:
@strekm according to @sayanh we are not impacted.

PR #10856, opened 2021-03-10
Update fn-ctrl so that it ignores user set annotations on fn's deployment

aerfio:
**Description**

- as in title
- add appropriate tests for that use case

**Related issue(s)**

See https://github.com/kyma-project/kyma/issues/10730


pPrecel (approved):
Good job 👍

PR #10889, opened 2021-03-15
Add readiness and startup probe to function container

aerfio:
**Description**

- as in title
- we might consider adding liveness probe too - this needs to be discussed


components/function-runtimes/python38/kubeless/kubeless.py
```diff
@@ -180,12 +180,22 @@ def preload():
...
+            requestlogger.ApacheFormatter(),
+        )
+    else:
+        loggedapp = app        
+    # end of modified section
+
 
+    quiet = 'KYMA_BOTTLE_QUIET_OPTION_DISABLED' not in os.environ
+    print(quiet)
+    print("ASDASD")
```
pPrecel, line 198: Logs from python functions looks like:
```
kubectl logs -n pprecel test3-cmtgb-78f8cc878-g4rsw function -f
True
ASDASD
```
Is `print("ASDASD")` necessary? And can we add another annotation to this line `print(quiet)`? Maybe, `print("quiet mode:", quiet)` looks better, or just remove this line too. If you decide to improve arguments for the `print` method, maybe we can implement something similar to the node functions too?

pPrecel: lol outdated 🥇

resources/serverless/charts/webhook/values.yaml
```diff
@@ -6,7 +6,7 @@ nameOverride:
 fullnameOverride:
 image:
   repository: "eu.gcr.io/kyma-project/function-webhook"
-  tag: "76ba4f03"
+  tag: "overriden by parent chart"
```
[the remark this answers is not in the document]
pPrecel: good point

pPrecel (dismissed):
LGTM! 🥇

m00g3n (approved):

PR #10898, opened 2021-03-16
Fix for helm hook for new eventing installation 

sayanh:
**Description**

- Issue was the post-install hook deleted all the deployments in kyma-integration ns in `pre-master-kyma-integration-k3s`. The problem was, where there are no applications, the event-service lists for all deployments in kyma-integration ns and eventually deletes them.

### DIFF
--- a/components/event-sources/upgrade-job/process/save_current_state_test.go
+++ b/components/event-sources/upgrade-job/process/save_current_state_test.go
@@ -6,6 +6,7 @@ import (
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
+	applicationconnectorv1alpha1 "github.com/kyma-project/kyma/components/application-operator/pkg/apis/applicationconnector/v1alpha1"
 	"github.com/sirupsen/logrus"
 
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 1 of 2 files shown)


components/event-sources/upgrade-job/process/save_current_state_test.go
```diff
@@ -147,4 +148,63 @@ func TestSaveCurrentState(t *testing.T) {
...
+		saveCurrentStateStep := NewSaveCurrentState(p)
+		expectedLabelsInCM := map[string]string{
+			"release": p.ReleaseName,
+		}
+		p.Steps = []Step{
+			saveCurrentStateStep,
+		}
+		err := p.Execute()
+		g.Expect(err).Should(gomega.BeNil())
+
```
marcobebway, line 170: Does it make sense to check `g.Expect(len(p.State.Applications.Items)).To(gomega.Equal(0))`?

sayanh: Actually, we are testing the contents of the `configmap`(which stores the state and does not contain applications) and not the state of `p`. Hence the state of `p` is not valuable in this context, IMO.

marcobebway (approved):

PR #10915, opened 2021-03-18
Fix SecurityContext for Postgresql

Tomasz-Smelcerz-SAP:
**Description**

Ensure non-privileged `securityContext` for ORY/Postgresql Pods.
Follow-up to: https://github.com/kyma-project/kyma/pull/10834

- Fix securityContext for containers
- Use `operator` - UID=11 instead of `root` in init containers

**Related issue(s)**
See also: #9550, https://github.com/kyma-project/kyma/pull/10834


strekm (approved):
/meow

PR #10950, opened 2021-03-23
nats clustersize

k15r:
**Description**

- increase the size of the kyma-cluster internal nats-cluster from 1 to 3 nodes in the production profile

**Related issue(s)**


berkanercan (approved):

radufa (dismissed):
Please change the profile value for "controller" to "nats-controller", cause now the overrides are applied to beb-controller.
Sollved in https://github.com/kyma-project/kyma/pull/10981

berkanercan:
> Please change the profile value for "controller" to "nats-controller", cause now the overrides are applied to beb-controller

I am fixing this in another ticket. We need to fix the deployment file as well.

radufa (approved):

PR #10954, opened 2021-03-23
Move pod-preset chart from cluster-essentials to service-catalog-addons

colunira:
**Description**

Cluster-essentials is not a good place to keep pod-preset chart, it is used in the service binding controller.

- Move pod-preset chart from cluster-essentials to service-catalog-addons

**Related issue(s)**

https://github.tools.sap/kyma/backlog/issues/1331


Tomasz-Smelcerz-SAP:
This PR is no longer valid. Service-catalog is being removed from Kyma and related files are being reorganized. Further cleanup will take place in Kyma 2.0 codebase.

Issue #9020, opened 2020-07-13
Sync Ory charts with ory/k8s repo


jakkab:
**Description**

Update Ory charts to reflect recent changes in the ory/k8s repo.

Caveats: 

- adopt the 'existing secret' mechanism (it should replace our custom logic, it is pretty similar anyways);
- find a way to access encryption keys (`.Values.hydra.hydra.config.secrets.system/cookie`) in our custom migration script (https://github.com/kyma-project/kyma/blob/master/resources/ory/files/secret-migration.sh). When Helm renders templates, those values are unset (see https://github.com/ory/k8s/blob/master/helm/charts/hydra/templates/_helpers.tpl#L107) to avoid saving unencrypted secrets to the Hydra configmap (encryption keys are part of the `config` tree in the values.yaml file).

Possible solutions:
- move the script to a dedicated subchart so that it gets rendered before the configmap in the Hydra chart unsets the values;
- ???

**Reasons**

Keeping Kyma up to date with ory/k8s repo to facilitate future updates.

strekm:
need to push it to 1.20

Tomasz-Smelcerz-SAP:
The latest release for `ory/k8s` containing ORY Helm charts is `v0.5.5`,  15 March 2021
This features:
- Hydra v1.8.5
-  Hydra-Maester v0.0.19
- Oathkeeper v0.38.5
- Oathkeeper-Maester v0.38.5

Tomasz-Smelcerz-SAP:
Charts are synced. Closing the issue

PR #10968, opened 2021-03-24
Sync ORY/Oathkeeper charts to the latest version

mjakobczyk:
**Description**

- update `oathkeeper` chart
- update `oathkeeper-maester` chart

**Related issue(s)**

Implements #9020

### DIFF
--- a/resources/ory/charts/oathkeeper/templates/ingress-api.yaml
+++ b/resources/ory/charts/oathkeeper/templates/ingress-api.yaml
@@ -32,8 +32,8 @@ spec:
         {{- range .paths }}
           - path: {{ . }}
             backend:
-              serviceName: {{ $fullName }}-api
-              servicePort: http
+              serviceName: {{ $fullName }}-api # TODO: change or leave?
+              servicePort: http # TODO: change or leave?
         {{- end }}
   {{- end }}
 {{- end }}
--- a/resources/ory/charts/oathkeeper/templates/ingress-api.yaml
+++ b/resources/ory/charts/oathkeeper/templates/ingress-api.yaml
@@ -31,9 +31,12 @@ spec:
         paths:
         {{- range .paths }}
           - path: {{ . }}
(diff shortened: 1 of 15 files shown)


resources/ory/charts/oathkeeper/templates/ingress-api.yaml
```diff
@@ -32,8 +32,8 @@ spec:
         {{- range .paths }}
           - path: {{ . }}
             backend:
-              serviceName: {{ $fullName }}-api
-              servicePort: http
+              serviceName: {{ $fullName }}-api # TODO: change or leave?
```
strekm, line 35: ? seems to me you have some doubts

resources/ory/charts/oathkeeper/templates/ingress-proxy.yaml
```diff
@@ -32,8 +32,8 @@ spec:
         {{- range .paths }}
           - path: {{ . }}
             backend:
-              serviceName: {{ $fullName }}-proxy
-              servicePort: http
+              serviceName: {{ $fullName }}-proxy # TODO: change or leave?
```
strekm, line 35: same as above

resources/ory/charts/oathkeeper/templates/ingress-api.yaml
mjakobczyk: ORY chart uses `Ingress` resource with version `extensions/v1beta1`. We use `networking.k8s.io/v1` on our side. As a result, in our chart the marked key seems to be no longer needed (IDE prompts that this key is not expected). Should we then consider removing it or not?

resources/ory/charts/oathkeeper/templates/ingress-proxy.yaml
mjakobczyk: The same reply as above.

strekm (approved):
/meow

PR #10987, opened 2021-03-25
Add busola-migrator component and redirecting logic

tgorgol:
**Description**

- Add busola-migrator component and redirecting logic


components/busola-migrator/internal/config/config.go
```diff
@@ -0,0 +1,32 @@
...
+	"github.com/vrischmann/envconfig"
+)
+
+type Config struct {
+	Port           int           `envconfig:"default=80"`
+	TimeoutRead    time.Duration `envconfig:"default=30s"`
+	TimeoutWrite   time.Duration `envconfig:"default=30s"`
+	TimeoutIdle    time.Duration `envconfig:"default=120s"`
+	BusolaURL      string        `envconfig:"default=https://busola.main.hasselhoff.shoot.canary.k8s-hana.ondemand.com"`
+	OIDCIssuerURL  string        `envconfig:"default=https://kyma.accounts.ondemand.com/"`
```
Wawrzyn321, line 17:
```suggestion
	OIDCIssuerURL  string        `envconfig:"default=https://kyma.accounts.ondemand.com"`
```

components/busola-migrator/README.md
```diff
@@ -0,0 +1,37 @@
...
+| **APP_BUSOLA_URL** | URL of the Busola cluster | `https://busola.main.hasselhoff.shoot.canary.k8s-hana.ondemand.com` |
+| **APP_OIDC_ISSUER_URL** | OIDC issuer url | `https://kyma.accounts.ondemand.com` |
+| **APP_OIDC_CLIENT_ID** | OIDC client id | `6667a34d-2ea0-43fa-9b13-5ada316e5393` |
+| **APP_OIDC_SCOPE** | OIDC scope | `openid` |
+| **APP_OIDC_USE_PKCE** | Should OIDC use PKCE | `false` |
+| **APP_STATIC_FILES_DIR** | Directory to look for the static webpage to serve | `./static` |
+| **OVERRIDE_BUSOLA_URL** | Optional override of url of the Busola cluster | None |
+| **OVERRIDE_OIDC_ISSUER_URL** | Optional override of OIDC issuer url | None |
+| **OVERRIDE_OIDC_CLIENT_ID** | Optional override of OIDC client id | None |
+| **OVERRIDE_OIDC_SCOPE** | Optional override of OIDC scope | None |
```
kazydek, line 36:
```suggestion
| **OVERRIDE_OIDC_SCOPE** | Optional override for the OIDC scope | None |
```

components/busola-migrator/README.md
```diff
@@ -0,0 +1,37 @@
...
+| **OVERRIDE_OIDC_USE_PKCE** | Optional override of Should OIDC use PKCE | None |
```
kazydek, line 37:
```suggestion
| **OVERRIDE_OIDC_USE_PKCE** | Optional override for the use of PKCE in OIDC  | None |
```

kazydek (dismissed):

components/busola-migrator/internal/config/config.go
```diff
@@ -0,0 +1,91 @@
...
+
+	cfg = applyOverrides(cfg, overrides)
+
+	return cfg
+}
+
+func getOverrides() configOverrides {
+	var overrides configOverrides
+
+	if val, ok := os.LookupEnv("OVERRIDE_BUSOLA_URL"); ok {
```
rJankowski93, line 52: Is it correct if some env is empty? 
What do you think about use `os.Getenv` and override env only if os env is not empty?

components/busola-migrator/internal/config/config.go
```diff
@@ -0,0 +1,91 @@
...
+		overrides.BusolaURL = ptr.String(val)
```
rJankowski93, line 53: Why you set this temp struct `config Overrides`? Can we set struct `Config` here?

components/busola-migrator/internal/busola/init.go
```diff
@@ -0,0 +1,63 @@
...
+    "disabledNavigationNodes": "",
+    "systemNamespaces": "istio-system knative-eventing knative-serving kube-public kube-system kyma-backup kyma-installer kyma-integration kyma-system natss kube-node-lease kubernetes-dashboard serverless-system"
+  },
+  "features": {
+    "bebEnabled": false
+  }
+}`
+
+func BuildInitURL(appCfg config.Config, kubeConfig *rest.Config) (string, error) {
+	if kubeConfig == nil {
```
rJankowski93, line 35: Is it necessary? 
If kubeConfig was empty line 22 from file `main.go` call `os.Exit(1)`

components/busola-migrator/internal/config/config.go
tgorgol: I didn't use Config struct and `envconfig` library because it doesn't allow nil values in case of unset variables, I needed a way to distinguish when the environment variable was set to empty string from when it wasn't set at all.

components/busola-migrator/internal/busola/init.go
tgorgol: I think it's worth checking if the variable is `nil` even if it's verified in some other place, before the function call, because there is always the possibility that someone changes the code in `main.go` and this might result in unexpected panic.

rJankowski93 (approved):

kazydek (approved):

pbochynski (approved):

PR #10988, opened 2021-03-25
Split up Logging chart rules  to respective subcharts

JeremyHarisch:
**Description**

- Prometheus-Alerting-Rules of logging chart have been split to sub-charts (fluentbit and Loki)
- Splitting up helps with the situation when fluentbit or Loki is disabled, all rules of active charts can be applied without having to deal with deactivating certain Rules, since if the sub-charts are deactivated, the corresponding rules will also not be applied

**Related issue(s)**

### DIFF
--- a/resources/logging/charts/fluent-bit/values.yaml
+++ b/resources/logging/charts/fluent-bit/values.yaml
@@ -130,6 +130,11 @@ updateStrategy: {}
 # Make use of a pre-defined configmap instead of the one templated here
 existingConfigMap: ""
 
+## Custom configuration for Kyma alerting rules
+kymaRules:
+  ## Adds the specified filter to NoOutputBytesProcessed alert for Fluent Bit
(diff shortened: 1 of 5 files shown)


resources/logging/charts/fluent-bit/values.yaml
```diff
@@ -130,6 +130,11 @@ updateStrategy: {}
 # Make use of a pre-defined configmap instead of the one templated here
 existingConfigMap: ""
 
+## Custom configuration for Kyma alerting rules
+kymaRules:
+  ## Adds the specified filter to NoOutputBytesProcessed alert for Fluent Bit
+  fluentBitAlertFilter: ""
+
```
skhalash, line 137: I'd prefer something like `prometheusRules` instead

skhalash (changes requested):

skhalash (approved):

PR #11006, opened 2021-03-29
Add properties to istio-ingress gateway to set "X-Forward-To" header

### LINKED ISSUE #7902: X-Forwarded-For handling in API Gateway v2
**Description**

Dear Colleagues,

in API Gateway v2 the indirection from istio via the ory components cause that the host header is now set to the K8S internal service name. Furthermore the X-Forwarded-* headers are set to misleading values. This causes components that use these values to produce external host names to not function anymore. Below is a real sample of the headers:

{
   "host":"logger.personservice.svc.cluster.local:8080",
   "user-agent":"Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_3) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/80.0.3987.122 Safari/537.36",
   "accept":"image/webp,image/apng,image/*,*/*;q=0.8",
   "accept-encoding":"gzip, deflate, br",
   "accept-language":"en-US,en;q=0.9",
   "cache-control":"no-cache",
   "certificate-data":"By=spiffe://cluster.local/ns/kyma-system/sa/oathkeeper-maester-account;Hash=0a7373aca126b0eff277e0e92cf4506450c17d745db97a724c6f8e15f85572fb;Subject=\"\";URI=spiffe://cluster.local/ns/istio-system/sa/istio-ingressgateway-service-account",
   "cookie":"_ga=GA1.2.1976884213.1550584065; rxVisitor=1583424650883EMG99P1EEMQIGV28522IC2H8F36RIN8S; dtLatC=1; dtSa=-; dtPC=15$350074451_535h-vGHPPPJDCJMDNMDORCTPANEMIMGEETJVN-0; dtCookie=v_4_srv_15_sn_727C6D32EA5EC2CA6BB0E09FD151CED8_perc_100000_ol_0_mul_1_app-3Aea7c4b59f27d43eb_1; rxvt=1584951910748|1584950045609",
   "dnt":"1",
   "pragma":"no-cache",
   "referer":"https://logger.{clusterhost}/",
   "sec-fetch-dest":"image",
   "sec-fetch-mode":"no-cors",
   "sec-fetch-site":"same-origin",
   "x-forwarded-for":"10.250.0.3, 127.0.0.1",
   "x-forwarded-proto":"https",
   "x-request-id":"c7bd09cd-8973-4c9f-a699-20a99a38387a",
   "content-length":"0",
   "x-forwarded-client-cert":"By=spiffe://cluster.local/ns/personservice/sa/default;Hash=de0fd670ff2214c09acb8ccc99e4ee724fde35e48217989d0134b60e78e4b3cd;Subject=\"\";URI=spiffe://cluster.local/ns/kyma-system/sa/oathkeeper-maester-account",
   "x-b3-traceid":"0dfdde37ff9924bfacf717d949c9ff8d",
   "x-b3-spanid":"161969ec2072d8fe",
   "x-b3-parentspanid":"7fee11bbbcb108c4",
   "x-b3-sampled":"0"
}

**Expected result**

"x-forwarded-for":"logger.{clusterhost}",
 "x-forwarded-proto":"https",

**Actual result**

   "x-forwarded-for":"10.250.0.3, 127.0.0.1",
   "x-forwarded-proto":"https",

**Steps to reproduce**

**Troubleshooting**

c0d1ngm0nk3y:
**Description**

- Modify `istio-ingressgateway` to set "X-Forwared-For" header
  - The annotation is only needed on AWS

**Related issue(s)**

Fixes #7902 

**Additional information**
https://kubernetes.io/docs/tutorials/services/source-ip/#source-ip-for-services-with-type-loadbalancer

### DIFF
--- a/resources/istio/files/istio-operator-cluster-evaluation.yaml
+++ b/resources/istio/files/istio-operator-cluster-evaluation.yaml
@@ -82,7 +82,10 @@ spec:
           requests:
             cpu: 25m
             memory: 32Mi
+        serviceAnnotations:
+          "service.beta.kubernetes.io/aws-load-balancer-type": "nlb"
         service:
+          externalTrafficPolicy: Local
           ports:
           - name: status-port
             port: 15021
--- a/resources/istio/files/istio-operator-cluster-production.yaml
+++ b/resources/istio/files/istio-operator-cluster-production.yaml
@@ -83,7 +83,10 @@ spec:
           requests:
             cpu: 100m
             memory: 128Mi
+        serviceAnnotations:
+          "service.beta.kubernetes.io/aws-load-balancer-type": "nlb"
         service:
+          externalTrafficPolicy: Local
           ports:
           - name: status-port
             port: 15021
--- a/resources/istio/files/istio-operator-cluster.yaml
+++ b/resources/istio/files/istio-operator-cluster.yaml
@@ -80,7 +80,10 @@ spec:
           requests:
             cpu: 100m
             memory: 128Mi
+        serviceAnnotations:
+          "service.beta.kubernetes.io/aws-load-balancer-type": "nlb"
         service:
+          externalTrafficPolicy: Local
           ports:
           - name: status-port
             port: 15021
--- a/resources/istio/files/istio-operator-minikube-evaluation.yaml
+++ b/resources/istio/files/istio-operator-minikube-evaluation.yaml
@@ -81,6 +81,7 @@ spec:
             cpu: 25m
             memory: 32Mi
         service:
+          externalTrafficPolicy: Local
           ports:
             - name: status-port
               port: 15021
--- a/resources/istio/files/istio-operator-minikube.yaml
+++ b/resources/istio/files/istio-operator-minikube.yaml
@@ -81,6 +81,7 @@ spec:
             cpu: 100m
             memory: 96Mi
         service:
+          externalTrafficPolicy: Local
           ports:
           - name: status-port
             port: 15021
--- a/tests/integration/api-gateway/gateway-tests/main_test.go
+++ b/tests/integration/api-gateway/gateway-tests/main_test.go
@@ -4,10 +4,13 @@ import (
 	"context"
 	"crypto/tls"
 	"encoding/base64"
+	"errors"
 	"fmt"
 	"log"
 	"math/rand"
+	"net"
 	"net/http"
+	"os"
 	"strings"
 	"testing"
 	"time"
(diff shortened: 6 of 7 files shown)


c0d1ngm0nk3y:
@strekm Do you have feedback on this PR?

c0d1ngm0nk3y:
@colunira Do you have feedback on this PR?

strekm:
@c0d1ngm0nk3y we provide overrides mechanism to achieve this kind of setup specific for cluster provider

c0d1ngm0nk3y:
> @c0d1ngm0nk3y we provide overrides mechanism to achieve this kind of setup specific for cluster provider

@strekm You mean for the aws annotation, right? Do you have a link to the documentation? But the `externalTrafficPolicy` makes sense as default, don't you think?

strekm:
@c0d1ngm0nk3y here you can read about istio overrides: https://kyma-project.io/docs/components/service-mesh/#configuration-configuration

c0d1ngm0nk3y:
The error seems not to be related.
```
namespaces "test-namespace-krszs" already exists
```
Hence I will re-trigger.

resources/istio/files/istio-operator-cluster-evaluation.yaml
```diff
@@ -82,7 +82,10 @@ spec:
           requests:
             cpu: 25m
             memory: 32Mi
+        serviceAnnotations:
+          "service.beta.kubernetes.io/aws-load-balancer-type": "nlb"
```
Tomasz-Smelcerz-SAP, line 86: This is strictly AWS-related setting. It should not be hardcoded in a file common to all environments. The proper way to do it:
- define aws-related settings somewhere in values.yaml
- in the aws-setting section prepare "enabled" flag with default value = false!
- if enabled embed the settings from values.yaml here.

resources/istio/files/istio-operator-cluster-evaluation.yaml
```diff
@@ -82,7 +82,10 @@ spec:
...
         service:
+          externalTrafficPolicy: Local
```
Tomasz-Smelcerz-SAP, line 88: This changes the behavior for ALL traffic. And it's not what we may want:

```denotes if this Service desires to route external traffic to node-local or cluster-wide endpoints. “Local” preserves the client source IP and avoids a second hop for LoadBalancer and Nodeport type services, but risks potentially imbalanced traffic spreading. “Cluster” obscures the client source IP and may cause a second hop to another node, but should have good overall load-spreading.```

If you really need that, consider adding another ingress gateway specific for the task. It should be included conditionally, disabled by default for deployments that don't need it.

File: resources/istio/files/istio-operator-cluster-production.yaml [hunk 3]
```diff
@@ -83,7 +83,10 @@ spec:
           requests:
             cpu: 100m
             memory: 128Mi
+        serviceAnnotations:
+          "service.beta.kubernetes.io/aws-load-balancer-type": "nlb"
```
Tomasz-Smelcerz-SAP, line 87: same comments as for `evaluation` profile

Tomasz-Smelcerz-SAP (changes requested):

resources/istio/files/istio-operator-cluster-evaluation.yaml
strekm: One more thing to add is that we try to keep Istio configuration as close as we can to default Istio configuration. By default this values is set to "Cluster".

c0d1ngm0nk3y: >  “Cluster” obscures the client source IP and may cause a second hop to another node, but should have good overall load-spreading.`

But for our use case, I think this is exactly what we don't want. We will look at the option with another gateway.

resources/istio/files/istio-operator-cluster-evaluation.yaml
c0d1ngm0nk3y: Thanks

krasimirdermendzhiev:
Hi all,
I have one question for Jaeger. If you have more services and what to see the trace. I can't see the trace to some service or external service. 
Distributed tracing/ Propagate headers - If the request has x-request-id you can go trough the spans. Applications must propagate the trace context between incoming and outgoing requests (x-request-id). Do you have some configuration for this?

c0d1ngm0nk3y:
I am not sure, if I would agree with keeping the `istio` defaults. Changing the defaults in a way that it improves the `kyma` use case, might make sense. And allowing the applications to "see" the IP of the caller sounds like a meaningful thing for a runtime.

But we will focus on other things first. If/when it will come up again, we can discuss how to tackle this.

Issue #11004, opened 2021-03-29
Avoid prematurely giving up of reconcilation loop


radufa:
**Description**
In some edge error cases, the reconciliation loop implemented by eventing controller for NATS could be prematurely terminated.
Here is an example of this behavior from an old version of eventing:
```
Only 2 retries within 1 second happened:

{"level":"info","ts":1616766959.560937,"logger":"reconciler.Subscription","msg":"received subscription reconciliation request","namespace":"test","name":"function-lastorder"}
{"level":"info","ts":1616766959.5610256,"logger":"reconciler.Subscription","msg":"Adding finalizer to subscription object","kind":"Subscription","name":"function-lastorder","namespace":"test","version":1}
{"level":"info","ts":1616766959.5774264,"logger":"reconciler.Subscription","msg":"received subscription reconciliation request","namespace":"test","name":"function-lastorder"}
[...]

PR #11010, opened 2021-03-30
Avoid swallowing errors in nats reconciler

radufa:
**Description**

- Avoid swallowing the original error in cases of successfully update of subscription status 

**Related issue(s)**

See also: #11004

### DIFF
--- a/components/eventing-controller/reconciler/subscription-nats/reconciler.go
+++ b/components/eventing-controller/reconciler/subscription-nats/reconciler.go
@@ -112,9 +112,7 @@ func (r *Reconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
 	// Check for valid sink
 	if err := r.assertSinkValidity(actualSubscription.Spec.Sink); err != nil {
 		r.Log.Error(err, "failed to parse sink URL")
-		err := r.syncSubscriptionStatus(ctx, actualSubscription, false, err.Error())
-		if err != nil {
-			r.Log.Error(err, "failed to sync subscription status")
+		if err := r.syncSubscriptionStatus(ctx, actualSubscription, false, err.Error()); err != nil {
 			return ctrl.Result{}, err
 		}
 		// No point in reconciling as the sink is invalid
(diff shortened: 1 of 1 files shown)


components/eventing-controller/reconciler/subscription-nats/reconciler.go
```diff
@@ -125,7 +123,9 @@ func (r *Reconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
 	err = r.natsClient.DeleteSubscription(desiredSubscription)
 	if err != nil {
 		log.Error(err, "failed to delete subscriptions")
-		err = r.syncSubscriptionStatus(ctx, actualSubscription, false, err.Error())
+		if err := r.syncSubscriptionStatus(ctx, actualSubscription, false, err.Error()); err != nil {
+			return ctrl.Result{}, err
```
k15r, line 127: the effect is correct, but imho the reason for the re-reconcile is off. here is seems to be reconciled as the subscriptionstatus could not be synced. But the original reason for reconciling again is that the deletion failed

components/eventing-controller/reconciler/subscription-nats/reconciler.go
```diff
@@ -148,18 +148,18 @@ func (r *Reconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
 	err = r.natsClient.SyncSubscription(desiredSubscription, r.eventTypeCleaner)
 	if err != nil {
 		r.Log.Error(err, "failed to sync subscription")
-		err = r.syncSubscriptionStatus(ctx, actualSubscription, false, err.Error())
+		if err := r.syncSubscriptionStatus(ctx, actualSubscription, false, err.Error()); err != nil {
+			return ctrl.Result{}, err
```
k15r, line 152: see previous comment

k15r (approved):

components/eventing-controller/reconciler/subscription-nats/reconciler.go
radufa: If the subscription status could not be saved, means that the status of the subscriptions previously saved is wrong. It makes the error analysis much difficult. Cause the reconciler should be idempotent, a new reconcile could perhaps solve the issue. This is let say a pattern... 
Of course, in the new call to reconcile(), the base action could succeed (in our case delete). 
From the perspective of reconciling calls, the error is important (!=nil) and not its type.

PR #11028, opened 2021-03-31
Bump image of eventing-controller and nats-controller 

radufa:
**Related issue(s)**

### DIFF
--- a/resources/eventing/charts/nats-controller/values.yaml
+++ b/resources/eventing/charts/nats-controller/values.yaml
@@ -10,7 +10,7 @@ image:
   # name is the name of the container image for the nats-eventing-controller
   name: "eventing-controller-nats"
   # tag is the container tag of the eventing-controller image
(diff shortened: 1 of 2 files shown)


resources/eventing/charts/nats-controller/values.yaml
```diff
@@ -10,7 +10,7 @@ image:
   # name is the name of the container image for the nats-eventing-controller
   name: "eventing-controller-nats"
   # tag is the container tag of the eventing-controller image
-  tag: "e2185713"
+  tag: "PR-11010"
```
marcobebway, line 13: Please use the final image SHA instead of the PR image.

Also, please update the image here as well `resources/eventing/charts/controller`.

radufa: Done

marcobebway (approved):

PR #11031, opened 2021-04-01
Add a flag to cluster-essentials CRD related resources

colunira:
**Description**

Introduce a flag to allow disabling CRD installation. This change is needed to ensure compatibility of our two ways of installation - kyma-installer and Hydroform installation library in which we want to install CRDs in a different way, but we still need other resources from cluster-essentials.

- Add a flag to cluster-essentials CRD related resources

**Related issue(s)**

yma/backlog/issues/1331


mjakobczyk (approved):

strekm (approved):
/meow

Issue #10154, opened 2020-12-07
Slow performance for python based serverless


valentinvieriu:
**Description**

Trying load testing the python hellow world produces very low results using the XL profile.
```
Python XL - profile 20 replicas using production profile
Running 2m test @ https://test-python.kyma.fun/
  12 threads and 128 connections
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency   137.74ms  140.89ms   1.20s    85.80%
    Req/Sec    25.44     15.21    90.00     61.41%
  Latency Distribution
     50%   81.74ms
     75%  158.68ms
     90%  350.39ms
     99%  659.27ms
  27131 requests in 1.67m, 4.74MB read
  Socket errors: connect 0, read 0, write 0, timeout 119
Requests/sec:    271.04
Transfer/sec:     48.52KB
```
compared with 

```
Node XL - profile 20 replicas using production profile
Running 2m test @ https://test-node.kyma.fun/
  12 threads and 128 connections
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency    47.30ms   42.70ms 959.29ms   90.77%
    Req/Sec   242.49     39.30   373.00     68.63%
  Latency Distribution
     50%   30.77ms
     75%   50.97ms
     90%   86.54ms
     99%  225.96ms
  289751 requests in 1.67m, 54.81MB read
Requests/sec:   2894.70
Transfer/sec:    560.71KB
```
**Expected result**

Python and nodejs should behave similar

**Actual result**

Python hello world is 10x to 100x slower than Nodejs
**Steps to reproduce**

- Create new python function, using ui.
- Create new nodejs function using ui
- compare requests per seconds from the 2. Nodejs is 10x to 100x faster than python

**Troubleshooting**

dbadura:
Tool used for benchmark:
https://github.com/wg/wrk

dbadura:
# Problem
Python function are very slow

# Testing mechanism
I used [Locust](locust.io) and wrote such test.
```python
from locust import HttpUser, task, constant

expected_string = "Hello World"

class Python38Test(HttpUser):
    wait_time = constant(1)

    @task
    def hello_word(self):
        with self.client.get("/", catch_response=True) as response:
            if expected_string not in response.text:
                response.failure("Got wrong response, expected {} ..., got: {}".format(expected_string,response.text))
```

Locust was used to test Nodejs12 and Python function.

# Available Resources
[...]

PR #11040, opened 2021-04-02
Improve performance of kubeless runtime

dbadura:
**Description**

- Improve the performance of python function
- Investigation is available [here](https://github.com/kyma-project/kyma/issues/10154#issuecomment-811858828)

**Related issue(s)**

#10154

### DIFF
--- a/components/function-runtimes/python38/kubeless/kubeless.py
+++ b/components/function-runtimes/python38/kubeless/kubeless.py
@@ -3,25 +3,37 @@
 import importlib
 import io
 import os
-import queue
 import sys
-
+import threading
+import stopit
 import bottle
 import prometheus_client as prom
-
+import queue
 
 # The reason this file has an underscore prefix in its name is to avoid a
 # name collision with the user-defined module.
+module_name = os.getenv('MOD_NAME')
+if module_name is None:
+    print('MOD_NAME have to be provided', flush=True)
+    exit(1)
 current_mod = os.path.basename(__file__).split('.')[0]
-if os.getenv('MOD_NAME') == current_mod:
-    raise ValueError(f'Module cannot be named {current_mod}')
+if module_name == current_mod:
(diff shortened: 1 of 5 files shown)


components/function-runtimes/python38/kubeless/kubeless.py
```diff
@@ -85,13 +89,19 @@ def metrics():
     return prom.generate_latest(prom.REGISTRY)
 
 
+@app.error(500)
+def exception_handler():
+    return 'Internet server error'
```
aerfio, line 94: did you mean "internal"?

components/function-runtimes/python38/kubeless/kubeless.py
```diff
@@ -102,26 +112,12 @@ def handler():
...
-                p.terminate()
-                p.join()
-                return bottle.HTTPError(408, "Timeout while processing the function")
-            else:
-                p.join()
-                if isinstance(res, Exception) and not isinstance(res, bottle.HTTPResponse):
-                    logging.error("Function returned an exception: %s", res)
-                    raise res
-                return res
+            return func(event, function_context)
```
aerfio, line 120: The way you changed this means we do not have timeout errors anymore, am I correct?

components/function-runtimes/python38/kubeless/kubeless.py
dbadura: Yea, It was old joke in compass. I will change it.

components/function-runtimes/python38/kubeless/kubeless.py
dbadura: I can add the timeout mechanism.

dbadura: @rJankowski93 and I managed the timeout issue

rJankowski93 (approved):

pPrecel (approved):

PR #11056, opened 2021-04-07
Fix volnurabilities in serverless images 

m00g3n:
**Description**

- fix function-runtime-nodejs12 image volnurabilities
- fix function-runtime-nodejs10 image volnurabilities
- fix function-runtime-python38 image volnurabilities
- fix function-build-init image volnurabilities
- fix function-controller-test image volnurabilities
- fix rafter-test image volnurabilities

**Related issue(s)**

### DIFF
--- a/components/function-runtimes/nodejs12/Dockerfile
+++ b/components/function-runtimes/nodejs12/Dockerfile
@@ -23,6 +23,9 @@
 # OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 # SOFTWARE.
 
+# We use this code to manually use newest alpine and nodejs.
+# Use official build when it will be available.
+
 FROM alpine:3.13.4
 
 ENV NODE_VERSION 12.22.1
--- a/components/function-runtimes/nodejs12/Dockerfile
+++ b/components/function-runtimes/nodejs12/Dockerfile
@@ -1,4 +1,128 @@
-FROM node:12.21.0-alpine3.12
+# Node 12 Alpine image from https://github.com/nodejs/docker-node/tree/main/12/alpine3.12
+#
+# The MIT License (MIT)
+#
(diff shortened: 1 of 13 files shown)


components/function-runtimes/nodejs12/Dockerfile
```diff
@@ -1,4 +1,128 @@
-FROM node:12.21.0-alpine3.12
+# Node 12 Alpine image from https://github.com/nodejs/docker-node/tree/main/12/alpine3.12
```
dbadura, line 1: I would add such comment:
`We use this code to manually use newest alpine and nodejs. Use official build when it will be available`

The same comment should be in nodejs10 and Python.

components/function-runtimes/nodejs12/docker-entrypoint.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/sh
+set -e
```
dbadura, line 2: Where it comes from?

tests/function-controller/Makefile
```diff
@@ -22,3 +23,8 @@ ensure-local:
 
 dep-status-local:
 	@echo "Go modules present in component - omitting."
+
+# workaround for fixed ENTRYPOINT in common makefile
+build-local:
```
dbadura, line 28: If you override the `ENTRYPOINT` above, this section is not needed.

dbadura (changes requested):
Bump the images.

components/function-runtimes/nodejs12/docker-entrypoint.sh
m00g3n: https://github.com/nodejs/docker-node/blob/main/12/alpine3.12/docker-entrypoint.sh

dbadura: I would add the link as a comment.

dbadura (approved):

PR #11058, opened 2021-04-07
Add latency metrics for the Event publishers

marcobebway:
**Description**

Add latency metrics for the Event publishers to the messaging server.


berkanercan (approved):
Looking good

components/event-publisher-proxy/pkg/handler/http/handler_test.go
```diff
@@ -346,6 +371,8 @@ func TestHandlerTimeout(t *testing.T) {
 	_ = resp.Body.Close()
 	if http.StatusInternalServerError != resp.StatusCode {
 		t.Fatalf("Test failed, want status code:%d but got:%d", http.StatusInternalServerError, resp.StatusCode)
+	} else {
```
sayanh, line 374: (nit): else is redundant

sayanh (approved):

components/event-publisher-proxy/pkg/handler/http/handler_test.go
marcobebway: Removed.

marcobebway:
Image updates should be done in this [PR](https://github.com/kyma-project/kyma/pull/10975).

Issue #11027, opened 2021-03-31
Improve istio-proxy-reset script


Tomasz-Smelcerz-SAP:
**Description**

AC:
- consider if the script should be in the installation scenario. Perhaps it's only necessary on upgrades?
- consider if the script failure should always also fail the entire Istio install/upgrade. Perhaps we can make this policy configurable via a parameter?
- every networking operation should have configurable retries
- improve replicaset handling, by adding a case for "parentless replicaset". The decision for now is to: delete pods.
- add case for replicationcontrollers. Delete the pods.
- filter our pods that are in "teminating" phase at the moment.
- consider what to do whit pods in other possible states different than Running and Terminating.
- for some objects it looks like kubectl rollout restart did nothing. Probably related to Progressing=false flag, Deployment is not ready, and available is not "1", this command do nothing. Think what to do in that case.

**Reasons**

We should make it more robust

PR #11067, opened 2021-04-08
Improvement/istio proxy reset

mjakobczyk:
**Description**

- polish `istio-proxy-reset` script, see Acs in the linked issue

**Related issue(s)**

Resolves https://github.com/kyma-project/kyma/issues/11027

### DIFF
--- a/resources/istio/files/istio-proxy-reset.sh
+++ b/resources/istio/files/istio-proxy-reset.sh
@@ -2,6 +2,7 @@
 
 cleanup() {
     if [[ -n "${REMOVE_PODS_FILE}" ]]; then
+        echo "Removing temporary file with pods data: ${REMOVE_PODS_FILE}"
         rm "${REMOVE_PODS_FILE}"
     fi
 }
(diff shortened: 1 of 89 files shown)


resources/istio/files/istio-proxy-reset.sh
```diff
@@ -1,40 +1,151 @@
...
+deletePod() {
+    local namespace=$1
+    local podName=$2
 
-for NS in $(kubectl get ns -l kyma-project.io/created-by=e2e-upgrade-test-runner -o name | cut -d '/' -f2); do
     if [[ "${dryRun}" == "false" ]]; then
-        kubectl delete rs -n "${NS}" --all
+        echo "    Deleting pod: ${namespace}/${podName}"
+        retry "${RETRIES_COUNT}" kubectl -n "${namespace}" delete pod "${podName}"
+        sleep 1
```
strekm, line 42: why do we need that sleep?

resources/istio/templates/istio-proxy-reset.yaml
```diff
@@ -64,10 +64,16 @@ spec:
...
+          - name: ISTIO_PROXY_IMAGE_PREFIX
             value: {{ .Values.kyma.proxyResetJob.commonIstioProxyImagePrefix }}
+          - name: ISTIO_PROXY_IMAGE_VERSION
+            value: {{ .Chart.Version }}
+          - name: RETRIES_COUNT
+            value: "5"
+          - name: DRY_RUN
+            value: "false"
+          - name: EXIT_CODE
+            value: "0"
```
colunira, line 76: Shouldn't this be exposed as {{ .Values... }} so it can be easily configured? Changing it now would require manual operations on a cluster or changing Kyma version (e.g. from 1.21 to master)

colunira (dismissed):
/meow

strekm (approved):
/meow

resources/istio/files/istio-proxy-reset.sh
Tomasz-Smelcerz-SAP: It is here to slow down Pod deletion. Otherwise, if the pods are deleted "too fast", all the Pods of some ReplicationController might be deleted before the controller spawns new ones and user face downtime.
Essentially, it's a "poor man's rollout" attempt. Remember that we delete Pods for objects where no "rollout" command exists. The problem is, proper value for this depends on the environment. In some it may be not necessary at all, in some it may be 5 seconds? Maybe we should make it configurable with a default value of zero?

strekm (approved):
/meow

Issue #11100, opened 2021-04-13
Adjust Loki default retention and volume size


a-thaler:
**Description**
The current default value of 1 day is very small for even debugging and development purpose. The in-cluster Loki stack must be seen as a non-HA, non-scalable solution and with that the defaults must be very sensitive.  In consequence of https://github.com/kyma-project/kyma/issues/11080 we analyzed once more running clusters having a bigger retention time in place.  Seeing the results, increasing the retention time in combination with increasing the default disk size is possible.

AC
- Increase default retention time to 5d
- Increase the default disk size to 30GB
- Validate the upgrade procedure, assure that it is running smooth. It can be acceptable that thenew disk size does not get applied as long as Loki itself will no hang on next re-start or whatever. If manual action is required to apply the new settings, think of a migration guide
- Validate settings on the weekly cluster

**Reasons**
Improved user experience

**Attachments**

a-thaler:
Solved with Kyma 1.23

PR #11132, opened 2021-04-15
Expand PVC instead of deleting it

rakesh-garimella:
**Description**

- We need pre-upgrade-job since we are changing the PVC size which is immutable hence helm upgrade would fail. So the job would delete the statefulset before upgrading.
- Also added migration guide.

**Related issue(s)**

#11100


docs/migration-guides/1.22-1.23.md
```diff
@@ -0,0 +1,43 @@
...
+### Out of box Loki volume size
+
+However if the persistence volume claim is out of box setting use following command to expand the volume:
+
+```
+kubectl delete statefulsets.apps -n kyma-system logging-loki
+
+kubectl patch persistentvolumeclaims -n kyma-system storage-logging-loki-0 --patch '{"spec": {"resources": {"requests": {"storage": "30Gi"}}}}'
+```
+> Volume exapansions are only possible if field `allowVolumeExpansion` is set to true. Please look [here](https://kubernetes.io/docs/concepts/storage/persistent-volumes/#expanding-persistent-volumes-claims) for more info.
```
skhalash, line 34:
```suggestion
Volume expansions are only possible if the field `allowVolumeExpansion` of the corresponding storage class is set to true. Please look [here]
```

docs/migration-guides/1.22-1.23.md
```diff
@@ -0,0 +1,43 @@
+# Migrate from 1.22 to 1.23
+
+## Increase the Loki volume size
+In Kyma 1.23.0, the size of the persistent volume claim has been increased from 10Gi to 30Gi
```
skhalash, line 4: I'd rephrase it a bit since a PVC doesn't have a size, it requests a volume of a certain size. 

``` suggestion
In Kyma 1.23.0, the Loki volume size requested by the persistent volume claim has been increased from 10Gi to 30Gi
```

skhalash: The same wording ( the PVC size) is used below

skhalash (changes requested):

skhalash (approved):
LGTM!

docs/migration-guides/1.22-1.23.md
```diff
@@ -0,0 +1,46 @@
...
+However if the persistence volume claim is out of box setting use following command to expand the volume:
+
+```
+kubectl delete statefulsets.apps -n kyma-system logging-loki
+
+kubectl patch persistentvolumeclaims -n kyma-system storage-logging-loki-0 --patch '{"spec": {"resources": {"requests": {"storage": "30Gi"}}}}'
+```
+>Volume expansions are only possible if the field `allowVolumeExpansion` of the corresponding storage class is set to true on the storage class. Please look [here]
+
+If the above process fails then the persistent volume claim template needs to be deleted. Use following command:
```
NHingerl:
```suggestion
If the process fails, delete the persistent volume claim template with the following command:
```

Which process - the one under "Out-of-box Loki volume size" or the one under "Custom Loki volume size" ?  Additionally, can we make clear how the users see that the operation wasn't successful and that they must use the following fix?

docs/migration-guides/1.22-1.23.md
```diff
@@ -0,0 +1,46 @@
...
+
+```
+kubectl delete statefulsets.apps -n kyma-system logging-loki
+
+kubectl delete persistentvolumeclaims -n kyma-system storage-logging-loki-0
+```
+> This action will delete the existing data.
```
NHingerl, line 14: Can we add the information how the users see that the overall operation was successful? Either here, at the end of the entire document, or after the respective instructions that work if there's no error.

And what is the next step? Line 7 announces that the instructions are a preparation for the upgrade, so can we say that the users can now proceed to do the upgrade to 1.23.0?

NHingerl (changes requested):

NHingerl:
Basically, if I understand it correctly, what we're trying to say is:

> With Kyma 1.23.0, default Loki volume size changes from 10Gi to 30Gi.
> Before you upgrade the Kyma version, you must set the desired persistent volume size to 30Gi:
> 
> ## Instructions if your current configuration uses more than than 30Gi
> 1.
> 2.
> 3.
> 
> ## Instructions if your current configuration uses less than 30Gi
> 1.
> 2.
> 3.
>
> Check whether <desired outcome>. Tada, now you can upgrade to Kyma 1.23.0!
> 
> If you get the error message "blabla" (or <some other indication that the operation failed>), you fix it by deleting the existing data with the following command: ...

docs/migration-guides/1.22-1.23.md
```diff
@@ -0,0 +1,23 @@
+# Migrate from 1.22 to 1.23
+
+## Increase the Loki volume size
+In Kyma 1.23.0, the Loki volume size requested by the persistent volume claim has been increased from 10Gi to 30Gi
+
+## Upgrade process
+During the upgrade process, the pre-upgrade job is patching the volume size so as to increase the size from 10Gi to 30Gi as the retention time for logs has been increased from 1 day to 5 days. However, the upgrade process would succeed if the field `allowVolumeExpansion` is set to `true` for the storage class. Please look [here](https://kubernetes.io/docs/concepts/storage/persistent-volumes/#expanding-persistent-volumes-claims) for more info.
```
NHingerl, line 7:
```suggestion
During the upgrade process, the pre-upgrade job increases the volume size from 10Gi to 30Gi. 
For the upgrade process to be successful, make sure that the field `allowVolumeExpansion` is set to `true` for the storage class. For more information, read the [Kubernetes documentation about Expanding Persistent Volumes Claims](https://kubernetes.io/docs/concepts/storage/persistent-volumes/#expanding-persistent-volumes-claims).
```

docs/migration-guides/1.22-1.23.md
```diff
@@ -0,0 +1,23 @@
...
+
+If the flag is not set on the storage class then the size of the volume would not change. Use following command to check the size of the volume
```
NHingerl, line 9:
```suggestion
If the flag is not set on the storage class, the size of the volume would not change. Use following command to check the size of the volume:
```

docs/migration-guides/1.22-1.23.md
```diff
@@ -0,0 +1,23 @@
...
+
+```
+kubectl get persistentvolumeclaims -n kyma-system storage-logging-loki-0
+```
+if the size is still the old size follow the steps below and then trigger update
```
NHingerl, line 14:
```suggestion
If the size is still the old size, delete the old data with the following command:
```

docs/migration-guides/1.22-1.23.md
```diff
@@ -0,0 +1,23 @@
...
+
+```
+kubectl delete statefulsets.apps -n kyma-system logging-loki
+
+kubectl delete persistentvolumeclaims -n kyma-system storage-logging-loki-0
+```
+> This action will delete the existing data.
```
NHingerl, line 21:
```suggestion
After the old data is deleted, trigger the Kyma update.
```

docs/migration-guides/1.22-1.23.md
```diff
@@ -0,0 +1,23 @@
...
+
+If there is custom volume size is set for the loki persistent volume claim, then evaluate if the current size would suffice for the logs for retention time of 5 days.
```
NHingerl, line 23:
```suggestion
Make sure that your custom volume size for the Loki persistent volume claim is sufficient for the new retention time of 5 days.
```

I put a similar hint right at the top, so I'm not sure if this is needed here again. Not sure if we want users to evaluate that right away, or at the end.

NHingerl (changes requested):

docs/migration-guides/1.22-1.23.md
```diff
@@ -0,0 +1,24 @@
...
+## Increase the Loki volume size
+In Kyma 1.23.0, the default Loki volume size requested by the persistent volume claim has been increased from 10Gi to 30Gi because the retention time for logs has been increased from 1 day to 5 days. 
+
+> If you are using a custom volume size for the Loki persistent volume claim, we recommend that you evaluate whether that volume size is sufficient to retain the logs for 5 days.
+
+## Upgrade process
+During the upgrade process, the pre-upgrade job increases the volume size from 10Gi to 30Gi. 
+For the upgrade process to be successful, make sure that the field `allowVolumeExpansion` is set to `true` for the storage class. For more information, read the [Kubernetes documentation about Expanding Persistent Volumes Claims](https://kubernetes.io/docs/concepts/storage/persistent-volumes/#expanding-persistent-volumes-claims).
+
+If the flag is not set on the storage class, the size of the volume would not change. Use following command to check the size of the volume:
```
NHingerl, line 12:
```suggestion
If the flag is not set on the storage class, the size of the volume would not change. Use the following command to check the size of the volume:
```

NHingerl (changes requested):
Missed this one earlier

NHingerl (approved):

skhalash (approved):

PR #11176, opened 2021-04-23
Fix random port generator for Event publisher tests

marcobebway:
**Description**

Currently, Event publisher tests use a random port generator which sometimes generates an already used port. We introduced a check mechanism to test if the generated port is used or not with a timeout based retry loop to return a free port.

### DIFF
--- a/components/event-publisher-proxy/testing/utils.go
+++ b/components/event-publisher-proxy/testing/utils.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"io"
 	"log"
+	"net"
 	"net/http"
 	"strconv"
 	"testing"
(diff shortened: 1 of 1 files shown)


components/event-publisher-proxy/testing/utils.go
```diff
@@ -163,11 +164,33 @@ func WaitForHandlerToStart(t *testing.T, healthEndpoint string) {
...
+	for {
+		select {
+		case <-tick.C:
+			{
+				port, err := generatePort()
+				if err != nil {
+					break
+				}
+
+				if err := check(port); err != nil {
```
themue, line 182: When ignoring the returned error a simple kind of `isPortUnused(port) bool` would be enough an more speaking (otherwise question would have to be what is checked).

components/event-publisher-proxy/testing/utils.go
```diff
@@ -163,11 +164,33 @@ func WaitForHandlerToStart(t *testing.T, healthEndpoint string) {
...
+	defer tick.Stop()
+
+	timeout := time.NewTimer(time.Minute)
+	defer timeout.Stop()
+
+	for {
+		select {
+		case <-tick.C:
+			{
+				port, err := generatePort()
```
themue, line 177: Would it make sense to integrate this retry-timeout-mimik directly into `generatePort() (int, error)`? So all callers would benefit of it.

marcobebway: If we integrated the retry loop into the `generatePort`, we'll lose retrying if the port is in use.

themue (approved):

components/event-publisher-proxy/testing/utils.go
marcobebway: Done.

PR #11189, opened 2021-04-26
Fix KO for eventing-publisher-nats

radufa:
**Description**

- Fix KO for publisher-nats

**Related issue(s)**


marcobebway:
(General comment)

As we'll unify the two publishers soon, I recommend to do this change under `components/event-publisher-proxy/config/event-publisher-proxy` instead of `components/event-publisher-proxy/config/event-publisher-nats`.

marcobebway (approved):

PR #11204, opened 2021-04-29
Improve perf python function

rJankowski93:
**Description**

- change timeout mechanism in python function 

**Related issue(s)**
#10154


components/function-runtimes/python38/kubeless/kubeless.py
```diff
@@ -106,21 +103,15 @@ def handler():
     func_calls.labels(method).inc()
     with func_errors.labels(method).count_exceptions():
         with func_hist.labels(method).time():
-            q = ctx.Queue()
-            p = ctx.Process(target=funcWrap, args=(q, event, function_context))
-            p.start()
-
+            que = Queue()
+            t = Thread(target=lambda q, arg1: q.put(funcWrap(arg1)), args=(que, event))
```
dbadura, line 107: I would rename `arg1` to `event`

rJankowski93: But the name `event` already exist in outer scope

Issue #11182, opened 2021-04-23
Best way to override monitoring-prometheus-istio recording_rules file 


paschdan:
What is the best way to override this file:

https://github.com/kyma-project/kyma/blob/3cfaf33c2c0b8f45268762ba6fe2e5521e5863b3/resources/monitoring/charts/prometheus-istio/values.yaml#L1222

The reason for me is, that some rules from the best practices are missing:

https://istio.io/v1.8/docs/ops/best-practices/observability/#workload-level-aggregation-via-recording-rules

Thanks in advance 
Paschdan

a-thaler:
Hi @paschdan, thanks for your feedback. I played a bit around but indeed found no acceptable way of adding own recording rules. That is mainly caused by the way how the override mechanism of the kyma installer is working at the moment and the fact that the relevant keys in the helm chart have a "." in the key name. That problem will be solved soon. We are working on a new way of installation supporting full values.yaml notation of helm.

Anyway, I looked into you actual problem of not having all recommended rules of istio covered in the default setup.

Indeed it looks to me that the linked best practice page is outdated. The actual exposed metrics are listed here: https://istio.io/latest/docs/reference/config/metrics/
There you can see that the metric for the received tcp bytes is called `istio_tcp_received_bytes_total` and is a counter and not as used in the recording rules a bucket with metrics `istio_tcp_received_bytes_total_count`, `istio_tcp_received_bytes_total_bucket` and `istio_tcp_received_bytes_total_sum`.

However, there are indeed metrics missing which are getting actually scraped. Mainly some `_sum` and `_count` counterparts to buckets. Also `istio_tcp_connections_opened_total` and `istio_tcp_connections_closed_total` is missing. These are missing as they are not used in the actual istio dashboards. As these metrics all have a quite high cardinality we try to reduce them as much as possible to keep the prometheus stable as much as possible. Still, if the reason for not having them aggregated is because of non-usage in the dashboards, we should not scrape them at all and should move them in the regexp for label dropping.

I will discuss it with @rakesh-garimella and come up with an improved setting.

paschdan:
Yes, we are interested on the missing histogram metrics, to add alerts, e.g. When the average response time passes a given value

```
rate(request_duration_sum[5m])
/
rate(request_duration_count[5m])
```

jeremyharisch:
Kyma CLI is now capable of overwriting values during deployment. This can be done using the `--values` or the `--values-file` flags.

I.e.:
[...]

PR #11209, opened 2021-05-03
added missing aggregation rules for istio

a-thaler:
**Description**

- Adds aggregation rules to the istio-prometheus which can be relevant and are recommended by istio.
- ...
- ...

**Related issue(s)**

Fixes https://github.com/kyma-project/kyma/issues/11182


a-thaler:
I tested the setup and it all looks file. However, it should not be merged till performance tests are done @rakesh-garimella

PR #11222, opened 2021-05-06
move nats crds to common crd folder

k15r:
**Description**

- nats crds need to be stored in common location

**Related issue(s)**

### DIFF
--- a/resources/cluster-essentials/files/natsclusters.nats.crd.yaml
+++ b/resources/cluster-essentials/files/natsclusters.nats.crd.yaml
@@ -16,20 +16,3 @@ spec:
     singular: natscluster
   scope: Namespaced
   version: v1alpha2
----
-apiVersion: apiextensions.k8s.io/v1beta1
-kind: CustomResourceDefinition
-metadata:
-  name: natsserviceroles.nats.io
-  annotations:
-    "helm.sh/hook": "crd-install"
-    "helm.sh/hook-delete-policy": "before-hook-creation"
-spec:
-  group: nats.io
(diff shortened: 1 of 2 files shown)


resources/cluster-essentials/files/natsclusters.nats.crd.yaml
```diff
@@ -16,20 +16,3 @@ spec:
     singular: natscluster
```
marcobebway, line 16: Why not move the `NatsCluster` CRD as well?

k15r: move where?

marcobebway (dismissed):

resources/cluster-essentials/files/natsclusters.nats.crd.yaml
marcobebway: Ah my mistake, it's already in the right place.

resources/cluster-essentials/files/natsserviceroles.nats.crd.yaml
```diff
@@ -0,0 +1,16 @@
+apiVersion: apiextensions.k8s.io/v1beta1
```
strekm, line 1: if this CRD is also valid for kyma 2.0 then it needs to be added to https://github.com/kyma-project/kyma/tree/main/installation/resources/crds to proper sub-folder.

strekm (approved):
/meow

marcobebway (approved):

PR #11242, opened 2021-05-10
Add region to fast integration KEB client

piotrmiskiewicz:
**Description**

- allow to configure provider region to KEB client in the fast integration test

**Related issue(s)**

### DIFF
--- a/tests/fast-integration/kyma-environment-broker/client.js
+++ b/tests/fast-integration/kyma-environment-broker/client.js
@@ -79,7 +79,7 @@ class KEBClient {
 
   async buildRequest(payload, endpoint, verb) {
     const token = await this.getToken();
-    let region = this.getRegion()
+    const region = this.getRegion();
     const url = `https://kyma-env-broker.${this.host}/oauth/${region}v2/${endpoint}`;
     const headers = {
       "X-Broker-API-Version": 2.14,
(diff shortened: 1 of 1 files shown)


tests/fast-integration/kyma-environment-broker/client.js
```diff
@@ -76,7 +79,8 @@ class KEBClient {
 
   async buildRequest(payload, endpoint, verb) {
     const token = await this.getToken();
-    const url = `https://kyma-env-broker.${this.host}/oauth/v2/${endpoint}`;
+    let region = this.getRegion()
```
crabtree, line 82:
```suggestion
    const region = this.getRegion();
```

tests/fast-integration/kyma-environment-broker/client.js
```diff
@@ -153,6 +157,14 @@ class KEBClient {
       return new Error(`error while deprovisioning SKR: ${err.toString()}`);
     }
   }
+
+  getRegion() {
+    let reg = this.region
```
crabtree, line 162: It may happen that `this.region` will be undefined (when the env variable is not present.) The condition in the if statement will return true, which I think is wrong. The concatenation will produce `undefined/`. The quickest fix for that case would be:
```
if (reg && reg != "")
```

crabtree (changes requested):

tests/fast-integration/kyma-environment-broker/client.js
```diff
@@ -153,6 +157,17 @@ class KEBClient {
...
+    if (!reg) {
+      return ""
+    }
+    if (reg && reg != "") {
+      reg += "/"
+    }
+    return reg
+  }
```
crabtree, line 170: I think this is more readable:
```suggestion
  getRegion() {
    if (this.region && this.region != "") {
      return `${this.region}/`;
    }
    return "";
  }
```

crabtree (approved):

PR #11244, opened 2021-05-10
Upgrade nats from 2.1.8 to 2.1.9

johgoe:
Use NATS 2.1.9 instead of 2.1.8

https://github.com/nats-io/nats-server/releases/tag/v2.1.9
https://github.com/nats-io/nats-operator/releases/tag/v0.8.2

After Version 0.8.0 the docker image repository was changed from connecteverything/nats-operator:0.8.0  to natsio/nats-operator:0.8.2

### DIFF
--- a/components/nats-operator/Dockerfile
+++ b/components/nats-operator/Dockerfile
@@ -1,4 +1,4 @@
-FROM connecteverything/nats-operator:0.7.6 as builder
+FROM natsio/nats-operator:0.8.2 as builder
 FROM alpine:3.13
 COPY --from=builder /usr/local/bin/nats-operator /usr/local/bin/nats-operator
 CMD ["nats-operator"]
--- a/resources/eventing/charts/nats/templates/40-cr.yaml
+++ b/resources/eventing/charts/nats/templates/40-cr.yaml
@@ -5,7 +5,7 @@ metadata:
   name: {{ include "nats.fullname" . }}
 spec:
   size: {{ .Values.clusterSize }}
-  version: "2.1.8"
+  version: "2.1.9"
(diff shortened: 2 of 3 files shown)


k15r:
Hi @johgoe ,

Thanks for preparing this PR.

johgoe:
It seems that the nats changes a lot between 2.1.x and 2.2.x (https://docs.nats.io/whats_new_22).  That's why I updated just to 2.1.9, but I agree if nats-2.2.4 works it would be better.

I already changed the docker image tag to PR-11244 and will try 2.2.4 later.

k15r (approved):
lgtm

johgoe:
@k15r I'm not really sure, why the test failed. It seems for me that it doesn't change the image, right? Can you explain the issue to me? Thanks in advance

sayanh (approved):

sayanh:
@johgoe Turns out `nats-operator` is using [kubernetes 1.15.2](https://github.com/nats-io/nats-operator/blob/master/go.mod#L43) as a dependency and that is causing the security scan to fail in the job as it is shown here:
```
"scan": {
        "status": "Vulns",
        "message": "There are some expired vulnerabilities",
        "protecodeID": "579046"
      } 
```

The security issue with Kubernetes 1.15.2 is https://nvd.nist.gov/vuln/detail/CVE-2020-8558. 

One way to make it green would be to fix nats-operator and use the patched release version here.

johgoe:
@sayanh I did open a new issue in the nats-operator project. 

Just to be safe: I there any security issue for current kyma versions because nats-operator 0.7.4 contains exactly the same dependencies but is delivered with the kyma installer.

johgoe:
Obsolete with merge of #12075

PR #11281, opened 2021-05-17
Migrate monitoring tests to JS

JeremyHarisch:


tests/fast-integration/skr-test/.env.template
```diff
@@ -1,7 +0,0 @@
-KEB_HOST=
```
pbochynski, line 1: Why did you remove it?

skhalash: Hm, I guess, just by mistake. Thanks, will revert it!

tests/fast-integration/Makefile
```diff
@@ -2,11 +2,13 @@
 ci:
 	npm install
 	npm test
+	npm run test-monitoring
```
pbochynski, line 5: It would be better to move the monitoring test to the test folder - it would be picked automatically, and you could skip that line.
But you can leave it as it is - I plan a bigger refactoring anyway.

tests/fast-integration/monitoring-test/monitoring-test.js
```diff
@@ -0,0 +1,102 @@
...
+
+const {
+  shouldIgnoreTarget,
+  shouldIgnoreAlert,
+  buildScrapePoolSet,
+  assertTimeSeriesExist: assertTimeSeriesExists,
+} = require("../monitoring/helpers");
+
+describe("Monitoring test", function () {
+  const suffix = genRandom(4);
```
pbochynski, line 28: appName, runtimeName, scenarioName, and runtimeID are not used. Copy & paste?

tests/fast-integration/monitoring-test/.env.template
```diff
@@ -0,0 +1,7 @@
+KEB_HOST=
```
pbochynski, line 1: Do you need it? It was used to provision Kyma with KEB. I don't see you do it.

pbochynski (changes requested):
Please remove the unused code. Test looks good :)

tests/fast-integration/monitoring-test/.env.template
skhalash: Removed

tests/fast-integration/monitoring-test/monitoring-test.js
skhalash: Removed

tests/fast-integration/Makefile
skhalash: I see, then I'd leave it as is since it's going to be reorganized anyway.

tests/fast-integration/monitoring-test/monitoring-test.js
```diff
@@ -0,0 +1,92 @@
...
+
+  after(() => {
+    cancelPortForward()
+  })
+
+  it("All Prometheus targets should be healthy", async () => {
+    let activeTargets = await getPrometheusActiveTargets();
+    let unhealthyTargets = activeTargets
+      .filter(t => !shouldIgnoreTarget(t) && t.health != "up")
+      .map(t => t.discoveredLabels.job);
```
shorim, line 43: can you add the other labels and the error message in the logs? and not only the job label.
We had before requests from other teams to include more information when the test for targets is failing, so that they more easily figure out the root cause for the failure.

shorim:
Is there a way to add retries for these tests? Without retries they sometimes failed randomly on the nightly cluster.
Or do you want to give it a try first on the pipeline without retries?

tests/fast-integration/monitoring/client.js
```diff
@@ -0,0 +1,58 @@
...
+} = require("../utils");
+
+let prometheusPort = 9090;
+
+function prometheusPortForward() {
+    return kubectlPortForward("kyma-system", "prometheus-monitoring-prometheus-0", prometheusPort);
+}
+
+async function getPrometheusActiveTargets() {
+    let responseBody = await get("/api/v1/targets?state=active");
```
shorim, line 15: I think we need to wrap the prometheus API requests in a try{}catch{} blocks in order to handle errors

skhalash: Hmm, the underlying `get` function will retry in case of an error and only fail (throw) when the max retry count exceeded. Do you think we need something additional?

tests/fast-integration/monitoring-test/readme.md
```diff
@@ -0,0 +1,11 @@
+# Monitoring test
+
+This test covers the monitoring components.
```
NHingerl, line 3: I'd personally like to see more details here, but I'll assume the actual testers have more context.
If they don't, we should add a little bit more info here :D

NHingerl (approved):

tests/fast-integration/monitoring-test/readme.md
skhalash: The readme file was added to be aligned with the rest of the tests. However, it's not really valuable. You can reason about the tests just by looking at the code.

tests/fast-integration/monitoring-test/monitoring-test.js
skhalash: Done

NHingerl (approved):

pbochynski (approved):

Issue #11128, opened 2021-04-14
Commerce-mock can not connect


philippe-lavoie:
**Description**

I'm following along the getting started and I try to connect the commerce-mock service with an application. I generate the token in the application and press 'Connect'. On the commerce-mock service, it says 'connected' but nothing shows up in the 'Remote API' section. On the Kyma side, the application has no connections.

**Expected result**

I should be able to connect commerce with Kyma.

**Actual result**

Connection fails

**Steps to reproduce**

Install 1.21.0 on AKS 

**Troubleshooting**

I raised this issue in Slack, no resolution so far.  I even destroyed AKS and created one with an older version of Kubernetes (1.18.14) instead of 1.19.17

a-thaler:
Just do double-check, have you registered any API via the mock after the connection got established? The "Remote API" tab shows the APIs registered on Kyma side.

PR #11291, opened 2021-05-17
Observability overview tweaks

skhalash:
**Description**

- Observability overview improvements after a discussion with @NHingerl 

**Related issue(s)**
See also #11128

### DIFF
--- a/docs/overview/51-telemetry-observability-overview.md
+++ b/docs/overview/51-telemetry-observability-overview.md
@@ -3,7 +3,7 @@ title: Telemetry and Observability
 type: Observability
 ---
 
-Out of the box, Kyma provides **telemetry** tools to collect and expose raw data, such as metrics, traces, and log data.
+Out of the box, Kyma provides tools to collect and expose **telemetry** data, such as metrics, traces, and log data.
 
 Of course, you'll want to view and analyse the data you're collecting. This is where **observability** comes in. 
 
--- a/docs/overview/51-telemetry-observability-overview.md
+++ b/docs/overview/51-telemetry-observability-overview.md
@@ -7,15 +7,15 @@ Out of the box, Kyma provides tools to collect and expose **telemetry** data, su
 
 Of course, you'll want to view and analyse the data you're collecting. This is where **observability** comes in. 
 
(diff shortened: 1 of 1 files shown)


docs/overview/51-telemetry-observability-overview.md
```diff
@@ -3,7 +3,7 @@ title: Telemetry and Observability
 type: Observability
 ---
 
-Out of the box, Kyma provides **telemetry** tools to collect and expose raw data, such as metrics, traces, and log data.
```
skhalash, line 6: I think it's important to be consistent with the terms we use here:

* Telemetry is data produced by applications that we want to observe. Metrics, logs, and traces are different kinds of telemetry
* AFAIK there is no official term for software that exposes telemetry. `Fluent Bit` is a log collector and in `Prometheus` world it's called an exporter.
* Observability is an ability to understand an application's behavior based on the telemetry it produces.

I like the term `Observability solution` used further in the docs ;)

docs/overview/51-telemetry-observability-overview.md
```diff
@@ -3,19 +3,19 @@ title: Telemetry and Observability
...
+Out of the box, Kyma provides tools to collect and expose **telemetry** data, such as metrics, traces, and log data.
 
 Of course, you'll want to view and analyse the data you're collecting. This is where **observability** comes in.
```
skhalash, line 8: I am not sure about the last sentence: This is where **observability** comes in. Maybe something about **observability tools** instead, @NHingerl?

docs/overview/51-telemetry-observability-overview.md
NHingerl: I agree! I like your suggested text.

docs/overview/51-telemetry-observability-overview.md
NHingerl, line 8:
```suggestion
Of course, you'll want to view and analyse the data you're collecting. This is where **observability** tools come in. 
```

NHingerl (changes requested):

NHingerl (approved):

PR #11295, opened 2021-05-18
Central gateway application registry

koala7659:
The implementation that does not allow Application Registry to register many services with the same name. Service name must be unique since the endpoint for proxying is of the following form: `{APP-NAME}/{SERVICE-NAME}/{TARGET-PATH}`. 

The validation is implemented with normalisation function that removes all white characters from provided service name and stores it as `displayName` in the application CRD. There are fixes for both functions create and update a service with API

**Tests**
Unit tests and Integration tests are provided


components/application-registry/internal/metadata/applications/util.go
```diff
@@ -172,6 +197,21 @@ func getServiceIndex(id string, app *v1alpha1.Application) int {
 	return -1
 }
 
+func normalizeName(displayName string) string {
```
everesio, line 200: Is the normalization function (how it should work) documented somewhere ?

everesio:
lgtm

components/application-registry/internal/metadata/applications/util.go
```diff
@@ -138,30 +138,55 @@ func replaceService(id string, app *v1alpha1.Application, service v1alpha1.Servi
...
 
 		return apperrors.AlreadyExists(message)
 	}
 
+	if serviceExistsWithNormalizedDisplayName(displayName, app) {
+		message := fmt.Sprintf("Service with name %s already exists", displayName)
+
+		return apperrors.WrongInput(message)
+	}
+
```
rafalpotempa, line 169: I don't understand why we return `apperrors.WrongInput()` and the reason is that the service already exists. The message suggests the same.

components/application-registry/internal/metadata/applications/repository_test.go
```diff
@@ -236,6 +236,56 @@ func TestCreateServices(t *testing.T) {
...
+		}
+		// when
+		err := repository.Create("production", newService)
+
+		// then
+		assert.Equal(t, apperrors.CodeWrongInput, err.Code())
+	})
+
+	t.Run("should not allow to create service if a service with the same normalized display name already exists", func(t *testing.T) {
+		// given
```
rafalpotempa, line 265: Why do we need to test `display name` and `normalized display name` separately?

koala7659: I agree both cases, can be reduces into single one - for testing the normalized display name. I will change that.

components/application-registry/internal/metadata/applications/util.go
koala7659: That's how new logic works. User should not be allowed to register the service with displayName that is not unique. This is considered as wrong input data (Status bad request 400)

rafalpotempa (dismissed):

components/application-registry/internal/metadata/applications/util.go
koala7659: I'm afraid not

franpog859 (approved):
LGTM

Issue #10539, opened 2021-02-08
Describe the difference between a Function and a microservice


kazydek:
**Description**

Describe the difference between a Function and a microservice.
We only mention it vaguely in [Getting Started guides](https://kyma-project.io/docs/master/root/getting-started#getting-started-overview) as follows:

`Then, you can create extensions for them by building a simple Function, a more complex microservice, or a mixture of those, depending on your use case complexity level.`

Still, we would need to document it in more detail, describe basic differences, use cases, and recommendations when to use which.

**Area**

Main concepts in docs

**Reasons**

Question on the Serverless channel: https://kyma-community.slack.com/archives/CD0K2SA7K/p1612566418008000

**Assignees**

@kyma-project/technical-writers

PR #11301, opened 2021-05-18
add first definition for microservice

NHingerl:
**Description**

- first draft for definition of a microservice
- ...
- ...

**Related issue(s)**

https://github.com/kyma-project/kyma/issues/10539


docs/glossary/glossary.md
```diff
@@ -9,11 +9,11 @@ Here's a list of the most commonly used terms:
...
+| Kyma Cluster        | A Kubernetes cluster with Kyma installed. A cluster is a set of worker machines, called nodes, that run containerized applications. Every cluster has at least one worker node.                                                                                                                                                                                                                                                                                                                                                                                                                                                                | [Kubernetes Components](https://kubernetes.io/docs/concepts/overview/components/) |
```
majakurcius, line 12: Or, alternatively, instead of repeating the cluster definition here, we could also just leave the first sentences and link to the Kubernetes definition of `cluster`, like this: 
```suggestion
| Kyma Cluster        | A [Kubernetes cluster](https://kubernetes.io/docs/reference/glossary/?fundamental=true#term-cluster) with Kyma installed.                                                                                                                                                                                                                                                                                                                                                                                                                                                                 | [Kubernetes Components](https://kubernetes.io/docs/concepts/overview/components/) |
```

docs/glossary/glossary.md
```diff
@@ -9,11 +9,11 @@ Here's a list of the most commonly used terms:
...
+| Function            | A simple code snippet that you can run without provisioning or managing servers. It implements the exact business logic you define. A Function is based on the Function custom resource and can be written in either Node.js or Python. A Function can perform a business logic of its own. You can also bind it to an instance of a service and configure it to be triggered whenever it receives a particular event type from the service or a call is made to the service's API. Functions are executed only if they are triggered by an event or an API call. |       |
```
majakurcius, line 13: I wonder if it makes sense to add this acronym here. 
On one hand, we don't use it again in the rest of the definition, but on the other, it feels to me like it reads easier with this `CR` there, and we already make the reader familiar with this abbr., used heavily through our docs. What do you think? 
```suggestion
| Function            | A simple code snippet that you can run without provisioning or managing servers. It implements the exact business logic you define. A Function is based on the Function custom resource (CR) and can be written in either Node.js or Python. A Function can perform a business logic of its own. You can also bind it to an instance of a service and configure it to be triggered whenever it receives a particular event type from the service or a call is made to the service's API. Functions are executed only if they are triggered by an event or an API call. |       |
```

docs/glossary/glossary.md
```diff
@@ -9,11 +9,11 @@ Here's a list of the most commonly used terms:
...
+| Microservice        | An architectural variant for extensions or applications, where you separate the tasks in smaller pieces that interact with each other as loosely coupled, independently deployable units of code. A failing microservice should not cause your whole application to fail. Microservices are packed in a container that is always running; it's idling if there is no load. The microservice should always be reachable even when the pods move around. Microservices typically communicate through APIs.                                                                                                                                                                   |       |
```
majakurcius, line 14:
```suggestion
| Microservice        | An architectural variant for extensions or applications, where you separate the tasks into smaller pieces that interact with each other as loosely coupled, independently deployable units of code. A failing microservice should not cause your whole application to fail. Microservices are packed in a container that is always running; it's idling if there is no load. The microservice should always be reachable even when the Pods move around. Microservices typically communicate through APIs.                                                                                                                                                                   |       |
```

docs/glossary/glossary.md
```diff
@@ -9,11 +9,11 @@ Here's a list of the most commonly used terms:
...
-| Namespace           | Namespaces organize objects in a cluster and provide a way to divide cluster resources. This way, several users can share a cluster but have access only to resources within the Namespace they have permissions for. This increases the security and organization of your cluster by dividing it into smaller units. Access to Namespaces in the Kyma environment depends on your permissions.                                                                                     |       |
+| Namespace           | Namespaces organize objects in a cluster and provide a way to divide cluster resources. This way, several users can share a cluster but have access only to resources within the Namespace they have permissions for. This increases the security and organization of your cluster by dividing it into smaller units. Access to Namespaces in the Kyma environment depends on your permissions.                                                                                     | [Kubernetes Namespaces](https://kubernetes.io/docs/concepts/overview/working-with-objects/namespaces/) |
```
majakurcius, line 16: The last sentence kind of feels like a repetition of `but have access only to resources within the Namespace they have permissions for.`, but it's nice and concise and I wouldn't like to see it go. 
How about this? 
```suggestion
| Namespace           | Namespaces organize objects in a cluster and provide a way to divide cluster resources. This way, several users can share a cluster but have access to resources restricted to specified Namespaces only. This increases the security and organization of your cluster by dividing it into smaller units. Access to Namespaces in the Kyma environment depends on your permissions.                                                                                     | [Kubernetes Namespaces](https://kubernetes.io/docs/concepts/overview/working-with-objects/namespaces/) |
```
It's just a suggestion, it doesn't bother me much, so if you'd rather keep the original, I'm fine with it too :)

docs/glossary/glossary.md
NHingerl: Thinking about it, maybe the term "Custom Resource (CR)" could even have its own entry, it's a very specific meaning that we're using in Kyma :D

docs/glossary/glossary.md
NHingerl: Shorter is (almost) always better!

majakurcius (approved):

PR #11336, opened 2021-05-20
[istio ingressgateway] add memory based HPA

### LINKED ISSUE #11303: Ingress gateway proxy are consuming a lot of mem when uploading multipart files on HTTP/2
**Description**
After exposing app via APIrule CR that allows uploading multipart file upload:
`curl --max-time 300 -i -X POST -H "Content-Type: multipart/form-data" -F "myFile=@500m-file" https://endpoint/upload`
the proxy is using large amount of memory and in some cases can get OOM-killed.
This does not occur when using HTTP/1.1 (check `--http1.1` option for curl)
There are multiple similar issues raised against Istio and Envoy that are not addressed.
https://github.com/istio/istio/issues/27373
https://github.com/envoyproxy/envoy/issues/13240
https://github.com/istio/istio/issues/25734
https://github.com/istio/istio/issues/24370

Kyma version: 1.20
Istio version: 1.8.2

piotrkpc:
Envoy proxy can use large amount of memory in some cases [0],
hence HPA based on CPU util alone is not enough and the deployment
should also scale based on memory util to prevent/minimize OOM kills.

[0] #11303 

Resolves #11303 

**Related issue(s)**

### DIFF
--- a/resources/istio/files/istio-kyma-validate-test.sh
+++ b/resources/istio/files/istio-kyma-validate-test.sh
@@ -76,8 +76,19 @@ function check_sidecar_injector() {
   log "    Automatic injection policy is enabled" green
 }
 
+function require_ingressgateway_hpa() {
+    echo "--> Checking istio-ingresgateway HPA"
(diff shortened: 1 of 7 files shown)


mjakobczyk (approved):
It looks like this average memory util has been applied in all the required places. Let's hope that 80% is fitting in this case! Make tests green and I'm alright. 👍

mjakobczyk (approved):

PR #11390, opened 2021-05-28
Set owner references for publisher deployment

radufa:
**Description**

- Dynamically set the owner reference for publisher deployment to be the eventing-controller deployment

**Related issue(s)**

See: https://github.com/orgs/kyma-project/projects/3#card-61593021

### DIFF
--- a/components/eventing-controller/reconciler/backend/reconciler.go
+++ b/components/eventing-controller/reconciler/backend/reconciler.go
@@ -18,6 +18,7 @@ import (
 	v1 "k8s.io/api/core/v1"
 	k8serrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/tools/record"
 	ctrl "sigs.k8s.io/controller-runtime"
@@ -549,6 +550,10 @@ func (r *Reconciler) CreateOrUpdatePublisherProxy(ctx context.Context, backend e
 		return nil, fmt.Errorf("unknown eventing backend type %q", backend)
 	}
 
+	if err := r.setPublisherOwnerReference(ctx, desiredPublisher); err != nil {
+		return nil, errors.Wrapf(err, "setting owner references")
+	}
+
 	err := r.Cache.Get(ctx, publisherNamespacedName, currentPublisher)
 	if err != nil {
 		if k8serrors.IsNotFound(err) {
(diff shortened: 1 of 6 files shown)


components/eventing-controller/reconciler/backend/reconciler_test.go
```diff
@@ -396,6 +424,24 @@ func ensureEventingBackendCreated(ctx context.Context, name, namespace string) {
 	}
 }
 
+func ensureControllerDeploymentCreated(ctx context.Context) *[]metav1.OwnerReference {
+	By(fmt.Sprintf("Ensuring an Eventing-Controller Deployment is created"))
```
themue, line 428: Why the `fmt.Sptrinf()` here?

themue (commented):
Only one small question.

components/eventing-controller/reconciler/backend/reconciler_test.go
radufa: Changed

components/eventing-controller/reconciler/backend/reconciler.go
```diff
@@ -548,6 +550,11 @@ func (r *Reconciler) CreateOrUpdatePublisherProxy(ctx context.Context, backend e
 		return nil, fmt.Errorf("unknown eventing backend type %q", backend)
 	}
 
+	if err := r.setPublisherOwnerReference(ctx, desiredPublisher); err != nil {
+		r.Log.Error(err, "setting owner references")
```
pxsalehi, line 554: Since Marco's PR is merged, and it includes updates to how tests log as well, I think all logs need to get updated.

components/eventing-controller/reconciler/backend/reconciler_test.go
```diff
@@ -202,6 +204,18 @@ var _ = AfterSuite(func() {
 })
 
 var _ = Describe("Backend Reconciliation Tests", func() {
+	var ownerReferences *[]metav1.OwnerReference
+
+	When("Creating a controller deployment", func() {
```
pxsalehi, line 209: Do we really need this `When`? Wouldn't it be enough to keep the `It`s that you have added to make sure every time a new publisher proxy deployment is create, it has the correct owner reference? Since the code you have added in the reconciler, anyway will look up the eventing-controller, if the controller is not running, setting the owner reference fails. Right? If so, then this `When` is already implied by the `It`s.

radufa: Yes, the `When` is more a documentation in this place.

components/eventing-controller/reconciler/backend/reconciler.go
radufa: Done

pxsalehi (approved):

themue (approved):

Issue #11321, opened 2021-05-20
If the system is working as expected (no switch), the log file is full with the same set of log messages:


radufa:
If the system is working as expected (no switch), the log file is full with the same set of log messages:
```
...
{"level":"info","ts":1621493705.130021,"logger":"reconciler.backend","msg":"Found secrets with BEB label","count":0}
{"level":"info","ts":1621493705.1300337,"logger":"reconciler.backend","msg":"Reconciling with backend as NATS"}
{"level":"info","ts":1621493705.1300404,"logger":"reconciler.backend","msg":"Found existing backend CR"}
{"level":"info","ts":1621493705.1300511,"logger":"reconciler.backend","msg":"No need to update existing backend CR"}
{"level":"info","ts":1621493705.1302183,"logger":"reconciler.backend","msg":"Created/updated backend CR"}
{"level":"info","ts":1621493705.130247,"logger":"reconciler.backend","msg":"trying to create/update eventing publisher proxy..."}
{"level":"info","ts":1621493705.1302927,"logger":"reconciler.backend","msg":"No need to update publisher proxy"}
{"level":"info","ts":1621493705.1302993,"logger":"reconciler.backend","msg":"Created/updated publisher proxy"}
{"level":"info","ts":1621493705.130316,"logger":"reconciler.backend","msg":"No need to update backend CR status"}
{"level":"info","ts":1621493705.1303806,"logger":"reconciler.backend","msg":"Found secrets with BEB label","count":0}
{"level":"info","ts":1621493705.130397,"logger":"reconciler.backend","msg":"Reconciling with backend as NATS"}
{"level":"info","ts":1621493705.130408,"logger":"reconciler.backend","msg":"Found existing backend CR"}
{"level":"info","ts":1621493705.130419,"logger":"reconciler.backend","msg":"No need to update existing backend CR"}
{"level":"info","ts":1621493705.1304278,"logger":"reconciler.backend","msg":"Created/updated backend CR"}
{"level":"info","ts":1621493705.1304355,"logger":"reconciler.backend","msg":"trying to create/update eventing publisher proxy..."}
{"level":"info","ts":1621493705.1304693,"logger":"reconciler.backend","msg":"No need to update publisher proxy"}
{"level":"info","ts":1621493705.1304784,"logger":"reconciler.backend","msg":"Created/updated publisher proxy"}
{"level":"info","ts":1621493705.1304882,"logger":"reconciler.backend","msg":"No need to update backend CR status"}
...
```

[...]

PR #11391, opened 2021-05-28
Align Eventing backend controller logger with the unified Kyma logger

marcobebway:
**Description**

Align Eventing backend controller logger with the unified Kyma logger.

> Note: This PR is tested here https://github.com/kyma-project/kyma/pull/11394.

**Related issue(s)**
https://github.com/kyma-project/kyma/issues/11321


alexandra-simeonova (approved):

components/eventing-controller/cmd/eventing-controller/main.go
```diff
@@ -2,93 +2,94 @@ package main
...
 	"github.com/kyma-project/kyma/components/eventing-controller/reconciler/backend"
 )
 
 func main() {
-	logger := ctrl.Log.WithName("setup")
-	scheme := runtime.NewScheme()
-	restCfg := ctrl.GetConfigOrDie()
+	setupLogger := ctrl.Log.WithName("setup")
+
+	opts := new(options.Options)
```
themue, line 23: While this is working too I would prefer a today more common `opts := options.New()`.

components/eventing-controller/cmd/eventing-controller/main.go
```diff
@@ -2,93 +2,94 @@ package main
...
+		setupLogger.Error(err, "failed to initialize logger")
+		os.Exit(1)
+	}
+	defer func() {
+		if err := logger.WithContext().Sync(); err != nil {
+			setupLogger.Error(err, "failed to flush logger")
+		}
+	}()
+
+	// set controller core logger
```
themue, line 40: Only an optical change, but we should change even comments to full sentences. Like here

```go
// Set controller core logger.
```

It's latest then needed if a controller has more than one sentence.

components/eventing-controller/log/log.go
```diff
@@ -0,0 +1,33 @@
...
+
+import (
+	"github.com/kyma-project/kyma/common/logging/logger"
+)
+
+type Logger struct {
+	*logger.Logger
+}
+
+// NewLogger returns a new logger with the given format and level
```
themue, line 11: As this is the only returned main type a `log.New()` would be enough.

components/eventing-controller/options/options.go
```diff
@@ -0,0 +1,63 @@
...
+	argNameMetricsAddr     = "metrics-addr"
+	argNameReconnectWait   = "reconnect-wait"
+	argNameReconcilePeriod = "reconcile-period"
+
+	// env
+	envNameLogFormat = "APP_LOG_FORMAT"
+	envNameLogLevel  = "APP_LOG_LEVEL"
+)
+
+type Options struct {
```
themue, line 23: I like the idea of an own options package a lot, like the log package. Only would also add an `options.New()`  here. Especially for once possibly needing executed logic when the options instance is created.

Also would add comments for all exported identifiers.

components/eventing-controller/log/log.go
```diff
@@ -0,0 +1,33 @@
+package log
```
themue, line 1: As calling the main type and all instances logger possibly this package should be called `logger` too.

themue (commented):
I do like the idea of those new packages for logger and options. Only have some cosmetically wishes.

pxsalehi (approved):

PR #11407, opened 2021-06-01
Add the Eventing backend type in the reconciler logs

marcobebway:
**Description**

Add the Eventing backend type in the reconciler logs.

### DIFF
--- a/components/eventing-controller/reconciler/backend/reconciler.go
+++ b/components/eventing-controller/reconciler/backend/reconciler.go
@@ -65,6 +65,9 @@ type Reconciler struct {
 	logger *logger.Logger
 	record record.EventRecorder
 	cfg    env.BackendConfig
+
+	// backendType is the type of the backend which the reconciler detects at runtime
+	backendType eventingv1alpha1.BackendType
 }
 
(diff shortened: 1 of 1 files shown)


themue (approved):

components/eventing-controller/reconciler/backend/reconciler.go
```diff
@@ -65,6 +65,9 @@ type Reconciler struct {
 	logger *logger.Logger
 	record record.EventRecorder
 	cfg    env.BackendConfig
+
+	// backendType is the type of the backend which the reconciler detects at runtime
+	backendType eventingv1alpha1.BackendType
```
pxsalehi, line 70: Why are we adding a new field? What happended with creating a sublogger in the beginning of the method instead? We only need one line on top of the two methods reconcileXXXBackend like this:
`log := r.namedLogger().With("backend", eventingv1alpha1.NatsBackendType)`
and in the rest of the method we just use `log`.

marcobebway: This filed is not exposed outside of the reconciler struct, and it was already computed later anyway.

We already have a common logger in the reconciler, introducing yet another sub-logger will force us to pass it to every function that needs to log something, and this in my opinion is not scalable.

pxsalehi (approved):

marcobebway:
retest

PR #11410, opened 2021-06-02
rafter, serverless: create Istio's `vs` instead of `apirule` during install

### LINKED ISSUE #10470: Deleting Kyma namespace not possible caused by blocking ORY finalizer
**Description**

Deletion of `kyma-system` namespace not possible because the ORY finaliser doesn't finish and namespace stays in status `Terminating`.

     $> kubectl get ns
     NAME               STATUS        AGE
     kyma-system        Terminating   105m

This requires a hack in the Kyma CLI to enable a proper uninstallation of Kyma.

**It is also part of this ticket to remove this hack after the finalizer issue is fixed** (see https://github.com/kyma-incubator/hydroform/pull/180/files#diff-7260007237ab7f121eee3278f895d798493a41670bd064270e2fb8936601d7fcR548).

**Expected result**

Deleting Kyma namespace has to be possible without failures.

**Actual result**

The namespace `kyma-system` stays endless in status `Terminating`.

**Steps to reproduce**
1. Provision a Kyma cluster with CLI: `kyma alpha provision k3s`
2. Deploy Kyma via CLI: `kyma alpha deploy`
3. Uninstall Kyma via CLI: `kyma alpha undeploy` or delete the namespace manually: `kubectl delete ns kyma-system`

**Troubleshooting**

Events in namespace `kyma-system` are:

      - lastTransitionTime: "2021-01-27T14:26:13Z"
        message: All content successfully deleted, may be waiting on finalization
        reason: ContentDeleted
        status: "False"
        type: NamespaceDeletionContentFailure
      - lastTransitionTime: "2021-01-27T14:26:13Z"
        message: 'Some resources are remaining: rules.oathkeeper.ory.sh has 2 resource
          instances'
        reason: SomeResourcesRemain
        status: "True"
        type: NamespaceContentRemaining
      - lastTransitionTime: "2021-01-27T14:26:13Z"
        message: 'Some content in the namespace has finalizers remaining: finalizer.oathkeeper.ory.sh
          in 2 resource instances'
        reason: SomeFinalizersRemain
        status: "True"
        type: NamespaceFinalizersRemaining

piotrkpc:
During uninstallation of Kyma race condition occurs and `ory-oathkeeper`
controller cannot remove it's finalizers if it will be delete
before it's resources. `rafter` and `serverless` components use
to relay on `apirule` during installation, but since they don't
use auth features so they could use Istio VirtualService instead
to avoid race conditions.

Resolves #10470


resources/rafter/charts/controller-manager/charts/minio/templates/virtualservice.yaml
```diff
@@ -0,0 +1,36 @@
...
+        allowOrigins:
+          - regex: .*
+      match:
+        - uri:
+            regex: /.*
+      route:
+        - destination:
+            host: {{ template "minio.fullname" . }}
+            port:
+              number: 9000
```
Tomasz-Smelcerz-SAP, line 36:
```suggestion
              number: {{ .Values.service.port }}
```

resources/serverless/templates/virtualservice.yaml
```diff
@@ -0,0 +1,36 @@
...
+        allowOrigins:
+          - regex: .*
+      match:
+        - uri:
+            regex: /.*
+      route:
+        - destination:
+            host: {{ template "registry-fullname" . }}
+            port:
+              number: 5000
```
Tomasz-Smelcerz-SAP, line 35:
```suggestion
              number: {{ .Values.global.dockerServicePort }}
```

resources/rafter/charts/controller-manager/charts/minio/templates/virtualservice.yaml
```diff
@@ -0,0 +1,36 @@
...
+          - Content-Type
+          - '*'
+        allowMethods:
+          - GET
+          - POST
+          - PUT
+          - DELETE
+        allowOrigins:
+          - regex: .*
+      match:
```
Tomasz-Smelcerz-SAP, line 29: Previously, an Oathkeeper rule was configured to only match `GET` requests on any path. Now we allow all methods. A `method` parameter should be added. See: https://istio.io/v1.9/docs/reference/config/networking/virtual-service/#HTTPMatchRequest

resources/serverless/templates/virtualservice.yaml
```diff
@@ -0,0 +1,36 @@
...
+          - '*'
+        allowMethods:
+          - GET
+          - HEAD
+          - POST
+          - PUT
+          - DELETE
+        allowOrigins:
+          - regex: .*
+      match:
```
Tomasz-Smelcerz-SAP, line 28: Previously, an Oathkeeper rule was configured to only match `GET` and `HEAD` requests on any path. Now we allow all methods. A `method` parameter should be added and perhaps another HTTPMatchRequest object. See: https://istio.io/v1.9/docs/reference/config/networking/virtual-service/#HTTPMatchRequest

Tomasz-Smelcerz-SAP (changes requested):

piotrkpc:
hey, code owners, ptal:
@derberg @m00g3n @aerfio @pPrecel @magicmatatjahu 
thanks

Tomasz-Smelcerz-SAP (approved):

PR #11252, opened 2021-05-11
Add omit empty for protocol and protocol settings for eventing-controller

sayanh:
**Description**

Changes proposed in this pull request:

- Add omit empty for protocol and protocol settings

PR #11411, opened 2021-06-02
Update Eventing Subscription CRD

marcobebway:
**Description**

Update Eventing Subscription CRD.

**Related PRs**
- Follow-up to https://github.com/kyma-project/kyma/pull/11252

### DIFF
--- a/installation/resources/crds/eventing/subscriptions.eventing.kyma-project.io.crd.yaml
+++ b/installation/resources/crds/eventing/subscriptions.eventing.kyma-project.io.crd.yaml
@@ -15,6 +15,10 @@ spec:
   scope: Namespaced
   versions:
   - name: v1alpha1
+    additionalPrinterColumns:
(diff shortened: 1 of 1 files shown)


marcobebway:
> Should we add kubebuilder tags to subscription type to have these generated?

Is this convention used in Kyma installer? Can you point me where is it used?

> Also, should we copy this also to cluster-essentials?

It is already there.

pxsalehi:
To make sure when we regenerate the CRDs we get the same output.
https://github.com/kyma-project/kyma/blob/245598f5d473e527aca2161e1aa2602ae117e041/components/eventing-controller/api/v1alpha1/eventingbackend_types.go#L46
https://github.com/kyma-project/kyma/blob/245598f5d473e527aca2161e1aa2602ae117e041/components/eventing-controller/api/v1alpha1/eventingbackend_types.go#L36

werdes72 (approved):

PR #11441, opened 2021-06-09
Static users bound to roles based on override

kwiatekus:
**Description**
With Kyma 2.0 user authentication is based on OIDC configuration. No static users should be provided by authentication proxy ( DEX ) and no static users should be bound to roles

- add `users.bindStaticUsers` override to controll binding of static users
- bind the users to roles by default in kyma 1.x installation mode
- disable static user binding in kyma 2.x installation mode


a-thaler (dismissed):

strekm (approved):
/meow

Issue #11439, opened 2021-06-08
NATS pods went missing


sayanh:
**Description**

Although `eventing-nats` was ready, the pods for nats went missing.
```
apiVersion: nats.io/v1alpha2
kind: NatsCluster
metadata:
  annotations:
    meta.helm.sh/release-name: eventing
    meta.helm.sh/release-namespace: kyma-system
  creationTimestamp: "2021-04-22T14:15:24Z"
  generation: 3
  labels:
    app.kubernetes.io/managed-by: Helm
  name: eventing-nats
  namespace: kyma-system
  resourceVersion: "149690235"
  selfLink: /apis/nats.io/v1alpha2/namespaces/kyma-system/natsclusters/eventing-nats
  uid: 6442783e-bb78-450c-89c9-67e536de5fa2
spec:
  natsConfig:
    debug: true
    trace: true
  pod:
    annotations:
      sidecar.istio.io/inject: "false"
    labels:
      app: nats
      app.kubernetes.io/managed-by: Helm
      helm.sh/chart: nats-1.0.0
      kyma-project.io/dashboard: eventing
    resources:
      limits:
        cpu: 500m
        memory: 1Gi
      requests:
        cpu: 400m
        memory: 512Mi
  size: 3
  version: 2.1.8
status:
  conditions:
  - reason: scaling cluster from 0 to 3 peers
    transitionTime: "2021-04-22T14:15:43Z"
    type: ScalingUp
  - reason: current state matches desired state
    transitionTime: "2021-04-22T14:16:15Z"
    type: Ready
  - reason: scaling cluster from 2 to 3 peers
    transitionTime: "2021-06-08T14:03:10Z"
    type: ScalingUp
  - reason: current state matches desired state
    transitionTime: "2021-06-08T14:03:17Z"
    type: Ready
  currentVersion: 2.1.8
  size: 3
```

Kyma cluster version: 1.22.0

Workaround was to restart NATS operator to create the 3 pods for NATS.

**Expected result**

**Actual result**

**Steps to reproduce**

**Troubleshooting**

PR #11452, opened 2021-06-10
Add pod antiaffinity to NATS cluster

pxsalehi:
**Description**

During K8s cluster upgrade, sometimes the NATS operator does not seem to be able to recover the NATS cluster.
https://github.tools.sap/kyma/backlog/issues/1549
#11439 

- Add "soft" pod-anti-affinity to the eventing nats cluster.
- Upgrade NATS operator version.

### DIFF
--- a/resources/eventing/charts/nats/templates/40-cr.yaml
+++ b/resources/eventing/charts/nats/templates/40-cr.yaml
@@ -22,3 +22,14 @@ spec:
   natsConfig:
     debug: true
     trace: true
+  template:
+    spec:
+      affinity:
+        podAntiAffinity:
+          preferredDuringSchedulingIgnoredDuringExecution:
+            - podAffinityTerm:
+                labelSelector:
+                  matchLabels:
+                    nats_cluster: eventing-nats
+                topologyKey: kubernetes.io/hostname
+              weight: 100


components/nats-operator/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM connecteverything/nats-operator:0.7.6 as builder
+FROM connecteverything/nats-operator:0.8.0 as builder
```
sayanh, line 1: https://github.com/nats-io/nats-operator/issues/323 will block the merging of this PR as of now.

radufa (approved):

radufa (approved):

components/nats-operator/Dockerfile
johgoe: The nats-operator image was renamed and is available already in version 0.8.2 but still with the same issues. see my pending PR https://github.com/kyma-project/kyma/pull/11244/

resources/eventing/charts/nats/templates/40-cr.yaml
```diff
@@ -22,3 +22,14 @@ spec:
   natsConfig:
     debug: true
     trace: true
+  template:
+    spec:
+      affinity:
+        podAntiAffinity:
+          preferredDuringSchedulingIgnoredDuringExecution:
```
radufa, line 29: "preferredDuringSchedulingIgnoredDuringExecution" will distribute the 3 nats instances on the 2 Kyma nodes (in a sdandard SKR installation). But during the draining of a node (caused by the upgrade of K8S version), all the 3 Nats instances are migrated to the only one node which is active.
 
After a new node is attaching  cluster, the active 3 Nats instances are NOT redistributed again between now the two active node but are remaining running on one node.
It seems that we have to find a better appraoch, but the fix is ok for now.

radufa (approved):

PR #11454, opened 2021-06-10
add certificate creation validation during install

piotrkpc:
**Description**

- add validation for certificate during installation and fail fast

**Related issue(s)**

### DIFF
--- a/resources/certificates/files/gardener.sh
+++ b/resources/certificates/files/gardener.sh
@@ -3,6 +3,8 @@ set -eo pipefail
 
 export HOME="/tmp"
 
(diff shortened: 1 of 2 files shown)


resources/certificates/files/gardener.sh
```diff
@@ -3,6 +3,8 @@ set -eo pipefail
 
 export HOME="/tmp"
 
+CERT_CHECK_TIMEOUT=5
```
mjakobczyk, line 6: It could be a configurable env at the `job.yaml` level definition.

piotrkpc: ptal at newest version.

mjakobczyk (approved):
/meow

Issue #11362, opened 2021-05-26
Fix Function CRD generation


tgorgol:
**Description**

There are currently two problems with generating CRDs for Function Controller:
1.  `make manifests` Makefile target which used to generate CRDs no longer works because `controller-gen-local` target it depended on was removed in [this PR](https://github.com/kyma-project/kyma/pull/10517).
2. When deprecating Node.js 10 runtime we decided to leave the enum value in Function CR's `status.runtime` property while removing it from the `spec.runtime` property, to allow existing Node.js 10 Functions to keep working. This change was made manually in generated CRD, so it will be lost after every re-generation of manifests. We should adjust the sources so that the generated CRDs don't require any manual changes.

**Expected result**

1. `make manifests` should work
2. generated CRDs shouldn't require any manual adjustments

PR #11455, opened 2021-06-10
Fix Function Controller manifests generation

tgorgol:
**Description**

- Fix Function Controller manifests generation

**Related issue(s)**
https://github.com/kyma-project/kyma/issues/11362

### DIFF
--- a/components/function-controller/Makefile
+++ b/components/function-controller/Makefile
@@ -14,6 +14,13 @@ IMG ?= $(APP_NAME):latest
 # Produce CRDs that work back to Kubernetes 1.11 (no version conversion)
 CRD_OPTIONS ?= "crd:trivialVersions=true,crdVersions=v1"
 
+# Get the currently used golang install path (in GOPATH/bin, unless GOBIN is set)
+ifeq (,$(shell go env GOBIN))
+GOBIN=$(shell go env GOPATH)/bin
+else
+GOBIN=$(shell go env GOBIN)
+endif
+
 pull-licenses-local:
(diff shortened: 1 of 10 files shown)


components/function-controller/Makefile
```diff
@@ -31,20 +38,36 @@ test-local:
...
+# download controller-gen if necessary
+.PHONY: controller-gen-local
+controller-gen-local:
+ifeq (, $(shell which controller-gen))
+	@{ \
+	set -e ;\
+	CONTROLLER_GEN_TMP_DIR=$$(mktemp -d) ;\
+	cd $$CONTROLLER_GEN_TMP_DIR ;\
+	go mod init tmp ;\
+	go get sigs.k8s.io/controller-tools/cmd/controller-gen@v0.2.9 ;\
```
pPrecel, line 63: Are there any breaking changes between version 0.2.9 and 0.2.4? I found in our `README.md` that we require `controller-gen (v0.2.4+)`. Maybe can we change this to `controller-gen (v0.2.9)`? Or is this not necessary?

pPrecel (approved):
LGTM

components/function-controller/Makefile
tgorgol: I can update the readme with the new version but looking at the [changelog of controller-tools](https://github.com/kubernetes-sigs/controller-tools/releases) I don't see any breaking changes from `0.2.4` to `0.2.9`. Used this version as it was the version our current CRDs were manually generated with in recent updates related to Kyma 2.0.

pPrecel (approved):

PR #11458, opened 2021-06-11
Bump Application Connectivity images

franpog859:
**Description**

- Bump Application Connectivity images


koala7659 (approved):
LGTM, hoever we should follow this PR with another one that will bump the images for tests (like application-operator-tests, runtime-agent-tests) to its newest versions

PR #11467, opened 2021-06-14
Reconnect to Nats in publisher-proxy

radufa:
**Description**

- Reconnect if the Nats connection was closed during pod migrations

**Related issue(s)**

### DIFF
--- a/components/event-publisher-proxy/pkg/nats/connect.go
+++ b/components/event-publisher-proxy/pkg/nats/connect.go
@@ -7,14 +7,40 @@ import (
 	"github.com/nats-io/nats.go"
 )
 
-// ConnectToNats returns a nats connection that is ready for use, or error if connection to the nats server failed
-func ConnectToNats(url string, retry bool, reconnects int, wait time.Duration) (*nats.Conn, error) {
-	connection, err := nats.Connect(url, nats.RetryOnFailedConnect(retry), nats.MaxReconnects(reconnects), nats.ReconnectWait(wait))
+type connectionData struct {
+	url        string
+	retry      bool
+	reconnects int
+	wait       time.Duration
+}
+
+type BackendConnection struct {
+	connectionData connectionData
+	Connection     *nats.Conn
+}
+
+// NewBackendConnection returns a new NewNatsConnection instance with the given NATS connection data.
+func NewBackendConnection(url string, retry bool, reconnects int, wait time.Duration) *BackendConnection {
+	bc := &BackendConnection{}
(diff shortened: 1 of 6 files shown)


components/event-publisher-proxy/pkg/nats/connect.go
```diff
@@ -7,14 +7,40 @@ import (
...
+}
+
+type BackendConnection struct {
+	connectionData connectionData
+	Connection     *nats.Conn
+}
+
+// NewBackendConnection returns a new NewNatsConnection instance with the given NATS connection data.
+func NewBackendConnection(url string, retry bool, reconnects int, wait time.Duration) *BackendConnection {
+	bc := &BackendConnection{}
```
themue, line 24:
```suggestion
	return &BackendConnection{
	    connectionData: connectionData{
	        url: url,
	        retry: retry,
	        reconnects: reconnects:,
	        wait: wait,
	    },
	}
```

components/event-publisher-proxy/pkg/sender/nats.go
```diff
@@ -7,7 +7,7 @@ import (
 	cenats "github.com/cloudevents/sdk-go/protocol/nats/v2"
 	cev2 "github.com/cloudevents/sdk-go/v2"
 	cev2event "github.com/cloudevents/sdk-go/v2/event"
-	"github.com/nats-io/nats.go"
+	pkgnats "github.com/kyma-project/kyma/components/event-publisher-proxy/pkg/nats"
```
themue, line 10: Possibly a different name than `pkgnats` as they are all somehow `pkg`. Maybe `epnats` like `cenats` above.

themue (commented):
Only two minor comments.

components/event-publisher-proxy/pkg/sender/nats.go
radufa: "pkgnats" was defined and used in the past, see:
https://github.com/kyma-project/kyma/blob/main/components/event-publisher-proxy/cmd/event-publisher-proxy/nats/nats.go#L13
https://github.com/kyma-project/kyma/blob/main/components/event-publisher-proxy/pkg/handler/nats/handler_test.go#L30
https://github.com/kyma-project/kyma/blob/main/components/event-publisher-proxy/pkg/sender/nats_test.go#L15

themue: Okay, so out of scope. 👌🏻

themue (approved):

PR #11480, opened 2021-06-15
bump controller-runtime version

pxsalehi:
**Description**

- Update the controller-runtime dependency

**Related issue(s)**


themue (approved):

components/eventing-controller/config/crd/bases/eventing.kyma-project.io_eventingbackends.yaml
```diff
@@ -4,7 +4,7 @@ apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   annotations:
-    controller-gen.kubebuilder.io/version: v0.4.0
+    controller-gen.kubebuilder.io/version: v0.2.5
```
sayanh, line 7: Let's upgrade to https://github.com/kubernetes-sigs/controller-tools/releases/tag/v0.6.0 if you are willing to.

Then you need to change the [Makefile](https://github.com/kyma-project/kyma/blob/main/components/eventing-controller/Makefile#L136) as well.

Finally, execute the `make copy-crds` target to sync CRDs in the resources folder.

components/eventing-controller/reconciler/backend/reconciler_test.go
```diff
@@ -416,8 +416,8 @@ var _ = Describe("Backend Reconciliation Tests", func() {
 					EventingReady:               utils.BoolPtr(false),
 					SubscriptionControllerReady: utils.BoolPtr(false),
 					PublisherProxyReady:         utils.BoolPtr(false),
-					BebSecretName:               "",
-					BebSecretNamespace:          "",
+					BebSecretName:               bebSecret1name,
```
sayanh, line 419: There should not be any bebsecret in the status as reconciler failed to stop the NATS controller. I think empty should be the correct value.

pxsalehi: We are using BEB as backend and the name/namespace is what is expected. Not ready status is due to not stopping the nats subscription controller. I think this also makes sense.

sayanh: Right, I overlooked, bebsecret name/namespace do get updated.

sayanh (dismissed):

sayanh (approved):

PR #11514, opened 2021-06-21
prepare kyma-gateway for move to certificates chart

a-thaler:
**Description**

- in order to move the kyma-gateway resource to the certificates chart, it needs to get the "keep" policy for one release so that it gets detached from the core chart on deletion of it
- ...
- ...

**Related issue(s)**


mjakobczyk (approved):
One remark - if this one is to be kept for only one release then make sure to somehow keep track on deleting it later!

Issue #10939, opened 2021-03-22
Enable installation without helm


pbochynski:
**Description**
It should be possible to install Kyma components without using helm for lifecycle management. 

Acceptance criteria:
- [ ] each component can be installed with the sequence: `helm template ... | kubectl apply -f -`
- [ ] all post/pre install/upgrade jobs should be removed (if it is hard to remove consider replacing it by a  custom logic in the controller)

**Reasons**
Current life cycle management is based on the imperative helm commands (install, upgrade). When an operation fails it is hard (often impossible) to resume it and bring the cluster to the target state. Helm features like atomic upgrades or rollbacks don't make any sense with multiple related components in Kyma.
The idea is to use helm only for templating and render target configuration based on default values, profiles and custom overrides. Such target configuration can be applied as a sequence of `kubectl apply` operations (create/patch) on kubernetes objects (native or our custom resources)

pbochynski:
Some charts generate invalid Yamls with duplicated keys (usually labels):
- fix for monitoring: https://github.com/kyma-project/kyma/pull/11515
- fix for eventing:

pbochynski:
You can install kyma without helm with the following commands:
```
cd tests/fast-integration
npm install
./kyma.js provision k3d
./kyma.js install --with-central-application-gateway --use-helm-template
```

kwiatekus:
We clearly would like to avoid migrating the logic of rafter helm jobs because rafter is scheduled to be removed.
I would like to clarify if rafter can be removed in `2.0-m1` milestone.

Rafter is used currently:
 - as a source for api specs rendering in console UI ( kyma 1.x ) - this is not the case in `2.0-m1`
 - in the process of API registration in application connector - would it be still the case in `2.0-m1` ? @janmedrek 
 
 CC : @pbochynski @PK85

janmedrek:
@kwiatekus there are 2 integration points regarding Rafter and Application Connector:
- Application Registry
- Compass Runtime Agent

We would need some time to drop the parts responsible for the integration, I would say `2.0-m2` is a safer option if we are to do it by ourselves (we are happy to see other teams contributing though!).

PK85:
[...]

PR #11528, opened 2021-06-23
Fix problem with duplicated keys (name label)

pbochynski:
Enables installation without helm (using helm template): https://github.com/kyma-project/kyma/issues/10939

### DIFF
--- a/resources/eventing/charts/publisher-proxy/templates/_helpers.tpl
+++ b/resources/eventing/charts/publisher-proxy/templates/_helpers.tpl
@@ -59,14 +59,6 @@ app.kubernetes.io/name: {{ include "publisher-proxy.fullname" . }}
 {{- end }}
 
 {{/*
-Labels for metrics service
-*/}}
-{{- define "publisher-proxy.metrics.labels" -}}
-{{ include "publisher-proxy.labels" . }}
-{{ include "publisher-proxy.peerauth.selectorLabels" . }}
-{{- end }}
-
-{{/*
 Selector labels for service monitor
 */}}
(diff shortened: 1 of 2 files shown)


resources/eventing/charts/publisher-proxy/templates/_helpers.tpl
```diff
@@ -58,14 +58,6 @@ Selector labels for peerauthentication
 app.kubernetes.io/name: {{ include "publisher-proxy.fullname" . }}
```
sayanh, line 58: Can you please remove `publisher-proxy.peerauth.selectorLabels` as it is redundant now?

pbochynski: I can't. It is still used here: https://github.com/kyma-project/kyma/blob/e386863bb4cd486cd323df6c6d01af10c4a87537/resources/eventing/charts/publisher-proxy/templates/peerauthentication.yaml#L9

pbochynski: I can't. It is still used here: https://github.com/kyma-project/kyma/blob/e386863bb4cd486cd323df6c6d01af10c4a87537/resources/eventing/charts/publisher-proxy/templates/peerauthentication.yaml#L9

pbochynski: I didn't want to do a bigger clean-up. Just remove duplicated keys to make the output YAML template parseable.

sayanh (approved):

PR #11533, opened 2021-06-23
Perform cleanup for Impersonate-Group header in request

mjakobczyk:
**Description**

Request coming to `apiserver-proxy` component can contain `impersonate-group` header. We manage to prepare it on our own, depending on the user's groups. In order to not pass any additional data, we should make sure that a cleanup is performed.

- before adding user's groups perform a cleanup on the header

**Related issue(s)**

### DIFF
--- a/components/apiserver-proxy/internal/proxy/proxy.go
+++ b/components/apiserver-proxy/internal/proxy/proxy.go
@@ -81,6 +81,7 @@ func (h *kubeRBACProxy) Handle(w http.ResponseWriter, req *http.Request) bool {
 	}
 
 	req.Header.Set("Impersonate-User", r.User.GetName())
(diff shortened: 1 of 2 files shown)


components/apiserver-proxy/internal/proxy/proxy.go
```diff
@@ -90,6 +91,10 @@ func (h *kubeRBACProxy) Handle(w http.ResponseWriter, req *http.Request) bool {
 	return true
 }
 
+func (h *kubeRBACProxy) cleanupImpersonateGroupHeader(req *http.Request) {
+	req.Header.Set("Impersonate-Group", "")
```
werdes72, line 95: How about `req.Header.Del`?

werdes72 (approved):

PR #11549, opened 2021-06-24
istio tracing config is set to 100

a-thaler:
**Description**

The istio tracing config is ambigious sampling rate settings leading to on overall effect of sampling all requests. That is way too much for a productive system, it was set to 1 in the past.

Furthermore, the config uses deprecated elements which prevents an effective override of the trace config

- removed outdated tracing config from all profiles
- configured the sampling rate with one setting only
- configured a sampling rate of 100 for the evaluation profile, otherwise 1

**Related issue(s)**

### DIFF
--- a/resources/istio/files/istio-operator-cluster-evaluation.yaml
+++ b/resources/istio/files/istio-operator-cluster-evaluation.yaml
@@ -129,8 +129,6 @@ spec:
       enabled: true
       k8s:
         env:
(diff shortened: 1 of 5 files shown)


cnvergence (approved):

cnvergence:
lgtm

mjakobczyk (approved):

Issue #10857, opened 2021-03-10
Refresh Kyma documentation to match current state and the roadmap


pbochynski:
**Description**

The target structure of the documentation:
- Overview 
   - value proposition
   - architecture
- Main concepts (areas) with purpose and high-level architecture
  - serverless
  - eventing
  - service consumption
  - application connectivity
  - API exposure
- Tasks:
  - deploy function
  - deploy microservice
  - send event
  - trigger workload with event
  - ...
  - to be defined

### ToDo

- [x] Review the existing content on the website [#10876](https://github.com/kyma-project/kyma/issues/10876) @pbochynski 
- [x] Come up with and implement the new doc structure (content) @klaudiagrz @NHingerl 
- [x] Implement the new docs structure (website) [#681](https://github.com/kyma-project/website/issues/681) @kwiatekus 
- [x] (Re)move or adjust the existing content:
    - ~Remove Rafter docs from the website~ [#10927](https://github.com/kyma-project/kyma/issues/10927) @kazydek 
    - ~Remove Helm Broker docs from the website~ [#10928](https://github.com/kyma-project/kyma/issues/10928) @klaudiagrz ✅ 
    - ~Remove Control Plane from the website~ [control-plane#676](https://github.com/kyma-project/control-plane/issues/676) @klaudiagrz 
    - Remove `minikube`
    - ~Adjust Serverless docs~ @kazydek ✅ [#11163](https://github.com/kyma-project/kyma/issues/11163) 
    - ~Adjust Eventing docs~ @alexandra-simeonova ✅ [#11200](https://github.com/kyma-project/kyma/issues/11200)
    - ~Adjust Observability docs~ @NHingerl [#11228](https://github.com/kyma-project/kyma/issues/11228)
    - ~Adjust old Service Management docs~ @klaudiagrz 
    - ~Adjust old Application Connector docs~ @majakurcius ✅ [#11622](https://github.com/kyma-project/kyma/issues/11622)
    - ~Adjust Security docs~ @mmitoraj ✅ 
    - ~Adjust Service Mesh docs~ @mmitoraj ✅ 
    - ~Adjust API exposure docs~ @mmitoraj ✅ 
    - ~Adjust Kyma general docs~
- [x] Update the content to meet Kyma 2.0 vision (Kyma 2.0 phase 2)  - the content is updated for Kyma 2.0, the rest of Service Management and Application Connector content will be updated after Service Catalog migration (2.X)
[...]

PR #11573, opened 2021-06-28
Improve overview headlines

NHingerl:
**Description**

- adjust the paragraph headlines so they aren't identical to the subtab titles
- rearrange the paragraphs so they match the order of the subtabs (alphabetical
- remove the `Type` metadata field in both files

**Related issue(s)**

https://github.com/kyma-project/kyma/issues/10857

### DIFF
--- a/docs/01-overview/02-main-areas/README.md
+++ b/docs/01-overview/02-main-areas/README.md
@@ -30,6 +30,8 @@ Kyma comes with tools that give you the most accurate and up-to-date monitoring,
 - For logging, Kyma uses [Loki](https://github.com/grafana/loki), a Prometheus-like log management system.
 - With the [Jaeger](https://github.com/jaegertracing) distributed tracing system, you can analyze the path of the requests sent to and from your service. This information helps you to optimize the latency and performance of your solution.
 
+For more details, see [Observability](./observability/51-telemetry-observability-overview.md).
+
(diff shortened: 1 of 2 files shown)


docs/01-overview/02-main-areas/README.md
```diff
@@ -1,28 +1,38 @@
...
+### What is Application Connectivity?
 
 - Simplifies and secures the connection between external systems and Kyma
 - Registers external events and APIs and simplifies the API usage
 - Provides asynchronous communication with services and Functions deployed in Kyma through events
 - Manages secure access to external systems
 - Provides monitoring and tracing capabilities to facilitate operational aspects
 
-### Service consumption
+### What is API exposure?
```
mmitoraj, line 17:
```suggestion
### What is API Exposure?
```
I'm not sure about the suggestion I'm making; however, all the other headlines are capitalized.

mmitoraj (approved):
I like the experimental linking to the Observability subtab and I think we could use it for all the sections :)

docs/01-overview/02-main-areas/README.md
NHingerl: I agree, we should be consistent one way or the other. 
The same is true for the subtab titles, we should align whether we want capitalisation or not.

PR #11634, opened 2021-07-06
Optimze HTTP header access in test

themue:
**Description**

Card for related Checkmarx link describes possible resource exhaustion while looking for the `Ce-Type` header. 

This header is now addressed directly.

### DIFF
--- a/components/event-publisher-proxy/pkg/handler/beb/handler_test.go
+++ b/components/event-publisher-proxy/pkg/handler/beb/handler_test.go
@@ -490,12 +490,12 @@ func validateApplicationName(appName string) testingutils.Validator {
 			return fmt.Errorf("expected the event-type [%s] to contain the application name [%s] for structured mode", event.Type(), appName)
 		}
 		// Default CE binary mode
-		value, ok := r.Header[testingutils.CeTypeHeader]
-		if !ok {
-			return errors.New("event-type header is not found")
+		value := r.Header.Get(testingutils.CeTypeHeader)
+		if len(value) == 0 {
+			return errors.New("event-type header is not found or empty")
 		}
-		if !strings.Contains(value[0], appName) {
-			return fmt.Errorf("expected the event-type [%s] to contain the application name [%s] for binary mode", value[0], appName)
+		if !strings.Contains(value, appName) {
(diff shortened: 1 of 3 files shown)


components/event-publisher-proxy/pkg/handler/beb/handler_test.go
```diff
@@ -487,18 +487,16 @@ func validateApplicationName(appName string) testingutils.Validator {
...
-		for k, v := range r.Header {
-			if strings.ToLower(k) != strings.ToLower(testingutils.CeTypeHeader) {
-				continue
-			}
-			if strings.Contains(v[0], appName) {
-				return nil
-			}
-			return errors.New(fmt.Sprintf("expected the event-type [%s] to contain the application name [%s] for binary mode", v[0], appName))
+		// Default CE binary mode
+		value, ok := r.Header[testingutils.CeTypeHeader]
```
sayanh, line 493: IMO r.Header.Get() fits well here as it takes of the case as well.

components/event-publisher-proxy/testing/utils.go
```diff
@@ -21,10 +21,10 @@ import (
 
 const (
 	// binary cloudevent headers
-	CeIDHeader          = "CE-ID"
-	CeTypeHeader        = "CE-Type"
-	CeSourceHeader      = "CE-Source"
-	CeSpecVersionHeader = "CE-SpecVersion"
+	CeIDHeader          = "Ce-Id"
```
sayanh, line 24: Why is this needed as headers are case-insensitive. ?

components/event-publisher-proxy/pkg/handler/beb/handler_test.go
themue: Thought about the same but not sure about empty headers vs non-existing headers. `Get()` would return an empty string. But might be okay here for the test, so I'll change.

components/event-publisher-proxy/testing/utils.go
themue: It's the Canonical MIME Header. But in case using `Get()` it's okay to set them to the lower-case ones as defined in the  standard. Here `Get()` changes them on the fly.

themue: Done.

sayanh (approved):

PR #11657, opened 2021-07-08
Restrict fluent-bit pod permissions

elchead:
**Description**
Fluentbit is distributed as a container which must run in the workload cluster as root user in order to collect logs from other containers on the same node. By getting access to that container you can get access to the node file system.

- as suggested [here](https://github.com/fluent/fluent-bit/issues/872): root access is needed but the permissions can be restricted to file access

**How it was tested:**
- the logs of the fluent-bit pod were checked
- in Grafana, the traffic in the logging board was compared to another cluster

**Related issue(s)**

https://github.tools.sap/kyma/backlog/issues/1560

### DIFF
--- a/resources/logging/charts/fluent-bit/values.yaml
+++ b/resources/logging/charts/fluent-bit/values.yaml
@@ -54,7 +54,8 @@ securityContext:
 service:
   type: ClusterIP
   port: 2020
-  labels: {}
+  labels: 
+    {}
   annotations:
     {}
     # prometheus.io/path: "/api/v1/metrics/prometheus"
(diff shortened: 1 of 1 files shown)


resources/logging/charts/fluent-bit/values.yaml
```diff
@@ -42,18 +42,19 @@ dnsConfig: {}
...
   # readOnlyRootFilesystem: true
   # runAsNonRoot: true
   # runAsUser: 1000
 
 service:
   type: ClusterIP
   port: 2020
-  labels:
-    {}
+  labels: {}
```
JeremyHarisch, line 57:
```suggestion
  labels: 
    {}
```

JeremyHarisch (approved):

Issue #11645, opened 2021-07-07
Eventing-controller does not reconcile events in Kyma 1.23.1


cnvergence:
**Description**
After a fresh install of Kyma 1.23.1 on Azure Gardener cluster with Kubernetes 1.18.18, we could observe long latencies in reconciling events by eventing-nats-controller.

It is istio-proxy that somehow blocks the communication between the eventing-controller and the Kubernetes API server.
No error logs were found in the istio-proxy sidecar nor any other component.

**Expected result**

As an example labeling eventing subscriptions normally works immediately, when executing `k label subscriptions.eventing.kyma-project.io -n tunas --all k=v2 --overwrite` we can see events in the logs of the eventing-nats-controller.

**Actual result**

With Kyma 1.23.1, after some time, executing `k label subscriptions.eventing.kyma-project.io -n tunas --all k=v2 --overwrite` results in no events processed by the controller and nothing is reported in the logs.

**Steps to reproduce**

Deploy k8s cluster with Kyma 1.23.1 and apply tests on eventing-nats-controller.

**Troubleshooting**

We could confirm that without an istio-proxy sidecar it works.
As another workaround, we could exclude the outbound traffic in istio-proxy by adding annotation:
`traffic.sidecar.istio.io/excludeOutboundIPRanges: "0.0.0.0/0"`

cnvergence:
After investigation we have found that this issue is not happening when we are not using the istio-proxy sidecar.
I could also replicate the issue on the clean Azure cluster on Gardener with Kyma 1.23.1
Then I focused on the reason why exactly the communication between API Server and eventing-controller is sometimes blocked without error messages. But in the debug logs of the sidecar I couldn’t see any errors or debug logs related to communication with API Server.
I could identify a few old issues on Istio that looked similar. [#1](https://github.com/istio/istio/issues/8696) and [#2](https://github.com/istio/istio/issues/12187 )that looked similar.
Yesterday, we could confirm that excluding internal IP address of the API Server with annotation on eventing-controller pod is a workaround for this
`traffic.sidecar.istio.io/excludeOutboundIPRanges: "100.64.0.1/32"`
Still we don’t know what exactly makes eventing-controller special as on Kyma 1.23.1 there is also service-catalog-controller-manager with Istio sidecar working fine. This could mean that the issue is not related to global settings of Istio.
Also Kyma 1.24.0 doesn’t seem to have this issue at all.

cnvergence:
[...]

PR #11662, opened 2021-07-09
Disable processing of egress traffic by Istio in eventing-nats-controller(1.23)

cnvergence:
**Description**
With Kyma 1.23 installed only on Azure clusters, we could observe that eventing-nats-controller was not able to communicate with API Server after ~5 minutes of being idle. So the behaviour is similar to this older [issue](https://github.com/kubernetes/client-go/issues/527).

Versions:
- k8s.io/client-go v0.19.3 in eventing-nats-controller
- Istio 1.9.5
- Kubernetes 1.18.18

- Bypass Istio for outbound traffic on eventing-nats-controller.

**Related issue(s)**

See also #11645

### DIFF
--- a/resources/eventing/charts/nats-controller/templates/deployment.yaml
+++ b/resources/eventing/charts/nats-controller/templates/deployment.yaml
@@ -12,6 +12,9 @@ spec:
     type: RollingUpdate
   template:
     metadata:
+      annotations:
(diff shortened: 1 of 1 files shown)


cnvergence:
Moving to 1.24

PR #9845, opened 2020-11-03
Separate application-connector gateway secret from kyma-gateway

Demonsthere:
**Description**

Changes proposed in this pull request:

- Change `kyma-gateway-certs-cacert` to `app-connector-certs`
- Create cert-sync deployment, which watches for changes in `kyma-gateway-certs` and synchronizes them with the new secret

**Related issue(s)**

Issue #9795, opened 2020-10-29
Unable to connect Applications mocks to Kyma after upgrade to istio-1.7


rakesh-garimella:
**Description**

While testing application mocks on kyma with istio-1.7 we see that the application mocks are not able to connect to kyma and they give following error:

```
Error: write EPROTO 139715443133288:error:14094418:SSL routines:ssl3_read_bytes:tlsv1 alert unknown ca:../deps/openssl/openssl/ssl/record/rec_layer_s3.c:1536:SSL alert number 48
```

Although its see sometimes we are able to connect but then it fails again during `register`.

**Expected result**

The application is able to connect to kyma without any issue.

**Actual result**

Connection of application to kyma fails.

**Steps to reproduce**

1. Install kyma with istio-1.7
2. Try to connect application mock.

**Troubleshooting**

Demonsthere:
Possible culprit: on a managed certificate environment (like gardener) the `kyma-gateway-certs` secret which holds TLS certs is managed by cert-manager. It creates a CA entry in the file, which is used by istio-ingressgateway instead of our CA, which is being held in `kyma-gateway-certs-cacert`

pbochynski:
https://istio.io/latest/docs/tasks/traffic-management/ingress/secure-ingress/#configure-a-mutual-tls-ingress-gateway
> Istio supports reading a few different Secret formats, to support integration with various tools such as cert-manager:
> - A TLS Secret with keys tls.key and tls.crt, as described above. For mutual TLS, a ca.crt key can be used.
> - A generic Secret with keys key and cert. For mutual TLS, a cacert key can be used.
> - A generic Secret with keys key and cert. For mutual TLS, a separate generic Secret named <secret>-cacert, with a cacert key. For example, httpbin-credential has key and cert, and httpbin-credential-cacert has cacert.

Demonsthere:
Yeah, we know that. The problem is, that what you have sent us is the order of priority Istio uses. If a ca.crt field is found in the httpbin-credential, Istio is not looking up httpbin-credential-cacert has cacert, which acts as a fallback

pbochynski:
Hmm. I have exactly such setup and it works. I have secret with 3 keys: tls.key, tls.crt and ca.crt (from lets encrypt). Another secret with suffix -cacert contains only cacert key. I am testing with istio 1.7.4. Both secrets are of type Opaque. First one generated by Gardener certificate manager.

Demonsthere:
[...]

PR #11678, opened 2021-07-13
Fixes in jobs managing Kyma certificates

koala7659:
Proposed changes:
- Remove cert sychronisation deployment
- Rename istio application connector gateway credential secret from app-connector-certs to kyma-gateway-certs-cacert
- Remove Connector Service App CA Certs Backup and Restore Jobs
- Remove job connector-service-migrate-ac-certs
- Add certificate migration functionality to Certs Setup Job and Compass Runtime Agent


docs/application-connector-compass-mode/08-11-rotate-root-ca.md
```diff
@@ -8,7 +8,7 @@ The Central Connector Service uses the Root certificate issued by the Certificat
 Two different components use the Root CA certificate. As a result, the certificate is stored in two separate Secrets:
 
   - The `connector-service-app-ca` Connector Service CA Secret responsible for signing certificate requests
-  - The `app-connector-certs` Istio Secret responsible for security in the Connector Service API
+  - The `kyma-gateway-certs-cacert` Istio Secret responsible for security in the Connector Service API
```
majakurcius, line 11: Note to self: change the names on the Kyma 2.0 branch.

docs/application-connector/08-11-rotate-root-ca.md
```diff
@@ -8,7 +8,7 @@ The Central Connector Service uses the Root certificate issued by the Certificat
 Two different components use the Root CA certificate. As a result, the certificate is stored in two separate Secrets:
 
   - The `connector-service-app-ca` Connector Service CA Secret responsible for signing certificate requests
-  - The `app-connector-certs` Istio Secret responsible for security in the Connector Service API
+  - The `kyma-gateway-certs-cacert` Istio Secret responsible for security in the Connector Service API
```
majakurcius, line 11: Note to self: change the names on the Kyma 2.0 branch.

docs/security/03-04-ingress-egress-traffic.md
```diff
@@ -78,7 +78,7 @@ The order differs depending on the mode:
   | Secret | ingress-tls-cert | `kyma-system` |
   | ConfigMap | net-global-overrides | `kyma-installer `|
   | Secret | kyma-gateway-certs | `istio-system` |
-  | Secret | app-connector-certs | `istio-system `|
+  | Secret | kyma-gateway-certs-cacert | `istio-system `|
```
majakurcius, line 81: Note to self: change the names on the Kyma 2.0 branch.

docs/service-mesh/10-02-no-access.md
```diff
@@ -40,11 +40,11 @@ If this solution doesn't work, you need to change the image of the Istio Ingress
     kubectl logs -n istio-system -l app=istio-ingressgateway -c ingress-sds
     ```
 
-5. In case of certificate-related issues, make sure `kyma-gateway-certs` and `app-connector-certs` Secrets are available in the `istio-system` Namespace and that they contain proper data. Run:
+5. In case of certificate-related issues, make sure `kyma-gateway-certs` and `kyma-gateway-certs-cacert` Secrets are available in the `istio-system` Namespace and that they contain proper data. Run:
```
majakurcius, line 43: Note to self: change the names on the Kyma 2.0 branch.

franpog859 (commented):
I installed it using `./kyma.js install --skip-components kiali,tracing --with-central-application-gateway --use-helm-template` command and tested it running `mocha test/2-commerce-mock.js -b`. All tests passed

everesio:
lgtm

franpog859 (commented):
I also tested it on Gardener + Azure. I installed the previous version, generated the certificates, and checked if they work using a new Application and the `v1/metadata/services` Application Registry endpoint. I upgraded Kyma to this PR's version, checked if proper components' images are there, and checked the certificate again. After that, I rotated the TLS Gardener certificates using [the CertificatesRevocation CR](https://github.com/gardener/cert-management#revoking-certificates-with-renewal) with the `renewal: true` property. I checked if the certificates are rotated and checked my certs again. Everything worked fine in this scenario 🎉🎉🎉 

I'm not submitting the `LGTM` review as I know that there is more to be tested

resources/compass-runtime-agent/values.yaml
```diff
@@ -39,8 +39,11 @@ compassRuntimeAgent:
         namespace: compass-system
     caCertificate:
       secret:
-        name: app-connector-certs
+        name: kyma-gateway-certs-cacert
```
Tomasz-Smelcerz-SAP, line 42: Why such name change? It implies that this is the main kyma-gateway certificate, while it's rather app-connector related cert. Isn't it?

akgalwas: Some time ago `Application Connectivity Certs Sync` deployment was introduced. The responsibility of the deployment was copying TLS certificates to `app-connector-certs` secret. The implications were as follows:
- the logic for creating `app-connector-certs` secret was implemented in `Application Connectivity Certs Sync` deployment and `Certs Setup Job`
- the logic for setting `app-connector-certs` secret was in two places (some keys were set by `Application Connectivity Certs Sync` and some were set by `Certs Setup Job`)

The rename reduced an overall complexity a lot:
- `Application Connectivity Certs Sync` deployment has been removed 
- `Certs Setup Job` is the only place where `Application Connectivity Certs` are managed
- `kyma-gateway-application-connector` Gateway uses both `kyma-gateway-certs` and `kyma-gateway-certs-cacert`

koala7659:
I have successfully tested those fixes in for KymaOS in following installation/upgrade scenarios:

* `global.applicationConnectorCa` and `global.applicationConnectorCaKey` overrides provided during installation.
* Overrides not provided during installation.
* `global.applicationConnectorCa` and `global.applicationConnectorCaKey` overrides provided during upgrade. Upgrading to the PR should use provided certificates.
* Overrides not provided during upgrade. Upgrading to the PR should preserve existing certificates.

Tested environments:

GKE (with own domain)
AKS (with own domain)

resources/compass-runtime-agent/values.yaml
Tomasz-Smelcerz-SAP: But it looks like it's incorrect.
Some thoughts, please correct me if I am wrong.
- Kyma Gateway is using `credentialName: kyma-gateway-certs` attribute
- It points to a secret with key and TLS certificate used to handle HTTPS traffic  to the cluster. It's a main "entry" for all external traffic (not related to application-connector)
- Kyma Gateway TLS mode is `SIMPLE` - it does not support Mutual TLS at this moment
- Configuring another secret with name `kyma-gateway-certs-cacert` is, according to Istio documentation, a way to provide Certificate necessary to validate client's certificates used in Mutual TLS.
- So one hand we're configuring the Kyma-Gateway to NOT use Mutual TLS (because it's mode is SIMPLE), on the other we're explicitly providing some Certificate that should be used for Mutual TLS...
- Istio may reject such configuration or treat that as mis-configuration and yield warnings, if it's not doing that at the moment
- Istio experts and other people trying to debug problems on the cluster will be certainly confused by such configuration.

If you need Mutual TLS for application-connector gateway, it should have it's own secret with a name distinct from the secret used by Kyma-Gateway.
Unless I deeply misunderstand how Mutual TLS in Istio works (it may be the case...) I think that what is proposed here is a "hack" that works only by accident - Istio should complain about misconfiguration of Kyma-Gateway.

mjakobczyk: I have to support @Tomasz-Smelcerz-SAP in this case.

We are now under construction of multiple domains. Each new domain, which is managed by client, requires a separate Gateway. As a result, the Gateway requires reference to the correct certificate for the domain. When clients are going to manage their domains with wildcards certificates there would always be a pair - Gateway and a Secret that comes from a Certificate. If something goes wrong with a network traffic it should be traceable which certificate fails. The approach that you introduce would suggest that generic Kyma-gateway has invalid cert while it is not true - it is your application connector. Whose responsibility would it be then to debug such problems? I have mixed feelings about this approach and wondered if there is some other way that can be taken instead.

akgalwas: Some facts:
- This PR reverts changes done in [this PR](https://github.com/kyma-project/kyma/pull/9845) as a fix for [this issue](https://github.com/kyma-project/kyma/issues/9795).
- According to the [discussion](https://github.com/kyma-project/kyma/issues/9795#issuecomment-719539250) in #9795 the secret was renamed from `kyma-gateway-certs-cacert` to `app-connector-certs` since there was an issue in Istio 1.7.1. What is important, according to the comment the Istio issue has been fixed in 1.7.4.
- The solution proposed in this solution has been tested thoroughly. Moreover it was working before Istio was upgraded to 1.7.1

Let me to answer your concerns.

> - So one hand we're configuring the Kyma-Gateway to NOT use Mutual TLS (because it's mode is SIMPLE), on the other we're explicitly providing some configuration - a Certificate - that should be used in Mutual TLS setup...
> - Istio may reject such configuration or treat that as mis-configuration and yield warnings, if it's not doing that at the moment
> - Istio experts and other people trying to debug problems on the cluster will be certainly confused by such configuration.

Let's consider `kyma-gateway`. Istio documentation describes `credentialName` as follows:

> For gateways running on Kubernetes, the name of the secret that holds the TLS certs including the CA certificates. Applicable only on Kubernetes. The secret (of type generic) should contain the following keys and values: key: <privateKey> and cert: <serverCert>. For mutual TLS, cacert: <CACertificate> can be provided in the same secret or a separate secret named <secret>-cacert. Secret of type tls for server certificates along with ca.crt key for CA certificates is also supported. Only one of server certificates and CA certificate or credentialName can be specified.


We have two gateways (`kyma-gateway` and `kyma-gateway-application-connector`) that share TLS certificates. I think that is ok.
As it comes to `kyma-gateway`, the type is `SIMPLE` so `kyma-gateway-certs-cacert` is simply ignored. I cannot find any information in the Istio documentation that proves this setup is incorrect.

akgalwas: > If something goes wrong with a network traffic it should be traceable which certificate fails. The approach that you introduce would suggest that generic Kyma-gateway has invalid cert while it is not true - it is your application connector. Whose responsibility would it be then to debug such problems? I have mixed feelings about this approach and wondered if there is some other way that can be taken instead.

I don't get your point. 
We have the following case:
- One gateway in SIMPLE mode, one gateway in MUTUAL mode
- Both share server certificate

So if server (TLS) certificate is incorrect then all requests will fail (no matter if accessed resource is protected with TLS or mTLS ). It will be quite easy to observe as TLS handshake will fail. In case CA Cert is incorrect only requests that access mTLS protected resources will fail.

koala7659: I'm supporting @akgalwas here. Please notice following arguments:

- The only reason we have now this separate secret and deployment to synchronise certificates is long gone fixed Istio bug. If this bug hadn't happened - nobody wouldn't have separated TLS credentials for both gateways into two secrets. 

- We never had separate TLS certs for different gateways. Kyma was designed to provide same (TLS) certificate to secure them both. In the current solution it was implemented by having two secrets with SAME (TLS) certificate data manually synchronised by separate deployment. This is overhead. We can do it in much simpler way without these intermediate elements.
  
-  In the current and proposed solution both gateways end with loading SAME (TLS) certificates data to secure connection. An additional layer of security is provided by mTLS mode in `app-connector-gateway` where a separate mTLS certificate is loaded from additional secret `kyma-gateway-certs-cacert` which is exactly how it is supposed to work.

- From Istio point of view such configuration is fine. In fact it is not recommended for WWW endpoints. Using same (TLS) certificate to secure multiple gateways [can cause 404 error](https://istio.io/latest/docs/ops/common-problems/network-issues/#404-errors-occur-when-multiple-gateways-configured-with-same-tls-certificate) 
This is a reason the when running `istioctl analyze` tool we can see the warning. 
`(Gateway kyma-gateway-application-connector.kyma-system) Duplicate certificate in multiple gateways [kyma-system/kyma-gateway-application-connector kyma-system/kyma-gateway] may cause 404s if clients re-use HTTP2 connections.
` 
But we are sure that for our case (we do not use HTTP2 and we have no browser interaction on `connector-gateway`)  both gateways will work fine.
 
- If you find the naming confusing, we can consider renaming secret `kyma-gateway-certs` into something more generic like `kyma-tls-certs`. In fact this SAME (TLS) certificate is used in both gateways.

franpog859 (approved):

mjakobczyk (approved):

majakurcius (approved):

janmedrek (approved):

docs/application-connector-compass-mode/08-11-rotate-root-ca.md
majakurcius: We no longer have this flow.

docs/application-connector/08-11-rotate-root-ca.md
majakurcius: We no longer have this flow.

PR #11680, opened 2021-07-14
added a psp specific for fluent-bit

a-thaler:
**Description**
To circumvent any read/write access to not needed folders on the node I added a PSP specific for fluent-bit covering the general kyma requirements but also adding restrictions on the hostpaths

- enabled specific PSP
- restricted host paths for fluent-bit
- restricted volume types
- restricted host ports

**Related issue(s)**

### DIFF
--- a/resources/logging/charts/fluent-bit/templates/psp.yaml
+++ b/resources/logging/charts/fluent-bit/templates/psp.yaml
@@ -12,7 +12,16 @@ spec:
   requiredDropCapabilities:
     - ALL
   volumes:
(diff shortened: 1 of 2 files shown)


chrkl:
The changes look reasonable to me. However, PSPs are deprecated since K8s 1.21 (https://kubernetes.io/blog/2021/04/06/podsecuritypolicy-deprecation-past-present-and-future/). So we would have to look for another solution at some time in future.

JeremyHarisch (approved):

PR #11703, opened 2021-07-19
Use the correct memory metrics to get memory usage in the container

rakesh-garimella:
**Description**

- Use `container_memory_working_set_bytes` instead of `container_memory_usage_bytes` as the former gives correct memory usage.
- Adjust the label `container_name` -> `container`

**Related issue(s)**


resources/monitoring/templates/grafana/kyma-dashboards/kyma-backends.yaml
```diff
@@ -110,9 +113,9 @@ data:
           "steppedLine": false,
           "targets": [
             {
-              "expr": "sum by(container) (container_memory_usage_bytes{namespace=\"$namespace\", pod=\"$pod\", container_name!=\"POD\",container!=\"\"})",
+              "expr": "sum by(name, container) (container_memory_working_set_bytes{namespace=\"$namespace\", pod=\"$pod\", container_name!=\"POD\",container!=\"\"})",
               "interval": "",
-              "legendFormat": "Current: {{`{{container_name}}`}}",
+              "legendFormat": "Current: {{`{{name}}`}}",
```
a-thaler, line 118: Shouldn't we display the container_name indeed. It is much prettier and usefull then this long name. Yes, there will be duplicates but they have different coloring.

resources/monitoring/templates/prometheus/rules-1.14/k8s.rules.yaml
```diff
@@ -57,8 +60,8 @@ spec:
           max by(namespace, pod, node) (kube_pod_info{node!=""})
         )
       record: node_namespace_pod_container:container_memory_swap
-    - expr: sum(container_memory_usage_bytes{job="kubelet", metrics_path="/metrics/cadvisor", image!="", container!="POD"}) by (namespace)
-      record: namespace:container_memory_usage_bytes:sum
+    - expr: sum(container_memory_working_set_bytes{job="kubelet", metrics_path="/metrics/cadvisor", image!="", container!="POD"}) by (namespace)
```
a-thaler, line 63: Are you sure we should change it in the community rules?

resources/monitoring/templates/grafana/dashboards-1.14/statefulset.yaml
```diff
@@ -193,7 +194,7 @@ data:
                         "tableColumn": "",
                         "targets": [
                             {
-                                "expr": "sum(container_memory_usage_bytes{job=\"kubelet\", metrics_path=\"/metrics/cadvisor\", cluster=\"$cluster\", namespace=\"$namespace\", pod=~\"$statefulset.*\"}) / 1024^3",
```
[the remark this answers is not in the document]
rakesh-garimella: The community chart is not updated upstream. I think making the change would keep the info consistent with rest of the charts.

resources/monitoring/templates/grafana/kyma-dashboards/kyma-backends.yaml
rakesh-garimella: will fix it

resources/monitoring/templates/prometheus/rules-1.14/k8s.rules.yaml
rakesh-garimella: I would like to make the change for consistency.

a-thaler (approved):

PR #11661, opened 2021-07-09
Update node exporter chart

elchead:
**Description**

IMPORTANT: image still needs to be updated in our registry

Changes proposed in this pull request:

- Update node exporter helm chart to image v1.1.2.(taken from [here](https://github.com/prometheus-community/helm-charts/tree/main/charts))
- ...
- ...

**What was tested**
- kyma1 install on cluster (Gardener: as5). Helm upgrade with new chart
- check pod logs
- check Grafana dashboard for node metrics (Kubernetes / Nodes)

**Related issue(s)**

PR #11707, opened 2021-07-19
fix of securityContext in node-exporter

a-thaler:
**Description**
With https://github.com/kyma-project/kyma/pull/11661/files a bug got introduced causing the chart to not render properly

- Renamed the value key to the old value podSecurityContext
- removed unneeded TODO comments
- ...

**Related issue(s)**


rakesh-garimella:
It would be nice to test it against kyma->kyma2 upgrade. Before the changes it was working fine with Kyma.

rakesh-garimella (approved):

PR #11712, opened 2021-07-20
Add readiness probes to OAuth2 proxies

skhalash:
**Description**

- Adds readiness probes. The upstream chart also includes one: https://github.com/oauth2-proxy/manifests/blob/main/helm/oauth2-proxy/templates/deployment.yaml

**Related issue(s)**


chrkl:
Let's add the same for the OAuth2 Proxy deployment in the Tracing and Kiali charts.

chrkl (approved):

Issue #11327, opened 2021-05-20
Enable tracing instrumentation in functions


kwiatekus:
**Description**
Enable function developers to trace request chains ( including functions) in tracing backend (jeager).
Employ open-telemetry standards and tooling in the functions so that manual instrumentation is possible (and even maybe auto-instrumentation). 

> As a function developer I want to have tracing SDK at hand so that I can use the trace context ( traceId, spanId ) in the manual instrumentation of the outgoing requests ( i.e calling another service via axios or sending messages back to the eventing ).

**AC**
 - [x] embed access to the tracer in the programming interface of the function ( via OpenTelemetry sdks for python and nodejs), 
 - [x] check the options for auto-instrumentation ( i.e in [NodeJS](https://github.com/open-telemetry/opentelemetry-js/tree/c0b021be5782ebee6a23015c992dc496b105dad2/getting-started) ) 
 - [x] If manual instrumentation is required for the outgoing requests - please add examples to the documentation
 - [x] SDK should comply with  protocols supported by jeager (zipkin or jeager protocol)
 - [x] Currently B3 ( [openzipkin](https://github.com/openzipkin/b3-propagation) ) trace headers are used by istio. We should support them.
 - [x] STRETCH - If possible we should also include w3s ( open standard) headers in the auto-instrumentation
  
**Reasons**
We should follow recommendations coming from open standards (see parent [issue](https://github.com/kyma-project/kyma/issues/11231)). By supporting tracing headers we give users tools to analyse request chains in the system.

**Hints**
We should try to accomplish that  as much as possible in the serverless runtimes (or serving layer, i.e express) so that  there is as little coding as possible required from the function developer who just expects some automaticity.
Check the NodeJS auto-instrumentation options (see [example](https://github.com/open-telemetry/opentelemetry-js/tree/c0b021be5782ebee6a23015c992dc496b105dad2/getting-started) where no manipulation of outgoing axios requests is necessary )

kwiatekus:
Follow up :
Enable users to  
 - creating new trace spans 
 - create custom tags 
 - make sure trace context is passed to the eventing as part of outgoing messages

kwiatekus:
We did some investigation for tracing instrumentation in nodejs - https://github.com/kyma-project/kyma/pull/11739
We have to fix problems:
 - function traces were not attached to the full execution trace
 - follow up on trace standards unififcation
We have to instrument python runtimes as well

kwiatekus:
Closing.
[...]

PR #11739, opened 2021-07-22
Enable tracing instrumentation in functions

m00g3n:
**Description**

- ...
- ...
- ...

**Related issue(s)**

See also: https://github.com/kyma-project/kyma/issues/11327


kwiatekus:
Current status in the opened PR:
 - code that registers instrumentation for Http and Express layer applied in nodejs runtime ( kubeless )
 - trace headers ( in b3 standard ) propagator applied
 
 Main trace span that originates from istio requests is chained properly on the trace stack in jeager
 Trace spans comming from express layer are not properly stacked on that same trace stack, but are detached ( and traceable by `unknown service` name in jeager )

Issue #10006, opened 2020-11-23
Create BEB webhooks with unique names


marcobebway:
**Description**

The EventingController creates BEB webhooks based on the Kyma subscription names only, which is not collision proof.

**Expected result**

Each Kyma subscription should have a BEB webhook with a unique name created for it.

**Actual result**

Multiple Kyma subscriptions with the same name in different namespaces share the same BEB webhook.

**Steps to reproduce**

1. Create a namespace `ns1`.
2. Create a Kyma subscription `sub1` inside the namespace `ns1`.
3. Create a namespace `ns2`.
4. Create a Kyma subscription `sub1` inside the namespace `ns2`.

sayanh:
Think about creating readable yet unique BEB webhook names.

k15r:
* UUids ??

pxsalehi:
The issue seems to be not webhooks, but the BEB subscriptions created on BEB. Since we only use the subscription name, the BEB subscription name will be the same for two Kyma subscriptions with the same name but in different namespaces. Which means the two Kyma subscriptions share the same BEB subscription and queue. Even if BEB allows this, the two Kyma subs will probably share the messages in the queue and each receives only a subset of the messages!

I tried to reproduce this. I could create two Kyma subs with the same name in different namespaces. They were both reported as ready. However, after a while they were both not ready and the reason was: `BEB Subscription not active`.

As for the solution, I'm not sure if we need UUIDs. As long as we also use the namespace name, it should be enough. Kubernetes wouldn't allow creating another Kyma sub with the same name in the same namespace anyway. So I'd suggest just adding 'namespaceName-` to the BEB subscription name.

pxsalehi:
After discussion with @marcobebway and @sayanh, we should still use a schema like `namespace-subName-uuid` since different Kyma clusters could be sharing the same BEB namespace and therefore create Kyma subs with the same K8s namespace and name.

A more important issue is what to do with existing subscriptions during an upgrade (where we migrate subscriptions based on the old naming schema to the new schema).

pxsalehi:
To solve the issue we need to address each of the following issues:
  - [ ] Change how subscription names are generated: We need to be able to reproduce the same sub name when reconciling BEB subs. Maybe the random UUID should go in another field in the sub CR (e.g. `id`) and be assigned the first time the Kyma sub is created. This would require introducing a (mutating) admission webhook for the subscriptions.
[...]

PR #11751, opened 2021-07-23
Ensure uniqueness of the BEB subscription name corresponding to a Kyma subscription

radufa:
**Description**

- The  BEB subscription name used for a Kyma subscription should be unique. 
- It's built from the concatenation of "shootName", Kyma subscription namespace and Kyma subscription name

**Related issue(s)**

See also: https://github.com/kyma-project/kyma/issues/10006


pxsalehi:
Will be continued in #11805

PR #11759, opened 2021-07-26
Increase memory limit for eval profile

sayanh:
**Description**

- Turns out for evaluation profile, eventing-controller needs more memory limit. Clusters show, `controller` container in eventing-controller pod uses around ~115 Mi
Screenshot 2021-07-26 at 14 40 05
Screenshot 2021-07-26 at 14 41 16


themue (approved):
Has this memory doubling been tested as limit?

Issue #9939, opened 2020-11-12
Eventing-Controller does not recreate BEB subscription when it is deleted manually via BEB


nachtmaar:
**Description**

Scenario:
- Create the following subscvription:

```yaml
apiVersion: eventing.kyma-project.io/v1alpha1
kind: Subscription
metadata:
  name: test-nachtmaar-1
  namespace: default
spec:
  protocol: BEB
  protocolsettings:
    exemptHandshake: true
    qos: 'AT-LEAST-ONCE'
    webhookAuth: <- to be filled out manually
  sink: "https://myapp.default.svc.cluster.local"
  filter:
    dialect: beb,
    filters:
    - eventSource:
        property: source
        type: exact
        value: /default/sap.kyma/kt1
      eventType:
        property: type
        type: exact
        value: tunas.ev2.poc.event1.v0
```

- Wait until it is ready
- Delete the Queue and Webhook on EMS Side

**Expected result**

The BEB Subscription is recreated

**Actual result**

The BEB Subscription is not found and the reconciler shows 404 not found errors.

```
[...]

PR #11807, opened 2021-07-30
Recreate BEB subscription if it was deleted from BTP side

radufa:
**Description**

- Recreate the BEB subscription if it was deleted from BTP cockpit

**Related issue(s)**

Fix https://github.com/kyma-project/kyma/issues/9939


components/eventing-controller/pkg/handlers/beb.go
```diff
@@ -116,8 +118,17 @@ func (b *Beb) SyncSubscription(subscription *eventingv1alpha1.Subscription, clea
 		// check if EMS subscription is the same as in the past
 		bebSubscription, err = b.getSubscription(sEv2.Name)
 		if err != nil {
-			log.Errorw("get BEB subscription failed", "error", err)
-			return false, err
+			log.Errorw("get BEB subscription failed", "bebSubscriptionName", sEv2.Name, "error", err)
```
k15r, line 121: bebSubscriptionName should also be a constant
same for error
```suggestion
			log.Errorw("get BEB subscription failed", BEB_SUBSCRIPTION_NAME, sEv2.Name, ERROR, err)
```

components/eventing-controller/pkg/handlers/beb.go
```diff
@@ -191,6 +202,29 @@ func (b *Beb) deleteCreateAndHashSubscription(subscription *types.Subscription,
...
+	}
+
+	// create a new EMS subscription
+	if err := b.createSubscription(subscription, log); err != nil {
+		log.Errorw("create BEB subscription failed", "bebSubscriptionName", subscription.Name, "error", err)
+		return nil, err
+	}
+
+	// get the new EMS subscription
+	bebSubscription, err := b.getSubscription(subscription.Name)
```
k15r, line 219: what does it mean if we fail here? we created a subscription successfully, but we cannot retrieve it. does it mean we just have to wait a bit, or is something really wrong on the backend side?

components/eventing-controller/pkg/handlers/beb.go
```diff
@@ -238,7 +272,7 @@ func (b *Beb) getSubscription(name string) (*types.Subscription, error) {
 		return nil, fmt.Errorf("get subscription failed: %v", err)
 	}
 	if resp.StatusCode != http.StatusOK {
-		return nil, fmt.Errorf("get subscription failed: %v; %v", resp.StatusCode, resp.Message)
+		return nil, fmt.Errorf("get subscription failed: %w; %v", errors.New(strconv.Itoa(resp.StatusCode)), resp.Message)
```
sayanh, line 275: Instead of doing the error.New twice, how about we keep this as is and create a 404 error constant to compare in [SyncSubscription:123](https://github.com/kyma-project/kyma/pull/11807/files#diff-17d39986bafe1b1562ce9db3c9fb4828533f6d1badfa43fc5476892b27dc88d1R123)

components/eventing-controller/pkg/handlers/beb.go
```diff
@@ -154,7 +165,7 @@ func (b *Beb) DeleteSubscription(subscription *eventingv1alpha1.Subscription) er
 func (b *Beb) deleteCreateAndHashSubscription(subscription *types.Subscription, cleaner eventtype.Cleaner, log *zap.SugaredLogger) (*types.Subscription, int64, error) {
 	// delete EMS subscription
 	if err := b.deleteSubscription(subscription.Name); err != nil {
-		log.Errorw("delete BEB subscription failed", "error", err)
+		log.Errorw("delete BEB subscription failed", "bebSubscriptionName", subscription.Name, "error", err)
```
sayanh, line 168: instead of adding bebSubscription name everywhere lets change the log object to include this key, `bebSubscriptionName `

components/eventing-controller/pkg/handlers/beb.go
```diff
@@ -249,7 +283,7 @@ func (b *Beb) deleteSubscription(name string) error {
 		return fmt.Errorf("delete subscription failed: %v", err)
 	}
 	if resp.StatusCode != http.StatusNoContent && resp.StatusCode != http.StatusNotFound {
-		return fmt.Errorf("delete subscription failed: %v; %v", resp.StatusCode, resp.Message)
+		return fmt.Errorf("delete subscription failed: %w; %v", errors.New(strconv.Itoa(resp.StatusCode)), resp.Message)
```
sayanh, line 286: Creating 2 new error types is unnecessary IMO.

components/eventing-controller/pkg/handlers/beb.go
radufa: Something is wrong on BEB side. We will try to redo the action at the next reconcile call.

components/eventing-controller/pkg/handlers/beb.go
radufa: The idea is to move from the old error handling based on string compare with the new one, exposed by the new "errors" package coming with Go 1.13 (or 1.14), I'm not sure. This will allow to have a deeper errors stack if calls are chained.

components/eventing-controller/pkg/handlers/beb.go
radufa: here is how the log is looking now:

{"level":"ERROR","timestamp":"2021-07-30T07:46:04Z","logger":"beb-handler","caller":"handlers/beb.go:121","message":"get BEB subscription failed","context":{"kind":"Subscription","version":1,"namespace":"test","name":"test","bebSubscriptionName":"test","error":"get subscription failed: 404; 404 Not Found"}}
{"level":"INFO","timestamp":"2021-07-30T07:46:04Z","logger":"beb-handler","caller":"handlers/beb.go:124","message":"Recreate the BEB subscription","context":{"kind":"Subscription","version":1,"namespace":"test","name":"test","bebSubscriptionName":"test"}}

We have, dependent of the calling point two subscriptions name:
- one for Kyma subscription: this is
```
"name":"test",
```
and one for beb subscription:
```
"bebSubscriptionName":"test"
```
I've put this, cause with the next PR, the two will be different and it's important to have in the log file the beb subscription name too.

radufa: Not in all places used to log messages, the beb subscription exists, but in all places the kyma subscription exists.
This is why the log object contains only the kyma subscription and the beb subscription name is added as an additional parameter were it is available.

components/eventing-controller/pkg/handlers/beb.go
radufa: True, but we have another issues where we should define these constants for the complete eventing-controller and not only for this particular file. Changing it later on a global level will be a better solution.

components/eventing-controller/pkg/handlers/beb.go
radufa: Yes, but this is conform with the new "errors" package added by golang.

components/eventing-controller/pkg/handlers/beb.go
radufa: Changed to use constants for log keys. Later we will refactor this to define global log keys for all eventing-controller packages, to allow running queries on logs

marcobebway:
> Please add tests for this change

Done.

components/eventing-controller/reconciler/subscription/reconciler_test.go
```diff
@@ -762,6 +762,98 @@ var _ = Describe("Subscription Reconciliation Tests", func() {
...
+				By("Checking Kyma Subscription ready condition to be true", func() {
+					getSubscription(kymaSubscription, ctx).Should(And(
+						reconcilertesting.HaveSubscriptionName(kymaSubscriptionName),
+						reconcilertesting.HaveSubscriptionReady(),
+					))
+				})
+			})
+
+			notfoundHandlerCalled := make(chan bool, 1)
+			Context("Simulate manual deletion of BEB Subscription", func() {
```
pxsalehi, line 818: I think we should not do this (playing with the Mock behaviour). The test should only rely on asserting simple REST calls that are made to the mock (ideally we have some stubs auto-generated from the OpenAPI spec of BEB). If the test needs more behaviour than simple call expectations, then we shouldn't use a mock, since it will be just relyinhg on some logic that we wrote and might not represent the actual system being tested.

pxsalehi (changes requested):

components/eventing-controller/reconciler/subscription/reconciler_test.go
marcobebway: Since this is not the main focus of this ticket, could you please create a ticket for this?

marcobebway: Since this is not the main focus of this ticket, could you please create a ticket for it?

sayanh (approved):

sayanh (approved):

sayanh (approved):

PR #11817, opened 2021-07-30
Parse smctl response for fast-integration tests

wozniakjan:
smctl returns 0 on failures so we need to check stdout too

**Description**

- smctl returns 0 even on failures, we need to parse the stdout to make sure it succeeded

**Related issue(s)**


voigt (approved):
/lgtm

PR #11827, opened 2021-08-02
Enable central gateway & validator by default

everesio:
- central application gateway and central validator should be enabled by default

### DIFF
--- a/resources/application-connector/values.yaml
+++ b/resources/application-connector/values.yaml
@@ -5,10 +5,10 @@ connection_token_handler:
   enabled: *connectorServiceEnabled
 
 central_application_connectivity_validator:
-  enabled: &centralApplicationConnectivityValidatorEnabled false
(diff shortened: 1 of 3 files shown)


resources/application-connector/values.yaml
```diff
@@ -73,9 +73,9 @@ global:
   event_service_integration_tests:
     version: "d6bbc47a"
   central_application_connectivity_validator:
-    version: "PR-11558"
+    version: "PR-11827"
   central_application_gateway:
-    version: "PR-11631"
+    version: "PR-11827"
```
franpog859, line 78: There are no changes in these components (only their charts are changed), therefore there are no new images

everesio: I removed this change

everesio:
Closing PR. Other PR will is created https://github.com/kyma-project/kyma/pull/11830

PR #11830, opened 2021-08-02
Enable central gateway & validator by default

everesio:

Enable central connectivity validatory and gateway based on  rafalpotempa/central-connectivity-validator branch.


franpog859 (commented):
I've seen `{"level":"error","log":{"message":"[TestResources: ApplicationOperatorUpgradeTest] End with error [start time: Tue, 03 Aug 2021 07:36:57 UTC, duration: 25.508059ms]: image versions are not upgraded: Application Connectivity Validator is not upgraded: deployments.apps \"operator-test-app-connectivity-validator\" not found","taskID":"44e55a31-4455-41a0-8852-bd95cf0a94ef","time":"2021-08-03T07:36:57.726Z"}}` in the `pre-main-kyma-gke-upgrade` job's logs. I think it won't work this way :( In previous version there is an old Validator and after the upgrade it's not (it's replaced with the Central one)

There must be a part in the Upgrade test that checks if all the deployments are upgraded properly. Not sure though. I think you'll need to dig into it deeper and check on how it truly works

franpog859 (approved):
LGTM

k15r (approved):

suleymanakbas91 (approved):

crabtree (approved):

pbochynski (approved):

Issue #11493, opened 2021-06-17
api-gateway tests  using dex installed from components


strekm:
**Description**

At this moment api-gateway tests are depending on dex installed as regular component from components.yaml file. as soon as Kyma 2.0 will be released we will start clean-up tasks including dex since it is not a part of kyma 2.0. Look for alternative way of dex installation or replacement for dex.

**Reasons**

Dex is not a part of kyma 2.0 and soon it will be removed. it shouldn't be a part of api-gateway tests

PR #11861, opened 2021-08-06
Replacing Dex with Ory Hydra login and consent app in api-gateway tests

cnvergence:
**Description**

- Since Kyma 2.0 won't include Dex, we need to change our JWT tests in api-gateway.

**Related issue(s)**

See also #11493

### DIFF
--- a/resources/ory/values.yaml
+++ b/resources/ory/values.yaml
@@ -151,7 +151,7 @@ oathkeeper:
           enabled: true
           config:
             jwks_urls:
-            - http://dex-service.kyma-system.svc.cluster.local:5556/keys
+            - http://ory-hydra-public.kyma-system.svc.cluster.local:4444/.well-known/jwks.json
             scope_strategy: wildcard
       authorizers:
         allow:
--- a/resources/ory/values.yaml
+++ b/resources/ory/values.yaml
@@ -151,6 +151,7 @@ oathkeeper:
           enabled: true
           config:
             jwks_urls:
+            - http://dex-service.kyma-system.svc.cluster.local:5556/keys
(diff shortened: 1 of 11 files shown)


tests/integration/api-gateway/gateway-tests/manifests/k8s/deployment.yaml
```diff
@@ -0,0 +1,28 @@
...
+      version: v1
+  template:
+    metadata:
+      labels:
+        app: ory-hydra-login-consent
+        version: v1
+    spec:
+      containers:
+        - name: login-consent
+          image: piotrkpc/hydra-login-consent:test_0.1
```
Tomasz-Smelcerz-SAP, line 19: We can't use such image. We have to provide something more "official". Perhaps a little project in kyma-incubator.

resources/ory/values.yaml
```diff
@@ -152,6 +152,7 @@ oathkeeper:
           config:
             jwks_urls:
             - http://dex-service.kyma-system.svc.cluster.local:5556/keys
```
Tomasz-Smelcerz-SAP, line 154: Is dex still required?

tests/integration/api-gateway/gateway-tests/manifests/k8s/deployment.yaml
cnvergence: Currently working on it. Here is the image for testing that pipelines work.

resources/ory/values.yaml
cnvergence: I thought that I would leave it in Ory for now - for the sake of old Kyma versions, where Dex is deployed

mjakobczyk (approved):

PR #11912, opened 2021-08-19
Implement gitv2

dbadura:
**Description**

- ...
- ...
- ...

**Related issue(s)**


components/function-controller/internal/git/go2git.go
```diff
@@ -0,0 +1,152 @@
...
+		return "", err
+	}
+
+	//tag
+	ref, err = repo.References.Dwim(options.Reference)
+	if err == nil {
+		return ref.Target().String(), nil
+	}
+	if !git2go.IsErrorCode(err, git2go.ErrNotFound) {
+		return "", errors.Wrap(err, "while lookup branch")
```
pPrecel, line 72: While lookup branch? :P Not tag? And maybe we should wrap other errors in the whole method

components/function-controller/internal/git/go2git.go
```diff
@@ -0,0 +1,152 @@
...
+	}
+	return "", errors.Errorf("Could find commit,branch or tag with given ref: %s", options.Reference)
```
pPrecel, line 74: `Could NOT find...` and please add space separator between the `,` and the `branch`

components/function-controller/internal/git/cloner.go
```diff
@@ -0,0 +1,27 @@
...
+
+import (
+	git2go "github.com/libgit2/git2go/v31"
+	"github.com/pkg/errors"
+)
+
+type git2goCloner struct {
+}
+
+func (g *git2goCloner) cloneRepo(options Options, outputPath string) (*git2go.Repository, error) {
```
pPrecel, line 11: What do you think about extracting the whole logic (`GetAuth` method and all `if`'s)? In this case, we'll not test this file anymore and this structure will be still functional

components/function-controller/internal/git/go2git.go
```diff
@@ -0,0 +1,152 @@
...
+	return func(url, username string, allowed_types git2go.CredentialType) (*git2go.Credential, error) {
+		return cred, nil
+	}
+}
+
+func sshCheckCallback() func(cert *git2go.Certificate, valid bool, hostname string) git2go.ErrorCode {
+	return func(cert *git2go.Certificate, valid bool, hostname string) git2go.ErrorCode {
+		return git2go.ErrOk
+	}
+}
```
pPrecel, line 146: Are these methods correlated with auth callbacks? Maybe we should move them to the `auth.go` file?

resources/serverless/templates/deployment.yaml
```diff
@@ -118,7 +118,7 @@ spec:
             {{ include "createEnv" ( dict "name" "APP_FUNCTION_BUILD_EXECUTOR_IMAGE" "value" .Values.containers.manager.envs.functionBuildExecutorImage "context" . ) | nindent 12 }}
             {{ include "createEnv" ( dict "name" "APP_FUNCTION_TARGET_CPU_UTILIZATION_PERCENTAGE" "value" .Values.containers.manager.envs.targetCPUUtilizationPercentage "context" . ) | nindent 12 }}
             # scan-image: {{ .Values.containers.manager.envs.functionBuildRepoFetcherImage.value }} # This line allows security scan tools to scan this image
-            {{ include "createEnv" ( dict "name" "APP_FUNCTION_BUILD_REPOFETCHER_IMAGE" "value" .Values.containers.manager.envs.functionBuildRepoFetcherImage "context" . ) | nindent 12 }}
+            {{ include "createEnv" ( dict "name" "APP_FUNCTION_BUILD_REPO_FETCHER_IMAGE" "value" .Values.containers.manager.envs.functionBuildRepoFetcherImage "context" . ) | nindent 12 }}
```
pPrecel, line 121: Good change :D I like it

components/function-controller/internal/git/go2git.go
```diff
@@ -0,0 +1,152 @@
...
+	return "", errors.Errorf("Could find commit,branch or tag with given ref: %s", options.Reference)
+}
+
+func (g *Git2GoClient) Clone(path string, options Options) (string, error) {
+	repo, err := g.cloneRepo(options, path)
+	if err != nil {
+		return "", errors.Wrap(err, "while cloning the repository")
+	}
+
+	oid, err := git2go.NewOid(options.Reference)
```
pPrecel, line 83: This `Reference` will be always a commit, right? Maybe we can change this name to less confusing? Only suggestion - It's not that important

components/function-controller/README.md
```diff
@@ -8,10 +8,13 @@ The Function Controller requires the following components to be installed:
...
 - [Istio](https://github.com/istio/istio/releases) (v1.4.3)
 - [Docker registry](https://github.com/docker/distribution) (v2.7.1)
-- [controller-gen](https://github.com/kubernetes-sigs/controller-tools/tree/master/cmd/controller-gen) (v0.2.9)
 
 ## Development
 
+To work with Function controller you need to have:
+- [libgit2-dev](https://libgit2.org/) (v1.1)
+- [controller-gen](https://github.com/kubernetes-sigs/controller-tools/tree/master/cmd/controller-gen) (v0.2.9)
+
```
pPrecel, line 17: 👍

components/function-controller/internal/controllers/serverless/build_resources.go
```diff
@@ -168,7 +168,7 @@ func buildRepoFetcherEnvVars(instance *serverlessv1alpha1.Function, gitOptions g
 		},
 		{
 			Name:  "APP_REPOSITORY_COMMIT",
-			Value: instance.Status.Repository.Reference,
+			Value: instance.Status.Commit,
```
pPrecel, line 171: 🥇

pPrecel (commented):
Really, really good work!

components/function-controller/internal/git/go2git.go
```diff
@@ -0,0 +1,152 @@
...
+	ref, err := g.lookupBranch(repo, options.Reference)
+	if err == nil {
+		return ref.Target().String(), nil
+	}
+	if !git2go.IsErrorCode(err, git2go.ErrNotFound) {
+		return "", err
+	}
+
+	//tag
+	ref, err = repo.References.Dwim(options.Reference)
```
pPrecel, line 67: OiD is only an object id. We should extract commit from the Head. Please look at the following example:
```
        log.Infof("Commit from dwimed tag: %s", YOUR_OID.String())

	err = repo.SetHeadDetached(YOUR_OID)
	if err != nil {
		log.Error(err)
	}
	head, err := repo.Head()
	if err != nil {
		log.Error(err)
	}
	commit, err := repo.LookupCommit(head.Target())
	if err != nil {
		log.Error(err)
	}

	log.Infof("Commit from head: %s", commit.Id().String())
```

components/function-controller/internal/git/go2git.go
dbadura: In this case yes, it's commit. It's also used in method `Lookup`, and reference can be `tag`, `branch` and `commit`

components/function-controller/README.md
```diff
@@ -8,10 +8,13 @@ The Function Controller requires the following components to be installed:
 
 - [Istio](https://github.com/istio/istio/releases) (v1.4.3)
 - [Docker registry](https://github.com/docker/distribution) (v2.7.1)
-- [controller-gen](https://github.com/kubernetes-sigs/controller-tools/tree/master/cmd/controller-gen) (v0.2.9)
 
 ## Development
 
+To work with Function controller you need to have:
```
klaudiagrz, line 14:
```suggestion
To develop the Function Controller, you need:
```

PR #11929, opened 2021-08-20
Add note about Kyma version support

### LINKED ISSUE #11928: Add note about Kyma version support
**Description**

The issue was raised by developers that we need to let customers/users know that Kyma patch fixes are not possible for older Kyma versions. Namely, any version older than "{LATEST} minus 2". See also: https://github.com/kyma-project/test-infra/blob/main/development/tools/jobs/releases/releases.go#L15 
A note in the Kyma README file could state that older versions are not supported. 

**Area**

* Documentation

**Reasons**

To make it clear before someone raises an issue with an older version that it's not possible to fix it anymore. 

**Assignees**

@kyma-project/technical-writers

**Attachments**

alexandra-simeonova:
**Description**

Fixes #11928 by adding a note that only the three latest Kyma versions are supported. 

- Add note about Kyma version support

**Related issue(s)**
#11928

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,7 +25,7 @@ Go to the [Kyma project website](https://kyma-project.io/) to learn more about o
 
 Install Kyma locally or on a cluster. See the [Installation guides](https://kyma-project.io/docs/root/kyma#installation-installation) for details.
 
-> **NOTE:** Make sure to install the latest Kyma version and keep it up to date by [upgrading Kyma](https://kyma-project.io/docs/#installation-upgrade-kyma). Please be aware we do not provide support or patch releases for Kyma versions older than the three latest ones. (For example, if the latest version is 1.24, only 1.24, 1.23, and 1.22 will be supported.)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -25,6 +25,8 @@ Go to the [Kyma project website](https://kyma-project.io/) to learn more about o
 
 Install Kyma locally or on a cluster. See the [Installation guides](https://kyma-project.io/docs/root/kyma#installation-installation) for details.
 
+> **NOTE:** Make sure to install the latest Kyma version and keep it up to date by [upgrading Kyma](https://kyma-project.io/docs/#installation-upgrade-kyma). Please be aware we do not provide support or patch releases for Kyma versions older than the three latest ones. (For example, if the latest version is 1.24, only 1.24, 1.23, and 1.22 will be supported.)
```
majakurcius, line 28:
```suggestion
> **NOTE:** Make sure to install the latest Kyma version and keep it up to date by [upgrading Kyma](https://kyma-project.io/docs/#installation-upgrade-kyma). Be aware that we do not provide support or patch releases for Kyma versions older than the three latest ones (for example, if the latest version is 1.24, only 1.24, 1.23, and 1.22 will be supported.)
```

majakurcius (approved):
👍

PR #11992, opened 2021-08-31
Fix webhook auth default config

sayanh:
**Description**

- Fix webhook auth default config where default config was overriden by Subscription webhook auth config

### DIFF
--- a/components/eventing-controller/pkg/handlers/beb.go
+++ b/components/eventing-controller/pkg/handlers/beb.go
@@ -98,7 +98,7 @@ func (b *Beb) SyncSubscription(subscription *eventingv1alpha1.Subscription, clea
 
 	// get the internal view for the ev2 subscription
 	var statusChanged = false
-	sEv2, err := getInternalView4Ev2(subscription, apiRule, b.WebhookAuth, b.ProtocolSettings, b.Namespace, b.SubNameMapper)
+	sEv2, err := getInternalView4Ev2(subscription, apiRule, *b.WebhookAuth, b.ProtocolSettings, b.Namespace, b.SubNameMapper)
 	if err != nil {
 		log.Errorw("get Kyma subscription internal view failed", ERROR_LOG_KEY, err)
 		return false, err
--- a/components/eventing-controller/pkg/handlers/utils.go
+++ b/components/eventing-controller/pkg/handlers/utils.go
@@ -129,7 +129,7 @@ func getQos(qosStr string) (types.Qos, error) {
 }
 
 func getInternalView4Ev2(subscription *eventingv1alpha1.Subscription, apiRule *apigatewayv1alpha1.APIRule,
-	defaultWebhookAuth *types.WebhookAuth, defaultProtocolSettings *eventingv1alpha1.ProtocolSettings,
+	defaultWebhookAuth types.WebhookAuth, defaultProtocolSettings *eventingv1alpha1.ProtocolSettings,
 	defaultNamespace string, nameMapper NameMapper) (*types.Subscription, error) {
 	emsSubscription, err := getDefaultSubscription(defaultProtocolSettings)
 	if err != nil {
@@ -195,7 +195,7 @@ func getInternalView4Ev2(subscription *eventingv1alpha1.Subscription, apiRule *a
 		}
 		auth.TokenURL = subscription.Spec.ProtocolSettings.WebhookAuth.TokenUrl
 	}
-	emsSubscription.WebhookAuth = auth
+	emsSubscription.WebhookAuth = &auth
 	return emsSubscription, nil
 }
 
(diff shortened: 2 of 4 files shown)


components/eventing-controller/pkg/handlers/utils_test.go
```diff
@@ -42,16 +42,20 @@ func TestGetInternalView4Ev2(t *testing.T) {
...
 		ClientSecret: "clientSecret",
 		TokenURL:     "tokenURL",
 	}
-
 	defaultNameMapper := NewBebSubscriptionNameMapper("my-shoot", 50)
 
+	bebSubEvents := eventingtesting.GetBEBEvents(types.Event{
+		Source: reconcilertesting.EventSource,
+		Type:   reconcilertesting.OrderCreatedEventType,
+	})
```
marcobebway, line 57: Why we need extra func `GetBEBEvents` when we can do:

```golang
bebSubEvents := types.Events{types.Event{
    Source: reconcilertesting.EventSource,
    Type:   reconcilertesting.OrderCreatedEventType,
}}
```

components/eventing-controller/pkg/handlers/utils_test.go
```diff
@@ -64,51 +68,23 @@ func TestGetInternalView4Ev2(t *testing.T) {
...
-				Type:         "oauth2",
-				GrantType:    "client_credentials",
-				ClientId:     "xxx",
-				ClientSecret: "xxx",
-				TokenUrl:     "https://oauth2.xxx.com/oauth2/token",
-				Scope:        []string{"guid-identifier"},
-			},
-		}
-		subscription.Spec.ProtocolSettings = givenProtocolSettings
+		subscription.Spec.ProtocolSettings = reconcilertesting.NewProtocolSettings(eventingtesting.WithContentMode, eventingtesting.WithExemptHandshake, eventingtesting.WithQOS, eventingtesting.WithWebhookAuth)
```
marcobebway, line 71: 👍

components/eventing-controller/pkg/handlers/utils.go
```diff
@@ -129,7 +129,7 @@ func getQos(qosStr string) (types.Qos, error) {
 }
 
 func getInternalView4Ev2(subscription *eventingv1alpha1.Subscription, apiRule *apigatewayv1alpha1.APIRule,
-	defaultWebhookAuth *types.WebhookAuth, defaultProtocolSettings *eventingv1alpha1.ProtocolSettings,
+	defaultWebhookAuth types.WebhookAuth, defaultProtocolSettings *eventingv1alpha1.ProtocolSettings,
```
marcobebway, line 132: How about the `ProtocolSettings`? Would changing it reflect in the default config as well?

components/eventing-controller/pkg/handlers/utils_test.go
sayanh: The intention was to reuse the method to reduce some code but it's just used only one hence removed.

components/eventing-controller/testing/test_helpers.go
```diff
@@ -133,6 +135,60 @@ func NewSubscription(name, namespace string, opts ...subOpt) *eventingv1alpha1.S
...
+}
+
+func WithQOS(protoSettings *eventingv1alpha1.ProtocolSettings) {
+	protoSettings.Qos = func() *string {
+		qos := "AT-LEAST_ONCE"
+		return &qos
+	}()
+}
+
+func WithWebhookAuth(protoSettings *eventingv1alpha1.ProtocolSettings) {
```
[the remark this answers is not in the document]
sayanh: done

components/eventing-controller/pkg/handlers/utils.go
sayanh: No, we are dealing with values for exemptHandshake and contentMode later.

components/eventing-controller/pkg/handlers/beb.go
```diff
@@ -98,7 +98,7 @@ func (b *Beb) SyncSubscription(subscription *eventingv1alpha1.Subscription, clea
 
 	// get the internal view for the ev2 subscription
 	var statusChanged = false
-	sEv2, err := getInternalView4Ev2(subscription, apiRule, b.WebhookAuth, b.ProtocolSettings, b.Namespace, b.SubNameMapper)
+	sEv2, err := getInternalView4Ev2(subscription, apiRule, *b.WebhookAuth, b.ProtocolSettings, b.Namespace, b.SubNameMapper)
```
radufa, line 101: Why could we not continue to use a "pointer" in the signature and instead we are copying (not deep copy) the object where the pointer points. The usual convention is to use in this case pointers (runtime performance and also avoiding of errors related to the "copy" of objects implementation in go) and only dereferencing the pointer at the moment of usage. At this point perhaps a "copy" of the object could be useful, like in our case.

radufa (approved):

components/eventing-controller/pkg/handlers/beb.go
sayanh: Sure, I'll make it a separate PR.

Issue #11179, opened 2021-04-23
Investigations on prometheus scraping without mtls


a-thaler:
**Description**
At the moment prometheus is running outside of the service-mesh. That sounds reasonable as it should observe the mesh and with that should not be part of it. Seeing istio demos and documentation, it appears to run outside of the mesh as well.

However, that conclusion results in a problem. Whenever prometheus should scrape an endpoint for metrics, it can not do that using mTLS. So usually the application needs to losen the requirements for client and needs to configure a permissive mode.

That is a not desired solution security wise and also not convenient.

Goal: Apply the recommended solution of istio a try by running a sidecar without intercepting the traffic and re-using the certificates inside prometheus.

Criterias:
- Documentation reflects the changes, also stating how and why prometheus is doing TLS by itself and not via envoy 
- Recommendation of istio is applied
- All unneded permissive modes for port of kyma components are removed

**Reasons**

**Attachments**
There seem to be a solution coming from istio itself by running a sidecar with prometheus which will not intercept any traffic but writes the certificates to a volume. That certificates prometheus will use for scraping the applications inside the mesh. See https://istio.io/latest/docs/ops/integrations/prometheus/#tls-settings

jeremyharisch:
The recommended solution of Istio was applied, meaning Prometheus is deployed including a sidecar without intercepting the traffic. Furthermore, Prometheus reuses the certificates of Istio to enable mTLS.

All possible kyma components, have been configured to scrape the metrics using mTLS. `Kiali` and `Grafana` will not use mTLS for now, due to some implementation issues. We need to have a second look at it in future.

The tutorial, as well as the documentation, have been adapted.


The pre-described changes will be done, by merging the following PRs in this given order:
1. [#11994](https://github.com/kyma-project/kyma/pull/11994)
2. [#182](https://github.com/kyma-project/examples/pull/182)
3. [#12013](https://github.com/kyma-project/kyma/pull/12013)

PR #11994, opened 2021-08-31
Enable Prometheus scraping using mTLS

jeremyharisch:
**Description**

Due to [this issue](https://github.com/kyma-project/kyma/issues/11179), we want to have a setup in which Prometheus runs outside of the Istio-Service-Mesh, but it should be possible that Prometheus scrapes the metrics using mTLS. 

Thus, we came up with the following solution:
- Deploy Prometheus with Istio sidecar-proxy. &#8594; SDS certificates are rotated and written to a volume that is shared with Prometheus.
- Sidecar of Prometheus does not intercept requests for Prometheus. &#8594; Prometheus does not have an active part in the service mesh. We have to do it in this way because, Prometheus’s model of direct endpoint access is incompatible with Istio’s sidecar proxy model.
- `PeerAuthentication` of Kyma components can be converted back to `STRICT`. Since, `STRICT` is the default value, the corresponding PeerAuthentication files are removed.
- The `ServiceMonitors` of each Kyma component needs the mounted TLS certificates for authentication.
- `insecureSkipVerify`  is set to true for each `ServiceMonitor` since Prometheus does not support `Istio secure naming`

**Related issue(s)**

#11179


resources/monitoring/templates/prometheus-operator/servicemonitor.yaml
```diff
@@ -10,17 +10,10 @@ metadata:
 spec:
   endpoints:
   {{- if .Values.prometheusOperator.tls.enabled }}
-  - port: https
```
a-thaler, line 13: why to remove it, just assure that the enabled setting is false by default

resources/monitoring/values.yaml
```diff
@@ -685,7 +689,15 @@ grafana:
 ##
 kubeApiServer:
   enabled: true
-  tlsConfig:
+  scheme: https
+
+    ## tlsConfig: TLS configuration to use when scraping the endpoint. For example if using istio mTLS.
+    ## Of type: https://github.com/coreos/prometheus-operator/blob/master/Documentation/api.md#tlsconfig
+  tlsConfig:
```
a-thaler, line 703: Does this really work? Have you checked the apiserver metrics?

resources/monitoring/charts/kube-state-metrics/values.yaml
```diff
@@ -94,7 +94,7 @@ tolerations: []
 
 # Annotations to be added to the pod
 podAnnotations:
-  sidecar.istio.io/inject: "false"
```
a-thaler: Please do not add a sidecar to kube-state-metrics

components/event-publisher-proxy/config/event-publisher-nats/120-peerauthentication.yaml
```diff
@@ -10,4 +10,4 @@ spec:
       app.kubernetes.io/instance: eventing
   mtls:
     mode: "PERMISSIVE"
----
```
a-thaler, line 13: Is there a good reason to keep it, to not have it STRICT?

a-thaler (changes requested):

components/event-publisher-proxy/config/event-publisher-nats/120-peerauthentication.yaml
jeremyharisch: Deleted it, it was not even used

resources/monitoring/values.yaml
jeremyharisch: Those certs are not used. Already checked it and works as expected

components/event-publisher-proxy/config/event-publisher-nats/120-peerauthentication.yaml
jeremyharisch: This PA is not used, will be removed

resources/monitoring/charts/kube-state-metrics/templates/servicemonitor.yaml
```diff
@@ -15,7 +15,8 @@ spec:
       {{- include "kube-state-metrics.selectorLabels" . | nindent 6 }}
   endpoints:
     - port: http
-      {{- if .Values.prometheus.monitor.honorLabels }}
+      scheme: http
+    {{- if .Values.prometheus.monitor.honorLabels }}
```
a-thaler, line 19: please revert the change :)

resources/monitoring/values.yaml
```diff
@@ -681,13 +685,23 @@ grafana:
...
+      certFile: /etc/prometheus/secrets/istio.default/cert-chain.pem
+      keyFile: /etc/prometheus/secrets/istio.default/key.pem
+      insecureSkipVerify: true
+
 ## Component scraping the kube api server
 ##
 kubeApiServer:
   enabled: true
-  tlsConfig:
+
```
a-thaler, line 699: please revert all changes here

a-thaler (approved):

piotrmiskiewicz (approved):

skhalash (approved):

sayanh (approved):

mjakobczyk (approved):
After checking it out and testing with Darek, it looks good to us 👍

apenen:
Hi! Could you apply this change also in version 1.x? Thanks.

Issue #11989, opened 2021-08-30
Event delivery for NATS is broken with multiple subscriptions


marcobebway:
**Description**

Not all Event subscribers receive Events when NATS is used as the Eventing backend.

**Steps to reproduce**

Can be reproduced from Kyma main branch (for example this [commit](https://github.com/kyma-project/kyma/commit/0cd4d2c7ad12ca9b5983e81d3d36a92c2048519b))

1. Install Kyma.
2. Make sure NATS is the default Eventing backend.
3. Create more than one function/subscriber with different sinks but all interested in the same EventType.
4. Publish one Event with the EventType mentioned above.
5. Check logs not all functions/subscribers receive the Event.

**Troubleshooting**

- Check Eventing controller logs, only one dispatch call is made to one of the subscribers.
- All created Kyma subscriptions should be ready.

radufa:
We have manually tested it during the tests of https://github.com/kyma-project/kyma/issues/11719 using Kyma 1.24.2 for OSS and 1.24.0 for SKR.
It seems that "main" is dirty now

radufa:
This issue is related to https://github.com/kyma-project/kyma/issues/11719

radufa:
Tested with Kyma 1.24.3 on SKR. It's working as expected. 
We should focus on "main" to fix it.

Issue #11719, opened 2021-07-20
The "multiplexer" eventing pattern doesn't work with the NATS backend


radufa:
**Description**

If two subscriptions are registered to the same filter, the event should be delivered to both subscriptions.
This is working for BEB but is not working for NATS

**Expected result**

**Actual result**

**Steps to reproduce**

To reproduce it, register two subscriptions having the same filter, but different names:
```
cat << EOF | k apply -f -
apiVersion: eventing.kyma-project.io/v1alpha1
kind: Subscription
metadata:
  name: test-noapp
  namespace: test
spec:
  filter:
    filters:
    - eventSource:
        property: source
        type: exact
        value: ""
      eventType:
        property: type
        type: exact
        value: sap.kyma.custom.noapp.order.created.v1
  protocol: ""
  protocolsettings: {}
  sink: http://test.test.svc.cluster.local
EOF
```
and
```
cat << EOF | k apply -f -
apiVersion: eventing.kyma-project.io/v1alpha1
kind: Subscription
metadata:
  name: test-noapp-1
  namespace: test
spec:
  filter:
    filters:
    - eventSource:
        property: source
        type: exact
        value: ""
      eventType:
        property: type
        type: exact
        value: sap.kyma.custom.noapp.order.created.v1
  protocol: ""
  protocolsettings: {}
  sink: http://anothertest.anothertest.svc.cluster.local
EOF
```
For BEB, the function receives two events, corresponding to the two subscriptions
For NATS only one subscriber receives the message.

**Troubleshooting**

k15r:
Can you help me understand why we would need to have this?

mfaizanse:
I am unable to reproduce this bug. I tried two cases with NATS:

CASE 1: Sink in same namespace
I created a function "test" and created two subscriptions with same filters in namespace "test".
I published an event, and the function received two events.

CASE 2: Sink in different namespace
I created a function "test" in namespace "test", and created a function "anothertest" in namespace "anothertest".
I created two subscriptions with same filters in namespace "test" with corresponding sinks to functions "test" and "anothertest"
I published an event, and both functions received the event.

Am I missing something in the steps to reproduce this?

radufa:
Tested on SKR 1.24.0 (AWS) too. For OSS it was tested with Kyma 1.24.2.
All is working as expected, it's not reproducible anymore.

radufa:
[...]

PR #12006, opened 2021-09-01
Fix event multiplexer pattern for NATS

radufa:
**Description**

- fix the event multiplexer pattern for NATS backend
- add tests to verify it

**Related issue(s)**

See also: #11719 #11989

### DIFF
--- a/components/eventing-controller/pkg/handlers/nats.go
+++ b/components/eventing-controller/pkg/handlers/nats.go
@@ -122,6 +122,7 @@ func (n *Nats) SyncSubscription(sub *eventingv1alpha1.Subscription, cleaner even
 		}
 
 		for i := 0; i < subsConfig.MaxInFlightMessages; i++ {
+			// queueGroupName must be unique for each subscription and subject
 			queueGroupName := createKeyPrefix(sub) + string(types.Separator) + subject
 			natsSub, subscribeErr := n.connection.QueueSubscribe(subject, queueGroupName, callback)
 			if subscribeErr != nil {
--- a/components/eventing-controller/pkg/handlers/nats.go
+++ b/components/eventing-controller/pkg/handlers/nats.go
@@ -122,7 +122,8 @@ func (n *Nats) SyncSubscription(sub *eventingv1alpha1.Subscription, cleaner even
 		}
 
 		for i := 0; i < subsConfig.MaxInFlightMessages; i++ {
-			natsSub, subscribeErr := n.connection.QueueSubscribe(subject, subject, callback)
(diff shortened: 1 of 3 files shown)


mfaizanse:
Hi,
* I have tested the PR on the cluster and the multiplexing is working. I used 3 functions and 3 subscriptions.
* I also reproduced the bug on kyma:main, and multiplexing is not working on main.
* Also, the unit test is failing without the fix, so thats good. And with new changes the unit test is passing.

components/eventing-controller/pkg/handlers/nats_test.go
```diff
@@ -180,6 +181,107 @@ func TestSubscription(t *testing.T) {
...
+
+	// Prepare event-type cleaner
+	application := applicationtest.NewApplication(eventingtesting.ApplicationNameNotClean, nil)
+	applicationLister := fake.NewApplicationListerOrDie(context.Background(), application)
+	cleaner := eventtype.NewCleaner(eventingtesting.EventTypePrefix, applicationLister, defaultLogger)
+
+	// Create 3 subscriptions having the same sink and the same event type
+	var subs [3]*eventingv1alpha1.Subscription
+	for i:=0; i< len (subs); i++ {
+		subs[i] = eventingtesting.NewSubscription("sub" + strconv.Itoa(i), "foo", eventingtesting.WithNotCleanEventTypeFilter)
```
k15r, line 233: nit: `fmt` is already imported, so 
```
fmt.Sprintf("sub-%d", i)
```
```suggestion
		subs[i] = eventingtesting.NewSubscription(fmt.Sprintf("sub-%d", i), "foo", eventingtesting.WithNotCleanEventTypeFilter)
```

components/eventing-controller/pkg/handlers/nats.go
```diff
@@ -122,7 +122,8 @@ func (n *Nats) SyncSubscription(sub *eventingv1alpha1.Subscription, cleaner even
 		}
 
 		for i := 0; i < subsConfig.MaxInFlightMessages; i++ {
-			natsSub, subscribeErr := n.connection.QueueSubscribe(subject, subject, callback)
+			queueGroupName := createKeyPrefix(sub) + string(types.Separator) + subject
```
k15r, line 125: Could we add a comment why we introduced this prefix?

radufa: Done

components/eventing-controller/pkg/handlers/nats_test.go
radufa: Done

k15r (dismissed):

k15r (approved):

PR #12033, opened 2021-09-03
fix/authpolicy to use enable log-ui

bdelage:
**Description**

- With the current authPolicy, istio will prevent log-ui to fetch/display logs from Loki

The https://log-ui.ingress.domain.name url is accessible. But "log-ui" cannot fetch any labels or logs from the loki endpoint.
In the istio-proxy sidecar logs in the loki pod we can see the following messages :

`logging-loki-0 istio-proxy [2021-09-03T14:45:31.835Z] "GET /api/prom/label/app/values HTTP/1.1" 403 - rbac_access_denied_matched_policy[none] - "-" 0 19 0 - "xxx.xxx.xxx.xxx, xxx.xxx.xxx.xxx ,xxx.xxx.xxx.xxx" "Mozilla/5.0 (Macintosh; Intel Mac OS X 10.15; rv:91.0) Gecko/20100101 Firefox/91.0" "e7010f2c-600b-49e1-85ed-8bac099772fa" "loki.ingress.domain.name" "-" inbound|3100|| - xxx.xxx.xxx.xxx:3100 xxx.xxx.xxx.xxx:0 outbound_.3100_._.logging-loki.kyma-system.svc.cluster.local -`

The field principals has a value only if a workload can identify itself via a certificate (it must have the istio proxy) during PeerAuthentication. And the field requestPrincipals is extracted from the jwt token during RequestAuthentication.

### DIFF
--- a/resources/logging/charts/loki/templates/kyma-additions/authorizationpolicy.yaml
+++ b/resources/logging/charts/loki/templates/kyma-additions/authorizationpolicy.yaml
@@ -18,6 +18,8 @@ spec:
   - from:
     - source:
         requestPrincipals: ["*"]
+    - source:
+        principals: ["*"]
     to:
     - operation:
         paths: ["/api/prom/query","/api/prom/label/*","/api/prom/tail"]


a-thaler:
Can you please describe the actual problem? We have automated tests proving that with a valid token you can access the endpoint. So what is the difference to "principal" only and why we should activate it?

a-thaler:
The "log-ui" is a view for the kyma console. It is not meant to be called by workload but by a browser. It requires to run as a frame inside the kyma console and will get the token injected from the logged-in user. In the kyma console you can navigate to the "Pods" section and "Show Logs" via the 3 dots behind an entry. The pop-up is the view coming from the "log-ui" component.

I assume you actually want to call Loki via API independent of any UI, right?
For that you should configure an additional AuthorizationPolicy where you explicitly define the serviceaccount of your workload.

I will close the PR. Feel free to contact me via Slack to discuss your scenario further to see what the best solution is. I don't want to expand the exposure of the Loki API to the outside by default.

PR #12044, opened 2021-09-06
Have a default value if env var is not defined

rakesh-garimella:
**Description**

-If there is no environment var defined for kyma major version then have a default value
**Related issue(s)**

### DIFF
--- a/tests/fast-integration/utils/index.js
+++ b/tests/fast-integration/utils/index.js
@@ -1041,11 +1041,19 @@ function genRandom(len) {
   return res;
 }
 
-function getEnvOrThrow(key, defValue = "") {
+function getEnvOrDefault(key, defValue = "") {
   if (!process.env[key]) {
     if (defValue != "") {
       return defValue
     }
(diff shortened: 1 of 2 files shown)


tests/fast-integration/utils/index.js
```diff
@@ -1041,8 +1041,11 @@ function genRandom(len) {
...
 }
 
-function getEnvOrThrow(key) {
+function getEnvOrThrow(key, defValue = "") {
   if (!process.env[key]) {
+    if (defValue != "") {
+      return defValue
+    }
     throw new Error(`Env ${key} not present`);
   }
```
skhalash, line 1050: I would not do this change. In this case, it does not throw if the env var is not set. Instead, it returns the def value, but only if it is set, which makes the API hard to understand. I would add another function instead, `getEnvOrDefault` that never throws and returns either the env var or default.

skhalash (changes requested):

skhalash (approved):

PR #12063, opened 2021-09-08
Add "check-code" for local publisher-proxy development

radufa:
**Description**

- Add a new target "code-check" which uses "check-imports", "fmt-local", "vet-local", "lint"

**Related issue(s)**

### DIFF
--- a/components/event-publisher-proxy/Makefile
+++ b/components/event-publisher-proxy/Makefile
@@ -1,6 +1,3 @@
-# Install lint on the local environment
-# curl -sfL https://install.goreleaser.com/github.com/golangci/golangci-lint.sh | bash -s -- -b $GOPATH/bin
-
 APP_NAME = event-publisher-proxy
 APP_PATH = components/$(APP_NAME)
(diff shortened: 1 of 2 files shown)


components/event-publisher-proxy/Makefile
```diff
@@ -26,8 +26,11 @@ dep-status-local:
 mod-verify-local:
 	GO111MODULE=on go mod verify
 
+# Use it before commit
+check-code: check-imports fmt-local vet-local lint
+
 test-local:
-	GO111MODULE=on go test ./...
+	GO111MODULE=on go test ./... -coverprofile cover.out
```
k15r, line 33: do we really need to always write the coverage to a file? Usually developers use an IDE to check the the coverage and these then also take care of generating the coverage for you

k15r (approved):

alexandra-simeonova (approved):

components/event-publisher-proxy/Makefile
radufa: Added only to have the same coverage behavior as in "eventing-controller"

PR #12096, opened 2021-09-14
Bump base images and versions for rafter and serverless

moelsayed:
**Description**

- ...
- ...
- ...

**Related issue(s)**

### DIFF
--- a/resources/rafter/values.yaml
+++ b/resources/rafter/values.yaml
@@ -13,7 +13,7 @@ tests:
 
   image:
     repository: "eu.gcr.io/kyma-project/rafter-test"
-    tag: "PR-11682"
+    tag: "PR-12096"
     pullPolicy: IfNotPresent
   disableConcurrency: true
   restartPolicy: Never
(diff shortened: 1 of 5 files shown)


resources/rafter/values.yaml
```diff
@@ -210,7 +210,7 @@ controller-manager:
         pullPolicy: IfNotPresent
 
     image:
-      tag: RELEASE.2021-06-17T00-10-46Z-05603b27
+      tag: RELEASE.2020-09-21T22-31-59Z-e85e5eaf
```
pPrecel, line 213: Why are we downgrading the image version? :O Is this because of the license?

moelsayed: I believe so, yes. This the image tag generated by TPI, so I am assuming that's the one we used last. @kwiatekus, @m00g3n  ?

pPrecel (approved):

resources/rafter/values.yaml
pPrecel: Everything looks ok. It's because of [this](https://github.com/moelsayed/third-party-images/commit/05603b27f3a2348c02ff1a92f6280d658df63fda#diff-27987557be8c3298ec9d912000851d8163da5062a8afad4fb99c17fa873c2893)
👍

Issue #12005, opened 2021-09-01
The status of NodeJS function changes to "deploying" when load test, and performance become to be poor


wh7577:
I installed Kyma 1.24.3 in azure, and created 3 nodes AKS using Standard_D8s_v3. When I test the performance of helloworld function (NodeJS default), the status of helloworld function will changed to "deploying" from time to time. And the performance become to be poor, from 30ms(when start) downgrade to 1000+ ms, and error "context deadline exceeded (Client.Timeout exceeded while awaiting headers)" shows in azure. 

AKS Kubernetes versions: 1.20.7
Kyma version : 1.24.3
Load Test tool : JMeter 5.4.1

**Expected result**
Status should be running and not change, the helloworld performance should be under 50 ms.

**Actual result**
Status changed to "deploying" from time to time, and the performance become to poor, from 30ms (when start) downgrade to 1000+ ms.

**Steps to reproduce**
1. create NodeJS function, set 3 replicas
2. perform load test by JMeter by 200 concurrent threads

**Troubleshooting**
See attached screendump.
![nodejs12-helloworld-become-to-deploying](https://user-images.githubusercontent.com/3860078/131604547-69d5a072-2674-46cc-9dd9-b78d362a2b18.png)
![nodejs12-helloworld-containers-not-ready](https://user-images.githubusercontent.com/3860078/131604549-9476d37a-9b25-4ebc-bc67-e98d2c45d1e0.png)
![nodejs12-helloworld-context-deadline-exceed](https://user-images.githubusercontent.com/3860078/131604551-d2dc6437-0973-4ede-9bfc-77b1f01db890.png)
![nodejs12-helloworld-function](https://user-images.githubusercontent.com/3860078/131604552-5537b0ba-d258-4cb0-8f9d-bb7631ff4674.png)
![nodejs12-helloworld-performance (3 pods)](https://user-images.githubusercontent.com/3860078/131604554-62b9a35e-c909-4264-92bd-5a224d30e569.png)
![nodejs12-helloworld-pods-config](https://user-images.githubusercontent.com/3860078/131604556-f1698129-d891-4218-8764-5069c3f8b4f4.png)
![nodejs12-helloworld-pod-warning-in-azure](https://user-images.githubusercontent.com/3860078/131604557-b275c71f-f07b-4e90-8ff1-db85a3fd7e4b.png)

kwiatekus:
Hi @wh7577 thanks for checking.
[...]

PR #12110, opened 2021-09-15
Introduce Liveness Probe

kwiatekus:
**Description**

- Increase timeout for function's readiness probe
- Introduce a liveness probe that would restart function pod instance only if healthz endpoint doesnt respond 3 times ( checked every 15 seconds )

**Related issue(s)**
https://github.com/kyma-project/kyma/issues/12005


pPrecel (approved):
LGTM. Good Job!

PR #12115, opened 2021-09-16
Move in-cluster eventing sub-test to the fixture

pxsalehi:
**Description**

- Move the binary mode in-cluster test to the fixture to ensure it is called in all fast-integration make targets.

**Related issue(s)**


wozniakjan (approved):
/lgtm

PR #12182, opened 2021-09-23
Make charts installable with helm template

pbochynski:
**Description**

- remove statements that prevent rendering virtual services with `helm template` command

**Related issue(s)**
https://github.com/kyma-project/kyma/issues/10939


strekm (approved):
/meow

Issue #12108, opened 2021-09-15
Common format for URLs of container images in values.yaml


adamwalach:
**Description**

This is a proposal to create a common format for defining URLs of container images used by our components.
Current state is like this (example from application connector):
```
global:
 ....
 containerRegistry:
    path: eu.gcr.io/kyma-project
 application_broker:
    version: "cddfbc5b"
 application_broker_eventing_migration:
    version: "a8a6bca9"
 application_connectivity_certs_s
.....
```
Proposed format:
```
global:
 ....
 containerRegistry: 
    path: eu.gcr.io/kyma-project
 images:
   application_broker:
      version: "cddfbc5b"
      name: "application-broker"
   application_broker_eventing_migration:
      version: "a8a6bca9"
      name: "application-broker-eventing-migration"
   k8s_tools:
      version: "20210817-8cdbfce3"
      name: "k8s_tools"
      directory: "tpi
 test-images:
   application_connector_tests:
       version: "PR-11939"
       name: "application_connector_tests"
```

Assumptions
- only one global registry for Kyma installation. Reason for that: simplifies the config file, and I've never seen a scenario for using multiple image registries at the same time (in Kyma)

Additional TODOs:
- helper function to glue together a full image URL based on a key
  format: `{{ .Global.containerRegistry.path }}/{{ .Global.images[key].directory}}/{{ .Global.images[key].name}}:{{ .Global.images[key].version }}`
- create a global definition listing all values YAMLs:
    ```
    definition.yaml
      - github.com/kyma-project/kyma/resources/application-connector/values.yaml
      - github.com/kyma-project/kyma/resources/apiserver-proxy/values.yaml
    ``` 

Open questions:
- do we need a separate field with test-images? maybe we could add something like "type:test" to the image definition
- maybe this also solves the problem of removing PR-images from main branch before release starts. I am not sure about that. We need to add a check that it's not allowed to use PR-image from a pull request that is still open

**Reasons**

Currently, it's almost impossible to gather a list of all images used by Kyma. That makes security scans and images promotion (during a release) very complicated and is error-prone. A strictly defined format proposed in this issue makes parsing very easy and provides an easy way to promote (change repo url and tag) for all images.

**Related issues**
- https://github.com/kyma-project/test-infra/issues/2453

k15r:
Even though you wrote there are no scenarios that need multiple registries we could easily default to the `containerRegistry` and allow to override it in the images configuration:
```
k8s_tools:
[...]

PR #12212, opened 2021-09-28
Updates to kiali and logging chart values.yaml

Ressetkk:
**Description**
Updated charts for logging and kiali based on suggestions from #12108
**Related issue(s)**

#12108


a-thaler (approved):

resources/kiali/values.yaml
```diff
@@ -298,7 +306,8 @@ kiali:
     # a single, specific Kiali version, thus this setting may have no effect depending on how the
     # operator itself was configured.
     #    ---
-      image_name: "eu.gcr.io/kyma-project/tpi/kiali"
+    #.   ATTENTION: this setting is will not be effective as it got replaced by the global.images.kiali setting.
```
NHingerl, line 309:
```suggestion
    #.   ATTENTION: This setting will not be effective. It has been replaced with the global.images.kiali setting. 
```

Are you sure this is the correct line for the Attention note? Why not put it right after line 300 so users can stop reading if they aren't curious?

resources/kiali/values.yaml
```diff
@@ -323,7 +332,8 @@ kiali:
     # a single, specific Kiali version, thus this setting may have no effect depending on how the
     # operator itself was configured.
     #    ---
-      image_version: "1.38.1-e0da5a68"
+    #.   ATTENTION: this setting is will not be effective as it got replaced by the global.images.kiali setting.
```
NHingerl, line 335:
```suggestion
    #.   ATTENTION: This setting will not be effective. It has been replaced with the global.images.kiali setting. 
```

Again, consider putting this right after line 321.

NHingerl (approved):

a-thaler (approved):

kwiatekus (approved):

Issue #12201, opened 2021-09-27
ORY Hydra lacks information about the domain


mjakobczyk:
**Description**

Ory Hydra should have an information about the domain. Currently it contains a proper but unusable in remote environments value: `127.0.0.1`. It should contain FQDN of public Hydra endpoint such as: `https://oauth2.my-domain.com`, where `my-domain` is the public cluster domain name (note that the name of Hydra's endpoint always starts with `oauth2`).

You can fetch the information about the configuration by running the following command:
```bash
curl https://oauth2.mst-2.goatz.shoot.canary.k8s-hana.ondemand.com/.well-known/openid-configuration
```

**Expected result**

**Actual result**

For now we get `127.0.0.1` (localhost) as a domain.

```json
{
  "issuer": "http://127.0.0.1:4444/",
  "authorization_endpoint": "http://127.0.0.1:4444/oauth2/auth",
  "token_endpoint": "http://127.0.0.1:4444/oauth2/token",
  "jwks_uri": "http://127.0.0.1:4444/.well-known/jwks.json",
  (...other config...)
}
```

**Steps to reproduce**

1. Install Kyma (either 1.X or 2.0) including Ory component.
2. Check your `<domain-name`.
3. Run the following command:
```bash
  curl https://oauth2.<domain-name>/.well-known/openid-configuration
```

**Troubleshooting**

Workaround is as it follows:
1. Edit the `ory-hydra` deployment:
```bash
kubectl -n kyma-system edit deployment ory-hydra
```
2. Change the value of the env variable: "URLS_SELF_ISSUER".
3. Apply the changes and wait for the `ory-hydra` pod to redeploy. It should take no more than a couple of minutes. After that you can verify if the changes were applied correctly.

raypinto:
Updated the URLS_SELF_ISSUER value in the deployment.yaml of ORY Hydra to obtain the domain name dynamically.
Tested the same on a Gardener cluster and locally on a k3d cluster and we observe that the Ory Hydra contains the actual domain name when kyma is installed with Ory.

```
Kyma is installed in version:	local
Kyma installation took:		0 hours 5 minutes
Kyma is running at:		rp-test1.goatz.shoot.canary.k8s-hana.ondemand.com
Kyma console:			not installed

Happy Kyma-ing! :)

➜  curl https://oauth2.rp-test1.goatz.shoot.canary.k8s-hana.ondemand.com/.well-known/openid-configuration
```
```json
[...]

PR #12219, opened 2021-09-29
Bug/ory url self issuer

raypinto:
**Description**

- URLS_SELF_ISSUER value to obtain the domain name dynamically on ory hydra installation

**Related issue(s)**
`See also #12201 `


strekm:
Hydra is serving localhost on /.well-known/openid-configuration as we did not want to expose this. we do not plan to support other OAuth flows with Hydra. Kyma team recommends using external Oauth2 / OIDC complaint services instead.
Cheers, Magda

PR #12229, opened 2021-09-29
Changes to the monitoring chart

Ressetkk:
**Description**

Updates to the monitoring chart values.yaml

**Related issue(s)**

#12108


resources/monitoring/charts/grafana/values.yaml
```diff
@@ -117,9 +114,6 @@ extraLabels: {}
 # priorityClassName:
 
 downloadDashboardsImage:
-  repository: curlimages/curl
```
a-thaler, line 120: we are not using that image/functionality, please just comment out and do not specify in the listing

resources/monitoring/charts/grafana/values.yaml
```diff
@@ -717,12 +705,6 @@ imageRenderer:
   enabled: false
   replicas: 1
   image:
-    # image-renderer Image repository
-    repository: grafana/grafana-image-renderer
-    # image-renderer Image tag
```
a-thaler, line 722: we are not using that image/functionality, please just comment out and do not specify in the listing. Please revert related changes of defining the image to have it more easy to update the chart

resources/monitoring/charts/prometheus-istio/values.yaml
```diff
@@ -47,8 +47,6 @@ alertmanager:
   ## alertmanager container image
   ##
   image:
-    repository: prom/alertmanager
```
a-thaler, line 50: we are not using that image/functionality, please just comment out and do not specify in the listing. Please revert related changes of defining the image to have it more easy to update the chart

resources/monitoring/charts/prometheus-istio/values.yaml
```diff
@@ -384,8 +380,6 @@ configmapReload:
     ## configmap-reload container image
     ##
     image:
-      repository: jimmidyson/configmap-reload
-      tag: v0.4.0
```
a-thaler, line 388: we are not using that image/functionality, please just comment out and do not specify in the listing. Please revert related changes of defining the image to have it more easy to update the chart

resources/monitoring/values.yaml
```diff
@@ -1223,9 +1290,6 @@ prometheusOperator:
     patch:
       enabled: false
       image:
-        repository: jettech/kube-webhook-certgen # This image is not activated in Kyma. This is kept here to keep parity with upstream charts
-        tag: v1.5.0
```
a-thaler, line 1227: please read the comment... please just keep it commented, the image is not actively used

resources/monitoring/charts/grafana/values.yaml
adamwalach: We can leave them as they are, but that means we need to add to them to the list of exceptions.
They will not be sent to security scanning pipeline and will not trigger any error on PR checks (these checks will be enabled on kyma repo after this PR is merged). 
The problem is that security tooling has no idea about flags that you use in helm charts, so it will be hidden even if you enable it. It might be noticed on our scans that run on development clusters, but there is no guarantee for that (depends on use case).

Halamix2: The image was restored

resources/monitoring/charts/grafana/values.yaml
Halamix2: The image was restored

resources/monitoring/charts/prometheus-istio/values.yaml
Halamix2: The image was restored

resources/monitoring/values.yaml
Halamix2: The image was restored

resources/monitoring/charts/prometheus-istio/values.yaml
Halamix2: The image was restored

resources/monitoring/charts/prometheus-istio/values.yaml
Halamix2: This change was reverted

skhalash (approved):

PR #12295, opened 2021-10-06
Fix gcloud and custom DB Ory install

cnvergence:
**Description**

Fixes a bug in installing Ory with custom DB or GCloud SQL. Such install will now fail as there are no PASSWORD_R_KEY and PASSWORD_R values passed.

- Replication password should be added only for Postgres databases.

**Related issue(s)**


mjakobczyk (approved):
LGTM 👍

Issue #11825, opened 2021-08-02
[Spike] Figure out a good structure for fast-integration tests


skhalash:
**Description**

**Reasons**

At the moment, fast-integration tests have no structure, which makes them hard to maintain. It's also unclear which guidelines to follow when adding a new test suite. The spike is about analyzing the pain points with fast-integration tests and presenting a possible solution, which will be broken down into follow-up tasks.

**Acceptance Criteria**

* Investigate how to make it possible to easily reuse test scenarios for install, upgrade tests for OSS and SKR Kyma
* Reduce the number of `make` targets by passing test scenarios, environment flags (compass+), etc. as parameters
* Investigate if we could use tags to solve the problems above
* Set up a linter https://github.com/kyma-project/kyma/issues/11370
* Improve directory structure
* Extract the `installer` module since it is not used by any tests
* Present the proposed solution to the team

**Attachments**

elchead:
Should we have a linter? Currently style is not consistent

Ressetkk:
Regarding test scenarios re-usability: Mocha tests work around one object that is passed between test cases. That means every test case has access to the fields under `this` directive, and those will be persistent around all child `describe()` and `it()` blocks. This thing is called *shared behaviours* and you can read more on the mocha's wiki. https://github.com/mochajs/mocha/wiki/Shared-Behaviours

Each set of tests can be defined as a function. Each tests retrieves the object to be tested using `this` directive. Let's say we have the following example the `foo` test will pass:
```js
function foo() {
  describe('foo', function () {
    it('foo', async function() {
      this.var.has.to.be.true;
    })
  })
}

describe('test-case', function () {
  before(async function() {
    this.var = true;
  })
  foo();
})
```
This approach should come in handy when re-using the tests are advised. Check my still-in-progress PRs that tackle this with the approach mentioned above #12514

lindnerby:
Closing this, since we introduced the following changes:
* Reduce the number of make targets 
* Introduce a linter
* Improve encapsulation of tests and commerce-mock
* Remove installer module
Still ideas to improve: utils.js is unmaintainable due to size, commerce-mock is too complex and should be rather separated for various scenarios.
For remaining improvements teams themself are responsible

PR #12514, opened 2021-10-29
Fast-integration hourly periodic for SKR nightly

Ressetkk:
**Description**
This PR addresses part of the #11825 regarding reusability of the test scenarios.
Mocha tests operate on a single `mocha` object that can store variables and access them using `this` clause. This allows defining test cases as functions to use them in tests that have shared behaviours and can define the same tests.

For SKR test scenarios I've introduced the `options` object which stores the options that can be easily overriden with custom values simply by using function parameters in constructor. This simplifies using the same test with multiple options without relying on a lot of environment variables, which may not be easy to read from code.

The provisioning and deprovisioning should not be a part of the test because this is step that prepares environment for testing so it's been moved to the `before()` hook. By moving the step to the hook it's easier to fail the test if there is something wrong with the preparation step and it does not execute any tests. Additionally the preparation step may be different for various environments despite having the same test scenarios.

OS Fast-integration test has also been modularised, but given the nature of this test nothing specific had to be introduced there, just wrapping it with a function was enough.

This PR also fixes the provisioning step of SKR nightly, which has been broken for some time because of the update of KCP CLI and KCPWRapper used in this scenario.

Also the test scenario for hourly SKR test execution has been added, but due to issues with the Compass infrastructure it will not be enabled.


tests/fast-integration/skr-nightly/test/test-nightly.js
```diff
@@ -0,0 +1,58 @@
...
+            runtime = runtimes.data[0];
+         }
+      });
+      it (`Initialize k8s client from nightly runtime`, async function () {
+         const shoot = await gardener.getShoot(runtime.shootName);
+         console.log(shoot.name, runtime.shootName);
+         initializeK8sClient({ kubeconfig: shoot.kubeconfig });
+      });
+   });
+   commerceMockTests();
```
pbochynski, line 56: You need to use compass test fixture like here:
https://github.com/kyma-project/kyma/blob/cfa81506876c3037048dedbf0f01c557ba127888/tests/fast-integration/skr-test/skr-test.js#L160-L168

Ressetkk: So, basically the tests that are defined in `test/` directory are not meant for SKR?

Ressetkk: In this case maybe refactoring the `skr-test` to be more re-usable will be better, in similar I did with the regular tests? Because right now the test cases are copy-pasted to the nightly provision job and run just after the SKR is provisioned.

PR #12535, opened 2021-11-03
React to OAuth2 secret changes in the Eventing controller 

### LINKED ISSUE #12516: Deletion of oauth2client breaks eventing in EventMesh mode
**Description**
If a user deletes the oauth2client our controller does not 
* either recreate the oauth2 client
* set subscription status ready to false

If the oauth2client gets recreated by the reconciler our controller still does not work as the clientcredentials in BEB mode have not been updated

**Expected behaviour**

* controller should own the oauth2client and watch it. 
* If recreation of the oauth2client happens then all subscriptions need to be updated.
* upgrade flow must still work: #11709

**finding**
* subscriptions were ready all the time even though they did not work
* restarting the controller did not fix the broken state even though now the secret did exist again
* fixing was possible by switching eventing to nats and then back to BEB  (switching to nats triggers a cleanup on BEB)

pxsalehi:
**Description**

- React to OAuth2 secret changes in the Eventing controller 

**Related issue(s)**

Closes #12516


components/eventing-controller/pkg/handlers/utils.go
```diff
@@ -261,6 +261,12 @@ func RemoveStatus(sub eventingv1alpha1.Subscription) *eventingv1alpha1.Subscript
 	return desiredSub
 }
 
+func SetStatusAsNotReady(sub eventingv1alpha1.Subscription) *eventingv1alpha1.Subscription {
```
radufa, line 264: "sub" should be a pointer to eventingv1alpha1.Subscription to avoid the "copy" of the actual parameter to the formal parameter during the function call. This "cop" is not a deep copy.

pxsalehi: I'm just trying to be consistent with the method above it `RemoveStatus`.

pxsalehi: To me it seemed that by passing by value, it makes it clear (and prevents) that the subscription is mutated.

radufa (approved):

PR #12539, opened 2021-11-04
[istio-conf] adjust helm settings to match Kyma 1 convention for tracing

piotrkpc:
In Kyma2 tracing was controlled by setting meshConfig.enableTracing while
in Kyma 1.x it's global.tracing.enabled. This changes the config, so both options are supported

### DIFF
--- a/resources/istio-configuration/templates/istio-operator.yaml
+++ b/resources/istio-configuration/templates/istio-operator.yaml
@@ -25,6 +25,8 @@ spec:
 {{- toYaml .Values.components.pilot.config | nindent 10}}
   meshConfig:
 {{- toYaml .Values.meshConfig | nindent 4 }}
+    enableTracing: {{ .Values.global.tracing.enabled }}
+
   values:
     global:
 {{- toYaml .Values.helmValues.global | nindent 8 }}
--- a/resources/istio-configuration/templates/istio-operator.yaml
+++ b/resources/istio-configuration/templates/istio-operator.yaml
@@ -24,8 +24,7 @@ spec:
       k8s:
 {{- toYaml .Values.components.pilot.config | nindent 10}}
   meshConfig:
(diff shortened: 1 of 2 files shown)


resources/istio-configuration/templates/istio-operator.yaml
```diff
@@ -24,7 +24,8 @@ spec:
       k8s:
 {{- toYaml .Values.components.pilot.config | nindent 10}}
   meshConfig:
-{{- toYaml .Values.meshConfig | nindent 4 }}
+{{- tpl (toYaml .Values.meshConfig | nindent 4) . | replace "'true'" "true" | replace "'false'" "false" }}
```
werdes72, line 27: what if it'll be "true" or `true` instead of 'true'

piotrkpc: If I understand correctly difference in Yaml between singl and double quotes it won't happen because this needs to be evaluated as string. Double quotes will attempt to evaluate special characters such as `\n` as new line instead just printing `\n` string

werdes72: I meant that it probably won't be replaced to boolean: replace "'true'" "true", will catch only `'true'`, but not `"true"`

piotrkpc: I agree but imo we don't need to care for this because it won't evaluate to `"true"` with double quotes.

Tomasz-Smelcerz-SAP: Perhaps I don't get your intention here, but it doesn't work for me:
```
> istio-configuration % helm template istio . --set global.tracing.enabled="'true'" | grep enableTracing
    enableTracing: 'true'
```
As you can see, the resulting value is a string: `'true'`, not a boolean `true`

Tomasz-Smelcerz-SAP (approved):

PR #12567, opened 2021-11-08
Add  a new Condition for BEB Webhook callback status

radufa:
**Description**

- Add  a new condition for BEB Webhook callback status
- The status "ready" of a BEB subscription checks now the status of the new condition too
- Avoid during the reconciler call to work with an outdated subscription object

**Related issue(s)**

### DIFF
--- a/components/eventing-controller/api/v1alpha1/subscription_lifecycle.go
+++ b/components/eventing-controller/api/v1alpha1/subscription_lifecycle.go
@@ -10,9 +10,9 @@ import (
 type ConditionType string
 
 const (
-	ConditionSubscribed         ConditionType = "Subscribed"
-	ConditionSubscriptionActive ConditionType = "Subscription active"
-	ConditionAPIRuleStatus      ConditionType = "APIRule status"
+	ConditionSubscribed              ConditionType = "Subscribed"
+	ConditionSubscriptionActive      ConditionType = "Subscription active"
+	ConditionAPIRuleStatus           ConditionType = "APIRule status"
 	ConditionSubscriptionWebhookCall ConditionType = "Webhook call status"
 )
 
@@ -150,7 +150,7 @@ func (s *SubscriptionStatus) IsConditionSubscribed() bool {
 func (s *SubscriptionStatus) IsConditionWebhookCall() bool {
 	for _, condition := range s.Conditions {
 		if condition.Type == ConditionSubscriptionWebhookCall &&
-				(condition.Status == corev1.ConditionTrue || condition.Status == corev1.ConditionUnknown)  {
+			(condition.Status == corev1.ConditionTrue || condition.Status == corev1.ConditionUnknown) {
 			return true
 		}
 	}
--- a/components/eventing-controller/api/v1alpha1/subscription_lifecycle.go
+++ b/components/eventing-controller/api/v1alpha1/subscription_lifecycle.go
@@ -10,10 +10,10 @@ import (
 type ConditionType string
 
 const (
-	ConditionSubscribed              ConditionType = "Subscribed"
-	ConditionSubscriptionActive      ConditionType = "Subscription active"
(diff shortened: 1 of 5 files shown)


components/eventing-controller/controllers/subscription/beb/reconciler.go
```diff
@@ -643,18 +657,25 @@ func (r *Reconciler) syncInitialStatus(ctx context.Context, subscription *eventi
 	}
 	// case: conditions are already initialized
 	if len(currentStatus.Conditions) >= len(expectedStatus.Conditions) && !updateReadyStatus {
```
k15r, line 659: is that even a valid check? if `currentStatus.Conditions  == [a,b,c]` and `expectedStatus.Conditions == [d,e]` that would pass, but clearly expected conditions would not be in currentStatus

how about always processing the conditions and not return early here?

components/eventing-controller/controllers/subscription/beb/reconciler.go
```diff
@@ -852,6 +873,28 @@ func (r *Reconciler) checkStatusActive(subscription *eventingv1alpha1.Subscripti
 	return false, retry, err
 }
 
+// checkLastFailedDelivery checks if LastFailedDelivery exists and if happened after LastSuccessfulDelivery
+func (r *Reconciler) checkLastFailedDelivery(subscription *eventingv1alpha1.Subscription) bool {
+	if len(subscription.Status.EmsSubscriptionStatus.LastFailedDelivery) > 0 {
+		var lastFailedDeliveryTime, LastSuccessfulDeliveryTime time.Time
+		var err error
+		if lastFailedDeliveryTime, err = time.Parse(time.RFC3339, subscription.Status.EmsSubscriptionStatus.LastFailedDelivery); err != nil {
+			r.namedLogger().Errorw("parse LastFailedDelivery failed", "error", err)
```
k15r, line 882: should we really hide this error here and assume everything is alright, by returning false?

components/eventing-controller/api/v1alpha1/subscription_lifecycle.go
```diff
@@ -36,6 +37,7 @@ const (
 	ConditionReasonAPIRuleStatusReady         ConditionReason = "APIRule status ready"
 	ConditionReasonAPIRuleStatusNotReady      ConditionReason = "APIRule status not ready"
 	ConditionReasonNATSSubscriptionActive     ConditionReason = "NATS Subscription active"
+	ConditionReasonSubscriptionWebhookCall    ConditionReason = "BEB Subscription webhook call"
```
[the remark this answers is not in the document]
radufa: changed

components/eventing-controller/controllers/subscription/beb/reconciler.go
```diff
@@ -643,18 +657,25 @@ func (r *Reconciler) syncInitialStatus(ctx context.Context, subscription *eventi
 	}
 	// case: conditions are already initialized
 	if len(currentStatus.Conditions) >= len(expectedStatus.Conditions) && !updateReadyStatus {
-		return nil
+		return false
+	} else if len(currentStatus.Conditions) > 0 {
```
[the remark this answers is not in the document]
radufa: true

components/eventing-controller/controllers/subscription/beb/reconciler.go
radufa: This was the old behavior, I didn't want to change it.
Changed now.

components/eventing-controller/controllers/subscription/beb/reconciler.go
radufa: This error will be reflected in the subscription status too. It's an indicator that we have to update the "Webhook call status" condition to reflect the changes in BEB

k15r (approved):

PR #12568, opened 2021-11-08
Extend kcp js wrapper

voigt:
**Description**

- adding routines for kyma upgrade

**Related issue(s)**

### DIFF
--- a/tests/fast-integration/kcp/client.js
+++ b/tests/fast-integration/kcp/client.js
@@ -1,6 +1,11 @@
 const execa = require("execa");
 const fs = require('fs');
-const {getEnvOrThrow, debug} = require("../utils");
+const {
+    getEnvOrThrow,
+    getEnvOrDefault,
+    debug,
+    wait
+} = require("../utils");
 
 class KCPConfig {
     static fromEnv() {
@@ -12,9 +17,11 @@ class KCPConfig {
             getEnvOrThrow("KCP_TECH_USER_PASSWORD"),
             getEnvOrThrow("KCP_OIDC_CLIENT_ID"),
             getEnvOrThrow("KCP_OIDC_CLIENT_SECRET"),
+            getEnvOrDefault("KCP_MOTHERSHIP_API_URL", "https://mothership-reconciler.cp.dev.kyma.cloud.sap/v1"),
+            getEnvOrDefault("KCP_KUBECONFIG_API_URL", "https://kubeconfig-service.cp.dev.kyma.cloud.sap"),
         );
     }
-    constructor(host, issuerURL, gardenerNamespace, username, password, clientID, clientSecret) {
+    constructor(host, issuerURL, gardenerNamespace, username, password, clientID, clientSecret, motherShipApiUrl, kubeConfigApiUrl) {
         this.host = host;
         this.issuerURL = issuerURL;
         this.gardenerNamespace = gardenerNamespace;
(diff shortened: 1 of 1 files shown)


tests/fast-integration/kcp/client.js
```diff
@@ -12,16 +17,20 @@ class KCPConfig {
             getEnvOrThrow("KCP_TECH_USER_PASSWORD"),
             getEnvOrThrow("KCP_OIDC_CLIENT_ID"),
             getEnvOrThrow("KCP_OIDC_CLIENT_SECRET"),
+            getEnvOrDefault("KCP_MOTHERSHIP_API_URL", "https://mothership-reconciler.cp.dev.kyma.cloud.sap/v1"),
```
voigt, line 20: Required by recent KCP-CLI; I thought it might make sense to fall back on *something*.

tests/fast-integration/kcp/client.js
```diff
@@ -81,13 +95,127 @@ class KCPWrapper {
...
+      }
+
     async exec(args) {
         try {
             const defaultArgs = [
                 `--config`, `${this.kcpConfigPath}`,
             ];
-            let output = await execa(`kcp`, args.concat(defaultArgs));
-            debug(output);
+            // debug([`>  kcp`, defaultArgs.concat(args).join(" ")].join(" "))
```
voigt, line 216: Leaving this here for later debugging...

tests/fast-integration/kcp/client.js
```diff
@@ -81,13 +95,127 @@ class KCPWrapper {
         return await this.exec(args);
     }
 
+    async version() {
+        const args = [`--version`];
+        return await this.exec(args);
+    }
+
+    async upgradeKyma (subaccount, kymaUpgradeVersion) {
+        const args = [`--verbose=6`, `upgrade`, `kyma`, `--version`, `"${kymaUpgradeVersion}"`, `--target`, `subaccount=${subaccount}`];
```
voigt, line 104: Will reduce verbosity later.

Ressetkk: I think the verbosity should be rather omitted and removed prior merging because it might print some sensitive information to the logs like access tokens, unless testing locally.

tests/fast-integration/kcp/client.js
Ressetkk: If uncommented, this call will print the `kcp` command with **all** of the flags used. That's a sensitive information and should be omitted.

tests/fast-integration/kcp/client.js
Ressetkk: I think it should throw an error regardless. For testing purposes we can add the env variable in the `proces.env.ENV_NAME = 'foo'` before invoking the command.

voigt: Fine from my side!

wozniakjan (approved):
/lgtm
/hold

I'm ok merging this, feel free to unhold at your convenience or sync with @Ressetkk

tests/fast-integration/kcp/client.js
voigt: I agree with you. Thats why I uncommented it, though I think it can be useful for debugging locally.

However, we still have the problem you mentioned: think of kcp login failing (for whatever reason) it will print out the entire command, including credentials.

I'm not exactly sure how to solve that, but we should consider this (maybe in another PR).

voigt:
Thanks @wozniakjan,
will merge this to speed up the pipeline task.
I'm in discussion with @Ressetkk offline about the last remark.

PR #12415, opened 2021-10-20
Replaced nats-operator with nats helm chart

mfaizanse:
**Description**

Changes proposed in this pull request:

- Replace nats operator with nats helm-chart
- Deletes nats-doctor container from eventing-controller deployment.
- configuration to enable/disable jetstream. By default its disabled.
- configurable storage backend (memory / file based)
- Added PSP for nats pods
- Removed old nats-operator CRDs in cluster-essentials which were used in the old Kyma installation

**Pre-merge checklist**
- [ ] Reconciler pre-action to delete old nats-operator is merged: https://github.com/kyma-incubator/reconciler/pull/431

**Related issue(s)**

PR #12579, opened 2021-11-09
Remove old nats-operator CRDs which were used in the old Kyma installation

VladislavPaskar:
These files were not used anymore since the new kyma installation were introduced, that's why they need to be removed.


mfaizanse:
I have incorporated these changes in this PR: https://github.com/kyma-project/kyma/pull/12415.

PR #12596, opened 2021-11-11
Bump eventing-controller's image

VladislavPaskar:
update to version 1.17

### DIFF
--- a/components/eventing-controller/go.mod
+++ b/components/eventing-controller/go.mod
@@ -1,6 +1,6 @@
 module github.com/kyma-project/kyma/components/eventing-controller
 
-go 1.16
+go 1.17
 
 require (
 	github.com/avast/retry-go/v3 v3.1.1
--- a/components/eventing-controller/go.mod
+++ b/components/eventing-controller/go.mod
@@ -27,3 +27,58 @@ require (
 	k8s.io/client-go v0.22.0
 	sigs.k8s.io/controller-runtime v0.9.6
 )
+
+require (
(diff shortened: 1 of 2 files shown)


sayanh (approved):

resources/eventing/values.yaml
```diff
@@ -8,7 +8,7 @@ global:
       version: PR-11920
     eventing_controller:
       name: eventing-controller
-      version: PR-12553
+      version: PR-12596
```
k15r, line 11: as this image was not created in the referenced image please use either the PR-Number of the other PR (the one that creates the new image) or if that one is already merged directly refer to the final commit sha of the merging commit of the pr that created the image

components/eventing-controller/go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/kyma-project/kyma/components/eventing-controller
 
-go 1.16
+go 1.17
```
k15r, line 3: please run 
`go mod tidy` and update the pr. switching to go 1.17 semantics introduced require directives for indirect dependencies to be added to the go.mod file

k15r (changes requested):

components/eventing-controller/go.mod
VladislavPaskar: done

resources/eventing/values.yaml
VladislavPaskar: done

k15r (approved):

sayanh (approved):

PR #12611, opened 2021-11-15
Replace the `registry` component

pPrecel:
**Description**

- because the `registry` that we are using in the area of serverless is no more developed ([see here](https://github.com/docker/distribution-library-image)) we decided to change the implementation of the registry to the fully compatible and not outdated [distrubution](https://github.com/distribution/distribution/tree/release/2.7).

**Related issue(s)**

https://github.com/kyma-project/third-party-images/pull/146

### DIFF
--- a/resources/serverless/charts/docker-registry/templates/deployment.yaml
+++ b/resources/serverless/charts/docker-registry/templates/deployment.yaml
@@ -47,7 +47,7 @@ spec:
 {{- end }}
       initContainers:
         - name: generate-htpasswd
-          image: "{{ include "imageurl" (dict "reg" .Values.global.containerRegistry "img" .Values.global.images.k8s_tools) }}"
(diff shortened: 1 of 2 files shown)


johgoe:
@pPrecel  Kyma CLI already installs a registry which comes with k3d. In environments like GCE or Azure a external registry from the cloud. Maybe it's better to just add a little bit extra provisioning to kyma cli and do not install a own registry at all.

pPrecel:
> @pPrecel Kyma CLI already installs a registry which comes with k3d. In environments like GCE or Azure a external registry from the cloud. Maybe it's better to just add a little bit extra provisioning to kyma cli and do not install a own registry at all.

@johgoe thanks for your suggestion. If I understand you correctly your proposition is not a suggestion to my changes in this PR. In this PR I'm only upgrading the image tag. If your proposition is still valid please create a ticket for it

kwiatekus (approved):

PR #12074, opened 2021-09-09
Fix: Do not override host header

szymon-nex:
**Description**

Changes proposed in this pull request:

- Remove overriding  HTTP `host` header in application-gateway proxyfactory

**Related issue(s)**
Fixes #11978

Issue #12617, opened 2021-11-16
CommerceMock tests on Kyma Open Source fail with latest version of application-gateway 


koala7659:
**Description**

After updating the image on `application-gateway` component in `values.yaml` file from version `PR-12181` to the latest one `4a206dab` I noticed that Fast Integration does not pass anymore. 

Additionally on prow following related jobs fail.

- pre-main-kyma-gardener-azure-alpha-eval
- pre-main-kyma-gardener-azure-alpha-prod
- pre-main-kyma-integration-k3d
- pre-main-kyma-integration-k3d-compass-dev

It is highly probably that we introduced regression in this component between versions `PR-12181` and `4a206dab`

Expected result
**All fast integration tests passed.**

Actual result

```
1) CommerceMock tests
       function should be reachable through secured API Rule:
     AssertionError: expected { event: {}, podName: 'lastorder-pl64l-57c6bcb96d-djzn7' } to have nested property 'order.totalPriceWithTax.value'
      at checkFunctionResponse (test/fixtures/commerce-mock/index.js:133:35)
      at processTicksAndRejections (node:internal/process/task_queues:96:5)
      at async Context.<anonymous> (test/1-commerce-mock.js:56:5)
```

koala7659:
Problem most likely caused by code change delivered in https://github.com/kyma-project/kyma/pull/12074/files

koala7659:
PR reversing changes from #12074 seems to work fine: https://github.com/kyma-project/kyma/pull/12625

PR #12625, opened 2021-11-16
Fix of application-gateway host regression

koala7659:
This fix removes host header fix #12074 that caused regression described by  #12617 in application-gateway component 

We will return to the host header problem after releasing Kyma 2.0


koala7659:
The fix seem to be working. Fast Integration Tests pass

Maladie (approved):

Issue #12573, opened 2021-11-09
Compass Runtime Agent should reconcile the mTLS secret in `istio-system`


janmedrek:
**Description**

Right now there is no component responsible for reconciliation of mTLS secret in `istio-system`, in order to provide a stable environment Compass Runtime Agent should handle that.

**Reasons**
We want our components to be fully reconciled.

**Acceptance Criteria**
- mTLS secret in `istio-system` is recreated if removed/modified

akgalwas:
The fix has been implemented. We are in progress of testing the [proposed solution](https://github.com/kyma-project/kyma/pull/12626).

PR #12626, opened 2021-11-16
Compass Runtime Agent should reconcile the mTLS secret in istio-system

akgalwas:
**Description**

- Compass Runtime Agent checks if secret containing CA root certificate exists ; if not it renews certificate to recreate missing secret

**Related issue(s)**
#12573


components/compass-runtime-agent/internal/compassconnection/connector.go
```diff
@@ -26,7 +26,7 @@ const (
 //go:generate mockery --name=Connector
 type Connector interface {
 	EstablishConnection(connectorURL, token string) (EstablishedConnection, error)
-	MaintainConnection(renewCert bool) (*certificates.Credentials, v1alpha1.ManagementInfo, error)
+	MaintainConnection(renewCert bool, credentialsExist bool) (*certificates.Credentials, v1alpha1.ManagementInfo, error)
```
franpog859, line 29: I think we should also regenerate the mock of the `Connector` interface

franpog859 (commented):
So as far as I understand the `MaintainConnection`, now we will renew the certificate and regenerate credentials also if the credentials are not present in the cluster. Is it correct? Sounds ok to me 😄 

I didn't do the manual test yet. Just checked the code

components/compass-runtime-agent/internal/compassconnection/connector.go
koala7659: Done

franpog859 (approved):

franpog859 (approved):

PR #12653, opened 2021-11-19
Extend KO dev environment

radufa:
**Description**

- Add PUBLISHER_IMAGE and PUBLISHER_IMAGE_PULL_POLICY for the KO environment, so that the PUBLISHER_IMAGE can be configured during the development (if KO is used)
 

**Related issue(s)**

### DIFF
--- a/components/eventing-controller/config/manager/manager.yaml
+++ b/components/eventing-controller/config/manager/manager.yaml
@@ -65,8 +65,6 @@ spec:
             value: example.com
           - name: WEBHOOK_TOKEN_ENDPOINT
             value: "https://oauth2.example-domain.com/oauth2/token"
-          - name: PUBLISHER_IMAGE
-            value: "eu.gcr.io/kyma-project/event-publisher-proxy:PR-12616"
           - name: PUBLISHER_IMAGE_PULL_POLICY
             value: Always
           - name: APP_LOG_FORMAT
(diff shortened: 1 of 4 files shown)


components/eventing-controller/config/manager/manager.yaml
```diff
@@ -65,6 +65,10 @@ spec:
             value: example.com
           - name: WEBHOOK_TOKEN_ENDPOINT
             value: "https://oauth2.example-domain.com/oauth2/token"
+          - name: PUBLISHER_IMAGE
+            value: "eu.gcr.io/kyma-project/event-publisher-proxy:PR-12616"
```
k15r, line 69: i know this is will not be easy, but think about a way to:
- either update this whenever a new epp is released (maybe a make target that takes care of it)
- or load automatically the latest image
- or just have ko compile it for you

radufa: Let's do it simple for now and we can have a better solution after the "big refactoring" of eventing-controller.
I've created a new issue to enhance the KO environment: https://github.com/kyma-project/kyma/issues/12666

k15r: The 3rd option is rather simple to implement and prevents developers from deploying an outdated version of the publisher proxy

radufa: Now ko is used to generate the publisher-proxy image from the local sources too. This image will be used by eventing-controller.

k15r (approved):

PR #12686, opened 2021-11-23
Turn off debug mode in skr 1x to 2 upgrade test

ksputo:
**Description**

- ...
- ...
- ...

**Related issue(s)**


wozniakjan (commented):
I think we rely on the `debug()` statements to be able to correlate failures with events from logs in elasticsearch

please make at least this a `console.log()` instead of `debug()`
https://github.com/kyma-project/kyma/blob/f654de606869f8aaba6787800add363d75f73c70/tests/fast-integration/skr-kyma-to-kyma2-upgrade/index.js#L65-L72
there might be other useful information currently also hidden by `debug()`

Issue #12606, opened 2021-11-12
Use a label to distinguish between different publisher proxy types


pxsalehi:
**Description**

Our [current approach](https://github.com/kyma-project/kyma/pull/12312) for figuring out the backend type of the existing publisher proxy in the cluster, relies on fetching the deployment and figuring our from the env vars, etc, what backend type it has. A simpler alternative, could be to add the proper labels to the publisher proxy when it is created by the eventing backend controller, and later simply, query using that label. 

As the publisher proxy deployment is watched by the eventing backend controller, any manual label changes on the deployment will be undone, by the backend controller.

VladislavPaskar:
Should work on that after reconciler gets more stable

radufa:
Will be implemented in two steps, without touching the "component" reconciler:
- first, just label the EEP deployment with the corresponding label
- second (later in a new version) cleanup the code and use only the label to identify (find) the EPP deployment.

VladislavPaskar:
The second part of the comment above is done here: https://github.com/kyma-project/kyma/issues/13656

PR #12706, opened 2021-11-25
Label the eventing-publisher-proxy deployment with the current backendType

VladislavPaskar:
**Description**

To check if the event publisher proxy is ready, we need check it's pods. For that currently we fetch all the possible vent publisher proxy pods and check these pods by comparing the env vars.  This approach was introduced here https://github.com/kyma-project/kyma/pull/12312

Easier would be to label the publisher with its backend type, so they it be fetched and distinguished by this label.

- label the eventing publisher proxy deployment with it's backendType 
- refactor the EPP deployment constructor using the functional options pattern

**Related issue(s)**
https://github.com/kyma-project/kyma/issues/12606


VladislavPaskar:
this PR requires change in reconciler, which cannot be performed before final kyma 2.x release, so it's blocked.

components/eventing-controller/pkg/deployment/publisher_deployment.go
```diff
@@ -22,10 +23,11 @@ const (
...
 	bebNamespacePrefix       = "/"
-	instanceLabelKey         = "app.kubernetes.io/instance"
-	instanceLabelValue       = "eventing"
-	dashboardLabelKey        = "kyma-project.io/dashboard"
-	dashboardLabelValue      = "eventing"
+	InstanceLabelKey         = "app.kubernetes.io/instance"
+	InstanceLabelValue       = "eventing"
+	DashboardLabelKey        = "kyma-project.io/dashboard"
+	DashboardLabelValue      = "eventing"
+	BackendTypeLabelKey      = "backendType"
```
nachtmaar, line 30: Kubernetes uses the prefix `app.kubernetes.io` ([source](https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/)).
|  Labels without a prefix are private to users.

Moreover, Kyma already comes with the `kyma-project.io` prefix. I suggest to either use this one or go one step further: `eventing.kyma-project.io`. What is your opinion on this ?

nachtmaar: Or we use `kyma-project.io/eventingBackend`. backendType is very generic and could mean a lot and be used by other pods in Kyma as well in the future 🤔

VladislavPaskar: I renamed the label name to `kyma-project.io/eventingBackend` as I think it's more suitable

nachtmaar (approved):

nachtmaar (approved):

nachtmaar (approved):

Issue #12461, opened 2021-10-26
Python functions fail to start with OOMKilled status


kwiatekus:
**Description**

Runtime profile XS is to small to run default hello world function.
Make sure that default runtime profile for python functions is set to `M` when running kyma in evaluation profile

**Expected result**

Smallest runtime profile (XS) should provide enough resources to run a  "hello world"- kind of function with no dependencies

**Actual result**

After creating an XS python function it fails with OOMKilled reason

**Steps to reproduce**

 - Deploy kyma 2 ( `kyma deploy` )
 - create couple of python functions with XS runtime profile
 - The function's deployment will fails to deliver a pod that is stable and is not crashing
 
 **Workaround**
Change the runtime profile of the function to M
![Screenshot 2021-11-15 at 12 00 47](https://user-images.githubusercontent.com/6783567/141770728-fe157352-2645-4f81-95b5-9e8544c608e0.png)

PR #12713, opened 2021-11-25
Add ability to add runtime resource presets

dbadura:
**Description**

- Add ability to set default presets per runtime
- Set profile M for python function in evaluation profile
**Related issue(s)**

See https://github.com/kyma-project/kyma/issues/12461

### DIFF
--- a/components/function-controller/pkg/apis/serverless/v1alpha1/function_defaults_test.go
+++ b/components/function-controller/pkg/apis/serverless/v1alpha1/function_defaults_test.go
@@ -2,8 +2,9 @@ package v1alpha1
 
 import (
 	"context"
-	"github.com/stretchr/testify/require"
 	"testing"
+
+	"github.com/stretchr/testify/require"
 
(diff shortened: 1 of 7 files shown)


components/function-controller/pkg/apis/serverless/v1alpha1/function_defaults_test.go
```diff
@@ -18,6 +20,12 @@ func TestSetDefaults(t *testing.T) {
 	zero := int32(0)
 	one := int32(1)
 	two := int32(2)
+
+	functionProfiles := `
+{
+	"python39": "L"
```
m00g3n, line 26: As agreed offline we need to align values with values in sub-chart and chart.

dbadura: I clean up the charts.

m00g3n (approved):

Issue #11768, opened 2021-07-27
Introduce Fast Integration Test for renewing and revoking a certificate


k15r:
**Description**

As part of `fast integration test` an applications certificate needs to be renewed and also be revoked

**Environment**
```OSS```

Disper:
We've removed ```SKR``` environment as we're not trying to test compass.

janmedrek:
we are removing the functionality from Kyma: https://github.com/kyma-project/kyma/issues/13718

PR #12718, opened 2021-11-26
Tests for renewing and revoking an application certificate

mvshao:
**Description**

As part of the `fast integration test`, an applications certificate needs to be renewed and also be revoked.

- adding tests for both cases, revoking and renewing the certificate

**Related issue(s)**
#11768


tests/fast-integration/package-lock.json
```diff
@@ -1,4446 +1,8 @@
 {
   "name": "kyma-js",
   "version": "0.0.1-alpha.12",
-  "lockfileVersion": 2,
+  "lockfileVersion": 1,
```
franpog859, line 4: I think we should stay with the second version of it. Did you try to upgrade `node.js` on your machine?

tests/fast-integration/test/1-commerce-mock.js
```diff
@@ -61,12 +69,26 @@ function commerceMockTests() {
...
+    it("CommerceMock should renew it's certificate", async function () {
+      await renewCommerceMockCertificate();
+    });
+    // call lambda and succeed
+    it("order.created.v1 event should trigger the lastorder function", async function () {
+      await sendEventAndCheckResponse();
+    });
+
+    // revoke certificate
+    it("should revoke Commerce Mock certificate", async function () {
```
franpog859, line 82: To see if revocated cert is really revocated, we should try to renew it. The renovation should fail in this case

tests/fast-integration/test/1-commerce-mock.js
```diff
@@ -61,12 +69,26 @@ function commerceMockTests() {
       await sendEventAndCheckResponse();
     });
 
+    // renew certificate
+    it("CommerceMock should renew it's certificate", async function () {
+      await renewCommerceMockCertificate();
+    });
+    // call lambda and succeed
```
franpog859, line 76: Maybe instead of commenting on the tests, we can enhance their names? For example `the event should trigger the function after the certificate renewal`

tests/fast-integration/test/fixtures/commerce-mock/index.js
```diff
@@ -375,6 +374,66 @@ async function connectCommerceMock(mockHost, tokenData) {
...
+
+async function revokeCommerceMockCertificate(){
+  let {cert, key} = await getCommerceMockCertFiles()
+   
+  const vs = await waitForVirtualService("mocks", "commerce-mock");
+  const mockHost = vs.spec.hosts[0];
+  const url = mockHost.replace(/(commerce.?)/,'');
+  const gateway = `https://gateway.${url}/v1/applications/certificates/revocations`;
+  const httpsAgent = new https.Agent({
+    //rejectUnauthorized: false, // curl -k
```
franpog859, line 418: Should it be there?

franpog859 (commented):
Looking at the jobs I see that the revocation test is not working:
```
Error: Error during revoking Commerce Mock certificate via Kyma connector service: Error: error:140AB18E:SSL routines:SSL_CTX_use_certificate:ca md too weak
    at convertAxiosError (utils/index.js:80:12)
    at revokeCommerceMockCertificate (test/fixtures/commerce-mock/index.js:433:11)
    at processTicksAndRejections (internal/process/task_queues.js:95:5)
    at async Context.<anonymous> (test/1-commerce-mock.js:83:7)
```
Looking into that

tests/fast-integration/test/1-commerce-mock.js
mvshao: When I tried to renew it I got 201 response from Axios, but when we take a look at the test pod log, the error message is included
`Error: Calling Certificate Renewal URL failed with status '403' and body '{\n \"code\": 403,\n \"error\": \"Certificate has been revoked.`

tests/fast-integration/test/fixtures/commerce-mock/index.js
mvshao: Yes, it should be there, because I need to remove the unnecessary `commerce` prefix. And finally, create a new `httpAgent` with the required credentials.

tests/fast-integration/test/1-commerce-mock.js
franpog859: Oh, that's bad :c We should enhance the Varkes response for such case then

tests/fast-integration/test/fixtures/commerce-mock/index.js
franpog859: I mean the comment :D If the comment should be there, I suggest making it more verbose because we may end up not knowing what's that about

tests/fast-integration/test/1-commerce-mock.js
franpog859: [Here](https://github.com/kyma-incubator/varkes/commit/87199ea1b912a4a69a4b5367d49b59e413fd45b8#) is the response part in Varkes. It looks like it doesn't propagate [the error that occurs while the renewal](https://github.com/kyma-incubator/varkes/blob/master/modules/app-connector/src/server/connection.ts#L193-L194)

mvshao:
One  test on the pipelines are failing due to that error 
`error:140AB18E:SSL routines:SSL_CTX_use_certificate:ca md too weak`

I also did a small investigation, and I observed that other tests are failing randomly, but the core of the problem is this same.
`Cannot read property {variableName} of undefined`

For more information check out the details of failing pipelines.

mvshao:
1. That error occurs every time.
2. Example of that error from `pre-main-kyma-integration-k3d` job and `the event should triger the fucntion after revoke of the certificate: Execute tests CommerceMock tests the event should triger the fucntion after revoke of the certificate` test

- `TypeError: Cannot read property 'data' of undefined`

mvshao:
For a couple of days, I have been trying to figure out what is the reason for the
`error:140AB18E:SSL routines:SSL_CTX_use_certificate:ca md too weak`
Here is a description of what was done and what was tested: 

- Extend the certificate and its key to 4096 bits - unsuccessful.

- Change Node.js version to `9.3.0` (in version >=10 they implemented strict SSL policy) - unsuccessful. Node version `9.3.0` doesn't support features already implemented in tests.

- Change the parameter `rejectUnauthorized` of the Axios httpsAgent to `true` or `false` - unsuccessful, the error occurs in both cases.

- Disable the SSL verification on Node.js side by using `process.env.NODE_TLS_REJECT_UNAUTHORIZED = '0'` - unsuccessful

mvshao:
Here I describe differences that I found, between Kyma on Gardener cluster and Kyma set-up by Prow.

_ex. Feature (Gardener / Prow)_

- OpenSSL version (1.1.1 / 1.1.1d)
- SSL security level (not provided / DEFAULT@SECLEVEL=2)

mvshao:
We are removing the certificates from Kyma. The test is irrelevant right now.

PR #12754, opened 2021-12-01
Expose istio /healthz/ready endpoint to public

life0215:
**Description**

- Expose istio /healthz/ready endpoint to public, so it's possible add external monitor for kyma runtime

**Related issue(s)**

### DIFF
--- a/resources/certificates/templates/virtualservice.yaml
+++ b/resources/certificates/templates/virtualservice.yaml
@@ -1,0 +1,19 @@
+apiVersion: networking.istio.io/v1beta1
+kind: VirtualService
+metadata:
+  name: istio-healthz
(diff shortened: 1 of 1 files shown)


strekm (changes requested):
please move this vs to istio-configuration component

cnvergence:
This is a valid point, deploying VS before the gateway would require some time for the Istio to update its state.
If it is not acceptable, I'm okay with leaving this VS in the certificates.

cnvergence (approved):

JackCheng01:
> Merging can be performed automatically once the requested changes are addressed.
Hi, @strekm  Can you help review again?

strekm (approved):

Issue #12743, opened 2021-11-30
Prometheus targets down on upgraded kyma2 cluster


a-thaler:
**Description**
When upgrading a Kyma1 cluster to Kyma 2, there are several prometheus targets being down. It affects all kyma components using mTLS based scraping having leftover peerauthentication resources.

The reasoning: Prometheus is doing the HTTPS call on its own as the sidecar is not intercepting calls. Such calls seem not to work on a permissive peerauthentication, it must be strict in order to accept https calls not initiated by a sidecar.

Grafana as an example workload running with a sidecar had a permissive policy in the past just for the scraping. With the switch to [mTLS](https://github.com/kyma-project/kyma/pull/11994) the peerauthentication resource got removed instead of switched to STRICT (as STRICT is the default).

The upgrade process is not deleting the resource as the reconciler is not supporting it. With that the permissive mode stays.

Solution:
Either assure that the cleanup script immediately cleans the resource after upgrade
Or keep the resource and set it to STRICT till deletion is supported with the reconciler.

Affected workloads:
- monitoring-grafana
- eventing-controller
- eventing-publisher-proxy
- logging-flluent-bit
- logging-loki
- ory-hydra-maester
- ory-oathkeeper-maester
- service-catalog-addons-service-binding-usage-controller

**Expected result**

**Actual result**

**Steps to reproduce**

**Troubleshooting**

PR #12760, opened 2021-12-01
Restore Istio PeerAuthentications (with strict mode)

skhalash:
**Description**

- [This PR](https://github.com/kyma-project/kyma/pull/11994) removes PeerAuthentication resources, which became unneeded after making Prometheus scrape the corresponding targets via HTTPS. However, when testing Kyma 1 to Kyma 2 upgrades using the Reconciler we found out that the TLS-scraped targets are down. The reason for that is the left-over PeerAuthentication resources with permissive mTLS mode set (Reconciler does not remove deleted resources when upgrading Kyma). In order to circumvent this problem, it was decided to restore the PeerAuthentication resources (with strict mode instead of permissive). 

**Related issue(s)**

Fixes https://github.com/kyma-project/kyma/issues/12743

### DIFF
--- a/resources/eventing/charts/controller/templates/peerauthentication.yaml
+++ b/resources/eventing/charts/controller/templates/peerauthentication.yaml
@@ -10,4 +10,4 @@ spec:
     matchLabels: {{- include "controller.peerAuth.selectorLabels" . | nindent 6 }}
   portLevelMtls:
     {{ .Values.metrics.config.port }}:
-      mode: "STRICT"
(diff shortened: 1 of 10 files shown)


k15r (approved):

chrkl (approved):

veichtj (approved):

wozniakjan (approved):
giving blind lgtm, it was explained to me that this is putting istio configuration back that was temporarily disabled.
/lgtm

PR #12766, opened 2021-12-01
disabled ingressgateway autoscaling for evaluation profile

a-thaler:
**Description**

For evaluation profile the HPA config for the ingressgateway was changed to maxReplica 1. Instead, the autoscaling should be simply disabled so that no HPA gets active.

- remove hpaSpec for evaluation profile
- disable autoscaling for evaluation profile
- ...

**Related issue(s)**

### DIFF
--- a/resources/istio-configuration/templates/istio-operator.yaml
+++ b/resources/istio-configuration/templates/istio-operator.yaml
@@ -44,6 +44,7 @@ spec:
     gateways:
       istio-ingressgateway:
         name: istio-ingressgateway
+        autoscaleEnabled: {{ index .Values "helmValues" "gateways" "istio-ingressgateway" "autoscaleEnabled" }}
         podAnnotations:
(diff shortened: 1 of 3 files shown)


resources/istio-configuration/templates/istio-operator.yaml
```diff
@@ -44,6 +44,7 @@ spec:
     gateways:
       istio-ingressgateway:
         name: istio-ingressgateway
+        autoscaleEnabled: {{ index .Values "helmValues" "gateways" "istio-ingressgateway" "autoscaleEnabled" }}
```
veichtj, line 47: Why we are using here the "index" function to determine the autoscale option?

veichtj (approved):
@a-thaler during retest I found out that if you want to downgrade your cluster from production profile to evaluation the hpa is deleted as expected. Nevertheless the ingress pods are not scaled down to one. 
As this is a very rare side case it is ok to go.
cc: @strekm

Issue #12426, opened 2021-10-21
Liveness probe for serverless controller does not signal downtime of reconciliation loop


kwiatekus:
**Description**

Improve liveness probing for serverless controller .

**Expected result**

When go routines crash, liveness probe doen't signals any problems and the reconciliation of functions remains broken

**Actual result**

In long running cluster ( v 1.24.7 ) it happens that go routines are hanged in case there are many git functions
In such case liveness probe should not be up so that kubernetes could restart the controller and reconciliation

**Steps to reproduce**

Install a kyma (1.24.7). Deploy  more than 10 git functions and leave the cluster.
The reconciliation of functions would eventually lead to a controller crash and the kubernetes would not restart it.

**Troubleshooting**

PR #12767, opened 2021-12-01
Add liveness probe implementation for serverless controller

dbadura:
**Description**

- add liveness probe which check if controller didn't stuck

**Related issue(s)**

https://github.com/kyma-project/kyma/issues/12426

### DIFF
--- a/components/function-controller/internal/controllers/serverless/function_reconcile.go
+++ b/components/function-controller/internal/controllers/serverless/function_reconcile.go
@@ -39,19 +39,21 @@ type FunctionReconciler struct {
 	config      FunctionConfig
 	scheme      *runtime.Scheme
 	gitOperator GitOperator
+	healthCh    chan bool
 }
 
-func NewFunction(client resource.Client, log logr.Logger, config FunctionConfig, gitOperator GitOperator, recorder record.EventRecorder) *FunctionReconciler {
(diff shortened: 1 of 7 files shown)


components/function-controller/internal/controllers/serverless/function_reconcile.go
```diff
@@ -81,6 +83,10 @@ func (r *FunctionReconciler) Reconcile(request ctrl.Request) (ctrl.Result, error
 	ctx, cancel := context.WithCancel(context.Background())
```
pPrecel, line 83: Please, consider adding  `context.WithTimeout` instead of `WithCancel`. It will prevent situations when the Controller is  overflowed because it received too many requests that he can't finish

components/function-controller/internal/controllers/serverless/function_reconcile.go
```diff
@@ -81,6 +83,10 @@ func (r *FunctionReconciler) Reconcile(request ctrl.Request) (ctrl.Result, error
...
 	defer cancel()
 
+	if IsHealthCheckRequest(request) {
+		r.healthCh <- true
```
pPrecel, line 87: If you would add `context.WithTimeout` please handle it here to not block routine because of overflowed channel

pPrecel (approved):

Issue #12719, opened 2021-11-26
Istio proxy reset does not restart all resources with sidecars


mjakobczyk:
**Description**

In Istio component Reconciler there is a istio proxy reset logic which automatically performs rollout or delete on resources that have pods with sidecars.

**Expected result**

After Istio upgrade / reconciliation Istio proxy reset should restart all the resources with sidecars.

**Actual result**

Not all the resources are correctly restarted.

`eventing-publisher-proxy` for some reason did not receive the newer sidecar. However, the rollout restart has been performed. Logs:

```text
{"level":"INFO","time":"2021-11-26T10:53:12.489+0100","message":"Rolled out Deployment/kyma-system/eventing-publisher-proxy"}
{"level":"INFO","time":"2021-11-26T10:53:12.489+0100","message":"Waiting for Deployment/kyma-system/eventing-publisher-proxy to be ready"}
```

```text
eventing-publisher-proxy-569b84875d-fnjnf.kyma-system              SYNCED   SYNCED   SYNCED   SYNCED   istiod-756d4969bf-rgvt7   1.11.4
```

The same applies to `tracing-jaeger`.

**Steps to reproduce**

1. Install `istio-configuration` component with Istio in version 1.11.4 using istioctl in version 1.11.4.

2. Upgrade `istio-configuration` component with Istio in version 1.20.0 using istioctl in version 1.11.4.

3. Check logs from the Istio component reconciler:

```logs
{"level":"INFO","time":"2021-11-26T10:53:07.467+0100","message":"Istio has been updated successfully"}
{"level":"INFO","time":"2021-11-26T10:53:07.467+0100","message":"Patching istio-sidecar-injector MutatingWebhookConfiguration..."}
{"level":"INFO","time":"2021-11-26T10:53:07.678+0100","message":"Patch has been applied successfully"}
{"level":"INFO","time":"2021-11-26T10:53:07.845+0100","message":"Retrieved 83 pods total from the cluster"}
{"level":"INFO","time":"2021-11-26T10:53:07.849+0100","message":"Found 30 matching pods"}
{"level":"INFO","time":"2021-11-26T10:53:10.488+0100","message":"Doing rollout for Deployment/kyma-system/service-catalog-catalog-webhook"}
{"level":"INFO","time":"2021-11-26T10:53:10.488+0100","message":"Doing rollout for Deployment/kyma-system/tracing-auth-proxy"}
[...]

PR #12769, opened 2021-12-02
Preserve Eventing Publisher proxy deployment annotations after reconciliation

marcobebway:
**Description**

The Eventing controller does not preserve the Eventing Publisher proxy annotations after reconciliation, which breaks rollout restart for the Eventing Publisher proxy deployment.

This PR introduces a fix for this behaviour by preserving the existing deployment annotations after reconciliation.

> Note: Not all annotations are preserved (e.g. "sidecar.istio.io/inject" is not preserved after reconciliation).

**Related issue(s)**

- https://github.com/kyma-project/kyma/issues/12719.


components/eventing-controller/controllers/backend/reconciler.go
```diff
@@ -640,6 +640,8 @@ func (r *Reconciler) CreateOrUpdatePublisherProxy(ctx context.Context, backend e
 	}
 
 	desiredPublisher.ResourceVersion = currentPublisher.ResourceVersion
+	desiredPublisher.Spec.Template.ObjectMeta.Annotations = currentPublisher.Spec.Template.ObjectMeta.Annotations
```
k15r, line 643: This check means, that we accept any annotation that was put on the template. 
Annotations such as the infamous `sidecar.istio.io/inject` must be set correctly as they could break our functionality.

Lets ignore specific annotations instead of this general approach

marcobebway: Done.

radufa (approved):

PR #12781, opened 2021-12-03
implement mutex in bebmock subscriptions

FriedrichWilken:
adding mutex to the subscriptions and requests of beb mock.


components/eventing-controller/testing/safesubscriptions.go
```diff
@@ -0,0 +1,53 @@
...
+import (
+	"strings"
+	"sync"
+
+	bebtypes "github.com/kyma-project/kyma/components/eventing-controller/pkg/ems/api/events/types"
+)
+
+// SafeSubscriptions encapsulates Subscriptions to provide mutual exclusion.
+type SafeSubscriptions struct {
+	mutex         *sync.RWMutex
```
k15r, line 12: is there a reason to have the RWMutex as a pointer? With the pointer it is possible to have two different SafeSubscriptions to share the same RWMutex. Having a concrete RWMutex in the struct prevents this. The only thing we need to ensure is, that all the methods are pointer receivers.

k15r (approved):

components/eventing-controller/controllers/subscription/beb/reconciler_test.go
```diff
@@ -792,17 +793,17 @@ var _ = Describe("Subscription Reconciliation Tests", func() {
...
 				var bebSubscription bebtypes.Subscription
-				Eventually(func() bool {
-					for r, payloadObject := range beb.Requests {
-						if reconcilertesting.IsBEBSubscriptionCreate(r, *beb.BEBConfig) {
-							bebSubscription = payloadObject.(bebtypes.Subscription)
-							receivedSubscriptionName := bebSubscription.Name
-							// ensure the correct subscription was created
-							return nameMapper.MapSubscriptionName(givenSubscription) == receivedSubscriptionName
-						}
+
```
k15r, line 796: That looks very familiar: see line 418++

k15r (commented):
cant we use https://onsi.github.io/gomega/#containelementelement-interface these matchers in our tests?

k15r (approved):

Issue #12810, opened 2021-12-07
Kyma 2.0 breaks eventing subscriptions


johgoe:
After upgrade from 1.24.8 to 2.0.0 the eventing-controller shows errors

kubectl logs -f -n kyma-system eventing-controller-7869d484b6-gg568

```
{"level":"ERROR","timestamp":"2021-12-07T21:21:06Z","logger":"nats-subscription-reconciler","caller":"nats/reconciler.go:186","message":"sink URL validation failed","context":{"kind":"Subscription","version":3,"namespace":"mynamespace","name":"search-service-recreate","error":"sink does not contain suffix: svc.cluster.local in the URL"}}
```

my subscription looks like

kubectl get subscriptions.eventing.kyma-project.io -o yaml -n mynamespace search-service-recreate 

```
apiVersion: eventing.kyma-project.io/v1alpha1
kind: Subscription
metadata:
  finalizers:
  - eventing.kyma-project.io
  name: search-service-recreate
  namespace: mxspf
spec:
  filter:
    filters:
    - eventSource:
        property: source
        type: exact
        value: ""
      eventType:
        property: type
        type: exact
        value: sap.kyma.custom.searchservice.collection.recreate.v1
  protocol: ""
  protocolsettings: {}
  sink: http://search-service.mynamespace.svc.cluster.local:8080/recreate
status:
  conditions:
  - lastTransitionTime: "2021-12-07T21:21:06Z"
    message: 'sink does not contain suffix: svc.cluster.local in the URL'
    reason: NATS Subscription active
    status: "False"
    type: Subscription active
  emsSubscriptionStatus: {}
  ready: false
```

Follow up:

- [x] https://github.com/kyma-project/kyma/issues/12812.

johgoe:
From my point of view beb is working because it trims the hostname before validation
https://github.com/kyma-project/kyma/blob/15cdcfaa822a730f2b5769c77dd999a36492b9dd/components/eventing-controller/controllers/subscription/beb/reconciler.go#L352

but in nats the trim is not available and the check can't handle ports and paths in the sink url
https://github.com/kyma-project/kyma/blob/15cdcfaa822a730f2b5769c77dd999a36492b9dd/components/eventing-controller/controllers/subscription/nats/reconciler.go#L294

k15r:
Follow up issue: https://github.com/kyma-project/kyma/issues/12812

johgoe:
[...]

PR #12811, opened 2021-12-07
eventing controller reconciler should validate only the host part in …

johgoe:
eventing controller reconciler should validate only the host part in sink url (#12810)


radufa (approved):

k15r:
Hi @johgoe ,

thanks for the quick fix. May i ask you to add a simple test to your PR to test that behaviour?

Issue #12521, opened 2021-11-02
Add serverless e2e test with different git providers


dbadura:
**Description**

It would be nice to have automated test which check if our implementation can authorize against different git providers.

**Reasons**
Every time we bump the `lib-git2` library or change to code, we have to manually test git integration.

** Acceptance Criteria

Create private repos with `Hello World` function in:
- github
- azure devops

The test should build and run `Hello World` function using basic auth, ssh key and bad credentials.
  
**Attachments**

kwiatekus:
The job history shows some problems
```
{"level":"error","msg":"Scenario git-auth-integration not exist","time":"2021-12-30T03:05:52Z"}
```
https://status.build.kyma-project.io/job-history/gs/kyma-prow-logs/logs/serverless-git-auth-integration-k3s-nightly

PR #12824, opened 2021-12-09
Add git-function authentication integration tests

moelsayed:
**Description**

- Add git-function authentication integration tests
- Change test chart to have configurable test suit. 

**Related issue(s)**

https://github.com/kyma-project/kyma/issues/12521


tests/function-controller/testsuite/scenarios/gitauth.go
```diff
@@ -0,0 +1,202 @@
...
+		MaxPollingTime:     cfg.MaxPollingTime,
+		InsecureSkipVerify: cfg.InsecureSkipVerify,
+		Log:                genericContainer.Log,
+		DataKey:            testsuite.TestDataKey,
+	}
+	steps := []step.Step{}
+	for _, testCase := range testCases {
+		testSteps, err := gitAuthFunctionTestSteps(genericContainer, testCase, poll)
+		if err != nil {
+			return nil, err
```
dbadura, line 97: Always `wrap` errors.

tests/function-controller/testsuite/gitops/gitrepository.go
```diff
@@ -7,3 +7,10 @@ func NoAuthRepositorySpec(url string) serverlessv1alpha1.GitRepositorySpec {
 		URL: url,
 	}
 }
+
+func AuthRepositorySpecWithURL(url string, repoAuth *serverlessv1alpha1.RepositoryAuth) serverlessv1alpha1.GitRepositorySpec {
```
dbadura, line 11: `RepositorySpec`? I think it's enough.

tests/function-controller/testsuite/gitops/function.go
```diff
@@ -5,13 +5,16 @@ import (
...
+		baseDir = "/"
+	}
 	return &function.FunctionData{
 		SourceType: serverlessv1alpha1.SourceTypeGit,
 		Body:       repoName,
 		Repository: serverlessv1alpha1.Repository{
-			BaseDir:   "/",
-			Reference: "master",
+			BaseDir:   baseDir,
+			Reference: "main",
```
dbadura, line 17: I think that we can add `reference` as parameter.

dbadura (changes requested):

tests/function-controller/testsuite/scenarios/gitauth.go
```diff
@@ -0,0 +1,202 @@
...
+	runtime            serverlessv1alpha1.Runtime
+	auth               *serverlessv1alpha1.RepositoryAuth
+	authSecretDataFunc func(testRepo) (map[string]string, error)
+}
+
+var testCases []testRepo = []testRepo{
+	{
+		name:             "github-func",
+		provider:         "Github",
+		url:              "git@github.com:moelsayed/pyhello.git",
```
dbadura, line 47: Remember to create repo for our team.

tests/function-controller/testsuite/gitops/gitrepository.go
moelsayed: I am not sure I understand? This is an external type that I am using, I am not defining it.

tests/function-controller/testsuite/scenarios/gitauth.go
```diff
@@ -0,0 +1,202 @@
...
+		return nil, errors.Wrapf(err, "while creating Git config")
+	}
+
+	data, err := tr.authSecretDataFunc(tr)
+	if err != nil {
+		return nil, err
+	}
+
+	return step.NewSerialTestRunner(genericContainer.Log, fmt.Sprintf("%s Function auth test", tr.provider),
+		teststep.CreateSecret(
```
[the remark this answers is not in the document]
moelsayed: Much easier to read, try it ;)

tests/function-controller/testsuite/scenarios/gitauth.go
```diff
@@ -0,0 +1,202 @@
...
+		}
+		if key == azureDevOpsUserEnvName {
+			data["username"] = value
+		}
+	}
+	return data, nil
+}
+
+func githupAuthSecretData(tr testRepo) (map[string]string, error) {
+	privateKey, ok := os.LookupEnv(tr.secretEnvKeys[0])
```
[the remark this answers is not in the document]
moelsayed: That makes perfect sense, I agree. I went with this way for two reasons:
- These configuration are provider specific, not even test case specific, which floating them will require modifying `main` and changing how we run tests every time we need to make changes or add a provider.
- I am trying to make it easy as possible to add a new provider. right now, it's matter of defining a `struct` for the repo and creating a small function to read the secrets. floating this configuration to main will complicate this process.

Also, technically, we do fail quickly, the only thing that happens before this is creating a k8s client. So, while I agree with you conceptually, I think that in this particular case, the current implementation is more suitable for us.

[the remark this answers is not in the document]
dbadura: I don't agree that you have to modify main.
You can create structure for configuration and at the begining of test initialize it with `env` settings using `envconfig.InitWithPrefix(&cfg, prefix)`

tests/function-controller/testsuite/gitops/gitrepository.go
dbadura: I mean, rename it to something like: `CreateGitRepoSpec`

tests/function-controller/testsuite/scenarios/gitauth.go
```diff
@@ -0,0 +1,206 @@
...
+		Verbose:     cfg.Verbose,
+		Log:         logf,
+	}, nil
+}
+func gitAuthFunctionTestSteps(genericContainer shared.Container, tr testRepo, poll poller.Poller) (step.Step, error) {
+	genericContainer.Log.Infof("Testing Git Function in namespace: %s", genericContainer.Namespace)
+
+	secret := secret.NewSecret(tr.auth.SecretName, genericContainer)
+
+	gitCfg, err := gitops.NewGitopsConfig(
```
dbadura, line 132: Do we need to create this?

tests/function-controller/testsuite/scenarios/gitauth.go
```diff
@@ -0,0 +1,206 @@
...
+			data["password"] = value
+		}
+		if key == azureDevOpsUserEnvName {
+			data["username"] = value
+		}
+	}
+	return data, nil
+}
+
+func githubAuthSecretData(tr testRepo) (map[string]string, error) {
```
dbadura, line 197: Is github different that other providers?

I think that we should have two methods to create two types of auth: `BasicAuth` and `SSHAuth` or we can even remove them

dbadura (changes requested):

resources/serverless/charts/k3s-tests/templates/k3s-test.yaml
```diff
@@ -35,8 +35,20 @@ spec:
       containers:
         - name: test
           image: "{{ include "imageurl" (dict "reg" .Values.global.containerRegistry "img" .Values.global.testImages.function_controller_test) }}"
+          {{- if .Values.testSuite }}
```
dbadura, line 38: Do we need this if?

tests/function-controller/testsuite/scenarios/gitauth.go
moelsayed: I added two providers:
- github using ssh key.
- azure devops using basic auth.

I am creating two different functions for the them because:
- content of the data is different based on the auth method.
- environment variables used to pass the credentials.

The idea was to have simple clear structure per provider to make it easy to add new providers later. I will wrap everything in a single function.

tests/function-controller/testsuite/gitops/gitrepository.go
moelsayed: it's consistent with the existing function  https://github.com/kyma-project/kyma/pull/12824/files#diff-5576881242e9ac4029b9741f8459b61cf50c8967d92940ef267b33b7e1f8fe9cR5

tests/function-controller/testsuite/scenarios/gitauth.go
moelsayed: Actually, no :+1:

resources/serverless/charts/k3s-tests/templates/k3s-test.yaml
```diff
@@ -35,8 +35,16 @@ spec:
       containers:
         - name: test
           image: "{{ include "imageurl" (dict "reg" .Values.global.containerRegistry "img" .Values.global.testImages.function_controller_test) }}"
-          args: ["serverless-integration"]
+          args: ["{{ .Values.testSuite }}"]
           env:
+          {{- if ( and (eq .Values.testSuite "git-auth-integration" ) (not (empty .Values.testSuite)) ) }}
+            - name: APP_GITHUB_AUTH_KEY
+              value: "{{ .Values.githubPrivateKey }}"
```
dbadura, line 42: I would use: `.Values.gitAuth.github.privateKey` and apply similar to the azure.

tests/function-controller/testsuite/scenarios/gitauth.go
```diff
@@ -0,0 +1,194 @@
...
+	return map[string]string{"key": auth.Key}
+}
+
+func getAzureDevopsTestcase(secretData map[string]string) testRepo {
+	return testRepo{name: "azure-devops-func",
+		provider:         "AzureDevOps",
+		url:              "https://kyma-wookiee@dev.azure.com/kyma-wookiee/kyma-function/_git/kyma-function",
+		baseDir:          "/code",
+		reference:        "main",
+		expectedResponse: "Hello Serverless",
```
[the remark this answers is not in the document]
moelsayed: That's an existing function example, I didn't create it. I am using it as is.

resources/serverless/charts/k3s-tests/templates/k3s-test.yaml
moelsayed: I think that would just complicate passing the overrides, since we generate them using the test script. I wouldn't do that unless there is really a need.

dbadura (approved):

Issue #12866, opened 2021-12-14
Specify NATS cluster label and cluster name for production profile


radufa:
**Description**
For NATS production profile, the distribution of pods / nodes is controlled by the "pod affinity rule" defined in:
https://github.com/kyma-project/kyma/blob/main/resources/eventing/charts/nats/values.yaml#L144
```
# Affinity for pod assignment
# ref: https://kubernetes.io/docs/concepts/configuration/assign-pod-node/#affinity-and-anti-affinity
affinity:
  podAntiAffinity:
    preferredDuringSchedulingIgnoredDuringExecution:
      - podAffinityTerm:
          labelSelector:
            matchLabels:
              nats_cluster: eventing-nats
          topologyKey: kubernetes.io/hostname
        weight: 100
```
As a consequence, a NATS instance must have the label
```
nats_cluster: eventing-nats
```
defined. 
Also each NATS instance should have the same cluster name "eventing-nats" instead of the host name.
```
[7] 2021/12/14 02:09:22.758355 [INF] Starting nats-server
[7] 2021/12/14 02:09:22.758421 [INF]   Version:  2.6.5
[7] 2021/12/14 02:09:22.758425 [INF]   Git:      [ea48105]
[7] 2021/12/14 02:09:22.758428 [INF]   Name:     eventing-nats-0
[7] 2021/12/14 02:09:22.758432 [INF]   ID:       NBC5TLIVT7I6SGJ76ONC367U5BQ4RVAJ4RHRMCU7M7RRRWZ3QUUDH5IV
[7] 2021/12/14 02:09:22.758436 [INF] Using configuration file: /etc/nats-config/nats.conf
[7] 2021/12/14 02:09:22.759937 [INF] Starting http monitor on 0.0.0.0:8222
[7] 2021/12/14 02:09:22.760025 [INF] Listening for client connections on 0.0.0.0:4222
[7] 2021/12/14 02:09:22.760281 [INF] Server is ready
[7] 2021/12/14 02:09:22.760306 [INF] Cluster name is t8uPxuz226MxE0fODNCLnt
[7] 2021/12/14 02:09:22.760309 [WRN] Cluster name was dynamically generated, consider setting one
[7] 2021/12/14 02:09:22.760335 [INF] Listening for route connections on 0.0.0.0:6222
[...]

Issue #12825, opened 2021-12-09
Event delivery fails on GKE nightly cluster


pbochynski:
**Description**

The in-cluster event test fails on nightly cluster

**Expected result**

Events are delivered.

**Actual result**

See logs from failed jobs: https://status.build.kyma-project.io/job-history/gs/kyma-prow-logs/logs/kyma-gke-nightly-fast-integration

**Steps to reproduce**
```
gcloud auth login
gcloud container clusters get-credentials nightly --region europe-west4 --project sap-kyma-prow-workloads
cd kyma-project/kyma/tests/fast-integration
npm test
```

**Troubleshooting**

The request to the function is logged:
```
Headers: { 'Content-Type': 'application/cloudevents+json', 'X-B3-Sampled': 1 }
Payload: {
  source: '/default/sap.kyma/tunas-prow',
  specversion: '1.0',
  eventtypeversion: 'v1',
  datacontenttype: 'application/json',
  id: 'event-structured-hc3o9',
  type: 'sap.kyma.custom.inapp.order.received.v1',
  data: { id: 'event-structured-hc3o9' }
}
In-app message sent. Publisher proxy response: 204
```
Publisher proxy logs:
```
time="2021-12-09T07:42:52Z" level=info msg="Event received id:[event-structured-hc3o9]"
time="2021-12-09T07:42:52Z" level=info msg="Sending event to NATS, id:[event-structured-hc3o9]"
time="2021-12-09T07:42:52Z" level=info msg="sent id:[event-structured-hc3o9], accepted: true"
time="2021-12-09T07:42:52Z" level=info msg="Event dispatched" duration="65.681µs" id=event-structured-hc3o9 responseBody= source=/default/sap.kyma/tunas-prow statusCode=204 type=sap.kyma.custom.inapp.order.received.v1
```

The eventing controller reports problems with deploying eventing-publisher-proxy. It looks like the latest version of the resource is not fetched before it is applied:
```
[...]

PR #12870, opened 2021-12-15
Fix the horizontal communication between NATS nodes in a NATS cluster (production profile)

radufa:
**Description**

- All NATS nodes in a production cluster should use the same cluster name
- The "podAntiAffinity" should work again

**Related issue(s)**

See also: 
https://github.com/kyma-project/kyma/issues/12866
https://github.com/kyma-project/kyma/issues/12825

### DIFF
--- a/resources/eventing/charts/nats/templates/statefulset.yaml
+++ b/resources/eventing/charts/nats/templates/statefulset.yaml
@@ -33,7 +33,9 @@ spec:
       labels:
         {{- include "nats.selectorLabels" . | nindent 8 }}
         {{- if .Values.statefulSetPodLabels }}
-        {{ toYaml .Values.statefulSetPodLabels | indent 8 }}
+        {{- range $key, $value := .Values.statefulSetPodLabels }}
+        {{ $key }}: {{ $value | quote }}
+        {{- end }}
         {{- end }}
     spec:
 {{- with .Values.securityContext }}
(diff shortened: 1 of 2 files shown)


resources/eventing/charts/nats/values.yaml
```diff
@@ -163,13 +163,16 @@ podAnnotations: {
 statefulSetAnnotations: {}
 
 # Labels to add to the pods of the NATS StatefulSet
-statefulSetPodLabels: {}
+statefulSetPodLabels: {
+  nats_cluster: eventing-nats
+}
```
k15r, line 168:
```suggestion
statefulSetPodLabels:
  nats_cluster: eventing-nats
```

resources/eventing/charts/nats/templates/statefulset.yaml
```diff
@@ -33,7 +33,9 @@ spec:
       labels:
         {{- include "nats.selectorLabels" . | nindent 8 }}
         {{- if .Values.statefulSetPodLabels }}
-        {{ toYaml .Values.statefulSetPodLabels | indent 8 }}
+        {{- range $key, $value := .Values.statefulSetPodLabels }}
+        {{ $key }}: {{ $value | quote }}
+        {{- end }}
```
k15r, line 38: is there no easier version to achieve the same, without manually quoting and iterating over the list of Labels?

resources/eventing/charts/nats/values.yaml
radufa: Done

resources/eventing/charts/nats/templates/statefulset.yaml
radufa: Done

k15r (approved):

PR #12872, opened 2021-12-15
Introduce priority class for DaemonSets (#12751)

### LINKED ISSUE #12720: Installation fails for fluent-bit daemonset as CPU is insufficient
**Description**
When starting the installation on a kubernetes cluster with a minWorkerNodes=1, all the components are getting installed on that one initial node. When it fills, kubernetes will spawn a second node where the rest of the components gets applied after a while.

If the logging component with the fluent-bit daemonset gets applied in a late stage, that first initial node is very full with workload and there is no CPU left to deploy a pod there. It will stay in pending state with the message "Insufficient CPU" and the installation times out. That situation do not recover on it's own.

**Expected result**
There are enough resources available, just the workload is distributed in a bad way. I expect that the installation succeeds and fluent-bit gets resources to run a pod on every node.

**Actual result**
One one node fluent-bit cannot start as other workloading is not rescheduling to different nodes.

**Steps to reproduce**
Install kyma on a cluster with initial one node only

**Troubleshooting**
Delete random pods manual from the initial node. They will get re-scheduled indeed on other nodes and fluent-bit can continue.

chrkl:
Description

Since the installation of Kyma components do not have a particular order, nodes might end up in the state where no sufficient resources for DaemonSet pods are left. This PR increases the priority of DaemonSet pods to get other pods evicted if necessary. Having a cluster-autoscaler, this would cause scaling up the nodes and the installation should be fine.

Introduce dedicated PriorityClass with higher priority for DaemonSets in kyma-system namespace
Related issue(s)
Fixes #12720

### DIFF
--- a/resources/cluster-essentials/templates/priorityclass-high.yaml
+++ b/resources/cluster-essentials/templates/priorityclass-high.yaml
@@ -1,0 +1,9 @@
+{{- if .Values.global.highPriorityClassName }}
+apiVersion: scheduling.k8s.io/v1
+kind: PriorityClass
+metadata:
+  name: {{ .Values.global.highPriorityClassName }}
+value: {{ .Values.global.highPriorityClassValue }}
+globalDefault: false
+description: "Global scheduling priority of Kyma DaemonSet components. Must not be blocked by unschedulable non-daemonset workloads."
+{{- end }}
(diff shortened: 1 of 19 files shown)


skhalash (approved):

wozniakjan (approved):
/lgtm

mfaizanse (approved):

mjakobczyk (approved):

a-thaler (approved):

Issue #12849, opened 2021-12-13
Serverless Benchmarking


kwiatekus:
**Description**

We want to gain better visibility in serverless performance and be able to detect positive/negative trends in serverless qualities.

We want to keep track of 2 different  flavours of performance (lifecycle, individual functions). 

 - [x] measure time between function creation time until pod ready (prio 1) - done [here](https://github.com/kyma-project/kyma/pull/12914)
 - [x] [individual runtimes](https://github.com/kyma-project/kyma/issues/12849#issuecomment-1019885908) (python, nodejs) - could be simple as collecting locust reports (prio 2)

**Execution plan**

 - run periodically on nightly:
   - deploy test fixtures (individual functions, function orchestrated in a chain) & apply load - run test suites jobs
 - collect average metrics after each test run
   -  using kyma observability tools and optionally locust reports
 - store resulting data to google cloud big query:
   - time to start and become ready
   - response time
   - requests per second vs response per second
   - error rate (200 vs 500)
 - plot the data on dashboards ( re-use grafana setup from test-infra )
 - alert that will inform us that performance is trending down (?) 

**AC**

 - runtime performance test should be independant on the runtime itself so that we get the same set of data for various types of runtimes and be able to compare them

**Reasons**

We want to measure if serverless refactoring is improving performance or causing performance losses

**Attachments**

**Related Issues**
https://github.com/kyma-project/kyma/issues/10154

---
<details>
    <summary>:warning: Funnel process - do not remove!</summary>
    </br>

Name|Value
:--- | :---
theme | Technical Debt
business value | 5
technical value | 8
user value | 2
open-source value | 5
effort | 
requested by | 
</details>

kwiatekus:
We have 2 flavours of performance (lifecycle vs runtime). We probably have to separate those:

Runtime
 - individual runtimes (python, nodejs) - could be simple as collecting locust raports
 - chain of functions - focus on workflow aspect - how long from initial trigger until the finish
 
 Function controller (lifecycle)
  - measure time between function creation time until pod ready

Phases (all automated):
1) Deploy cluster (nightly)
2) Deploy Test Fixtures (functions)
3) Export Data (i.e locust raport & prometheus metrics)
4) Storing in S3 bucket

Consumption (human)
App that fetches the time records
1) present the charts

dbadura:
Shouldn't we also gather metrics for function updates?

kwiatekus:
Phase 2 PRs - benchmarking individual runtimes
[...]

PR #12914, opened 2021-12-21
Add functions startup metrics

moelsayed:
**Description**

- Add functions startup metrics
- Add serverless metrics dashboard

**Related issue(s)**

https://github.com/kyma-project/kyma/issues/12849


pPrecel (approved):
LGTM

dbadura (changes requested):
Bump image

kwiatekus:
Screenshot 2022-01-04 at 16 01 30


I wonder if the user is actually able to observe changes in a given time span ( say 1 week).
If yes, maybe a time based chart would be better than gauges.

But on the other hand, if the values stay constant, then gauges would serve it purpose better. 
Its a metric that is rather observed in a long time context ( by us, not users ) taking into account serverless implementation changes

kwiatekus (approved):

PR #12918, opened 2021-12-21
Fix NATS servers discovery

marcobebway:
**Description**

Fix NATS servers discovery by making the NATS service headless.

> This is a revert of NATS service changes done here 9a289bc4cd115e0a2ae81853a6e9baa9401212aa.

**Related issue(s)**

- https://github.com/kyma-project/kyma/issues/12825


VladislavPaskar (approved):

marcobebway:
Tested this change with Kyma CLI and it does not break when upgrading the Eventing components from release **2.0.0** to **PR-12918**.

```
$ kyma version

Kyma CLI version: stable-4-g48faef7
```

PR #12919, opened 2021-12-21
Increase memory limits for a few monitoring components

skhalash:
**Description**

We found out that `kube-state-metrics` and `prometheus-operator` containers were in the CrashLoopBackOff state, constantly OOMKilled. The memory consumption actually grew over the limits upon application startup and then stabilized. As a workaround, we will increase memory limits for those 2 containers.

- Increase memory limits for `kube-state-metrics` and `prometheus-operator`

**Related issue(s)**
https://github.com/kyma-incubator/reconciler/issues/592

### DIFF
--- a/resources/monitoring/values.yaml
+++ b/resources/monitoring/values.yaml
@@ -1408,7 +1408,7 @@ prometheusOperator:
   resources:
     limits:
       cpu: 300m
-      memory: 500Mi
+      memory: 1Gi
     requests:
       cpu: 10m
       memory: 90Mi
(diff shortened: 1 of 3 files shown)


resources/monitoring/values.yaml
```diff
@@ -1408,7 +1408,7 @@ prometheusOperator:
   resources:
     limits:
       cpu: 300m
-      memory: 500Mi
+      memory: 1Gi
```
a-thaler, line 1411: If you do the big increase here, you can remove it from production profile

skhalash: Makes sense, will do

a-thaler (approved):

PR #12951, opened 2021-12-29
 Add tests for valid sink with path (#12810)

johgoe:
Add a new test case to avoid the case of #12810


FriedrichWilken (approved):
Looks good. Thank's for your contribution, @johgoe.

PR #12952, opened 2021-12-30
Add a troubleshooting guide about a misconfigured resource

### LINKED ISSUE #12950: Troubleshooting guide - unreachable resources
**Description**

Provide a general troubleshooting guide for Kyma advising users to delete an unreachable resource to have it reinstalled by the reconciler.

Details: Generally, all resources in `kyma-system`, `istio-system`, and other cluster-wide resources with the `reconciler.kyma-project.io/managed-by-reconciler-disclaimer` annotation should be managed by the reconciler. However, the reconciler does not support all use cases. When a user modifies a resource, breaks something, and then the resource is not reachable because the reconciler cannot help, the workaround is to delete the resource. After deleting it, the reconciler reinstalls the resource. 

**Reasons**

To have accurate troubleshooting for Kyma users.

**Assignees**

@kyma-project/technical-writers

mmitoraj:
**Description**

- Added a  troubleshooting guide about a misconfigured resource

**Related issue(s)**
Resolves #12950

### DIFF
--- a/docs/04-operation-guides/troubleshooting/01-resources-misconfigured.md
+++ b/docs/04-operation-guides/troubleshooting/01-resources-misconfigured.md
@@ -6,7 +6,7 @@ All resources in the Kyma-native Namespaces with the `reconciler.kyma-project.io
 
 ## Symptom
 
-A resource is not working as expected. It could be either invalid, unavailable, etc.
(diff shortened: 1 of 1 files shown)


strekm (approved):
/meow

klaudiagrz (approved):

PR #12997, opened 2022-01-06
Retry reconciling subscriptions on NATS disconnect

### LINKED ISSUE #12930: Event delivery for existing subscriptions is broken when NATS server is restarted
**Description**

Event delivery for existing subscriptions is broken when NATS server is restarted.

**Expected result**

Event delivery for existing subscriptions should continue to be working after NATS server is restarted.

**Steps to reproduce**

1. Scale down the NATS StatefulSet replicas to one (or use the Eventing [evaluation](https://github.com/kyma-project/kyma/blob/8238d01397fa9abb08d0ff0913f620676b83f795/resources/eventing/profile-evaluation.yaml#L21) profile).
2. Create Kyma subscriptions and their corresponding subscriber(s).
3. Publish Events.
4. Check that Events are reaching the subscriber(s).
5. Delete the NATS server pod.
6. Publish Events.
7. The subscriber(s) will not receive Events from already existing subscriptions.

**Troubleshooting**

When the NATS server pod is restarted, the Eventing controller does not recreate the NATS subscriptions on the new NATS server instance.

pxsalehi:
**Description**

If the NATS server is down and the NATS subscription reconciler exceeds the retry
attempts configured, the Kyma subscription would still be reported as ready, even
though no messages could be delivered. The PR fixes this by forcing reconciliation
of the Kyma subscriptions after reconnection attempts have been exhausted. This
would move the subscriptions to a not ready state, which would cause periodic reconciliation
of the subscriptions until NATS is available again.

**Related issue(s)**

Closes #12930

### DIFF
--- a/components/eventing-controller/controllers/subscription/nats/reconciler.go
+++ b/components/eventing-controller/controllers/subscription/nats/reconciler.go
@@ -8,6 +8,8 @@ import (
 	"reflect"
 	"strings"
 
+	"github.com/nats-io/nats.go"
+
 	"github.com/kyma-project/kyma/components/eventing-controller/controllers/events"
 
 	k8serrors "k8s.io/apimachinery/pkg/api/errors"
@@ -49,6 +51,8 @@ type Reconciler struct {
 	recorder         record.EventRecorder
 	eventTypeCleaner eventtype.Cleaner
 	sinkValidator
+	// This channel is used to enqueue a reconciliation request for a subscription
+	customEventsChannel chan event.GenericEvent
 }
 
 var (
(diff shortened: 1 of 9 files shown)


components/eventing-controller/controllers/subscription/nats/reconciler.go
```diff
@@ -49,43 +51,65 @@ type Reconciler struct {
...
+	subs, err := r.getAllKymaSubscriptions(context.Background())
+	if err != nil {
+		// NATS reconnect attempts are exceeded, and we cannot reconcile subscriptions! If we ignore this,
+		// there will be no future chance to retry connecting to NATS!
+		panic(err)
 	}
+	r.enqueueReconciliationForSubscriptions(subs)
 }
 
-func (r *Reconciler) SetupWithManager(mgr ctrl.Manager) error {
```
mfaizanse, line 85: Is there any reason to remove this method? Isn't it currently referenced in [here](https://github.com/kyma-project/kyma/blob/9aaa0ee225fa06d62b5c58ce17093156c15374fe/components/central-application-connectivity-validator/cmd/centralapplicationconnectivityvalidator/centralapplicationconnectivityvalidator.go#L136)?

pxsalehi: I hope not! :-) Otherwise, it shouldn't build! This was not used. I think it is just left over from before, when the subscription controller were used separately. This is int he NATS reconciler file. What you referenced is returned from a `NewController` call, I think.

mfaizanse (approved):

PR #13000, opened 2022-01-07
Create loki-stats-dashboard

Pranav-SA:
**Description**

- Adds a new statistics loki-dashboard to observe pod/application-wise log generation.

**Related issue(s)**
kyma#12728


skhalash:
Comments regarding the dashboard:
* As far as I remember we agreed to drop the cluster-wide error rate counter from the Loki dashboard, however, I can still see it.
* The volume usage pane seems broken, could you take a look?
* Would it be possible to use `bytes_rate` or `bytes_over_time ` instead of `count_over_time`? The amount of data per second is more important than the number of lines/second.
* What makes chunks flushed/sec so important that we have it on the first row of the Loki dashboard. How can we use this information?

skhalash (approved):

PR #12691, opened 2021-11-24
Disable istio for SC webhook

piotrmiskiewicz:
**Description**

Changes proposed in this pull request:
- disable Service Catalog webhook istio injection

**Related issue(s)**

PR #13017, opened 2022-01-10
Disable istio for service catalog webhook

wozniakjan:
**Description**

This is a followup to https://github.com/kyma-project/kyma/pull/12691, it seems to have been rendered without the annotation.


ksputo (approved):
lgtm

PR #13046, opened 2022-01-12
Fix skr-svcat-migration-test updateSKR() call

szwedm:
**Description**

- replace runtimeName with skr.shoot.name in updateSKR() args

**Related issue(s)**


wozniakjan (approved):
/lgtm

PR #13050, opened 2022-01-12
Avoid the deletion of NATS-subscription (on NATS server) at each reconciliation call

mfaizanse:
**Description**

- ...
- ...
- ...

**Related issue(s)**


radufa (changes requested):

components/eventing-controller/pkg/handlers/nats.go
```diff
@@ -125,8 +128,32 @@ func (n *Nats) SyncSubscription(sub *eventingv1alpha1.Subscription, cleaner even
 		cleanSubjects = append(cleanSubjects, subject)
 	}
 
+	// if subscription filters are modified, then delete subscriptions from NATS
+	// so that new subscriptions are created accordingly
+	if len(sub.Status.CleanEventTypes) != len(cleanSubjects) ||
+		!reflect.DeepEqual(sub.Status.CleanEventTypes, cleanSubjects) {
```
k15r, line 134: do we care for order here? Imho we should not recreate the subscriptions just because the order changed

mfaizanse: It will be covered in the follow up [issue](https://github.com/kyma-project/kyma/issues/13108)

components/eventing-controller/pkg/handlers/nats.go
```diff
@@ -125,8 +128,32 @@ func (n *Nats) SyncSubscription(sub *eventingv1alpha1.Subscription, cleaner even
...
+
+		log.Infow(
+			"deleting subscriptions on NATS because subscription filters are modified",
+			"oldSubjects", sub.Status.CleanEventTypes,
+			"newSubjects", cleanSubjects,
+		)
+
```
radufa, line 141: At the creation of a new subscription this log message appears. Perhaps using it on debug log level with a changed message could be better:
```
{"level":"INFO","timestamp":"2022-01-19T14:24:56Z","logger":"event-type-cleaner","caller":"eventtype/clean.go:67","message":"clean event-type","context":{"prefix":"sap.kyma.custom","type":"sap.kyma.custom.noapp.order.created.v1","before":"sap.kyma.custom.noapp.order.created.v1","after":"sap.kyma.custom.noapp.order.created.v1"}}
{"level":"INFO","timestamp":"2022-01-19T14:24:56Z","logger":"nats-handler","caller":"handlers/nats.go:136","message":"deleting subscriptions on NATS because subscription filters are modified","context":{"kind":"Subscription","version":2,"namespace":"test","name":"test1","oldSubjects":[],"newSubjects":["sap.kyma.custom.noapp.order.created.v1"]}}
{"level":"INFO","timestamp":"2022-01-19T14:24:56Z","logger":"event-type-cleaner","caller":"eventtype/clean.go:67","message":"clean event-type","context":{"prefix":"sap.kyma.custom","type":"sap.kyma.custom.noapp.order.created.v1","before":"sap.kyma.custom.noapp.order.created.v1","after":"sap.kyma.custom.noapp.order.created.v1"}}
```

mfaizanse: updated

radufa (approved):

PR #10251, opened 2020-12-22
NATS Subscription Controller

montaro:
Description

Changes proposed in this pull request:

- Add a controller which should create eventing infrastructure in Nats for `subscriptions.eventing.kyma-project.io`

Related issue(s)

See https://github.com/kyma-project/kyma/issues/9862

Issue #9766, opened 2020-10-26
Optimize number of reconciliation calls in eventing-controller


nachtmaar:
Currently the number of reconciliations for one Subscription is very high because the status of the Subscription is changed multiple times (and written to etcd). An alternative to this is to only update the status at the end of the reconciliation. 

Please ensure that status changes are not lost (e.g. if the end of the reconcile method is never reached).

This approach will also help us to separate in the future the "K8S controlling" part, which will remain inside of eventing-controller and the "subscriptions manager (first BEB and later NATS)" which  will be moved outside of eventing-controller.

**Remarks**
-  [ ] check if code improvements can be implemented as well, or create a new ticket for such changes

Target: 
- aim for 1 update for subscription plus one for the status subresource

**Acceptance**
- [x] #13025
- [x] #13026

radufa:
Refactoring the code should also try to use the K8S library instead of an external one (hashstructure package) to calculate the deep hash value, see:
Kubenetes controller https://github.com/kubernetes/kubernetes/blob/master/pkg/controller/controller_utils.go#L1130-L1145.

berkanercan:
1. Optimize calls to api server
2. Optimize the number of request to BEB (has higher prio)

k15r:
estimate is just for the BEB part

k15r:
- [ ] https://github.com/kyma-project/kyma/issues/12032
- [ ] https://github.com/kyma-project/kyma/issues/11988

radufa:
Here is a comment which should be also solved: https://github.com/kyma-project/kyma/pull/13070/files#r784688541

radufa:
For NATS controller, the PR: https://github.com/kyma-project/kyma/pull/13070 optimizes the reconcilations calls too, r3ducing them at the minimum.

PR #13070, opened 2022-01-14
Fix endless reconciliation calls

radufa:
**Description**

- Fix endless reconciliation calls if the controller runs under "heavy load"
- Cleanup the code referring to subscription status update
- Reduce the numbers of reconciliation calls
- Shutdown NATS embedded test server in all test functions

**Related issue(s)**

### DIFF
--- a/components/eventing-controller/controllers/subscription/nats/reconciler.go
+++ b/components/eventing-controller/controllers/subscription/nats/reconciler.go
@@ -241,18 +241,6 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 		return ctrl.Result{}, nil
 	}
 
-	// Clean up the old subscriptions
-	if err := r.Backend.DeleteSubscription(subscription); err != nil {
-		log.Errorw("delete subscription failed", "error", err)
-		if err := r.syncSubscriptionStatus(ctx, subscription, false, false, err.Error()); err != nil {
-			if k8serrors.IsConflict(err) {
-				return ctrl.Result{Requeue: true}, nil
-			}
-			return ctrl.Result{}, err
-		}
-		return ctrl.Result{}, err
-	}
-
 	// Synchronize Kyma subscription to NATS backend
 	subscriptionStatusChanged, err := r.Backend.SyncSubscription(subscription, r.eventTypeCleaner)
 	if err != nil {
--- a/components/eventing-controller/controllers/subscription/nats/reconciler.go
+++ b/components/eventing-controller/controllers/subscription/nats/reconciler.go
@@ -5,7 +5,6 @@ import (
 	"fmt"
 	"net/url"
 	"os"
-	"reflect"
 	"strings"
 
 	"github.com/nats-io/nats.go"
@@ -173,34 +172,24 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 
 	r.namedLogger().Debugw("received subscription reconciliation request", "namespace", req.Namespace, "name", req.Name)
 
-	actualSubscription := &eventingv1alpha1.Subscription{}
-	result := ctrl.Result{}
+	cachedSubscription := &eventingv1alpha1.Subscription{}
 
 	// Ensure the object was not deleted in the meantime
-	err := r.Client.Get(ctx, req.NamespacedName, actualSubscription)
+	err := r.Client.Get(ctx, req.NamespacedName, cachedSubscription)
 	if err != nil {
 		return ctrl.Result{}, client.IgnoreNotFound(err)
 	}
 
 	// Handle only the new subscription
-	desiredSubscription := actualSubscription.DeepCopy()
+	subscription := cachedSubscription.DeepCopy()
 
 	// Bind fields to logger
-	log := utils.LoggerWithSubscription(r.namedLogger(), desiredSubscription)
+	log := utils.LoggerWithSubscription(r.namedLogger(), subscription)
 
-	if !desiredSubscription.ObjectMeta.DeletionTimestamp.IsZero() {
-		// Clean up the subscriptions from NATS
-		if err := r.Backend.DeleteSubscription(desiredSubscription); err != nil {
-			log.Errorw("delete subscription failed", "error", err)
-			if err := r.syncSubscriptionStatus(ctx, actualSubscription, false, err.Error()); err != nil {
-				return ctrl.Result{}, err
-			}
-			return ctrl.Result{}, err
-		}
-
+	if !subscription.ObjectMeta.DeletionTimestamp.IsZero() {
 		// The object is being deleted
-		if utils.ContainsString(desiredSubscription.ObjectMeta.Finalizers, Finalizer) {
-			if err := r.Backend.DeleteSubscription(desiredSubscription); err != nil {
+		if utils.ContainsString(subscription.ObjectMeta.Finalizers, Finalizer) {
+			if err := r.Backend.DeleteSubscription(subscription); err != nil {
 				log.Errorw("delete subscription failed", "error", err)
 				// if failed to delete the external dependency here, return with error
 				// so that it can be retried
@@ -208,48 +197,70 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 			}
 
 			// remove our finalizer from the list and update it.
-			desiredSubscription.ObjectMeta.Finalizers = utils.RemoveString(desiredSubscription.ObjectMeta.Finalizers, Finalizer)
-			if err := r.Client.Update(ctx, desiredSubscription); err != nil {
-				events.Warn(r.recorder, desiredSubscription, events.ReasonUpdateFailed, "Update Subscription failed %s", desiredSubscription.Name)
+			subscription.ObjectMeta.Finalizers = utils.RemoveString(subscription.ObjectMeta.Finalizers, Finalizer)
+			if err := r.Client.Update(ctx, subscription); err != nil {
+				events.Warn(r.recorder, subscription, events.ReasonUpdateFailed, "Update Subscription failed %s", subscription.Name)
 				log.Errorw("remove finalizer from subscription failed", "error", err)
+				if k8serrors.IsConflict(err) {
+					return ctrl.Result{Requeue: true}, nil
+				}
 				return ctrl.Result{}, err
 			}
 			log.Debug("remove finalizer from subscription succeeded")
-			return ctrl.Result{}, nil
 		}
-	}
-
-	// Check for valid sink
-	if err := r.sinkValidator(ctx, r, desiredSubscription); err != nil {
-		log.Errorw("sink URL validation failed", "error", err)
-		if err := r.syncSubscriptionStatus(ctx, actualSubscription, false, err.Error()); err != nil {
-			return ctrl.Result{}, err
-		}
-		// No point in reconciling as the sink is invalid
+		// Stop reconciliation as the object is being deleted
 		return ctrl.Result{}, nil
 	}
 
 	// The object is not being deleted, so if it does not have our finalizer,
 	// then lets add the finalizer and update the object. This is equivalent
 	// registering our finalizer.
-	if !utils.ContainsString(desiredSubscription.ObjectMeta.Finalizers, Finalizer) {
-		desiredSubscription.ObjectMeta.Finalizers = append(desiredSubscription.ObjectMeta.Finalizers, Finalizer)
-		if err := r.Update(context.Background(), desiredSubscription); err != nil {
+	if !utils.ContainsString(subscription.ObjectMeta.Finalizers, Finalizer) {
+		subscription.ObjectMeta.Finalizers = append(subscription.ObjectMeta.Finalizers, Finalizer)
+		if err := r.Update(context.Background(), subscription); err != nil {
 			log.Errorw("add finalizer to subscription failed", "error", err)
+			if k8serrors.IsConflict(err) {
+				return ctrl.Result{Requeue: true}, nil
+			}
 			return ctrl.Result{}, err
 		}
 		log.Debug("add finalizer to subscription succeeded")
-		result.Requeue = true
+		return ctrl.Result{Requeue: true}, nil
 	}
 
-	if result.Requeue {
-		return result, nil
+	// Check for valid sink
+	if err := r.sinkValidator(ctx, r, subscription); err != nil {
+		log.Errorw("sink URL validation failed", "error", err)
+		if err := r.syncSubscriptionStatus(ctx, subscription, false, false, err.Error()); err != nil {
+			if k8serrors.IsConflict(err) {
+				return ctrl.Result{Requeue: true}, nil
+			}
+			return ctrl.Result{}, err
+		}
+		// No point in reconciling as the sink is invalid
+		return ctrl.Result{}, nil
 	}
 
-	_, err = r.Backend.SyncSubscription(desiredSubscription, r.eventTypeCleaner)
+	// Clean up the old subscriptions
+	if err := r.Backend.DeleteSubscription(subscription); err != nil {
+		log.Errorw("delete subscription failed", "error", err)
+		if err := r.syncSubscriptionStatus(ctx, subscription, false, false, err.Error()); err != nil {
+			if k8serrors.IsConflict(err) {
+				return ctrl.Result{Requeue: true}, nil
+			}
+			return ctrl.Result{}, err
+		}
+		return ctrl.Result{}, err
+	}
+
+	// Synchronize Kyma subscription to NATS backend
+	subscriptionStatusChanged, err := r.Backend.SyncSubscription(subscription, r.eventTypeCleaner)
 	if err != nil {
 		log.Errorw("sync subscription failed", "error", err)
-		if err := r.syncSubscriptionStatus(ctx, desiredSubscription, false, err.Error()); err != nil {
+		if err := r.syncSubscriptionStatus(ctx, subscription, false, false, err.Error()); err != nil {
+			if k8serrors.IsConflict(err) {
+				return ctrl.Result{Requeue: true}, nil
+			}
 			return ctrl.Result{}, err
 		}
 		return ctrl.Result{}, err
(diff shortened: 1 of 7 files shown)


components/eventing-controller/controllers/subscription/nats/reconciler.go
```diff
@@ -144,135 +143,156 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
...
 			return ctrl.Result{}, nil
 		}
 	}
 
+	// The object is not being deleted, so if it does not have our finalizer,
+	// then lets add the finalizer and update the object. This is equivalent
+	// registering our finalizer.
+	if !utils.ContainsString(subscription.ObjectMeta.Finalizers, Finalizer) {
+		subscription.ObjectMeta.Finalizers = append(subscription.ObjectMeta.Finalizers, Finalizer)
+		if err := r.Update(context.Background(), subscription); err != nil {
```
k15r, line 189: is it worth introducing yet another update just to set the finalizer? We could continue reconciling the resource and just add the finalizer in the final object.

radufa: It will be solved in the more general context, I've added it: https://github.com/kyma-project/kyma/issues/9766#issuecomment-1013015225
The idea of this PR was to be similar with https://github.com/kyma-project/kyma/pull/13001 and not to diverge too much

components/eventing-controller/controllers/subscription/nats/reconciler.go
```diff
@@ -173,135 +172,156 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
...
 	// Handle only the new subscription
-	desiredSubscription := actualSubscription.DeepCopy()
+	subscription := cachedSubscription.DeepCopy()
 
 	// Bind fields to logger
-	log := utils.LoggerWithSubscription(r.namedLogger(), desiredSubscription)
+	log := utils.LoggerWithSubscription(r.namedLogger(), subscription)
 
-	if !desiredSubscription.ObjectMeta.DeletionTimestamp.IsZero() {
+	if !subscription.ObjectMeta.DeletionTimestamp.IsZero() {
```
marcobebway, line 189: I think we are missing a return statement before this code block ends.

components/eventing-controller/controllers/subscription/nats/reconciler.go
```diff
@@ -173,135 +172,156 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
...
 			// remove our finalizer from the list and update it.
-			desiredSubscription.ObjectMeta.Finalizers = utils.RemoveString(desiredSubscription.ObjectMeta.Finalizers, Finalizer)
-			if err := r.Client.Update(ctx, desiredSubscription); err != nil {
-				events.Warn(r.recorder, desiredSubscription, events.ReasonUpdateFailed, "Update Subscription failed %s", desiredSubscription.Name)
+			subscription.ObjectMeta.Finalizers = utils.RemoveString(subscription.ObjectMeta.Finalizers, Finalizer)
+			if err := r.Client.Update(ctx, subscription); err != nil {
+				events.Warn(r.recorder, subscription, events.ReasonUpdateFailed, "Update Subscription failed %s", subscription.Name)
 				log.Errorw("remove finalizer from subscription failed", "error", err)
-				return ctrl.Result{}, err
+				if k8serrors.IsConflict(err) {
```
marcobebway, line 204: What about other errors? why do we ignore them?

components/eventing-controller/controllers/subscription/nats/reconciler.go
```diff
@@ -173,135 +172,156 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
...
+					return ctrl.Result{Requeue: true}, nil
+				}
 			}
 			log.Debug("remove finalizer from subscription succeeded")
```
marcobebway, line 208: This log now is not accurate, because line 204 `if k8serrors.IsConflict(err)` only checks conflicts, but other errors might stil happen.

components/eventing-controller/controllers/subscription/nats/reconciler.go
```diff
@@ -173,135 +172,156 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
...
 	desiredConditions := make([]eventingv1alpha1.Condition, 0)
-	conditionAdded := false
+	conditionContained := false
+	conditionsUpdated := false
 	condition := eventingv1alpha1.MakeCondition(eventingv1alpha1.ConditionSubscriptionActive,
 		eventingv1alpha1.ConditionReasonNATSSubscriptionActive, corev1.ConditionFalse, message)
 	if isNatsSubReady {
 		condition = eventingv1alpha1.MakeCondition(eventingv1alpha1.ConditionSubscriptionActive,
-			eventingv1alpha1.ConditionReasonNATSSubscriptionActive, corev1.ConditionTrue, "")
+			eventingv1alpha1.ConditionReasonNATSSubscriptionActive, corev1.ConditionTrue, "") // TODO use message instead of ""
```
marcobebway, line 285: Will we resolve/need this TODO?

components/eventing-controller/controllers/subscription/nats/reconciler.go
radufa: Great! 
This is a very old issue introduced in January 2021, in this PR:
https://github.com/kyma-project/kyma/pull/10251/files#diff-34abfaea5dbe6be277421ffd2a2b5a7af2dfe19436d22ffab9bc8c3880dcc34bR94
This is why at the deletion of a subscription CR, in some cases, we had multiple reconcile() callbacks..

radufa: done

components/eventing-controller/controllers/subscription/nats/reconciler.go
radufa: done

components/eventing-controller/controllers/subscription/nats/reconciler.go
radufa: done

components/eventing-controller/controllers/subscription/nats/reconciler.go
radufa: done

components/eventing-controller/controllers/subscription/nats/reconciler.go
```diff
@@ -173,139 +172,201 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
...
-	return result, nil
+	return ctrl.Result{}, nil
 }
 
 // syncSubscriptionStatus syncs Subscription status
-// subsConfig is the subscription configuration that was applied to the subscription. It is set only if the
-// isNatsSubReady is true.
-func (r *Reconciler) syncSubscriptionStatus(ctx context.Context, sub *eventingv1alpha1.Subscription, isNatsSubReady bool, message string) error {
-	desiredSubscription := sub.DeepCopy()
+func (r *Reconciler) syncSubscriptionStatus(ctx context.Context, sub *eventingv1alpha1.Subscription, isNatsSubReady bool, forceUpdateStatus bool, message string) error {
```
nachtmaar, line 271: This method is really **hard to read** and **understand**. As far as I see there is only **one condition** which is used for NATS: `eventingv1alpha1.ConditionSubscriptionActive`

Can't we simplify the method ?
- there is no need for a loop, there is only one condition
- we could create the `desired` conditions and then `compare` them to the `current` conditions (of course without comparing `LastTransitinTime`) and update if both are not equal.

components/eventing-controller/controllers/subscription/nats/reconciler.go
```diff
@@ -173,139 +172,201 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
...
+			eventingv1alpha1.ConditionReasonNATSSubscriptionActive, corev1.ConditionTrue, message)
 	}
 	for _, c := range sub.Status.Conditions {
 		var chosenCondition eventingv1alpha1.Condition
 		if c.Type == condition.Type {
-			// take given condition
-			chosenCondition = condition
-			conditionAdded = true
+			if !conditionContained {
+				if equalsConditionsIgnoringTime(c, condition) {
```
nachtmaar, line 285: 👍 for refactoring this comparison into a method

components/eventing-controller/controllers/subscription/nats/reconciler.go
```diff
@@ -354,46 +415,15 @@ func defaultSinkValidator(ctx context.Context, r *Reconciler, subscription *even
...
 }
 
-func (r *Reconciler) namedLogger() *zap.SugaredLogger {
-	return r.logger.WithContext().Named(reconcilerName)
+// equalsConditionsIgnoringTime checks if two conditions are equal, ignoring lastTransitionTime
+func equalsConditionsIgnoringTime(a, b eventingv1alpha1.Condition) bool {
+	if a.Type == b.Type && a.Status == b.Status && a.Reason == b.Reason && a.Message == b.Message {
+		return true
+	}
+	return false
```
nachtmaar, line 428: This can be simplified:

```suggestion
	return a.Type == b.Type && a.Status == b.Status && a.Reason == b.Reason && a.Message == b.Message
```

components/eventing-controller/pkg/handlers/nats.go
```diff
@@ -215,11 +222,10 @@ func (n *Nats) DeleteSubscription(sub *eventingv1alpha1.Subscription) error {
 			"subject", s.Subject,
 		)
 
-		if strings.HasPrefix(key, subKeyPrefix) {
+		if subKeyPrefix == createKymaSubscriptionNamespacedName(key, s).String() {
```
nachtmaar, line 225: We agreed to  add a unit test for this as part of this issue https://github.com/kyma-project/kyma/issues/13161

components/eventing-controller/controllers/subscription/nats/reconciler.go
```diff
@@ -173,139 +172,201 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
...
+			}
 			return ctrl.Result{}, err
 		}
 		return ctrl.Result{}, err
 	}
 	log.Debug("create NATS subscriptions succeeded")
 
 	// Update status
-	if err := r.syncSubscriptionStatus(ctx, desiredSubscription, true, ""); err != nil {
+	if err := r.syncSubscriptionStatus(ctx, subscription, true, subscriptionStatusChanged, ""); err != nil {
```
nachtmaar, line 259: Thy is the message (last argument to function syncSubscriptionStatus) empty ?

components/eventing-controller/controllers/subscription/nats/reconciler.go
```diff
@@ -173,139 +172,201 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
...
+	// Synchronize Kyma subscription to NATS backend
+	subscriptionStatusChanged, err := r.Backend.SyncSubscription(subscription, r.eventTypeCleaner)
 	if err != nil {
 		log.Errorw("sync subscription failed", "error", err)
-		if err := r.syncSubscriptionStatus(ctx, desiredSubscription, false, err.Error()); err != nil {
+		if err := r.syncSubscriptionStatus(ctx, subscription, false, false, err.Error()); err != nil {
+			if k8serrors.IsConflict(err) {
+				return ctrl.Result{Requeue: true}, nil
+			}
 			return ctrl.Result{}, err
```
nachtmaar, line 252: While I really like the idea of using `ctrl.Result{Requeue: true}` whenever there is a k8s conflict error, I see that it bloats the code up a lot. I don't see the case that the caller of `syncSubscriptionStatus` wants to perform custom error handling. So maybe we should move this to the callee instead and save some lines of code to reduce the size of the methods. What is your take on this ?

components/eventing-controller/controllers/subscription/nats/reconciler_test.go
```diff
@@ -254,6 +255,83 @@ func testCleanEventTypes(id int, eventTypePrefix, natsSubjectToPublish, eventTyp
...
 
+// testUpdateSubscriptionStatus tests if the reconciler can create and update the Status of a Subscription as expected.
+func testUpdateSubscriptionStatus(id int, eventTypePrefix, natsSubjectToPublish, eventTypeToSubscribe string) bool {
+	return When("updating the clean event types in the Subscription status", func() {
+		It("should mark the Subscription as ready", func() {
+			//  set default expectations
+			defaultCondition := eventingv1alpha1.MakeCondition(
+				eventingv1alpha1.ConditionSubscriptionActive,
+				eventingv1alpha1.ConditionReasonNATSSubscriptionActive,
+				v1.ConditionTrue, "")
```
nachtmaar, line 266: Why is the message empty ?

components/eventing-controller/controllers/subscription/nats/reconciler_test.go
```diff
@@ -254,6 +255,83 @@ func testCleanEventTypes(id int, eventTypePrefix, natsSubjectToPublish, eventTyp
...
+			subscriptionName := fmt.Sprintf(subscriptionNameFormat, id)
+			optFilter := reconcilertesting.WithEmptyFilter
+			optWebhook := reconcilertesting.WithWebhookForNats
+			optConditions := reconcilertesting.WithMultipleConditions
+			subscription := reconcilertesting.NewSubscription(subscriptionName, namespaceName, optFilter, optWebhook, optConditions)
+			reconcilertesting.WithValidSink(namespaceName, subscriberSvc.Name, subscription)
+
+			ensureSubscriptionCreated(ctx, subscription)
+
+			Context("A Subscription with muktiple conditions of the same type", func() {
```
nachtmaar, line 292: nit: small typo
```suggestion
			Context("A Subscription with multiple conditions of the same type", func() {
```

components/eventing-controller/controllers/subscription/nats/reconciler_test.go
```diff
@@ -254,6 +255,83 @@ func testCleanEventTypes(id int, eventTypePrefix, natsSubjectToPublish, eventTyp
...
+				// the nats subject list to publish to; these are supposed to be equal to the cleanEventTypes
+				natsSubjectsToPublish := []string{
+					fmt.Sprintf("%s0", natsSubjectToPublish),
+				}
+				// the filter that are getting added to the subscription
```
nachtmaar, line 297: nit: plural
```suggestion
				// the filters that are getting added to the subscription
```

components/eventing-controller/controllers/subscription/nats/reconciler_test.go
```diff
@@ -254,6 +255,83 @@ func testCleanEventTypes(id int, eventTypePrefix, natsSubjectToPublish, eventTyp
 	})
 }
 
+// testUpdateSubscriptionStatus tests if the reconciler can create and update the Status of a Subscription as expected.
+func testUpdateSubscriptionStatus(id int, eventTypePrefix, natsSubjectToPublish, eventTypeToSubscribe string) bool {
```
nachtmaar, line 259: What is the main purpose of this test ? As I understand it you only want to make sure that then there are multiple conditions, only one conditions ends up in the status. Is that correct ? If yes then I would prefer to reduce the test to only test this behaviour instead of also dealing with cleaned event types.

components/eventing-controller/controllers/subscription/nats/reconciler.go
```diff
@@ -173,139 +172,201 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
...
-		if err := r.syncSubscriptionStatus(ctx, actualSubscription, false, err.Error()); err != nil {
-			return ctrl.Result{}, err
-		}
-		// No point in reconciling as the sink is invalid
+		// Stop reconciliation as the object is being deleted
 		return ctrl.Result{}, nil
 	}
 
 	// The object is not being deleted, so if it does not have our finalizer,
 	// then lets add the finalizer and update the object. This is equivalent
```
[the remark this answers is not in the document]
radufa: done

components/eventing-controller/controllers/subscription/nats/reconciler.go
radufa: done

components/eventing-controller/controllers/subscription/nats/reconciler.go
radufa: the convention was to write a message in case of errors

components/eventing-controller/controllers/subscription/nats/reconciler_test.go
radufa: cause for "true" no message was used till now (see my other comment)

components/eventing-controller/controllers/subscription/nats/reconciler_test.go
radufa: done

components/eventing-controller/controllers/subscription/nats/reconciler_test.go
radufa: Cause the PR changes the "saving" of "cleaned event types" and the "configuration" into the subscription status, it tests them too.

components/eventing-controller/controllers/subscription/nats/reconciler_test.go
radufa: done

components/eventing-controller/controllers/subscription/nats/reconciler.go
radufa: As discussed, the original design was to be ready to have multiple conditions which should also "explain" why the global "ready" status is false (like in the BEB case). It could happen that we will add more conditions for JS backend.
Agree, we should redesign this piece of code..

components/eventing-controller/controllers/subscription/nats/reconciler.go
radufa: handle it in a new helper method

nachtmaar (approved):

PR #13081, opened 2022-01-14
add cleanEventTypes to print columns

FriedrichWilken:
**Description**

The cleanEventType were added to print column of the CRD so this property gets displayed in Busola.

(+ some minor general refactoring)

### DIFF
--- a/components/eventing-controller/config/crd/bases/eventing.kyma-project.io_subscriptions.yaml
+++ b/components/eventing-controller/config/crd/bases/eventing.kyma-project.io_subscriptions.yaml
@@ -23,6 +23,9 @@ spec:
     - jsonPath: .metadata.creationTimestamp
       name: Age
       type: date
+    - jsonPath: .status.cleanEventTypes
+      name: CleanEventTypes
+      type: string
     name: v1alpha1
     schema:
       openAPIV3Schema:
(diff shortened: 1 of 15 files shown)


components/eventing-controller/pkg/handlers/eventtype/clean.go
```diff
@@ -34,15 +34,15 @@ type cleaner struct {
 	logger            *logger.Logger
 }
 
-// compile-time check
+// compile-time check of interface comliance
```
k15r, line 37:
```suggestion
// compile-time check of interface compliance
```

components/eventing-controller/config/crd/bases/eventing.kyma-project.io_subscriptions.yaml
```diff
@@ -23,6 +23,9 @@ spec:
     - jsonPath: .metadata.creationTimestamp
       name: Age
       type: date
+    - jsonPath: .status.cleanEventTypes
+      name: CleanEventTypes
+      type: string
```
k15r, line 28: Are we sure this is not an array of strings?

components/eventing-controller/controllers/subscription/nats/reconciler_test.go
```diff
@@ -61,24 +61,77 @@ type testCase func(id int, eventTypePrefix, natsSubjectToPublish, eventTypeToSub
...
 var (
 	reconcilerTestCases = []testCase{
-		testCreateDeleteSubscription,
-		testCreateSubscriptionWithValidSink,
-		testCreateSubscriptionWithInvalidSink,
-		testCreateSubscriptionWithEmptyProtocolProtocolSettingsDialect,
-		testChangeSubscriptionConfiguration,
-		testCreateSubscriptionWithEmptyEventType,
-		testCleanEventTypes,
+	//	testCreateDeleteSubscription,
```
k15r, line 64: ? why are those tests disabled?

Issue #13029, opened 2022-01-10
NATS EPP does not retry failed publish request after NATS server restarts


pxsalehi:
**Description**

While investigating #12930, I found out that once the NATS server is up again, the eventing publisher proxy fails with `HTTP/1.1 502 Bad Gateway` the first time I publish an event. A second publish request to the EPP succeeds. It seems we have a way of reconnecting to NATS, however we are not retrying upon failure, and the client has to retry. 

**Expected result**

If possible, the EPP should reconnect to NATS and retry publishing the event.

**Actual result**

**Steps to reproduce**

To reproduce, you can follow the same approach as #12930.

**Troubleshooting**

marcobebway:
This ticket is blocked until we decide which approach to use for the fix.

marcobebway:
We decided to implement a proper readiness probe for publisher that depends on the readiness of NATS server(s).

PR #13120, opened 2022-01-19
Implement readiness probe for NATS publisher based on connection status of NATS server(s)

marcobebway:
**Description**

Implement readiness probe for the NATS publisher proxy based on the connection status of the NATS server(s).

**Related issue(s)**
- https://github.com/kyma-project/kyma/issues/13029


components/event-publisher-proxy/pkg/sender/nats.go
```diff
@@ -18,48 +18,45 @@ type GenericSender interface {
 	Send(context.Context, *cev2event.Event) (int, error)
 }
 
-// BebMessageSender is responsible for sending messages over HTTP.
+// NatsMessageSender is responsible for sending messages over HTTP.
```
k15r, line 21: 😄

k15r:
the error message sent to the user in case the connection to nats is not working could be improved:
```
Kyma application API responded with status 500 { "status": 500, "message": "connection status not connected: RECONNECTING" }```

FriedrichWilken (changes requested):
there is an issue when port forwarding to epp, while having not nats backend up: when sending an event it gets a 2xx status, while it should be 4xx or 5xx.

marcobebway:
> there is an issue when port forwarding to epp, while having not nats backend up: when sending an event it gets a 2xx status, while it should be 4xx or 5xx.

Thanks for reporting this. I pushed a [commit](https://github.com/kyma-project/kyma/pull/13120/commits/1fc7ebeaaaa6b346bf172231bcb00691accbe436) for the fix.

marcobebway:
> there is an issue when port forwarding to epp, while having not nats backend up: when sending an event it gets a 2xx status, while it should be 4xx or 5xx.

Done.

FriedrichWilken (approved):

Issue #11686, opened 2021-07-14
Fix all infos & warnings from istioctl analyze command


strekm:
**Description**

Clear out an output of `istioctl analyze -n kyma-system` and `istioctl analyze -n kyma-integration` commands. At this moment there are bunch of infos about wrongly named ports and few warnings related to annotations. 
[kyma-integration.txt](https://github.com/kyma-project/kyma/files/6815584/kyma-integration.txt)
[kyma-system.txt](https://github.com/kyma-project/kyma/files/6815585/kyma-system.txt)

**Reasons**

Improve visibility of istio related issues, improve troubleshooting & debugging

**Attachments**
[kyma-integration.txt](https://github.com/kyma-project/kyma/files/6815588/kyma-integration.txt)
[kyma-system.txt](https://github.com/kyma-project/kyma/files/6815589/kyma-system.txt)

mjakobczyk:
It seems that some of the port names connected with `Tracing` can not be changed on Kyma side as they are defined in [`jaeger-operator` itself](https://github.com/jaegertracing/jaeger-operator/blob/master/pkg/service/collector.go#L62).
As a result, following infos would not be addressed:
```bash
Info [IST0118] (Service tracing-jaeger-agent.kyma-system) Port name config-rest (port: 5778, targetPort: 5778) doesn't follow the naming convention of Istio port.
Info [IST0118] (Service tracing-jaeger-collector-headless.kyma-system) Port name c-tchan-trft (port: 14267, targetPort: 14267) doesn't follow the naming convention of Istio port.
Info [IST0118] (Service tracing-jaeger-collector.kyma-system) Port name c-tchan-trft (port: 14267, targetPort: 14267) doesn't follow the naming convention of Istio port.
```

cnvergence:
For duplicate certificate in the application-connector gateway, we won't push any changes
```bash
Warning [IST0138] (Gateway kyma-system/kyma-gateway-application-connector) Duplicate certificate in multiple gateways [kyma-system/kyma-gateway-application-connector kyma-system/kyma-gateway] may cause 404s if clients re-use HTTP2 connections.
```
Please see the explanation [here](https://github.com/kyma-project/kyma/pull/11678#discussion_r691354677)

cnvergence:
The same goes for the tracing-operator
```
Warning [IST0107] (Deployment kyma-system/tracing-jaeger) Misplaced annotation: sidecar.istio.io/inject can only be applied to Pod
[...]

Issue #12977, opened 2022-01-04
Reconciler fails to upgrade jaeger operator service


pbochynski:
**Description**
The job [periodic-main-kyma-incubator-reconciler-kyma1-kyma2-upgrade](https://status.build.kyma-project.io/job-history/gs/kyma-prow-logs/logs/periodic-main-kyma-incubator-reconciler-kyma1-kyma2-upgrade) fails with such log message:
```
{"level":"DEBUG","time":"2022-01-04T07:48:54.795Z",
"message":"Bookkeeper evaluated reconciliation (schedulingID:6f8ec1bc-f300-4f00-bb65-6fec5dcd0ba5--4ab42bee-e187-4cb0-b166-1808765b6dda) for cluster '6f8ec1bc-f300-4f00-bb65-6fec5dcd0ba5' to cluster status 'reconciling': Done=CRDs[2],cluster-essentials[3],certificates[3],application-connector[4],monitoring[4],cleaner[1],istio-configuration[3],api-gateway[4],service-catalog-addons[4],kiali[4],ory[4],eventing[4],logging[4],service-catalog[4],helm-broker[4],rafter[4],cluster-users[4],
serverless[4] / Error= / New= / Running=tracing[4][error: kubeClient failed to update Service 'tracing-jaeger-operator' (namespace: kyma-system): 
All attempts fail:\n
#1: cannot patch \"tracing-jaeger-operator\" with kind Service: Service \"tracing-jaeger-operator\" is invalid: spec.ports[1].name: Duplicate value: \"metrics\"\n
#2: cannot patch \"tracing-jaeger-operator\" with kind Service: Service \"tracing-jaeger-operator\" is invalid: spec.ports[1].name: Duplicate value: \"metrics\"\n
#3: cannot patch \"tracing-jaeger-operator\" with kind Service: Service \"tracing-jaeger-operator\" is invalid: spec.ports[1].name: Duplicate value: \"metrics\"]"}
```

skhalash:
Fixed in https://github.com/kyma-project/kyma/pull/12985, the jobs are green again.

PR #13175, opened 2022-01-25
Fix Tracing port names in services to comply with Istio convention

mjakobczyk:
**Description**

We should follow [official Istio documentation about port naming convention](https://istio.io/latest/docs/reference/config/analysis/ist0118/). After editing both services there is a similar spec which indicates that the used protocol is TCP.

```yaml
...
spec:
  ...
  ports:
  - name: tcp-metrics
    port: 8383
    protocol: TCP
    targetPort: 8383
  ...
```

- fix port names in `tracing-jaeger-operator` service to include protocol prefix
- fix port names in `tracing-jaeger-metrics` service to include protocol prefix
- fix port names in `zipkin` service to include protocol prefix
- get rid of the following errors:

```bash
Info [IST0118] (Service tracing-jaeger-operator.kyma-system) Port name metrics (port: 8383, targetPort: 8383) doesn't follow the naming convention of Istio port.
Info [IST0118] (Service tracing-jaeger-metrics.kyma-system) Port name metrics (port: 14269, targetPort: 14269) doesn't follow the naming convention of Istio port.
Info [IST0118] (Service zipkin.kyma-system) Port name jaeger-collector-zipkin (port: 9411, targetPort: 9411) doesn't follow the naming convention of Istio port.
```

**Related issue(s)**

See: https://github.com/kyma-project/kyma/issues/11686

### DIFF
--- a/resources/tracing/templates/kyma-additions/metrics-service.yaml
+++ b/resources/tracing/templates/kyma-additions/metrics-service.yaml
@@ -7,7 +7,7 @@ metadata:
     app.kubernetes.io/name: {{ include "jaeger-operator.fullname" . }}-jaeger-metrics
 spec:
   ports:
-  - name: http-metrics
+  - name: http-jaeger-metrics
     port: 14269
     protocol: TCP
     targetPort: 14269
--- a/resources/tracing/templates/kyma-additions/metrics-service.yaml
+++ b/resources/tracing/templates/kyma-additions/metrics-service.yaml
@@ -7,7 +7,7 @@ metadata:
     app.kubernetes.io/name: {{ include "jaeger-operator.fullname" . }}-jaeger-metrics
 spec:
   ports:
-  - name: tcp-metrics
+  - name: http-metrics
     port: 14269
     protocol: TCP
     targetPort: 14269
--- a/resources/tracing/templates/kyma-additions/metrics-service.yaml
+++ b/resources/tracing/templates/kyma-additions/metrics-service.yaml
@@ -7,7 +7,7 @@ metadata:
     app.kubernetes.io/name: {{ include "jaeger-operator.fullname" . }}-jaeger-metrics
 spec:
   ports:
-  - name: metrics
+  - name: tcp-metrics
     port: 14269
     protocol: TCP
     targetPort: 14269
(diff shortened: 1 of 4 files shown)


resources/tracing/templates/kyma-additions/metrics-service.yaml
```diff
@@ -7,7 +7,7 @@ metadata:
     app.kubernetes.io/name: {{ include "jaeger-operator.fullname" . }}-jaeger-metrics
 spec:
   ports:
-  - name: metrics
+  - name: tcp-metrics
```
chrkl, line 10: Guess it's a http connection, right? Iirc correctly, having a http- prefix allows Istio to account proper http response codes in the metrics and visualize in Kiali.

mjakobczyk: Yes, sure - TCP is simply more generic. If we want to be specific we can call it `http` also. Changing 👍

chrkl: Looks like you also have to change the port names in the ServiceMonitors to make the tests happy.

resources/tracing/templates/service.yaml
```diff
@@ -6,7 +6,7 @@ metadata:
 {{ include "jaeger-operator.labels" . | indent 4 }}
 spec:
   ports:
-  - name: metrics
+  - name: tcp-metrics
```
chrkl, line 9: Let's change this as well

chrkl:
Please test that your PR would not trigger [this](https://github.com/kyma-project/kyma/issues/12977) again. Thanks!

resources/tracing/templates/kyma-additions/zipkin-service.yaml
```diff
@@ -7,7 +7,7 @@ metadata:
     app.kubernetes.io/component: zipkin 
 spec:
   ports:
-  - name: jaeger-collector-zipkin
+  - name: tcp-jaeger-collector-zipkin
```
chrkl, line 10: We figured out that zipkin is also a http based protocol. Please change this prefix as well.

chrkl (approved):

PR #13189, opened 2022-01-26
Clean up NATS handler unit tests

pxsalehi:
**Description**

This is a follow up PR to cleanup the NATS handler unit tests, after changes in https://github.com/kyma-project/kyma/pull/12997

- Fix race condition in `testing/subscriber.go` (at least now `go test -race` doesn't complain for the handler tests!)
- Reuse existing utility functions to simplify and shorten the unit tests to improve readability
- Try to follow the same order and naming in the unit tests

**Related issue(s)**

### DIFF
--- a/components/eventing-controller/pkg/handlers/nats_test.go
+++ b/components/eventing-controller/pkg/handlers/nats_test.go
@@ -105,7 +105,7 @@ func TestSubscription(t *testing.T) {
 	natsBackend := NewNats(natsConfig, env.DefaultSubscriptionConfig{MaxInFlightMessages: defaultMaxInflight}, nil, defaultLogger)
 	g.Expect(natsBackend.Initialize(env.Config{})).Should(Succeed())
 
-	subscriber, _ := startSubscriber()
+	subscriber := startSubscriber()
 	defer subscriber.Shutdown()
 	g.Expect(subscriber.IsRunning()).To(BeTrue())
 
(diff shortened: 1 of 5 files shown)


components/eventing-controller/pkg/handlers/nats_test.go
```diff
@@ -1072,96 +660,72 @@ func TestRetryUsingCESDK(t *testing.T) {
...
 	// After reconnect, event delivery should work again
 	ev2data := "newsampledata"
-	g.Expect(SendEventToNATS(natsClient, ev2data)).To(Succeed())
+	g.Expect(SendEventToNATS(natsBackend, ev2data)).Should(Succeed())
 	expectedEv2Data := fmt.Sprintf("\"%s\"", ev2data)
-	g.Expect(subscriber.CheckEvent(expectedEv2Data, subscriberCheckURL)).To(Succeed())
+	g.Expect(subscriber.CheckEvent(expectedEv2Data)).Should(Succeed())
 }
 
 func startNATSServer() (*server.Server, int) {
```
marcobebway, line 731: Do we still need to return the int NATS port?

pxsalehi: It's used in one test, and we can't get it back from the server.

components/eventing-controller/pkg/handlers/nats_test.go
```diff
@@ -1172,22 +736,20 @@ func startNATSServer() (*server.Server, int) {
 
 func startSubscriber() (*eventingtesting.Subscriber, int) {
```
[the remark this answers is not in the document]
pxsalehi: This is also used in one test. But since we can change the subscriber struct, I have instead exported the port in the struct, and removed it here.

marcobebway (approved):
/lgtm

Issue #13059, opened 2022-01-13
Change reconciliation period based on resource status


k15r:
## Description

Subscriptions should be reconciled more often if they are not ready. 
If a subscription is not ready due to a non existing sink the default 10 minute reconciliation period is not optimal. 

Consider this scenario:
* service and subscription are created at the same time
* subscription gets reconciled, but the service is not ready yet, but will be ready few seconds later
* subscription will not be ready
* 10 mins later the subscription will be reconciled again. This time everything is ready. Subscription will turn ready

## Proposed solution
* `not ready` subscriptions should be reconciled more often. Ideally every few seconds in the beginning, then with exponential backoff 
* `ready` subscriptions should be reconciled with a constant delay in order to react to other state changes

nachtmaar:
Maybe this can help: https://github.com/kubernetes-sigs/kubebuilder/issues/460 (did not check the content yet)

pxsalehi:
Reading the `controller-runtime` code, to me it seems, that this is what the default queue of the controller does, as long as the reconcile function returns error or `Requeue: true`. 
https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/internal/controller/controller.go#L311-L334
This is also pointed out by the issue that @nachtmaar shared.

The only difference between returning error or `Requeue: true`, seems to be logging the error and counting it into controller metrics. So as long as we just return error from the reconciler when we cannot mark the subscription as ready, this should work. IMO, this makes the code more readable in comparison to selectively returning `Requeue: true`.

pxsalehi:
For BEB, as I see in the BEB subscription controller, it seems that we do return error whenever we cannot mark the subscription as ready. However, I'd rather verify/document it with a unit test as planned also for NATS. For this kind of unit test, some minor refactoring would help to reduce mocking or making it easier to mock the controller dependencies. However, as currently a general refactoring of the BEB subscription controller is in progress, I'll do this once we have finished https://github.com/kyma-project/kyma/issues/13026

k15r:
already addressed by https://github.com/kyma-project/kyma/pull/13070

PR #13205, opened 2022-01-28
Subscriptions reconciliation should be requeued if subscription cannot be marked as ready

pxsalehi:
**Description**

See https://github.com/kyma-project/kyma/issues/13059#issuecomment-1023989101

- In the NATS reconciler, return an error whenever we cannot set the subscription status to ready.

**Related issue(s)**

This PR addresses https://github.com/kyma-project/kyma/issues/13059 for the NATS reconciler. If we agree on this way of addressing the issue, I'll create another PR for BEB.

### DIFF
--- a/components/eventing-controller/controllers/subscription/nats/reconciler.go
+++ b/components/eventing-controller/controllers/subscription/nats/reconciler.go
@@ -225,21 +225,21 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 	// Check for valid sink
 	if err := r.sinkValidator(ctx, r, subscription); err != nil {
 		log.Errorw("sink URL validation failed", "error", err)
-		if err := r.syncSubscriptionStatus(ctx, subscription, false, false, err.Error()); err != nil {
-			return checkIsConflict(err)
+		if syncErr := r.syncSubscriptionStatus(ctx, subscription, false, false, err.Error()); err != nil {
+			return ctrl.Result{}, syncErr
 		}
-		// No point in reconciling as the sink is invalid
-		return ctrl.Result{}, nil
+		// No point in reconciling as the sink is invalid, return latest error to requeue the reconciliation request
+		return ctrl.Result{}, err
 	}
 
 	// Synchronize Kyma subscription to NATS backend
-	subscriptionStatusChanged, err := r.Backend.SyncSubscription(subscription, r.eventTypeCleaner)
-	if err != nil {
-		log.Errorw("sync subscription failed", "error", err)
-		if err := r.syncSubscriptionStatus(ctx, subscription, false, false, err.Error()); err != nil {
-			return checkIsConflict(err)
+	subscriptionStatusChanged, syncErr := r.Backend.SyncSubscription(subscription, r.eventTypeCleaner)
+	if syncErr != nil {
+		log.Errorw("sync subscription failed", "error", syncErr)
+		if err := r.syncSubscriptionStatus(ctx, subscription, false, false, syncErr.Error()); err != nil {
+			return ctrl.Result{}, err
 		}
-		return ctrl.Result{}, err
+		return ctrl.Result{}, syncErr
 	}
 	log.Debug("create NATS subscriptions succeeded")
 
--- a/resources/eventing/values.yaml
+++ b/resources/eventing/values.yaml
@@ -5,7 +5,7 @@ global:
   images:
     eventing_controller:
       name: eventing-controller
-      version: PR-13189
+      version: PR-13205
       pullPolicy: "IfNotPresent"
     publisher_proxy:
       name: event-publisher-proxy


components/eventing-controller/controllers/subscription/nats/reconciler.go
```diff
@@ -225,21 +225,21 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
...
 		log.Errorw("sink URL validation failed", "error", err)
-		if err := r.syncSubscriptionStatus(ctx, subscription, false, false, err.Error()); err != nil {
-			return checkIsConflict(err)
+		if syncErr := r.syncSubscriptionStatus(ctx, subscription, false, false, err.Error()); err != nil {
+			return ctrl.Result{}, syncErr
 		}
-		// No point in reconciling as the sink is invalid
-		return ctrl.Result{}, nil
+		// No point in reconciling as the sink is invalid, return latest error to requeue the reconciliation request
+		return ctrl.Result{}, err
```
nachtmaar, line 232: I did read your comment on the [issue](https://github.com/kyma-project/kyma/issues/13059#issuecomment-1023989101).
Seems this really uses exponential backoff.  
Did you test it with the example from the issue and can confirm it ?

Out of curiosity: Did you dig into the code to the lines where the exponential backoff is implemented?
I came accross [this](https://github.com/kubernetes/client-go/blob/4fc3881b7ffe4a5c1c7c3cffc7dd9d9667989730/util/workqueue/default_rate_limiters.go#L41) but am not sure whether this is really used by the controller.

components/eventing-controller/controllers/subscription/nats/reconciler.go
```diff
@@ -225,21 +225,21 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 	// Check for valid sink
 	if err := r.sinkValidator(ctx, r, subscription); err != nil {
 		log.Errorw("sink URL validation failed", "error", err)
-		if err := r.syncSubscriptionStatus(ctx, subscription, false, false, err.Error()); err != nil {
-			return checkIsConflict(err)
+		if syncErr := r.syncSubscriptionStatus(ctx, subscription, false, false, err.Error()); err != nil {
+			return ctrl.Result{}, syncErr
```
nachtmaar, line 229: The method `checkIsConflict` was implemented in a previous [PR](https://github.com/kyma-project/kyma/pull/13070/files):

```go
func checkIsConflict(err error) (ctrl.Result, error) {
	if k8serrors.IsConflict(err) {
		// Requeue the Request to try to reconcile it again
		return ctrl.Result{Requeue: true}, nil
	}
	return ctrl.Result{}, err
}
```
As you described in the issue already, there is no fundamental difference between 
`return ctrl.Result{}, err` and `return ctrl.Result{Requeue: true}, nil`. 

The latter will miss these two lines:

```go
ctrlmetrics.ReconcileErrors.WithLabelValues(c.Name).Inc()
log.Error(err, "Reconciler error")
```

So the basic question is if we want to treat `k8serrors.IsConflict` as an expected error. If yes, then we don't need to log it out and don't need to increase the metrics for errored reconciliations.

My personal opinion is that it is nice to not bloat the logs of the reconciler with the conflict error. That is just noise too my eyes then debugging something.

What is your take ?

nachtmaar: cc: @radufa since you implemented the change.

components/eventing-controller/controllers/subscription/nats/reconciler.go
pxsalehi: Regarding testing it, I haven't. But I have seen the back-off behaviour work when we return error, since that is a normal flow for our reconciler that sometimes returns errors. Frankly, it is hard to write a test, and even if we do, we are testing controller-runtime.

Yes, that is also where I ended up, checking the behaviour of the default queue.

components/eventing-controller/controllers/subscription/nats/reconciler.go
pxsalehi: I agree that as it is, here `checkIsConflict` doesn't really make much of a difference aside from causing the controller runtime to also log the error. In some cases I've removed it (when subscription is not ready). In other cases, I'm not sure if it can be removed.

If you're only concern is bloated logs (which I also dislike), we need to do bigger clean ups. Currently, whenever we encounter an error in the reconciler, we log it and we return an error. Returning an error would also get logged by the controller runtime. So I think, we shouldn't log just the error, if we have no other extra information (we could return an enriched/wrapped error and leave the logging to the controller runtime).

mfaizanse:
lgtm

components/eventing-controller/controllers/subscription/nats/reconciler.go
radufa: >My personal opinion is that it is nice to not bloat the logs of the reconciler with the conflict error. That is just noise too my eyes then debugging something.

Agree, for the "helping the customer" perspective, is good to have in the log file only "real errors". The customer is reading the log file and send it somehow to us for the future analysis. We have tried to clean-up the log file (in main), so that under "info" only errors should be logged after the controller is stabilized.

nachtmaar: @pxsalehi I agree that logging an error and returning it to controller-runtime to log it as well makes no sense. Your proposed solution makes sense to me :) Would you like to create an issue for it ? :)

Thanks for explaining the pattern you used for removing checkIsConflict. I did not see it before.

I think in both cases we have a log line which states the actual error. So independent of if we (1) log it ourselves or (2) let controller-runtime log it I think we should avoid logging the conflict error in case of a kubernetes conflict.

nachtmaar: @radufa I think we should add a comment to describe the `advantage` of using `checkIsConflict`.

components/eventing-controller/controllers/subscription/nats/reconciler.go
pxsalehi: To be honest, IMO, that would be unnecessary. We are testing mostly sink validation logic. The only thing this PR is changing is the return value of the `Reconcile` method in a couple of places. If you really insist on having tests for that, I'd suggest the following:

All we want to verify is that under certain conditions, a single call of `Reconcile` returns error. This can, and should IMO, be done with a simple unit test. We'd need to mock the backend, sink validator and K8s client, and make sure the `Reconcile` call returns error when it should, e.g., when backend has returned error, or sink validator has returned error. The details of the backend and the sink validator are irrelevant in this case, since they are just dependencies of the reconciler.

The problem with using `envtest` in this case is that: 1) we are adding more and more tests for small cases and complicating that test suite more than it is. 2) In the example you explained, we have no control over when or how often is the reconcile method called, and it is affected by the changes to the subscription CRD. E.g. when in step two the subscription status has been updated as not ready, wouldn't that trigger another reconciliation? IMO, `envtest` tests should be reserved for cases when we care about the eventual state of the object (in this case the subscription), not the behaviour/return type of our reconciler code.

With unit test, we could call the Reconcile method exactly one time, they are much faster, and we could test all the cases rather than just one. The problem with the unit test is that there will be a lot of mocking.

Do you agree that this is more suitable for a unit test? Should I try that?

pxsalehi: As agreed, here is an issue to do this later. Since currently our reconciler tests are 'under construction'!
https://github.com/kyma-project/kyma/issues/13218

components/eventing-controller/controllers/subscription/nats/reconciler.go
pxsalehi: A reminder to clean up the duplicate error log entries: https://github.com/kyma-project/kyma/issues/13219

pxsalehi:
pre-main-kyma-gardener-gcp-eventing-upgrade

pxsalehi:
pre-main-kyma-skr-eventing

components/eventing-controller/controllers/subscription/nats/reconciler.go
nachtmaar: I talked to @pxsalehi in person and I totally agree a unit test is better than an integration test here.

nachtmaar (approved):

Issue #13207, opened 2022-01-28
Adjust resource configuration for "S" and "XS" profiles for python runtimes


kwiatekus:
**Description**
Current resource limits for XS and S profiles are not sufficient to accommodate hello world python application.
The lowest possible profile is M and S and XS are useless.

Please adjust the profiles so that:
 - XS becomes limits from current M profile
 - Adjust resources ( linearly ) for each profile  : S, M, L, XL

kwiatekus:
@pPrecel thank you for fixing this. It works for me now.
Please adjust the documentation [here](https://kyma-project.io/docs/kyma/latest/05-technical-reference/svls-09-available-presets/)

PR #13226, opened 2022-01-31
Adjust function profiles

pPrecel:
**Description**

- Adjust function profiles due to XS and S is already useless. My propositions: 
- - now default preset is `XS`
- - old `M` becomes `XS`
- - old `L` becomes `S`
- - old `XL` becomes `M`
- - new `L` is double bigger than new `M`
- - new `XL` is double bigger than new `L`

**Related issue(s)**

https://github.com/kyma-project/kyma/issues/13207

### DIFF
--- a/resources/serverless/charts/webhook/values.yaml
+++ b/resources/serverless/charts/webhook/values.yaml
@@ -116,22 +116,22 @@ values:
             "limitMemory": "256Mi"
           },
           "M": {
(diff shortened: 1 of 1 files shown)


Cortey (approved):
Works both in Busola and via Kyma CLI

dbadura (approved):

Issue #13182, opened 2022-01-26
Missing datacontenttype CE attribute results in stringified payloads sent via NATS


kwiatekus:
**Description**

I noticed that event data payload type is not resolved and equivalent CE `datacontenttype ` fields  are not set.
In case of NATS backend this results in situations that json payloads get stringified.

From the function developer perspective it is unconvinient. I'd like to be able to re-use the same function code regardless of the eventing backend selected. Right now I have to manually specify `datacontenttype` in case of json obejcts

**Expected result**
I'd expect the event payload data type is recognised at the [utility function](https://github.com/kyma-project/kyma/blob/19830ec3d3a36e924186e8c55001dc83a3b37a1c/components/function-runtimes/nodejs14/kubeless/kubeless_rt/lib/ce.js#L56)
I'd expect the incoming `event.data` to be of same type (preferably json object, not stringified object) 

**Actual result**

In case NATS is used, and datacontenttype is not set explicitly, the payload gets stringified

```js
module.exports = {
    main: function (event, context) {
...
    const eventType = "sap.kyma.custom.aaa.bbb.ccc.v1"
    var eventOut=event.buildResponseCloudEvent("123124",eventType,{"foo": "bar"});
    eventOut.source="kyma"
    eventOut.specversion="1.0"
    // eventOut.datacontenttype="application/json"

    event.publishCloudEvent(eventOut);
...
    }
}
```

pxsalehi:
@kwiatekus Thanks for creating the issue. I am not able to reproduce this using `main`. I have tried the following function:
```
module.exports = {
    main: function (event, context) {
        console.log('Type of event.data: ', typeof event.data)
        console.log(event.data)
    }
}
```

And for each backend I publish an event that has the active backend in the event data. For example:
```
curl -k -i --data @<(cat <<EOF
    {
        "source": "/default/sap.kyma/tunas-develop",
        "specversion": "1.0",
        "eventtypeversion": "v1",
        "data": {"current-backend":"NATS"},
        "datacontenttype": "application/json",
        "id": "759815c3-b142-48f2-bf18-c6502dc0998f",
        "type": "sap.kyma.custom.myapp.order.created.v1"
    }
EOF
    ) -H "Content-Type: application/cloudevents+json" "http://localhost:8080/publish"
```

This is what I'll see in the function log:
```
Type of event.data:  object
{ 'current-backend': 'BEB' }

Type of event.data:  object
{ 'current-backend': 'NATS' }
```

[...]

PR #13263, opened 2022-02-02
Adding the missing Cloud Event attribute

Cortey:
**Description**
Setting the `datacontenttype` attribute to the correct value prevents json payloads from getting stringified.

**Related issue(s)**
See also #13182

### DIFF
--- a/components/function-runtimes/python39/kubeless/kubeless.py
+++ b/components/function-runtimes/python39/kubeless/kubeless.py
@@ -116,6 +116,13 @@ class Event:
             headers = {"Content-Type": "application/cloudevents+json"}
             )
     
+    def resolvedatatype(event_data):
+        match type(event_data)
+            case 'dict':
+                return 'application/json'
+            case 'str':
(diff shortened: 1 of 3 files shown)


components/function-runtimes/python39/kubeless/kubeless.py
```diff
@@ -116,6 +116,13 @@ def publishCloudEvent(self, data):
             headers = {"Content-Type": "application/cloudevents+json"}
             )
     
+    def resolvedatatype(event_data):
+        match type(event_data)
+            case 'dict':
+                return 'application/json'
+            case 'str':
+                return 'text/plain'
```
pPrecel, line 124: is it possible to receive type other than dict or str? what in a situation like this? This question is for two other switch sentences too

Cortey: There is no such possibility thus we don't need a default case.

pPrecel (approved):

PR #12964, opened 2022-01-03
Dockerfile lint 3

ammarlakis:
**Description**

Changes proposed in this pull request:

- Fixing linting errors in Dockerfiles. Patch 3 includes the components:
  - components/iam-kubeconfig-service/Dockerfile
  - components/kyma-operator/Dockerfile
  - components/service-binding-usage-controller/Dockerfile
  - components/uaa-activator/Dockerfile
  - tests/integration/api-gateway/Dockerfile
  - tests/integration/apiserver-proxy/Dockerfile
  - tests/integration/dex/Dockerfile
  - tests/rafter/Dockerfile
  - tests/service-catalog/Dockerfile
  - tools/event-subscriber/Dockerfile
  - tools/gitserver/Dockerfile

**Related issue(s)**
Related to #12854

Issue #12854, opened 2021-12-13
Update Dockerfiles to fix Hadolint warning


adamwalach:
**Description**

Fix dockerfiles in Kyma repo and implement scanning of test-infra Dockerfiles with [Hadolint](https://github.com/hadolint/hadolint)

**Reasons**

We need to make sure that our Dockerfiles conform to [dockerfile best practices](https://docs.docker.com/develop/develop-images/dockerfile_best-practices/).

**Acceptance Criteria**

- Dockerfiles in the Kyma repo are compliant :)
- new job triggered on every Dockerfile change 
- pr is blocked if Dockerfile is not compliant

PR #13274, opened 2022-02-03
Dockerfile linting

ammarlakis:
**Description**

- Fixing linting errors in Dockerfiles.

**Related issue(s)**

Split from #12964 
Related to #12854


tests/service-catalog/Dockerfile
```diff
@@ -15,11 +15,13 @@ RUN CGO_ENABLED=0 go test -tags=acceptance -c ./test -o /servicecatalog.test
 FROM alpine:3.10
 RUN apk add --no-cache curl
 
+WORKDIR /app
+
 COPY --from=builder env-tester.bin /go/bin/env-tester.bin
 COPY --from=builder ./servicecatalog.test .
```
wozniakjan, line 21: do you know if it matters that `servicecatalog.test` moved inside of the container from `/servicecatalog.test` to `/app/servicecatalog.test`?

wozniakjan (approved):
/lgtm
/hold

feel free to unhold if the below mentioned comment doesn't cause any problems

tests/service-catalog/Dockerfile
ammarlakis: no since `.` means current work directory, which is set to `/app`.

wozniakjan: I know, just wondering if some app expects it to be hardcoded path. Feel free to unhold at your convenience.

PR #13313, opened 2022-02-08
Dockerfile-lint-3-2

ammarlakis:
**Description**

- Fixing linting errors in Dockerfiles.

**Related issue(s)**

Split from #12964 
Related to #12854

### DIFF
--- a/tools/gitserver/Dockerfile
+++ b/tools/gitserver/Dockerfile
@@ -1,7 +1,11 @@
 FROM alpine:3.12
 
-RUN apk update &&\
-  apk add git-daemon apache2 apache2-utils
+FROM alpine:3.15
+
+RUN apk add --no-cache git-daemon apache2 apache2-utils && \
+    apk upgrade --no-cache --repository=http://dl-cdn.alpinelinux.org/alpine/edge/community \
+    expat
+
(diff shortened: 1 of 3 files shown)


tools/gitserver/Dockerfile
```diff
@@ -1,7 +1,11 @@
 FROM alpine:3.12
 
-RUN apk update &&\
-  apk add git-daemon apache2 apache2-utils
+FROM alpine:3.15
+
+RUN apk add --no-cache git-daemon apache2 apache2-utils && \
+    apk upgrade --no-cache --repository=http://dl-cdn.alpinelinux.org/alpine/edge/community \
```
dbadura, line 6: Do we need this?

ammarlakis: yes, it has a critical vulnerability according to protecode scan so I had to update it to the latest, and the latest version is in edge repo.

dbadura: Maybe I am blind, but I don't see that after adding repo `edge` we install something from this repo.
Also this image is used only in tests.

ammarlakis: see next line, I'm upgrading `expat` library, which was already there (I guess a dependency of apache2) to the latest version.

Issue #13366, opened 2022-02-11
Recover Loki from storage exhaustion


a-thaler:
Continue finding a solution so that we have active guards in place so that Loki will not crash when reaching its limits.

a-thaler:
Here the basic idea outlined as shell script: https://github.com/a-thaler/kyma/commit/92aae13476da20e7942c15317074bb474aef024f

a-thaler:
Most important to mention: as mentioned in the docu of Loki, when deleting the chunks the index will stay. The only negative side effect which I could observe is that: queries will fail which are reaching back into the time range where chunk deletion happened. That is not the best user experience in grafana, anyway it is much better then causing an unrecoverable crash loop.

PR #13369, opened 2022-02-12
Sidecar for loki to perform retention by size

a-thaler:
**Description**

- adds a sidecar for loki to delete oldest chunks dependent on volume size
- ...
- ...

**Related issue(s)**

https://github.com/kyma-project/kyma/issues/13366

### DIFF
--- a/resources/logging/charts/loki/files/chunkCleaner.sh
+++ b/resources/logging/charts/loki/files/chunkCleaner.sh
@@ -1,0 +1,36 @@
+#!/usr/bin/env bash
+set -e
+
+function updateCurrentSize() {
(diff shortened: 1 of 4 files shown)


resources/logging/charts/loki/files/chunkCleaner.sh
```diff
@@ -0,0 +1,36 @@
+#!/usr/bin/env bash
+set -e
+
+function updateCurrentSize() {
+    CURRENT_SIZE_KB=$(df /data | tail -n 1 | awk '{print $3}')
```
chrkl, line 5: In my tests, `'{print $4}'` took the right column

chrkl: No, it's correct like this. You want the used bytes.

PR #13390, opened 2022-02-14
Nats sync initial status

raypinto:
**Description**

- ...
- ...
- ...

**Related issue(s)**


components/eventing-controller/pkg/handlers/nats.go
```diff
@@ -101,40 +100,40 @@ func newCloudeventClient(config env.NatsConfig) (cev2.Client, error) {
...
+			return []string{}, err
 		}
 		cleanSubjects = append(cleanSubjects, subject)
 	}
+	return cleanSubjects, nil
+}
 
+// SyncSubscription synchronizes the given Kyma subscription to NATS subscription.
+// The returned bool acts as a marker for changed subscription status.
+func (n *Nats) SyncSubscription(sub *eventingv1alpha1.Subscription, _ ...interface{}) (bool, error) {
```
FriedrichWilken, line 127: I looks like the returned bool is always going to be `false`. I wonder if we can get rid of it. It looks like it is required to fulfil the Interface. 

On the other hand, wherever `SyncSubscription()` is used the returned value gets discarded anyhow. 

The only place where it is used, I think, is [here](https://github.com/raypinto/kyma/blob/7328a7c8ac2cc98cf45196e2490449a6e49f8e7e/components/eventing-controller/controllers/subscription/nats/reconciler.go#L224)
and then [here](https://github.com/raypinto/kyma/blob/7328a7c8ac2cc98cf45196e2490449a6e49f8e7e/components/eventing-controller/controllers/subscription/nats/reconciler.go#L235) where it is used together with the initial status bool. So, it might be okay to get rid of the bool here, right?

raypinto: That's correct! I wanted to get rid of the bool, but since it's required for the interface and its being used in BEB, I let it be. Probably I could discard the return value, as it doesn't add value. What do you think?

FriedrichWilken: Right, I forgot to look what BEB does with the bool! But yes, then don't change the signature to keep the interface implementation but discard the return value as well.

FriedrichWilken (approved):

Issue #13316, opened 2022-02-08
Allow enabling the feature flag in the Eventing E2E ProwJobs


pxsalehi:
We should be able to run our E2E ProwJobs with the feature flag for JetStream enabled. This would allow running tests with JetStream backend on the PRs. The jobs could be triggered via `pjtester`. To make that work, we'd need to:

1. Find out how can we set an ENV var in `pjtester`.
2. If the above is possible, we'd need to adjust our ProwJobs to enable JetStream during installation. If it is not possible, we'd need to find a different way.

Issue #13251, opened 2022-02-02
Setup feature flag and scaffolding for a JetStream backend


pxsalehi:
The feature flag should enable JetStream in the deployed NATS cluster, and start a JetSream Subscription manager instead of the existing NATS subscription manager.

The JetStream subscription manager can for now be empty, and raise an error when receiving a reconciliation request.

We could also check whether it is possible to parametrise our Prow jobs so that we can call our PR validation jobs with JetStream enabled.

PR #13394, opened 2022-02-15
Clear subscription status in JetStream reconciler

pxsalehi:
**Description**

While testing https://github.com/kyma-project/kyma/issues/13316, I found out that the "empty" JetStream reconciler (done in https://github.com/kyma-project/kyma/issues/13251) should at lease make sure subscription status is cleared, as otherwise, during an upgrade with feature flag enabled, the subscription would still be reported as ready. 

Marking the subscription as not ready makes the tests fail quicker, and it is clear why the test fails (i.e. the subscription is not ready).

### DIFF
--- a/components/eventing-controller/cmd/eventing-controller/main.go
+++ b/components/eventing-controller/cmd/eventing-controller/main.go
@@ -42,19 +42,32 @@ func main() {
 	// Set controller core logger.
 	ctrl.SetLogger(zapr.NewLogger(ctrLogger.WithContext().Desugar()))
 
-	// Add schemes.
+	// Instantiate and initialize all the subscription managers.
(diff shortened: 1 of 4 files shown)


components/eventing-controller/cmd/eventing-controller/main.go
```diff
@@ -42,19 +42,32 @@ func main() {
 	// Set controller core logger.
 	ctrl.SetLogger(zapr.NewLogger(ctrLogger.WithContext().Desugar()))
 
-	// Add schemes.
```
pxsalehi, line 45: Changes in this file are just to add the AddToScheme call for JetStream within the same `if`.

PR #13405, opened 2022-02-16
Always wait for subscriptions in Eventing tests

pxsalehi:
**Description**

- Always wait for the subscriptions to be ready. In the `pre-main-kyma-gardener-gcp-eventing-upgrade` ProwJob this is missing, since we run tests after upgrade which could mark subscriptions as not ready (e.g. upgrading with JetStream feature flag enabled).
- The `pre-main-kyma-gardener-gcp-eventing-upgrade` takes quite long, and if any tests fail we would have to rerun the whole job. Using the `--bail` flag, stops the job quicker if some tests fail.

**Related issue(s)**


VladislavPaskar (approved):
thanks for this enhancement

lindnerby (approved):

jeremyharisch (approved):

PR #13409, opened 2022-02-16
Reimplementing NATS integration tests with plain Go

FriedrichWilken:
This PR replaces the existing integration tests --which make use of the ginkgo framework-- with ones that use plain golang table tests.

The old tests, for the better part followed one of two patterns:
1. create a subscription and check if it gets reconciled as expected.
2. create a subscription and check if it gets reconciled as expected then change it and check if gets reconciled as expected, again.

The new tests do exactly this, but in table tests. The goal is, to make it easier to create new test cases and to change existing ones.

The old tests also tested against a scenario of an empty prefix. This, for now is moved into an extra test. This needs to be implemented as a fast integration test.


components/eventing-controller/controllers/subscription/nats/integration_test.go
```diff
@@ -0,0 +1,837 @@
...
+		{
+			name: "create and delete",
+			subscriptionOpts: []reconcilertesting.SubscriptionOpt{
+				reconcilertesting.WithFilter(reconcilertesting.EventSource, reconcilertesting.OrderCreatedEventTypeNotClean),
+				reconcilertesting.WithWebhookForNATS(),
+				reconcilertesting.WithSinkURLFromSvc(ens.subscriberSvc),
+			},
+			expect: expect{
+				k8sSubscription: []gomegatypes.GomegaMatcher{
+					reconcilertesting.HaveCondition(reconcilertesting.DefaultCondition("")),
```
nachtmaar, line 114: nit: While this is completely true, what is important here is that the condition is ready. I wonder if we should put more emphasis on this fact and rename it to `DefaultReadyCondition ` ? 🤔 
What is your opinion on this ?

```suggestion
					reconcilertesting.HaveCondition(reconcilertesting.DefaultReadyCondition("")),
```

What is a bit confusing to me is that a DefaultCondition does not have a default `message` (`""` in this case).

components/eventing-controller/controllers/subscription/nats/integration_test.go
```diff
@@ -0,0 +1,837 @@
...
+					reconcilertesting.HaveSubsConfiguration(configDefault(ens.defaultSubscriptionConfig.MaxInFlightMessages)),
+				},
+				natsSubscription: []gomegatypes.GomegaMatcher{
+					reconcilertesting.BeNotNil(),
+					reconcilertesting.BeValid(),
+					reconcilertesting.HaveSubject(reconcilertesting.OrderCreatedEventType),
```
nachtmaar, line 120: nit: out-of-scope because you did not introduce it, but I am interested in your opinion.

If `OrderCreatedEventTypeNotClean ` is the not clean version, wouldn't it be easier to read if `OrderCreatedEventType ` is named `OrderCreatedEventTypeClean` ? 
The reader directly sees: input is **not clean**, output is **clean**. So cleaning worked :)

components/eventing-controller/controllers/subscription/nats/integration_test.go
```diff
@@ -0,0 +1,837 @@
...
+				reconcilertesting.WithSinkURLFromSvc(ens.subscriberSvc),
+			},
+			expect: expect{
+				k8sSubscription: []gomegatypes.GomegaMatcher{
+					reconcilertesting.HaveCondition(
+						eventingv1alpha1.MakeCondition(
+							eventingv1alpha1.ConditionSubscriptionActive,
+							eventingv1alpha1.ConditionReasonNATSSubscriptionNotActive,
+							v1.ConditionFalse, nats.ErrBadSubject.Error(),
+						),
```
nachtmaar, line 140: nit: There is a lot of nesting going on. We could refactor the condition to a function with name `NatsConditionWithErrorBadSubject` 🤔 

```suggestion
						reconcilertesting.NatsConditionWithErrorBadSubject
```

Which version do you prefer ?

components/eventing-controller/controllers/subscription/nats/integration_test.go
```diff
@@ -0,0 +1,837 @@
...
+					),
+				},
+			},
+		},
+
+		{
+			name: "invalid sink; misses 'http' and 'https'",
```
nachtmaar, line 147: this test is reading very well :) input: invalid URL, output: invalid sink condition and k8s event. I do like it 👍

components/eventing-controller/controllers/subscription/nats/integration_test.go
nachtmaar: This is done for `conditionInvalidSink` as well in other tests in this file :)

components/eventing-controller/controllers/subscription/nats/integration_test.go
```diff
@@ -0,0 +1,837 @@
...
+		{
+			name: "valid sink",
+			subscriptionOpts: []reconcilertesting.SubscriptionOpt{
+				reconcilertesting.WithFilter("", reconcilertesting.OrderCreatedEventTypeNotClean),
+				reconcilertesting.WithSinkURL(validSinkURL(ens)),
+			},
+			expect: expect{
+				k8sSubscription: []gomegatypes.GomegaMatcher{
+					reconcilertesting.HaveCondition(reconcilertesting.DefaultCondition("")),
+					reconcilertesting.HaveSubscriptionReady(),
```
nachtmaar, line 256: nit: `reconcilertesting.HaveCondition(reconcilertesting.DefaultCondition(""))` does also test that `reconcilertesting.HaveSubscriptionReady()`.

We should use it consistently (either always both statements or only the first statement). There are some calls to `		reconcilertesting.HaveCondition(reconcilertesting.DefaultCondition("")),` without `reconcilertesting.HaveSubscriptionReady(),`

An idea from another comment is to rename `DefaultCondition ` to `DefaultReadyCondition`. This covers both lines.

```suggestion
					reconcilertesting.HaveCondition(reconcilertesting.DefaultReadyCondition("")),
```

components/eventing-controller/controllers/subscription/nats/integration_test.go
```diff
@@ -0,0 +1,837 @@
...
+				k8sSubscription: []gomegatypes.GomegaMatcher{
+					reconcilertesting.HaveCondition(reconcilertesting.DefaultCondition("")),
+				},
+				natsSubscription: []gomegatypes.GomegaMatcher{
+					reconcilertesting.BeNotNil(),
+					reconcilertesting.BeValid(),
+					reconcilertesting.HaveSubject(reconcilertesting.OrderCreatedEventType),
+				},
+			},
+			shouldTestDeletion: true,
```
nachtmaar, line 308: Why do we sometimes test the deletion and sometimes not ? 🤔 Can't we just always test that the deletion works ?

components/eventing-controller/controllers/subscription/nats/integration_test.go
```diff
@@ -0,0 +1,837 @@
...
+	err := ens.k8sClient.Update(ctx, subscription)
+	g.Expect(err).Should(gomega.BeNil())
+}
+
+func createSubscription(ctx context.Context, g *gomega.GomegaWithT, ens *testEnsemble, subscriptionOpts ...reconcilertesting.SubscriptionOpt) (*eventingv1alpha1.Subscription, string) {
+	subscriptionName := fmt.Sprintf(subscriptionNameFormat, ens.testID)
+	ens.testID++
+	subscription := reconcilertesting.NewSubscription(subscriptionName, ens.subscriberSvc.Namespace, subscriptionOpts...)
+	subscription = createSubscriptionInK8s(ctx, ens, subscription, g)
+	return subscription, subscriptionName
```
nachtmaar, line 569: ultra nit: `subscriptionName` is already part of the returned `subscription`. There is no need to return it.

This would mean we need to write calls like this (example):
```go
testSubscriptionOnK8s(ctx, g, ens, subscription, subscriptionName, tc.expect.k8sSubscription...)
```

to:


```go
testSubscriptionOnK8s(ctx, g, ens, subscription, subscription.Name, tc.expect.k8sSubscription...)
```

I think adding only a dot between subscription and Name is not a big deal. So IMHO there is no need to return `subscriptioName`.

nachtmaar (commented):
Here are my comments for the first round of review. In do prefer the table-driven tests using testing.T instead of ginkgo. 
I left a lot of nits, but I think it is worth discussing it because it could further improve readability. 

I will soon have a second look at it. I did not check `TestChangeSubscription ` yet.

components/eventing-controller/controllers/subscription/nats/integration_test.go
```diff
@@ -0,0 +1,837 @@
...
+			},
+			expectBefore: expect{
+				k8sSubscription: []gomegatypes.GomegaMatcher{
+					reconcilertesting.HaveCondition(reconcilertesting.DefaultCondition("")),
+					reconcilertesting.HaveSubsConfiguration(configDefault(ens.defaultSubscriptionConfig.MaxInFlightMessages)),
+				},
+			},
+			changeSubscription: func(subscription *eventingv1alpha1.Subscription) {
+				eventTypes := []string{
+					uncleanEventType("0"),
```
nachtmaar, line 353: out-of-scope of your PR: IMHO the function would be more intuitive if it would be called `(make)UncleanOrderCreatedEventType("0")`. Same for the `cleanEventType ` function.

What is your opinion on this ?

components/eventing-controller/controllers/subscription/nats/integration_test.go
```diff
@@ -0,0 +1,837 @@
...
+					reconcilertesting.HaveCleanEventTypes([]string{
+						cleanEventType("0"),
+						cleanEventType("1"),
+					}),
+				},
+			},
+		},
+
+		{
+			name: "clean event types; change filters",
```
nachtmaar, line 373: nit: the event types are unclean then the subscription is created:

```suggestion
			name: "unclean event types; change filters",
```

components/eventing-controller/controllers/subscription/nats/integration_test.go
```diff
@@ -0,0 +1,837 @@
...
+			testEventsOnK8s(ctx, g, ens, tc.expectBefore.k8sEvents...)
+			testSubscriptionOnNATS(g, ens, subscriptionName, tc.expectBefore.natsSubscription...)
+
+			tc.changeSubscription(subscription)
+			updateSubscriptionOnK8s(ctx, g, ens, subscription)
+
+			testSubscriptionOnK8s(ctx, g, ens, subscription, subscriptionName, tc.expectAfter.k8sSubscription...)
+			testEventsOnK8s(ctx, g, ens, tc.expectAfter.k8sEvents...)
+			testSubscriptionOnNATS(g, ens, subscriptionName, tc.expectAfter.natsSubscription...)
+			testDeletionOnK8s(ctx, g, ens, subscription, tc.shouldTestDeletion)
```
nachtmaar, line 525: nit: This is a perfect fit for the [GWT pattern](https://github.com/kyma-project/community/blob/main/concepts/eventing-code-guidelines.md#gwt-given-when-then-pattern):

```suggestion
			// given
			subscription, subscriptionName := createSubscription(ctx, g, ens, tc.subscriptionOpts...)

			testSubscriptionOnK8s(ctx, g, ens, subscription, subscriptionName, tc.expectBefore.k8sSubscription...)
			testEventsOnK8s(ctx, g, ens, tc.expectBefore.k8sEvents...)
			testSubscriptionOnNATS(g, ens, subscriptionName, tc.expectBefore.natsSubscription...)
			
			// when
			tc.changeSubscription(subscription)
			updateSubscriptionOnK8s(ctx, g, ens, subscription)

			// then
			testSubscriptionOnK8s(ctx, g, ens, subscription, subscriptionName, tc.expectAfter.k8sSubscription...)
			testEventsOnK8s(ctx, g, ens, tc.expectAfter.k8sEvents...)
			testSubscriptionOnNATS(g, ens, subscriptionName, tc.expectAfter.natsSubscription...)
			testDeletionOnK8s(ctx, g, ens, subscription, tc.shouldTestDeletion)
```

components/eventing-controller/controllers/subscription/nats/integration_test.go
```diff
@@ -0,0 +1,837 @@
...
+	var testCases = []struct {
+		name               string
+		subscriptionOpts   []reconcilertesting.SubscriptionOpt
+		expectBefore       expect
+		changeSubscription func(subscription *eventingv1alpha1.Subscription)
+		expectAfter        expect
+		shouldTestDeletion bool
+	}{
+		{
+			name: "clean event types; add filters to subscription without filters",
```
nachtmaar, line 339: I am trying to find a better name for the test. What is does it adding unclean event types and expecting that they are cleaned later in the status. The current name does not reflect this 🤔 However it is hard to find a good name 😃

Some ideas:

```suggestion
			name: "adding unclean event types eventually leads to cleaned event types",
```

```suggestion
			name: "eventually, adding unclean event types to the spec leads to cleaned event types in the status",
```

```suggestion
			name: "unclean event types in the spec are eventually clean inside the status",
```

components/eventing-controller/controllers/subscription/nats/integration_test.go
```diff
@@ -0,0 +1,837 @@
...
+				)
+			},
+			expectAfter: expect{
+				k8sSubscription: []gomegatypes.GomegaMatcher{
+					reconcilertesting.HaveCondition(reconcilertesting.DefaultCondition("")),
+					reconcilertesting.HaveSubsConfiguration(configDefault(ens.defaultSubscriptionConfig.MaxInFlightMessages)),
+					reconcilertesting.HaveSubscriptionReady(),
+					reconcilertesting.HaveCleanEventTypes([]string{reconcilertesting.OrderCreatedEventType}),
+					gomega.Not(reconcilertesting.HaveCondition(reconcilertesting.MultipleDefaultConditions()[0])),
+					gomega.Not(reconcilertesting.HaveCondition(reconcilertesting.MultipleDefaultConditions()[1])),
```
nachtmaar, line 505: Ensuring that the conditions are not present can be already done in the `expectBefore` function

components/eventing-controller/controllers/subscription/nats/integration_test.go
```diff
@@ -0,0 +1,837 @@
...
+			name: "resolve multiple conditions",
+			subscriptionOpts: []reconcilertesting.SubscriptionOpt{
+				reconcilertesting.WithEmptyFilter(),
+				reconcilertesting.WithWebhookForNATS(),
+				reconcilertesting.WithMultipleConditions(),
+				reconcilertesting.WithSinkURLFromSvc(ens.subscriberSvc),
+			},
+			expectBefore: expect{
+				k8sSubscription: []gomegatypes.GomegaMatcher{
+					reconcilertesting.HaveCleanEventTypes(nil),
```
nachtmaar, line 486: What does this do ? Checking that there are no clean event types inside the status ? Is it just me or is this not obvious ? 🤔 I am interested in your opinion.

nachtmaar: What about `reconcilertesting.HaveEmptyCleanEventTypes` ?

components/eventing-controller/controllers/subscription/nats/integration_test.go
```diff
@@ -0,0 +1,837 @@
...
+					reconcilertesting.BeNotNil(),
+					reconcilertesting.HaveSubject(cleanEventType("")),
+					reconcilertesting.BeValid(),
+				},
+			},
+			shouldTestDeletion: true,
+		},
+
+		{
+			name: "resolve multiple conditions",
```
nachtmaar, line 477: I wonder why this is not tested inside `TestCreateSubscription` instead. I don't see the need to change something in order to test that multiple conditions are cleaned up and that there is only one condition at the end.

nachtmaar: The cleaning of `OrderCreatedEventTypeNotClean` can be tested without testing the multiple conditions IMHO.

components/eventing-controller/testing/test_helpers.go
```diff
@@ -65,6 +66,18 @@ const (
 
 type APIRuleOption func(r *apigatewayv1alpha1.APIRule)
 
+func GetFreePort() (port int, err error) {
+	var a *net.TCPAddr
+	if a, err = net.ResolveTCPAddr("tcp", "localhost:0"); err == nil {
+		var l *net.TCPListener
+		if l, err = net.ListenTCP("tcp", a); err == nil {
+			defer l.Close()
```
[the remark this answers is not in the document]
FriedrichWilken: I actually started o use the `GetFreePort` func :D. I also added err handling via gomega
```
	natsPort, err := reconcilertesting.GetFreePort()
	g.Expect(err).ToNot(gomega.HaveOccurred())
	ens := setupTestEnsemble(ctx, reconcilertesting.EventTypePrefix, g, natsPort)
```

since the  `metricsPort` is not needed anywhere else it now gets created where metrics gets started.

components/eventing-controller/testing/test_helpers.go
```diff
@@ -65,6 +66,18 @@ const (
 
 type APIRuleOption func(r *apigatewayv1alpha1.APIRule)
 
+func GetFreePort() (port int, err error) {
```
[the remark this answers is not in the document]
FriedrichWilken: I tried ... 😄 . will take that doc, thanks.

components/eventing-controller/testing/matchers.go
```diff
@@ -237,3 +238,25 @@ func HaveValidBEBNamespace(bebNamespaceKey, namespace string) gomegatypes.Gomega
 		return false
 	}, BeTrue())
 }
+
+//
+// NATS subscriptions
+//
+
+func BeValid() gomegatypes.GomegaMatcher {
```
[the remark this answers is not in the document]
FriedrichWilken: renamed them so the names reflect that they work with nats subscriptions

components/eventing-controller/controllers/subscription/nats/integration_test.go
```diff
@@ -0,0 +1,837 @@
...
+	k8sEvents        []v1.Event
+	natsSubscription []gomegatypes.GomegaMatcher
+}
+
+//TestUnavailableNATSServer tests if a subscription is reconciled properly when the NATS backend is unavailable.
+func TestUnavailableNATSServer(t *testing.T) {
+	natsPort := 4220
+	ctx := context.Background()
+	g := gomega.NewGomegaWithT(t)
+	ens := setupTestEnsemble(ctx, reconcilertesting.EventTypePrefix, g, natsPort, 7070)
```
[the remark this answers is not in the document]
FriedrichWilken: well, couldt I even just defer cancel where it gets created and actually move it out of the ensemble struct?

components/eventing-controller/controllers/subscription/nats/integration_test.go
FriedrichWilken: true, removed the Ready and renamed to `DefaultReadyCondition `

components/eventing-controller/controllers/subscription/nats/integration_test.go
FriedrichWilken: I am also not sure. I carried this over from the original tests. My hunch was, that the deletion originally only was done to  prevent collisions with subscription names due to the old tests setup. But since i can't say for sure, I kept it. Further, my arg for not doing it always, that this restructure was supposed to be a step away from test always everything. I was also half-in-favor for always deleting, but came decided to go the current way.

components/eventing-controller/controllers/subscription/nats/integration_test.go
FriedrichWilken: agree

components/eventing-controller/controllers/subscription/nats/integration_test.go
FriedrichWilken: mhh, I guess I again created `uncleanEventType` with the idea of c#'s properties. I am also not completely happy with this.

components/eventing-controller/controllers/subscription/nats/integration_test.go
FriedrichWilken: changed it to `cleanEventTypes` to clarify what this is about.

components/eventing-controller/controllers/subscription/nats/integration_test.go
FriedrichWilken: added

components/eventing-controller/controllers/subscription/nats/integration_test.go
FriedrichWilken: I just renamed them to `cleanEventTypes` since this is what they try to test. the fact that you didn't get that is proof, that my naming was terrible :D

components/eventing-controller/controllers/subscription/nats/integration_test.go
FriedrichWilken: no, I think the idea of the original author was, that these conditions disappear after updating the subscription

components/eventing-controller/controllers/subscription/nats/integration_test.go
FriedrichWilken: I see your point. The problem is that we have to distinguish between `nil` and `[]string{}`. The latter I would call `EmptyCleanEventType` while the first can be better described as `HaveCleanEventTypes(nil)`.

components/eventing-controller/controllers/subscription/nats/integration_test.go
FriedrichWilken: this test case is special. it was the last one i moved over from the old tests, because I also struggled with understanding, what the goal of this test actually was.

FriedrichWilken: we can sync on this, if you want

components/eventing-controller/controllers/subscription/nats/integration_test.go
FriedrichWilken: forgot to add here, that after a conversation with the po, i made deletion tests obligatory

components/eventing-controller/testing/test_helpers.go
```diff
@@ -65,6 +66,20 @@ const (
...
+	if a, err = net.ResolveTCPAddr("tcp", "localhost:0"); err == nil {
+		var l *net.TCPListener
+		if l, err = net.ListenTCP("tcp", a); err == nil {
+			defer l.Close()
+			return l.Addr().(*net.TCPAddr).Port, nil
+		}
+	}
+	return
+}
+
```
nachtmaar, line 82: This suggestion will return the error from `l.Close` as well

```suggestion
// GetFreePort determines a free port on the host. It does so by delegating the job to net.ListenTCP.
// Then providing a port of 0 to net.ListenTCP, it will automatically choose a port for us.
func GetFreePort() (port int, err error) {
	var a *net.TCPAddr
	if a, err = net.ResolveTCPAddr("tcp", "localhost:0"); err == nil {
		var l *net.TCPListener
		if l, err = net.ListenTCP("tcp", a); err == nil {
			port := l.Addr().(*net.TCPAddr).Port
			err = l.Close()
			return port, err
		}
	}
	return
}
```

components/eventing-controller/controllers/subscription/nats/integration_test.go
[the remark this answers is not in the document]
FriedrichWilken: it turns out, that this would make everything more complicated. the current solution is the best compromise. :)

components/eventing-controller/testing/test_helpers.go
FriedrichWilken: i take that

components/eventing-controller/controllers/subscription/nats/integration_test.go
FriedrichWilken: this is for a followup pr

nachtmaar (approved):

PR #13412, opened 2022-02-16
Add missing method to CORS for release 1.24

### LINKED ISSUE #11126: CORS allowMethods missing methods configured in API rule
**Description**

After configuring an API rule with a PATCH method, browser PATCH requests would fail due to CORS error:  `PATCH is not allowed by Access-Control-Allow-Methods in preflight response`. Further investigation revealed that the associated istio virtual-service configuration was missing the `PATCH` in the `allowMethods` array.

```yaml
spec:
  ...
  http:
  - corsPolicy:
      allowHeaders:
      - Authorization
      - Content-Type
      - '*'
      allowMethods:
      - GET
      - POST
      - PUT
      - DELETE
      # Note the missing PATCH here
      allowOrigins:
      - regex: .*
    match:
    - uri:
        regex: /v1/.*
  ...
```

Adding this manually resolved the issue. Note it needed to be manually added to all of the relevant paths.

**Expected result**

Expected that the methods configured in the API rule would be allowed by CORS in the virtual service.

**Actual result**

The configured methods were missing from the vs.

**Steps to reproduce**

1. Create a service pointing to a deployment that will accept a patch request
2. Configure an API rule for that service which accepts the PATCH method
3. Make a PATCH request from a browser and observe the request is blocked due to CORS policy

**Troubleshooting**

GET, POST, PUT, and DELETE calls were all successful but PATCH calls were blocked.

dariusztutaj:
**Description**
Adding missing PATCH method to CORS in api-gsateway

**Related issues:**
Fixes #11126 in release 1.24

### DIFF
--- a/resources/api-gateway/values.yaml
+++ b/resources/api-gateway/values.yaml
@@ -59,6 +59,7 @@ config:
       - "POST"
       - "PUT"
       - "DELETE"
+      - "PATCH"
     allowHeaders: # "*" is not yet supported by all browsers
       - "Authorization"
       - "Content-Type"


dariusztutaj:
Decision was made to add this feature only to main branch

PR #13414, opened 2022-02-17
backport: enable containerd support in fluent-bit

a-thaler:
**Description**

- ...
- ...
- ...

**Related issue(s)**

https://github.com/kyma-project/kyma/pull/12554


a-thaler:
All options could be set via overrides as well and indeed a dedicated backport is not mandatory, so we will skip that

Issue #12889, opened 2021-12-16
Separate GIT related logic from function reconciliation - Propose refactoring


dbadura:
We want to divide the function-controller to remove git related logic (cloning) from reconciling of functions to separate controller and reduce the cloning operations to bare minimum.

It can be achieved by two steps:
1. Create git-controller which will follow the algorithm presented here: ![git_flow_git_controller](https://user-images.githubusercontent.com/45110612/146395689-7fa345ae-3731-4b56-a534-340619d62f93.png).
The function will be triggered to rebuild. This can be achieved by watching git repo from function using [watchers](https://book.kubebuilder.io/reference/watching-resources.html), and changed algorithm can look like this: ![git_flow_fn_controller](https://user-images.githubusercontent.com/45110612/146395613-3edb46c1-94c1-458a-ae6f-272cc8bbb31e.png)

This will allow the function to rebuild just after git check of git repository.
The git repository will have configurable recheck duration.

The pros of this approach are:
- removing the cloning from function controller will decrease the amount of git cloning. Cloning wil be done once in git-controller, unlike now that is made for every function.
- the function controller won't have any git library reference.

To consider:
- Fetching last commit can be improved by using cache dir and force fetching latest changes.
- We should have ability to check last commit in given directory. Function can be in one repository in different directory on the same branch. We should avoid situation like this, that functions are rebuilt because one of them changed on branch.

2.  Introduction of shared volume, which can be used to cache: git repositories, inline functions (small projects can be created).
This volume can be used for building for kaniko instead of using repo fetcher (additional clone operation) and will result in reduction of  git cloning to the bare minimum.

To consider:
- The volume can be replaced by fetch API inside the cluster.

The second step is WIP.

See: https://github.com/kyma-project/kyma/issues/12527

hwinkel:
is that not very similar to what flux-cd or tekton already does ?

pPrecel:
> is that not very similar to what flux-cd or tekton already does ?

Probably yes, but flux-cd and tekton have been adding a lot more functionalities and functions - I think we should not introduce another tool and use his one feature only.

pPrecel:
There is no need to introduce a new label for git functions. For filter functions, we can use field selectors instead of label selectors.

dbadura:
Instead of sending event to fn reconciler, we can use [annotations](https://kubernetes.io/docs/concepts/overview/working-with-objects/annotations/) and annotate the function to trigger the reconciliation.

[...]

PR #13421, opened 2022-02-17
Serverless new cr

dbadura:
**Description**

- Proposal for new serverless cr

**Related issue(s)**

See: https://github.com/kyma-project/kyma/issues/12889


components/serverless/api/v1beta1/function_types.go
```diff
@@ -0,0 +1,206 @@
...
+type Source struct {
+	// +optional
+	ArtifactSource *ArtifactSource `json:"artifactSource"`
+	//TODO: we should check if it is possible
+	// +optional
+	CRDSource *CRDSource `json:"crdSource"`
+	// +optional
+	GitRepositorySource *GitRepositorySource `json:"gitRepositorySource"`
+	// +optional
+	InlineSource *InlineSource `json:"inlineSource"`
```
moelsayed, line 43: I think the `Source` suffix is redundant there.

components/serverless/api/v1beta1/function_types.go
```diff
@@ -0,0 +1,206 @@
...
+	Name      string `json:"name"`
+	Namespace string `json:"namespace"`
+}
+
+type CRDSource struct {
+	Api        string `json:"api"`
+	Version    string `json:"version"`
+	Kind       string `json:"kind"`
+	Name       string `json:"name"`
+	Namespaces string `json:"namespaces"`
```
moelsayed, line 61: Do you thing it makes sense to use `ObjectMeta` and `TypeMeta` for these fields instead?

components/serverless/api/v1beta1/function_types.go
```diff
@@ -0,0 +1,206 @@
...
+
+type ArtifactSource struct {
+	//TODO: maybe it's worth do distinguish between local and external, as we know the internal URL
+	URL     string `json:"url"`
+	BaseDir string `json:"baseDir"`
+	// +optional
+	Credentials *Credentials `json:"credentials"`
+}
+
+type ConfigMapSelector struct {
```
moelsayed, line 76: I don't think `selector` is the correct term here?

components/serverless/api/v1beta1/function_types.go
```diff
@@ -0,0 +1,206 @@
...
+	Password v1.SecretKeySelector `json:"password"`
+}
+
+type Oauth struct {
+	OauthURL string               `json:"oatuhURL"`
+	User     v1.SecretKeySelector `json:"user"`
+	Password v1.SecretKeySelector `json:"password"`
+}
+
+type Template struct {
```
moelsayed, line 105: What is this for?

components/serverless/api/v1beta1/function_types.go
```diff
@@ -0,0 +1,206 @@
...
+	RuntimeProfile *string `json:"runtimeProfile"`
+
+	// +optional
+	Resources *v1.ResourceRequirements `json:"resources,omitempty"`
+
+	// +optional
+	BuildProfile *string `json:"buildProfile"`
+
+	// +optional
+	BuildResources *v1.ResourceRequirements `json:"buildResources,omitempty"`
```
moelsayed, line 138: How do these fields work together?

moelsayed (commented):
Just a a few questions

components/serverless/api/v1beta1/function_types.go
dbadura: You're right, it should be Ref.

components/serverless/api/v1beta1/function_types.go
dbadura: This is for defining annotations and labels on deployment of function.

components/serverless/api/v1beta1/function_types.go
dbadura: Can those objects be used to reference particular crd?
I looked at them and they carry many information which I think we don't need.

components/serverless/api/v1beta1/function_types.go
dbadura: I wanted to make it similar to the [k8s volume source](https://github.com/kubernetes/api/blob/master/core/v1/types.go#L49)
Generally you have to choose only one of them.

dbadura: You can choose predefined profile or create custom one.

components/serverless/api/v1beta1/function_types.go
dbadura: Removed

components/serverless/api/v1beta1/function_types.go
dbadura: Could you give me a example of what those 2 objects expose and can be helpful?

components/serverless/api/v1beta1/function_types.go
dbadura: What about this:
```
build:
 - profile
 - resources
```
```
runtime:
 - profile
 - resources
```
And only one of those options should be filled.

dbadura: New idea
```
requirements:
 -build:
    -profile
    -resources
 -runtime
  - profile
  - resources
  ```

Comment (moelsayed) [reply components/serverless/api/v1beta1/function_types.go]: I like the latest idea more. And you are right about the `runtime` name. I would also suggest `deployment` or `instance` for this.

File: components/serverless/api/v1beta1/function_types.go (hunk 2 above)
Comment (moelsayed) [reply components/serverless/api/v1beta1/function_types.go:61]: Nothing comes to my mind at this point.

File: components/serverless/api/v1beta1/function_types.go (hunk 5 above)
Comment (dbadura) [reply components/serverless/api/v1beta1/function_types.go]: What do you think about `function`?
```
requirements/resourceConfig:
 -build:
    -profile
    -resources
 -function
  - profile
  - resources
  ```

dbadura: I've updated the cr.

PR #13432, opened 2022-02-18
Report function pod status in fast-integration SVCAT migration test

wozniakjan:
**Description**

The redundant check for pod items length resulted in a suboptimal error message
```
  1) SKR SVCAT migration test
       Should check if pod presets injected secrets to functions containers:
     TypeError: Cannot read property 'metadata' of undefined
      at Object.checkPodPresetEnvInjected (skr-svcat-migration-test/test-helpers.js:80:32)
      at processTicksAndRejections (internal/process/task_queues.js:95:5)
      at async Context.<anonymous> (skr-svcat-migration-test/skr-svcat-migration-test.js:103:5)
```

This should now correctly report the status of the pod even if there is correctly one but not in `Running`.

**Related issue(s)**

### DIFF
--- a/tests/fast-integration/skr-svcat-migration-test/test-helpers.js
+++ b/tests/fast-integration/skr-svcat-migration-test/test-helpers.js
@@ -63,12 +63,10 @@ async function getFunctionPod(functionName) {
     }
     sleep(10000);
   }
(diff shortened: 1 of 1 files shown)


tillknuesting (approved):
lgtm

Issue #13321, opened 2022-02-08
Handle enabling the feature flag in the Eventing component reconciler


pxsalehi:
Once we merge the JetStream feature flag in #13251, we also need to address how this can be enabled using the Kyma reconciler. As of https://github.com/kyma-project/kyma/pull/13259, I have tested the Kyma reconciler using Kyma CLI unstable version `main-9feb3a3`.

- Upgrade from `main` (https://github.com/kyma-project/kyma/commit/75da62ce99a9bc773759afb3e76da28de715f574, a 2.x version w/o the feature flag) to `PR-13259` (with feature flag enableJetStream=false): works w/o any reconciler changes required. The only place the feature flag shows up is in the eventing-controller deployment env vars.

- Installing `main` (enableJetStream=true): works w/o any reconciler changes required. It also correctly brings up a JetStream cluster.

- Upgrading from  `main` (enableJetStream=false) to `main` (enableJetStream=true): It correctly updates the eventing-controller deployment and the NATS ConfigMap (`eventing-nats-config`), however, you'd need to restart the NATS servers (or the StatefulSet scaled to 0 and back up) to have JetStream enabled. I think, upgrading from `main` w/o any feature flag directly to enabled feature flag would also require the restart, as the NATS cluster is already running.

Depending on how and when we plan to release the JetStream backend, most likely we would require some pre-action in the eventing component reconciler that would perform the NATS cluster restart.

VladislavPaskar:
Unfortunately I couldn't reproduce  the `false->true` upgrade scenario, but found out that the JS-specific data is not being removed from the Nats StatefulSet and Pod(s) during the `true->false` scenario. I added a PR, which should fix that problem: https://github.com/kyma-incubator/reconciler/pull/864

pxsalehi:
Which part couldn't you reproduce? You mean after enabling the flag, no restart of the NATS server (i.e. recreating the pod) was required? If that is the case, then we would not need any change on the reconciler.

This issue is about enabling the flag. Do we need to support disabling it?

Why are you removing the StatefulSet? I think removing the pod, or scaling down and then up should be enough.

VladislavPaskar:
I couldn't reproduce the part when provisioning from `enableJetStream=false` to `enableJetStream=true`, which is the third bullet point in the issue. In that case provisioner was adding the Jetstream related data to the StatefulSet (in our case that is the js secret to the pod's envs). 

[...]

PR #13496, opened 2022-02-24
Add env to Nats StatefulSet to distinguish if jetstream is enabled

VladislavPaskar:
**Description**

- add env var to nats container for reconciler to be able to recongnize that whether the jetstream chart changes were applied to the nats stateful set or not 
- there is a reconciler PR, which depends on this change: https://github.com/kyma-incubator/reconciler/pull/864

**Related issue(s)**
https://github.com/kyma-project/kyma/issues/13321

### DIFF
--- a/resources/eventing/charts/nats/templates/statefulset.yaml
+++ b/resources/eventing/charts/nats/templates/statefulset.yaml
@@ -131,6 +131,10 @@ spec:
           value: {{ include "nats.clusterAdvertise" . }}
 
         {{- if .Values.global.features.enableJetStream }}
+        - name: JS_ENABLED
+          value: "true"
+        {{- end }}
+        {{- if .Values.global.features.enableJetStream }}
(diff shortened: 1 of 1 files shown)


pxsalehi (changes requested):
Why an environment variable? Why not a label or an annotation? Is the env var used by the NATS server?

VladislavPaskar:
> Why an environment variable? Why not a label or an annotation? Is the env var used by the NATS server?

That was my first thought too, but I made it env variable next to some over JS-related env variable in order to be sure that their rendering via charts will be the same.

Let's take the disabling the JS-feature scenario. Using the latest reconciler the eventing controller's label was set to from `true` to `false`. The configmap's JS-related data got removed, but the JS-related env variable in the statefulSet's podTemplate remained. That's why I added an env variable as an indicator if all the JS-data got removed.

PR #13505, opened 2022-02-25
remove var name and import name collisions

FriedrichWilken:
This PR resolves name conflicts between variables and imported packages.


components/eventing-controller/controllers/subscription/beb/unit_test.go
```diff
@@ -8,7 +8,7 @@ import (
 	"github.com/pkg/errors"
 
 	kymalogger "github.com/kyma-project/kyma/common/logging/logger"
-	"github.com/kyma-project/kyma/components/eventing-controller/logger"
+	eventinglogger "github.com/kyma-project/kyma/components/eventing-controller/logger"
```
mfaizanse, line 11: (nit) should we use camel case? `eventingLogger`?

mfaizanse (approved):

PR #13506, opened 2022-02-25
rename files for NATS reconciler tests

FriedrichWilken:
To have more [logical named](https://scene-si.org/2019/04/15/next-level-go-testing/#public-vs-private-tests-apis) test files this PR renames the NATS reconciler test files.

That's it. That's the PR.


pxsalehi (changes requested):
Why not rename the package of `.../beb/reconciler_test.go` to `beb_test`?

FriedrichWilken:
> Why not rename the package of `.../beb/reconciler_test.go` to `beb_test`?

we have 
`reconciler.go` and `equality.go` 
and for each of them we have tests
`reconciler_internal_test.go`, `reconciler_test.go` and equality_test.go`
so I think with the current naming pattern it's clear what file contains what tests.

pxsalehi:
> > Why not rename the package of `.../beb/reconciler_test.go` to `beb_test`?
> 
> in `beb` we have `reconciler.go` with its tests in `reconciler_internal_test.go` and `reconciler_test.go` `equality.go` with its tests in `equality_test.go` so I think with the current naming pattern it's clear what file contains what tests.
> 
> What do you think?

But in `.../nats/reconciler_test.go`, the package is `nats_test`. Why aren't we following the same pattern?

FriedrichWilken:
> > > Why not rename the package of `.../beb/reconciler_test.go` to `beb_test`?
> > 
> > 
> > in `beb` we have `reconciler.go` with its tests in `reconciler_internal_test.go` and `reconciler_test.go` `equality.go` with its tests in `equality_test.go` so I think with the current naming pattern it's clear what file contains what tests.
> > What do you think?
> 
> But in `.../nats/reconciler_test.go`, the package is `nats_test`. Why aren't we following the same pattern?

we use only `nats_test` because we restructured these tests very recently. this still needs do be done in the future for beb.

pxsalehi (dismissed):

pxsalehi (approved):

VladislavPaskar (approved):

Issue #13386, opened 2022-02-14
Verify apiserver metrics before 2.1 rollout


a-thaler:
**Description**
validate that we have not increased prometheus footprint by latest prom upgrade

apiserver metrics are talking 1/3 of all metrics, so we need to be sure that we have not increased the amount by unused rules and metrics. 

**Criterias**
- There is only one apiserver dashboard left showing meaningful like latencies (not strange SLO/SLA calculations)
- Rules are active required by the dashboard and essential alerts (do we need kubeApiserverSlos: true?)

dennis-ge:
Findings:
- The only apiserver metric with high utilization was `apiserver_request_duration_seconds_bucket`. We reduced the footprint of this metric by only considering some selected buckets and not all (initially around 30 buckets, now only 6). This was done in the first PR. 
- The apiserver dashboard used the previous mentioned metric and some other metrics with low count (no recording rules are used in the dashboard)
- The `kubeApiserverSlos` rules were disabled with this PR as they include SLO/SLA calculations
- The second PR reduced the footprint of the second highest metric in Prometheus, the `rest_client_request_latency_seconds_bucket` metric. We reduce the footprint of this metric two ways:
    - Reduce amount of buckets for the metric
    - Only consider metric data from specific k8s API groups (core, apps, batch and one special API group, e.g., jaegertracing.io for the tracing component)
- Moreover, the telemetry component got upgraded in the second PR to a newer controller-runtime where the metric is optional and disabled by default.

PR #13508, opened 2022-02-28
reduce apiserver metrics footprint

dennis-ge:
**Description**

- disable apiserver slo metrics
- reduce amount of buckets for `apiserver_request_duration_seconds_bucket` metric. Preferred to allowlisting 6 fixed buckets instead of denylisting all other buckets with an ugly regex (e.g. `'apiserver_request_duration_seconds_bucket;(0\.[0235689].*|[0-9]\.[1-9]?5.*|[2-4|6-9]|[1-6][05])`)

**Related issue(s)**
Relates to #13386


chrkl (approved):
Please check with @a-thaler if we should add this to the release notes

PR #13522, opened 2022-03-01
Refactor reconciler test functions

VladislavPaskar:
**Description**

Currently we have two very similar `NewSubscription` constructors in our `components/eventing-controller/testing/test_helpers.go` and `components/eventing-controller/controllers/subscription/nats/unit_test.go` files. Also with have configuration options for Subscription in both nats and test_helpers files. 

The idea or this PR is to refactor them in order to make the code more readable and structured. 

- move configuration options for Subscription into one place
- change the nats-reconciler specific subscription constructor for usage of an already existing code

### DIFF
--- a/components/eventing-controller/controllers/subscription/nats/unit_test.go
+++ b/components/eventing-controller/controllers/subscription/nats/unit_test.go
@@ -1,14 +1,15 @@
 // This file contains unit tests for the NATS subscription reconciler.
-// It uses the testing.T and gomega testing libraries to perform the assertions.
+// It uses the testing.T and stretchr/testify libraries to perform assertions.
 // TestEnvironment struct mocks the required resources.
 package nats
 
 import (
 	"context"
-	"path/filepath"
 	"strings"
 	"testing"
 	"time"
+
+	"github.com/stretchr/testify/require"
 
 	kymalogger "github.com/kyma-project/kyma/common/logging/logger"
 	eventingv1alpha1 "github.com/kyma-project/kyma/components/eventing-controller/api/v1alpha1"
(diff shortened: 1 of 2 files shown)


raypinto:
According to the testing guidelines, we would like to get rid of ginko and gomega for unit tests. How about we do it as part of this refactoring for this file?

VladislavPaskar:
> Looks good! According to the testing guidelines, we would like to get rid of ginko and gomega for unit tests. How about we do it as part of this refactoring for this file?

good idea, I removed them and made some other things smoother

raypinto:
> > Looks good! According to the testing guidelines, we would like to get rid of ginko and gomega for unit tests. How about we do it as part of this refactoring for this file?
> 
> good idea, I removed them and made some other things smoother

Thank you. As discussed, looks like the duplicated code is back again with the new commit. Would suggest to remove it and use the logic as in the first commit.

components/eventing-controller/controllers/subscription/nats/unit_test.go
```diff
@@ -29,81 +30,84 @@ import (
...
 
-const namespaceName = "test"
+const (
+	subscriptionName = "testSubscription"
+	namespaceName    = "test"
+)
 
 var (
 	defaultSubsConfig = env.DefaultSubscriptionConfig{MaxInFlightMessages: 1, DispatcherRetryPeriod: time.Second, DispatcherMaxRetries: 1}
 	testEnv           *envtest.Environment
```
raypinto, line 40: is this really needed?

components/eventing-controller/controllers/subscription/nats/unit_test.go
```diff
@@ -457,76 +458,45 @@ func setupTestEnvironment(t *testing.T) *TestEnvironment {
...
-	return func(sub *eventingv1alpha1.Subscription) {
-		sub.Status.Ready = status
-	}
-}
-func WithFinalizers(finalizers []string) SubscriptionOpt {
-	return func(sub *eventingv1alpha1.Subscription) {
-		sub.ObjectMeta.Finalizers = finalizers
-	}
+// NewTestSubscription creates a test subscription
+func NewTestSubscription(opts ...controllertesting.SubscriptionOpt) *eventingv1alpha1.Subscription {
```
raypinto, line 474: I was wondering if this is needed (you can use the call within the function directly). But I think from a readability point of view, its good!

components/eventing-controller/controllers/subscription/nats/unit_test.go
```diff
@@ -196,94 +200,93 @@ func Test_syncSubscriptionStatus(t *testing.T) {
...
 		{
 			name:                  "With no existing svc in the cluster, service has the wrong name",
 			givenSubscriptionSink: "https://eventing-nats.test.svc.cluster.local:8080",
 			givenSvcNameToCreate:  "eventing-nats",
-			expectedErrString:     "",
+			wantErrString:         "",
 		},
 	}
 	for _, tC := range testCases {
 		testCase := tC
```
raypinto, line 283: I think this is an additional assignment that is not needed. You can write `testCase` instead of `tC` in the above line. Same comment for all occurrences in this file.

components/eventing-controller/controllers/subscription/nats/unit_test.go
```diff
@@ -1,5 +1,5 @@
 // This file contains unit tests for the NATS subscription reconciler.
-// It uses the testing.T and gomega testing libraries to perform the assertions.
+// It uses the testing.T testing library to perform assertions.
```
raypinto, line 2:
```suggestion
// It uses the testing.T and stretchr/testify testing libraries to perform assertions.
```

components/eventing-controller/controllers/subscription/nats/unit_test.go
```diff
@@ -457,76 +458,45 @@ func setupTestEnvironment(t *testing.T) *TestEnvironment {
 	}
 }
 
-func createFakeClient(g *GomegaWithT) client.WithWatch {
+func createFakeClient(t *testing.T) client.WithWatch {
 	testEnv = &envtest.Environment{
```
raypinto, line 462: I removed this assignment and the tests looked good. Probably this is not necessary here?

components/eventing-controller/controllers/subscription/nats/unit_test.go
VladislavPaskar: Yes, as I explained to in yesterday this is needed for the readability purpose

components/eventing-controller/controllers/subscription/nats/unit_test.go
VladislavPaskar: Yes, because I want to use constants instead of magic strings

components/eventing-controller/controllers/subscription/nats/unit_test.go
VladislavPaskar: https://stackoverflow.com/questions/68559574/using-the-variable-on-range-scope-x-in-function-literal-scopelint

components/eventing-controller/controllers/subscription/nats/unit_test.go
VladislavPaskar: I don't really understand what you removed, that the tests remained green ? Here I am registering our eventing crds including subscription into the testing scheme. Without that you get something like `no kind is registered for the type v1alpha1.Subscription`

components/eventing-controller/controllers/subscription/nats/unit_test.go
VladislavPaskar: At the point of time it was written, the IDE was complaining about it, strangely not anymore, I will remove it

components/eventing-controller/controllers/subscription/nats/unit_test.go
VladislavPaskar: Yes, you are right, `testEnv` is not needed here, as these are unit tests, we mock here everything and we are not using k8s API server

components/eventing-controller/controllers/subscription/nats/unit_test.go
VladislavPaskar: updated

components/eventing-controller/controllers/subscription/nats/unit_test.go
VladislavPaskar: Yes, the `golangci-lint` is complaining about ` Using the variable on range scope `testCase` in function literal` in that case

components/eventing-controller/controllers/subscription/nats/unit_test.go
VladislavPaskar: I see now, that you mentioned the `testEnv` variable, it is removed now

VladislavPaskar:
> I just remembered: Could you also update the eventing-controller image tag?

As disscused in the team, we don't do this for the tests

raypinto (approved):

PR #13535, opened 2022-03-02
Fix pre-main-skr-kyma-eventing

VladislavPaskar:
**Description**

- do not register a compass scenario with an already used name, which broke the pipeline in the past
- ...
- ...

**Related issue(s)**


VladislavPaskar:
close in favour of https://github.com/kyma-project/kyma/pull/13574

PR #13571, opened 2022-03-08
Add correlation ID to requests in compass runtime agent 

krasish:
**Description**
As a step to improve the integration with [compass](https://github.com/kyma-incubator/compass), correlation IDs can be very useful for easier troubleshooting of issues with provisioning/resyncing. With this PR, a request ID per each reconciliation loop of the compass connection controller is generated and propagated (as the "x-request-id" HTTP header) to all HTTP requests sent to Compass. 
The format of the request-id is the following: 
```<runtime-id>_<generated-uuid>```

- Add correlation ID transport to go HTTP clients used across modules
- Introduce a context function argument to all Interfaces which end up as an HTTP request to CMP
- Create a context for each reconciliation loop and populate it with a correlation ID. Pass the context where necessary.
- Regenerate mocks & adapt unit tests

### CODE REVIEW

File: components/compass-runtime-agent/internal/compassconnection/package_test.go [hunk 1]
```diff
@@ -106,8 +110,18 @@ var (
 	}
 
 	kymaModelApps = []kymaModel.Application{{Name: "App-1", ID: "abcd-efgh"}}
-
-	operationResults = []kyma.Result{{ApplicationName: "App-1", ApplicationID: "abcd-efgh", Operation: kyma.Create}}
+	testCtx       = correlation.SaveCorrelationIDHeaderToContext(context.Background(), str.Ptr(correlation.RequestIDHeaderKey), str.Ptr("test-correlation-id-123"))
+
+	operationResults    = []kyma.Result{{ApplicationName: "App-1", ApplicationID: "abcd-efgh", Operation: kyma.Create}}
+	requestIDCtxMatcher = func(ctx context.Context) bool {
```
koala7659, line 116: With new matching function I see there are two different conventions used for matching mocked parameters in thests. I think both matching unctions should be uniformed and used in the same way.  So we can declare this function here with `mock.MatchedBy`

components/compass-runtime-agent/internal/compassconnection/package_test.go
```diff
@@ -68,7 +72,7 @@ var (
 			}
 		}
 		// only check if x-request-id header was set
-		return input[RequestIDHeader] != ""
+		return true
```
koala7659, line 75: We should remove RequestIDHeader processing from that iteration. It is not used anymore for request identification. This can be also simplified, since we use only one header in the request `ConnectorTokenHeader`.

koala7659 (changes requested):
Please see my comments in the code. Unfortunately I did not find the reason why fast integration tests fail with your PR.

components/compass-runtime-agent/internal/compassconnection/connector.go
```diff
@@ -25,8 +27,8 @@ const (
```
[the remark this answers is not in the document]
krasish: Done ✅

components/compass-runtime-agent/go.mod
```diff
@@ -57,6 +57,7 @@ require (
 	github.com/spf13/pflag v1.0.5 // indirect
 	github.com/stretchr/objx v0.3.0 // indirect
 	github.com/vektah/gqlparser/v2 v2.1.0 // indirect
+	github.com/vektra/mockery v1.1.2 // indirect
 	github.com/xeipuuv/gojsonpointer v0.0.0-20190905194746-02993c407bfb // indirect
```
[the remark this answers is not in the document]
krasish: Done ✅

components/compass-runtime-agent/internal/compassconnection/package_test.go
krasish: Done ✅

components/compass-runtime-agent/internal/compassconnection/package_test.go
krasish: Done ✅

components/compass-runtime-agent/cmd/main.go
```diff
@@ -102,8 +107,21 @@ func main() {
 	compassConnectionSupervisor, err := controllerDependencies.InitializeController()
 	exitOnError(err, "Failed to initialize Controller")
```
koala7659, line 109: This is just a test code for POC. Please notice that configProvider.GetRuntimeConfig() may just fail while compass-runtime-agent is waiting for configuration secret to be created. 

In the case of failing k3d jobs the configuration secret `compass-agent-configuration` from which you read runtime id is provided in the testing suite AFTER Kyma is fully installed on k3d. As a result  compass-runtime-agent should assume that this secret is missing and just wait until it is ready.

components/compass-runtime-agent/internal/compassconnection/init.go
```diff
@@ -56,7 +56,16 @@ func (config DependencyConfig) InitializeController() (Supervisor, error) {
 		config.RuntimeURLsConfig,
 		config.ConnectionDataCache)
```
koala7659, line 58: This is just a POC code to test if the job will succeed. Please see my previous comment.

koala7659 (commented):
Please see my comments about reading runtimeID with Provider interface

Issue #13114, opened 2022-01-19
Make istio-configuration resposible only for managing istio installation


strekm:
**Description**

Istio kyma component has few responsibilities: manage istio and apply kyma istio related resources. that proves to be difficult to maintain in istio specific reconciler ([more details](https://github.com/kyma-incubator/reconciler/issues/543)). Idea is to split responsibilities into 2 kyma components: 
- istio holding istiooperator CR with custom reconciler only resposnible for managing istio installation and related actions (e.g. restart of sidecars). this component should stay as pre-requisite component. 
- istio-resources containing all istio related resources like grafana dashboards, mtls conf, gateway, istio healthz VS and so on. this component should be regular component. this part is out of scope of this task.

Istio-configuration component should not contain any istio related resources. only part of this chart should be istiooperator CR.

AC:
- [x] istio-configuration component name changed to istio
- [x] monitoring and mtls conf is not part of istio chart
- [x] docu updated
- [x] RN created
- [x] CODEOWNERS updated
- [x] components.yaml adjusted
- [ ] config for KCP adjusted if necessary

**Reasons**

having well defined, single responsibility components

**Attachments**

part of https://github.com/kyma-incubator/reconciler/issues/543

strekm:
configuration in management-plane will be handled with https://github.tools.sap/kyma/backlog/issues/2453

PR #13603, opened 2022-03-10
Rename istio-configuration to istio in components.yaml

cnvergence:
**Related issue(s)**

See also https://github.com/kyma-project/kyma/issues/13114


werdes72 (approved):
/woof

PR #13636, opened 2022-03-16
Handle nats-server restart for Jetstream

raypinto:
**Description**

- Added checks for recreating streams in case of memory storage (when the connection is lost)
- Enabled JetStream server to use File storage for streams

**Related issue(s)**


resources/eventing/charts/nats/values.yaml
```diff
@@ -115,7 +115,7 @@ nats:
     #                          #
     ############################
     fileStorage:
-      enabled: false
+      enabled: true
```
raypinto, line 118: Unless this is enabled, setting [this](https://github.com/kyma-project/kyma/blob/fff53022d89875a3539677120bcc4dd9702a299c/resources/eventing/charts/controller/values.yaml#L82) to 'file' does not do anything.

components/eventing-controller/pkg/handlers/jetstream.go
```diff
@@ -142,11 +149,8 @@ func (js *JetStream) SyncSubscription(subscription *eventingv1alpha1.Subscriptio
 	for _, subject := range subscription.Status.CleanEventTypes {
 		jsSubKey := js.GenerateJsSubKey(subject, subscription)
 
-		if js.conn.Status() != nats.CONNECTED {
-			if err := js.Initialize(js.connClosedHandler); err != nil {
-				log.Errorw("reset JetStream connection failed", "status", js.conn.Stats(), "error", err)
-				return err
-			}
+		if err := js.checkJetStreamConnection(log); err != nil {
```
mfaizanse, line 152: (nit) we are checking it on line 104, so we don't need to check it again in the same function, IMO?

components/eventing-controller/pkg/handlers/jetstream.go
```diff
@@ -101,11 +101,18 @@ func (js *JetStream) Initialize(connCloseHandler ConnClosedHandler) error {
...
 	subKeyPrefix := createKeyPrefix(subscription)
+	if err := js.checkJetStreamConnection(log); err != nil {
+		return err
+	}
 
 	// check if there is any existing JetStream subscription in global list
 	// which is not anymore in this subscription filters (i.e. cleanSubjects).
 	// e.g. when filters are modified.
 	for key, jsSub := range js.subscriptions {
+		if !jsSub.IsValid() {
```
mfaizanse, line 112: First check if the jssub is related to this kyma subscription and then make changes in the global list of jetstream subs.

components/eventing-controller/pkg/handlers/jetstream_test.go
```diff
@@ -967,6 +967,10 @@ func TestSubscription_JetStreamServerRestart(t *testing.T) {
 	jsBackend := testEnvironment.jsBackend
 	defer testEnvironment.natsServer.Shutdown()
 	defer testEnvironment.jsClient.natsConn.Close()
+
+	// use File storage to restore streams
+	jsBackend.config.JSStreamStorageType = JetStreamStorageTypeFile
```
mfaizanse, line 972: (nit) lets have another test with `in-memory` mode?

resources/eventing/charts/nats/values.yaml
```diff
@@ -106,7 +106,7 @@ nats:
     #                           #
     #############################
     memStorage:
-      enabled: true
+      enabled: false
```
mfaizanse, line 109: shouldn't it be also true?

resources/eventing/charts/nats/values.yaml
pxsalehi: I'm not so sure about that! What do you mean "doesn't do anything"? It configures the stream correctly, right? That's what it is supposed to do.

Switching the value you linked to, is for using memory or file storage for the stream (configuring a JetStream stream). When you set the value to `file` it will write to a file in `/data` (I think). The question is, what is mounted at `/data` ? If nothing, then you are writing to a file, but in the container, which means an ephemeral layer, which will be gone after a restart. 

So, yes, if you want to write to a file on a PV, then you should set that value to true. I think, it is just a matter of documentation. If you check the evaluation vs. production profile, these flagged are set. 

By permanently setting that flag to true, you will always create a PVC which would use a PV, even if you are using in-memory stream!

raypinto: The streams are created correctly but are not persisted after the restart, since we are writing to a file in the container which is ephemeral. I assume the point of using filestorage is to persist streams and hence we should created a PVC.

The question is should we set this value globally, so that when we use want streams (in the controller) to be created in memory, we do not create this PVC either?

components/eventing-controller/pkg/handlers/jetstream.go
raypinto: done

components/eventing-controller/pkg/handlers/jetstream_test.go
raypinto: Yes fair point.

resources/eventing/charts/nats/values.yaml
raypinto: On second thought, according to [this line](https://github.com/kyma-project/kyma/blob/9d03ae8cbb67ebab5202ae327294a43c99831b4d/resources/eventing/charts/nats/templates/configmap.yaml#L46), I'm not sure if the storage directory is even created unless the value is set to true.

raypinto: After discussing with the team, I've moved the value to enable the storage at the global level. This would create the PVC only when we enable file storage for streams.

components/eventing-controller/pkg/handlers/jetstream_test.go
raypinto: Added two more tests.

PR #13689, opened 2022-03-22
Consistent sink validation in BEB and NATS

### LINKED ISSUE #13354: Sink Validation behaviour is different in NATS and BEB
**Description**

If the sink validation fails in NATS, the `ready` status of the kyma subscription becomes `false`, but for BEB - the `ready` status remains `true` (if it was previously true)

**Expected result**

The subscription exhibits similar behaviour with both NATS and BEB backends

**Actual result**

NATS and BEB show different behaviours

**Steps to reproduce**

- Create a valid subscription as below with NATS backend

```
cat << EOF | kubectl apply -f -
apiVersion: eventing.kyma-project.io/v1alpha1
kind: Subscription
metadata:
  name: test-noapp
  namespace: test
spec:
  filter:
    filters:
    - eventSource:
        property: source
        type: exact
        value: ""
      eventType:
        property: type
        type: exact
        value: sap.kyma.custom.noapp.order.created.v1
  protocol: ""
  protocolsettings: {}
  sink: http://test.test.svc.cluster.local
EOF
```

- Once the subscription is active, edit the existing subscription and change the sink value to `test.test.svc.cluster.local`
- The ready status is now false
- Repeat the same steps with BEB backend. In this case the ready status of the subscription remains true.

**Troubleshooting**

pxsalehi:
**Description**

- Reuse sink validator instead of the duplicate code. The existing code in BEB which was doing sink validation would issue a "skippable" error when the scheme was missing from the sink.

**Related issue(s)**

Closes #13354

### DIFF
--- a/components/eventing-controller/testing/test_helpers.go
+++ b/components/eventing-controller/testing/test_helpers.go
@@ -374,9 +374,8 @@ func WithOrderCreatedFilter() SubscriptionOpt {
 	return WithFilter(EventSource, OrderCreatedEventType)
 }
 
-func WithInvalidSink() SubscriptionOpt {
-	return WithSinkURL("invalid")
-
+func WithSinkMissingScheme(svcNamespace, svcName string) SubscriptionOpt {
+	return WithSinkURL(fmt.Sprintf("%s.%s.svc.cluster.local", svcName, svcNamespace))
 }
 
 // WithValidSink is a SubscriptionOpt for creating a subscription with a valid sink that itself gets created from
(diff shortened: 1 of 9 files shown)


components/eventing-controller/testing/test_helpers.go
```diff
@@ -374,9 +374,8 @@ func WithOrderCreatedFilter() SubscriptionOpt {
 	return WithFilter(EventSource, OrderCreatedEventType)
 }
 
-func WithInvalidSink() SubscriptionOpt {
```
pxsalehi, line 377: After the test changes, this was unused, and therefore I removed it.

mfaizanse (approved):

Issue #13625, opened 2022-03-14
Check if sending a NACK for an event is counted towards redeliver count in JetStream consumer


mfaizanse:
**Description**

Currently, in the JetStream backend, if the dispatcher fails to deliver the event to the sink then it sends a NACK to the jetstream-server so that JetStream sent the event again. We need to check if the NACK is counted towards the redeliver count and does JetStream stops to redeliver this event after the redelivery count is exhausted. Otherwise, it will keep on redeliver the same event forever.

mfaizanse:
##### Findings:
- NACKing a msg is counted towards the redeliver count as mentioned in the JetStream [doc](https://docs.nats.io/nats-concepts/jetstream/consumers#maxdeliver).
- If there is not `maxDeliver` config defined when creating a consumer, then JetStream will try to redeliver any NACKed msg infinite number of times.

##### Suggestion:
- Currently, we are not defining the `maxDeliver` config for consumers in eventing-controller, thats why our dispatcher will receive the NACKed msg again infinite number of times and it will keep on retrying to deliver the msg to the sink.
  - [Option 1] We can define a `maxDeliver` for the consumers, but one point-to-ponder is that JetStream will redeliver the msg instantly without any delay. Thus, the `maxDeliver` would be exceeded too easily. (as the sink could take some time to be ready or be healthy).
    - There is an open [issue](https://github.com/nats-io/nats-server/issues/2042) in NATS to introduce `ExponentialBackoffPolicy` when re-delivering a NACKed msg. This would be helpful in our case if we NACK the msg when the dispatcher fails to deliver the msg to the sink.
  - [Option 2] Other option would be use the [`AckWait`](https://docs.nats.io/nats-concepts/jetstream/consumers#ackwait) for consumers and to not `NACK` the msg if the dispatcher fails to deliver the msg to the sink. In this case, the ACK wait timeout will happen after some time and JetStream will then redeliver the msg.

mfaizanse:
There is a `NumDelivered` field in msg metadata, which may tell us the deliver attempt number.
```
       // define callback for consumer
	callback := func(m *nats.Msg) {
		msgMetaData, err := m.Metadata()
		if err != nil {
			log.Fatalln(err)
		}
		fmt.Printf("msgMetaData.NumDelivered %d\n", msgMetaData.NumDelivered)
	}
```

mfaizanse:
[...]

PR #13693, opened 2022-03-22
Added MaxDeliver config to JetStream consumer

mfaizanse:
**Description**

- Set MaxDeliver to 100 for JetStream consumer.
- Set AckWait to 30 seconds for JetStream consumer.
- Changed the dispatcher to not NAK the msg if it fails to deliver it to the sink, so that it timeouts after AckWait period and then redeliver. It will keep on trying for 100*30 seconds (50 minutes)

**Related issue(s)**

- https://github.com/kyma-project/kyma/issues/13625


raypinto (commented):
Looks good! 
(nit) Maybe a test to show that the maxRedeliver works as expected? If you think it's just a jetstream feature and the test doesn't add value, feel free to drop it.

raypinto (approved):

PR #13703, opened 2022-03-23
Implement cleanup for JetStream

raypinto:
**Description**

- Cleanup logic for JetStream implemented

**Related issue(s)**


components/eventing-controller/pkg/subscriptionmanager/jetstream/jetstream_test.go
```diff
@@ -0,0 +1,154 @@
...
+func (c *jetStreamSubMgrMock) Start(_ env.DefaultSubscriptionConfig, _ subscriptionmanager.Params) error {
+	return nil
+}
+
+func (c *jetStreamSubMgrMock) Stop(_ bool) error {
+	return nil
+}
+
+func TestCleanup(t *testing.T) {
+	jsSubMgr := jetStreamSubMgrMock{}
```
nachtmaar, line 42: Is `jsSubMgr` required for this test ? From what I see it only servers as a container for `Client` and `Backend`.

components/eventing-controller/pkg/subscriptionmanager/jetstream/jetstream_test.go
```diff
@@ -0,0 +1,154 @@
...
+
+func (c *jetStreamSubMgrMock) Start(_ env.DefaultSubscriptionConfig, _ subscriptionmanager.Params) error {
+	return nil
+}
+
+func (c *jetStreamSubMgrMock) Stop(_ bool) error {
+	return nil
+}
+
+func TestCleanup(t *testing.T) {
```
nachtmaar, line 41: The test is really long. Can it be split up into smaller and reusable functions ? For example refactoring out the actual test setup as in the [guidelines](https://github.com/kyma-project/community/blob/main/concepts/eventing-code-guidelines.md#separate-test-setup-from-actual-test).

It all boils down to something like this:

```
	// given
	// init something
	
	// when
	err = cleanup(jsSubMgr.Backend, jsSubMgr.Client, defaultLogger.WithContext())

	// then
	// ensure jetstream subscription is deleted
```

components/eventing-controller/pkg/subscriptionmanager/jetstream/jetstream_test.go
```diff
@@ -0,0 +1,154 @@
...
+	wantSubStatus := eventingv1alpha1.SubscriptionStatus{}
+	require.Equal(t, wantSubStatus, gotSub.Status)
+
+	// test JetStream subscriptions/consumers are gone
+	info, err = jsClient.StreamInfo(envConf.JSStreamName)
+	require.NoError(t, err)
+	require.Equal(t, info.State.Consumers, 0)
+
+	// send again an event and check subscriber, check subscriber should fail
+	err = handlers.SendEventToJetStream(jsBackend, data)
```
nachtmaar, line 148: I wonder what the bare minimum is what the test should actually test. It feels the test is doing too much at once for me 🤔 

For me the bare minimum is that the jetstream subscription was deleted and the subscription status got updated. As done in lines 135-145.

However, testing that sending an event to a not existing jetstream subscription does not work is not necessary here I feel (it could be another test case maybe).

What is your opinion on this ?

components/eventing-controller/pkg/subscriptionmanager/jetstream/jetstream.go
```diff
@@ -102,9 +107,62 @@ func (sm *SubscriptionManager) Start(defaultSubsConfig env.DefaultSubscriptionCo
...
+	}
+
+	// clean all status.
+	isCleanupSuccessful := true
+	for _, v := range subs.Items {
+		sub := v
+		subKey := types.NamespacedName{Namespace: sub.Namespace, Name: sub.Name}
+		log := logger.With("key", subKey.String())
+
+		desiredSub := handlers.RemoveStatus(sub)
```
nachtmaar, line 150: out-of-scope since you did not introduce it: I think `RemoveStatus` is the wrong name because it is actually overriding the status with the defaults. 

Maybe like this ? 🤔 

```suggestion
		desiredSub := handlers.ResetStatusToDefaults(sub)
```

components/eventing-controller/pkg/subscriptionmanager/jetstream/jetstream.go
```diff
@@ -102,9 +107,62 @@ func (sm *SubscriptionManager) Start(defaultSubsConfig env.DefaultSubscriptionCo
...
+// clean removes all JetStream artifacts.
+func cleanup(backend handlers.JetStreamBackend, dynamicClient dynamic.Interface, logger *zap.SugaredLogger) error {
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+
+	var ok bool
+	var jsBackend *handlers.JetStream
+	if jsBackend, ok = backend.(*handlers.JetStream); !ok {
+		err := errors.New("convert backend handler to JetStream handler failed")
+		logger.Errorw("no JetStream backend exists", "error", err)
```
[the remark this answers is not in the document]
raypinto: Good point. I think it's redundant.

components/eventing-controller/pkg/subscriptionmanager/jetstream/jetstream.go
```diff
@@ -102,9 +107,62 @@ func (sm *SubscriptionManager) Start(defaultSubsConfig env.DefaultSubscriptionCo
...
+	}
+
+	subs, err := handlers.ToSubscriptionList(subscriptionsUnstructured)
+	if err != nil {
+		return errors.Wrapf(err, "convert subscriptionList from unstructured list failed")
+	}
+
+	// clean all status.
+	isCleanupSuccessful := true
+	for _, v := range subs.Items {
```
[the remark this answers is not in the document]
raypinto: If I do, I get the lint error: `G601: Implicit memory aliasing in for loop.`

components/eventing-controller/pkg/subscriptionmanager/jetstream/jetstream.go
raypinto: It's actually removing the Status and replacing it with an empty Subscription Status which has just the Ready field. I don't think defaults is the right term, but I can't think of anything better. So I'll take the suggestion 👍

components/eventing-controller/pkg/subscriptionmanager/jetstream/jetstream_test.go
raypinto: And it's just mocking the behaviour and calls of the actual `SubscriptionManager`. Honestly, I don't think its needed either. We can discuss this.

components/eventing-controller/pkg/subscriptionmanager/jetstream/jetstream.go
```diff
@@ -102,9 +107,62 @@ func (sm *SubscriptionManager) Start(defaultSubsConfig env.DefaultSubscriptionCo
...
+		desiredSub := handlers.RemoveStatus(sub)
+		if err := handlers.UpdateSubscriptionStatus(ctx, dynamicClient, desiredSub); err != nil {
+			isCleanupSuccessful = false
+			log.Errorw("update NATS subscription status failed", "error", err)
+		}
+
+		// clean subscriptions from JetStream.
+		if jsBackend != nil {
+			if err := jsBackend.DeleteSubscription(&sub); err != nil {
+				isCleanupSuccessful = false
```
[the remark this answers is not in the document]
raypinto: IMO Failure to delete a subscription from the server should not prevent switching of the backends. The real question is if we switch for good, and the server is shutdown (yet to be implemented), will this error message or the condition help?

components/eventing-controller/pkg/subscriptionmanager/jetstream/jetstream_test.go
raypinto: I'll try to cleanup👍

components/eventing-controller/pkg/subscriptionmanager/jetstream/jetstream_test.go
raypinto: Sending an event to a not existing jetstream subscription is just an extension to be absolutely sure that the cleanup was done. If there are no consumers, there's no way the event can be consumed. If you also feel that this is not needed, I can remove it.

components/eventing-controller/pkg/subscriptionmanager/jetstream/jetstream.go
nachtmaar: Mhh, I know what you mean 🤔. Defaulting is more the defaulting in Go which is associated with non pointer types.

components/eventing-controller/pkg/subscriptionmanager/jetstream/jetstream_test.go
nachtmaar: We discussed it internally and decided it makes sense to have it, especially because the stream might get deleted later and checking the consumers is then not possible anymore.

nachtmaar (approved):

components/eventing-controller/pkg/subscriptionmanager/jetstream/jetstream_test.go
```diff
@@ -0,0 +1,193 @@
...
+
+	// create NATS subscription
+	err := jsBackend.SyncSubscription(testSub)
+	require.NoError(t, err)
+	return testSub
+}
+
+func testEventing(testEnv *TestEnvironment) error {
+	// make sure subscriber works
+	err := testEnv.subscriber.CheckEvent("")
```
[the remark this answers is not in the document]
raypinto: As discussed, it's more of a readiness probe. I've updated the comment as well.

components/eventing-controller/pkg/subscriptionmanager/jetstream/jetstream_test.go
```diff
@@ -0,0 +1,193 @@
...
+	cleanedSubjects, _ := handlers.GetCleanSubjects(testSub, eventtype.CleanerFunc(cleaner))
+	testSub.Status.CleanEventTypes = jsBackend.GetJetStreamSubjects(cleanedSubjects)
+
+	// create NATS subscription
+	err := jsBackend.SyncSubscription(testSub)
+	require.NoError(t, err)
+	return testSub
+}
+
+func testEventing(testEnv *TestEnvironment) error {
```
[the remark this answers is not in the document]
raypinto: Agree. Updated.

components/eventing-controller/pkg/subscriptionmanager/jetstream/jetstream_test.go
```diff
@@ -0,0 +1,193 @@
...
+
+const (
+	data                  = "sampledata"
+	subscriptionName      = "test"
+	subscriptionNamespace = "test"
+)
+
+var expectedDataInStore = fmt.Sprintf("%q", data)
+
+func TestCleanup(t *testing.T) {
```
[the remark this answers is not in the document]
raypinto: Thank you!

nachtmaar (approved):

Issue #13213, opened 2022-01-28
[Telemetry Operator] Simple Status on CRD


a-thaler:
**Description**
A `LogPipeline` must indicate crucial status information so that the user knows if the desired state is reached, what potential problems blocked it, what implied attributes are relevant. All needs to be reflected in the CRD so that also a UI can display it nicely.

**Goal**
Have a proper status for LogPipelines which are indicating:
- a state - ready, pending
- conditions - listing when a state was achieved

skhalash:
Implemented status conditions similar to standard Kubernetes resources (pods, replicasets, deployments, etc.).

PR #13727, opened 2022-03-25
Add log pipeline status conditions

skhalash:
**Description**

- Replace deleting FluentBit pods with patching the FluentBit daemon set (aka rollout restart)
- Introduce LogPipeline status conditions to track the lof pipeline status changes
- Fix minor bug in parser synching
- Add fast-integration test 

**Related issue(s)**
https://github.com/kyma-project/kyma/issues/13213


tests/fast-integration/telemetry-test/telemetry-test.js
```diff
@@ -89,7 +111,8 @@ describe('Telemetry Operator tests', function() {
 
   it('Should create valid LogPipeline with HTTP output plugin', async () => {
     await k8sApply(logPipelineCR, telemetryNamespace);
-    await waitForDaemonSet(fluentBitName, telemetryNamespace, 30000);
+    await waitForLogPipelineStatusCondition('logpipeline-test', 'Pending', 20000);
```
chrkl, line 114: I am wondering if there is a chance that we make this call too late and miss the `Pending` state. This could potentially make the test flaky.

chrkl (changes requested):
Well done! Only a few minor findings.

tests/fast-integration/telemetry-test/telemetry-test.js
skhalash: We can rely on the fact that the transition from Pending to Running takes time if in the case of a one-node cluster. We actually caught a bug in our code when the transition was happening too fast. But I agree, let us observe how "stable" this assumption is.

lindnerby (approved):

Issue #13546, opened 2022-03-03
Use a better consumer name when creating subscriptions on JetStream


mfaizanse:
**Description**

As `Spaces, tabs, period (.), greater than (>) or asterisk (*) are prohibited` to be used in the consumer names when creating a subscription on JetStream, currently we are encoding (i.e. base64) the consumer name before we pass it to JetStream, so that the JetStream do not complain about invalid characters in consumer name. 

We need to find a better approach for consumer names, as according to JetStream [guidelines](https://docs.nats.io/running-a-nats-service/nats_admin/jetstream_admin/naming), it is recommended to limit the consumer name to max 32 characters:
```
consumer name, so a generally safe approach would be to keep names under 32 characters.
```
Right now, the base64 encoded string can easily exceed the limit of 32 characters, so we need to use a better approach for consumer names, in order to avoid any unexpected behaviour by JetStream.

#### Acceptance criteria:
- [ ] Find the solution for consumer names
- [ ] Implement the solution and tests.

PR #13744, opened 2022-03-25
Create JetStream consumers with the recommended name length

marcobebway:
**Description**

Create JetStream consumers with the recommended name length mentioned [here](https://docs.nats.io/running-a-nats-service/nats_admin/jetstream_admin/naming).

**Related issue(s)**

- https://github.com/kyma-project/kyma/issues/13546

### DIFF
--- a/components/eventing-controller/pkg/handlers/jetstream.go
+++ b/components/eventing-controller/pkg/handlers/jetstream.go
@@ -2,6 +2,8 @@ package handlers
 
 import (
 	"context"
+	"crypto/md5" // #nosec
+	"encoding/hex"
 	"fmt"
 	"net/http"
 	"strings"
@@ -10,16 +12,15 @@ import (
 
 	cev2 "github.com/cloudevents/sdk-go/v2"
 	cev2protocol "github.com/cloudevents/sdk-go/v2/protocol"
-	"github.com/kyma-project/kyma/components/eventing-controller/pkg/tracing"
-	"github.com/kyma-project/kyma/components/eventing-controller/utils"
-	"k8s.io/apimachinery/pkg/types"
+	"github.com/nats-io/nats.go"
+	"github.com/pkg/errors"
+	"go.uber.org/zap"
 
 	eventingv1alpha1 "github.com/kyma-project/kyma/components/eventing-controller/api/v1alpha1"
 	"github.com/kyma-project/kyma/components/eventing-controller/logger"
 	"github.com/kyma-project/kyma/components/eventing-controller/pkg/env"
-	"github.com/nats-io/nats.go"
-	"github.com/pkg/errors"
-	"go.uber.org/zap"
+	"github.com/kyma-project/kyma/components/eventing-controller/pkg/tracing"
+	"github.com/kyma-project/kyma/components/eventing-controller/utils"
 )
 
 var _ JetStreamBackend = &JetStream{}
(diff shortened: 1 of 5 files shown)


components/eventing-controller/pkg/handlers/jetstream.go
```diff
@@ -45,12 +47,49 @@ type JetStreamBackend interface {
...
+// NamespacedName returns the Kubernetes namespaced name.
+func (s SubscriptionSubjectIdentifier) NamespacedName() string {
+	return s.namespacedName
+}
+
+// computeConsumerName returns JetStream consumer name of the given subscription and subject.
+// It uses the crypto/md5 lib to return a string of 32 characters as recommended by the JetStream
+// documentation https://docs.nats.io/running-a-nats-service/nats_admin/jetstream_admin/naming.
+func computeConsumerName(subscription *eventingv1alpha1.Subscription, subject string) string {
+	cn := subscription.Namespace + separator + subscription.Name + separator + subject
```
mfaizanse, line 77: (nit) is it better to use `fmt.sprintf()`?

components/eventing-controller/pkg/handlers/jetstream.go
```diff
@@ -45,12 +47,49 @@ type JetStreamBackend interface {
...
+	h := md5.Sum([]byte(cn)) // #nosec
+	return hex.EncodeToString(h[:])
+}
+
+// computeNamespacedName returns Kubernetes namespaced name of the given subscription.
+func computeNamespacedName(subscription *eventingv1alpha1.Subscription) string {
+	return subscription.Namespace + separator + subscription.Name
```
mfaizanse, line 84: (nit) is it better to use `fmt.sprintf()`?

components/eventing-controller/pkg/handlers/jetstream_test.go
```diff
@@ -1199,3 +1202,170 @@ func setupTestEnvironment(t *testing.T) *TestEnvironment {
...
+	}
+	for _, tc := range testCases {
+		tc := tc
+		t.Run(tc.name, func(t *testing.T) {
+			t.Parallel()
+			gotInstanceEqual := reflect.DeepEqual(tc.givenIdentifier1, tc.givenIdentifier2)
+			assert.Equal(t, tc.wantEqual, gotInstanceEqual)
+
+			gotConsumerNameEqual := tc.givenIdentifier1.ConsumerName() == tc.givenIdentifier2.ConsumerName()
+			assert.Equal(t, tc.wantEqual, gotConsumerNameEqual)
```
mfaizanse, line 1299: (nit) is `require.equal()` preferable?

components/eventing-controller/pkg/handlers/jetstream.go
marcobebway: As discussed offline, we'll keep it like that because simple string concatenation is more performant that `fmt.sprintf()`.

components/eventing-controller/pkg/handlers/jetstream.go
marcobebway: As discussed offline, we'll keep it like that because simple string concatenation is more performant that `fmt.sprintf()`.

components/eventing-controller/pkg/handlers/jetstream_test.go
marcobebway: As discussed offline, we'll keep it like that because `t.Run` would run all test cases anyway and `require.Equal` would not have the desired effect here.

mfaizanse (dismissed):

mfaizanse (approved):

Issue #13422, opened 2022-02-17
Remove Service Catalog area from Kyma opensource


PK85:
**Description**

With Kyma 2.2 release we plan to remove Service Catalog area from this project.

**Reasons**

- hyperscaler cloud providers discontinued work on their Service Brokers. They support operators which will allow to provision services.
- k8s Service Catalog is dead, no active development.
- service binding usage feature causes issues when prod-preset controller is down and end-application pod is restarting. Result: application has no environment variables injected.

**AC**

- [x] 1) remove all related charts and components. See what sc-migrator reconciler is cleaning up from Kyma cluster after migration, link: https://github.com/kyma-incubator/reconciler/pull/722
- [x] 2) double check documentation to not mention service catalog
- [x] 3) double check if all the test are adjusted and green
- [x] 4) Final release note about SC removal, that it's really gone now

klaudiagrz:
Please remember to include the note about SC removal in the release notes 🙂

wozniakjan:
> TODO: check how much less resources (CPU and memory) Kyma will use after the service catalog removal.

@PK85, here are some statistics from an idle freshly provisioned cluster
```
$ k top pod service-catalog-catalog-controller-manager-7d96d47bbb-p9lxm
NAME                                                          CPU(cores)   MEMORY(bytes)
service-catalog-catalog-controller-manager-7d96d47bbb-p9lxm   3m           81Mi
```

```
$ k top pod service-catalog-catalog-webhook-5d7487cb95-zfvmk
NAME                                               CPU(cores)   MEMORY(bytes)
service-catalog-catalog-webhook-5d7487cb95-zfvmk   1m           11Mi
```

```
$ k top pod service-catalog-addons-service-binding-usage-controller-5cnw7ch
NAME                                                              CPU(cores)   MEMORY(bytes)
service-catalog-addons-service-binding-usage-controller-5cnw7ch   3m           87Mi
```

we also set some requests and limits on these
```
$ k get deployment service-catalog-addons-service-binding-usage-controller service-catalog-catalog-controller-manager service-catalog-catalog-webhook -o json | jq -c '.items[].spec.template.spec.containers[] | {"name":.name,"resources":.resources}'
{"name":"service-binding-usage-controller","resources":{"limits":{"cpu":"100m","memory":"96Mi"},"requests":{"cpu":"50m","memory":"48Mi"}}}
{"name":"healthcheck","resources":{}}
[...]

PR #13752, opened 2022-03-28
Document post 2.3 cleanup script

kwiatekus:
**Description**

- Document 2.2->2.3 cleanup script 

**Related issue(s)**
https://github.com/kyma-project/kyma/issues/13422

### DIFF
--- a/docs/assets/2.1-2.2-fix-delete-orphaned-resources.sh
+++ b/docs/assets/2.1-2.2-fix-delete-orphaned-resources.sh
@@ -1,0 +1,30 @@
+#!/usr/bin/env bash
+set -e
+
+echo "
+Deleting application broker"
+
+echo "Framefrogs please contribute"
+exit 1
+
+echo "
+Deleting rafter"
+
+kubectl -n kyma-system delete crd assetgroups.rafter.kyma-project.io
+kubectl -n kyma-system delete crd assets.rafter.kyma-project.io
+kubectl -n kyma-system delete crd buckets.rafter.kyma-project.io
+kubectl -n kyma-system delete crd clusterassetgroups.rafter.kyma-project.io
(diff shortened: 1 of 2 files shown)


kwiatekus:
@kyma-project/technical-writers should group those migration guides into one parent node?
👇 this won't scale, right ?
Screenshot 2022-03-28 at 13 32 47

docs/assets/2.1-2.2-fix-delete-orphaned-resources.sh
```diff
@@ -0,0 +1,30 @@
...
+echo "
+Deleting application broker"
+
+echo "Framefrogs please contribute"
+exit 1
+
+echo "
+Deleting rafter"
+
+kubectl -n kyma-system delete crd assetgroups.rafter.kyma-project.io
```
wozniakjan, line 13: do we want to make the script idempotent? As implemented with `set -e`, if it fails later for any reason, the fact it deleted the first CRD, every execution will fail as well because attempting to delete nonexisting resource with kubectl returns a non-zero exit code.

```bash
$ k get crd abcdefg
Error from server (NotFound): customresourcedefinitions.apiextensions.k8s.io "abcdefg" not found

$ echo $?
1
```

kwiatekus:
New PR : https://github.com/kyma-project/kyma/pull/14095

docs/assets/2.1-2.2-fix-delete-orphaned-resources.sh
kwiatekus: thx. I applied the fix in the new [PR](https://github.com/kyma-project/kyma/pull/14095).
I created a new one because this branch name is wrong (its going to be rollout in 2.3)

Issue #13290, opened 2022-02-07
Cleanup custom Kyma dashboards and alerts


skhalash:
**Acceptance criteria**
* Reassess existing Kyma Grafana dashboards and Prometheus alerts and delete the unused ones (or the ones overlapping with the upstream dashboards/alerts)

PR #13758, opened 2022-03-28
Cleanup custom kyma alerts

Pranav-SA:
**Description**

- Remove redundant custom kyma alerts. 
- Update ignore alerts list.
- Remove `apiserver_request_latencies_bucket` references.

**Related issue(s)**
#13290


resources/monitoring/templates/prometheus/kyma-rules/prometheus.rules.yaml
```diff
@@ -1,19 +0,0 @@
...
-metadata:
-  name: {{ printf "%s-%s" (include "kube-prometheus-stack.fullname" .) "kyma-prometheus.rules" | trunc 63 | trimSuffix "-" }}
-  labels:
-    app: {{ template "kube-prometheus-stack.name" . }}
-{{ include "kube-prometheus-stack.labels" . | indent 4 }}
-spec:
-  groups:
-  - name: prometheus.rules
-    rules:
-    - alert: ScrapeLimitForTargetExceeded
```
a-thaler, line 12: We will need that rule to inform the user about reaching the limit

a-thaler:
What is with the "Kyma / Kubernetes / Pods" Dashboard? I think that also should be cleaned up?

resources/monitoring/templates/prometheus/kyma-rules/prometheus.rules.yaml
Pranav-SA: Indeed, thanks I'll add it back.

skhalash (approved):

PR #13764, opened 2022-03-28
Switch from go-get to go-install and bump linter version

raypinto:
**Description**

- Remove deprecated `go get` and use `go install` instead
- Bump golangci-lint to 1.45.2 as go1.18 is supported from v1.45.0.

**Related issue(s)**

### DIFF
--- a/components/eventing-controller/Makefile
+++ b/components/eventing-controller/Makefile
@@ -92,9 +92,9 @@ undeploy: ## Undeploy controller from the K8s cluster specified in ~/.kube/confi
 	kustomize build config/default | kubectl delete -f -
 
 CONTROLLER_GEN = $(PROJECT_DIR)/bin/$(shell uname)/controller-gen
-ARCH := $(shell uname)
+OS := $(shell uname)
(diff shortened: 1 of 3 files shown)


components/eventing-controller/Makefile
```diff
@@ -92,8 +92,9 @@ undeploy: ## Undeploy controller from the K8s cluster specified in ~/.kube/confi
 	kustomize build config/default | kubectl delete -f -
 
 CONTROLLER_GEN = $(PROJECT_DIR)/bin/$(shell uname)/controller-gen
+ARCH := $(shell uname)
```
k15r, line 95: i know this was in there before, but can we rename it from ARCH to OS as it is not the architecture, but the os we are interested in

raypinto: done

clebs (changes requested):
Please replace the deprecated linters:
golint -> revive
scopelint -> exportloopref

clebs (approved):

k15r (approved):

PR #13775, opened 2022-03-29
Pass Jetstream environment variables to eventing publisher proxy

VladislavPaskar:
**Description**

In order to publish the events via EPP in a jetstream mode, we need to propagate the jetstream config from the eventing controller to eventing publisher proxy.

- pass the js config as env variables to EPP
- add new test cases for epp deployment

**Related issue(s)**


components/eventing-controller/pkg/env/backend_config.go
```diff
@@ -18,19 +18,26 @@ type BackendConfig struct {
...
+	LimitsMemory      string `envconfig:"PUBLISHER_LIMITS_MEMORY" default:"128Mi"`
+	PriorityClassName string `envconfig:"PUBLISHER_PRIORITY_CLASS_NAME" default:""`
+
+	// jetstream
+	EnableJetStreamBackend  bool   `envconfig:"ENABLE_JETSTREAM_BACKEND" default:"false"`
+	JSStreamName            string `envconfig:"JS_STREAM_NAME" default:"kyma"`
+	JSStreamStorageType     string `envconfig:"JS_STREAM_STORAGE_TYPE" default:"memory"`
+	JSStreamRetentionPolicy string `envconfig:"JS_STREAM_RETENTION_POLICY" default:"interest"`
+	JSStreamMaxMessages     int64  `envconfig:"JS_STREAM_MAX_MSGS" default:"-1"`
+	JSStreamMaxBytes        int64  `envconfig:"JS_STREAM_MAX_BYTES" default:"-1"`
```
mfaizanse, line 40: What is the purpose of passing the extra fields: `JSStreamStorageType`, `JSStreamRetentionPolicy`, `JSStreamMaxMessages` and `JSStreamMaxBytes` to EPP? Is it needed in EPP to initialise the jetStream connection or anyplace else?

components/eventing-controller/utils/jetstream_utils.go
```diff
@@ -24,7 +24,7 @@ func toJetStreamStorageType(s string) (nats.StorageType, error) {
 	return nats.MemoryStorage, fmt.Errorf("invalid stream storage type %q", s)
 }
 
-func toJetStreamRetentionPolicy(s string) (nats.RetentionPolicy, error) {
+func ToJetStreamRetentionPolicy(s string) (nats.RetentionPolicy, error) {
```
mfaizanse, line 27: (nit) Why do we need to move this file from `pkg/handlers` to `utils`? Can't we keep it in `pkg/handlers` and just export the methods? IMO, `utils` directory is for things more generic for the whole codebase, JetStream specific helpers should be in the `pkgs`.

components/eventing-controller/pkg/env/backend_config.go
VladislavPaskar: I pass them in order to match the jetstream config fully in the EPP when publishing the events. There I will not only check for the jetstreamName, but also for the storageType, RetentionPolicy and other properties.

components/eventing-controller/utils/jetstream_utils.go
VladislavPaskar: I moved them into `pkg/utils`. The reason I exposed them is because I want to reuse them in the EPP

components/eventing-controller/pkg/env/backend_config.go
VladislavPaskar: I checked once again, seems that stream name is unique for jetstream, so we need to pass only the stream name

components/eventing-controller/utils/jetstream_utils.go
VladislavPaskar: https://github.com/kyma-project/kyma/pull/13775#discussion_r838231841

components/eventing-controller/pkg/deployment/publisher_deployment.go
```diff
@@ -282,6 +283,9 @@ func getBEBEnvVars(publisherConfig env.PublisherConfig) []v1.EnvVar {
 }
 
 func getNATSEnvVars(publisherConfig env.PublisherConfig) []v1.EnvVar {
+	// reading the NATS backend config in order to get Jetstream related env variables
+	natsConfig := env.GetNatsConfig(10, 3*time.Second)
```
VladislavPaskar, line 287: the `10` and `*time.Second` are default values, and they stay for `maxReconnects` and `reconnectWait` which are not relevant for epp, so we can go with these default. values.

components/eventing-controller/pkg/deployment/publisher_deployment_test.go
```diff
@@ -1,69 +1,242 @@
...
+	for _, val := range affinityLabels {
+		assert.Equal(t, val, PublisherName)
+	}
+}
+
+// natsBackendAssertions checks the BEB-specific data was set in the NewBEBPublisherDeployment
+func bebBackendAssertions(t *testing.T, deployment appsv1.Deployment) {
+	container := deployment.Spec.Template.Spec.Containers[0]
+
+	// check nats-specific env variables
```
mfaizanse, line 234: fix the comment to `// check beb-specific env variables`

components/eventing-controller/pkg/deployment/publisher_deployment_test.go
```diff
@@ -1,69 +1,242 @@
...
+	}{
+		{
+			name:                  "NATS should be set properly after calling the constructor",
+			givenBackend:          "NATS",
+			wantBackendAssertions: natsBackendAssertions,
+		},
+		{
+			name:                  "BEB should be set properly after calling the constructor",
+			givenBackend:          "BEB",
+			wantBackendAssertions: bebBackendAssertions,
```
mfaizanse, line 40: The test name is `TestNewNATSDeployment` and we are also testing the BEB case in it, so please make it consistent :)

components/eventing-controller/pkg/deployment/publisher_deployment_test.go
```diff
@@ -1,69 +1,242 @@
...
+	assert.Equal(t, jsEnabled.Value, "true")
+
+	// check the affinity was set
+	affinityLabels := deployment.Spec.Template.Spec.Affinity.PodAntiAffinity.PreferredDuringSchedulingIgnoredDuringExecution[0].PodAffinityTerm.LabelSelector.MatchLabels
+	for _, val := range affinityLabels {
+		assert.Equal(t, val, PublisherName)
+	}
+}
+
+// natsBackendAssertions checks the BEB-specific data was set in the NewBEBPublisherDeployment
```
mfaizanse, line 230: fix the comment to `// bebBackendAssertions checks the BEB-specific...`

mfaizanse (approved):

PR #13785, opened 2022-03-30
Unit tests for NATS Reconcile()

### LINKED ISSUE #13218: Add unit test to test/document when should the NATS reconciler return an error
**Description**
After investigating https://github.com/kyma-project/kyma/issues/13059, we have decided to return error when the Kyma subscription is not ready, so that the reconciliation request is requeued. To verify this we need to add unit tests that check the `Reconcile` call returns error when it should. For details please see: 
https://github.com/kyma-project/kyma/pull/13205#discussion_r795437028

**Related Issues**
https://github.com/kyma-project/kyma/issues/13161

pxsalehi:
**Description**

- Unit test for NATS reconciler return values
- Unexport test util/structs
- Fix two benign bugs that ignore cleaner/validator errors

**Related issue(s)**

Closes #13218

### DIFF
--- a/components/eventing-controller/controllers/subscription/nats/reconciler.go
+++ b/components/eventing-controller/controllers/subscription/nats/reconciler.go
@@ -214,7 +214,7 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 	// Check for valid sink
 	if err := r.sinkValidator.Validate(subscription); err != nil {
 		log.Errorw("sink URL validation failed", "error", err)
-		if syncErr := r.syncSubscriptionStatus(ctx, subscription, false, statusChanged, err.Error()); err != nil {
+		if syncErr := r.syncSubscriptionStatus(ctx, subscription, false, statusChanged, err.Error()); syncErr != nil {
 			return ctrl.Result{}, syncErr
 		}
 		// No point in reconciling as the sink is invalid, return latest error to requeue the reconciliation request
--- a/components/eventing-controller/controllers/subscription/nats/reconciler.go
+++ b/components/eventing-controller/controllers/subscription/nats/reconciler.go
@@ -205,7 +205,7 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 	statusChanged, err := r.syncInitialStatus(subscription, log)
 	if err != nil {
 		log.Errorw("sync initial status failed", "error", err)
-		if syncErr := r.syncSubscriptionStatus(ctx, subscription, false, statusChanged, err.Error()); err != nil {
+		if syncErr := r.syncSubscriptionStatus(ctx, subscription, false, statusChanged, err.Error()); syncErr != nil {
 			return ctrl.Result{}, syncErr
 		}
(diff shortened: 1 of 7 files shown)


components/eventing-controller/controllers/subscription/nats/reconciler.go
```diff
@@ -205,7 +205,7 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 	statusChanged, err := r.syncInitialStatus(subscription, log)
 	if err != nil {
 		log.Errorw("sync initial status failed", "error", err)
-		if syncErr := r.syncSubscriptionStatus(ctx, subscription, false, statusChanged, err.Error()); err != nil {
+		if syncErr := r.syncSubscriptionStatus(ctx, subscription, false, statusChanged, err.Error()); syncErr != nil {
```
marcobebway, line 208: 👍

components/eventing-controller/controllers/subscription/nats/reconciler_internal_test.go
```diff
@@ -307,23 +310,142 @@ func Test_syncInitialStatus(t *testing.T) {
...
+		tc := testCase
+		t.Run(tc.name, func(t *testing.T) {
+			reconciler := testCase.givenReconcilerSetup()
+			r := ctrl.Request{NamespacedName: types.NamespacedName{
+				Namespace: tc.givenSubscription.Namespace,
+				Name:      tc.givenSubscription.Name,
+			}}
+			res, err := reconciler.Reconcile(context.Background(), r)
+			req.Equal(res, tc.wantReconcileResult)
+			req.Equal(err, tc.wantReconcileError)
```
marcobebway, line 427: Since `req` is configured with the `t` value from the parent test, I think we would not see the diff of what was expected vs what was computed for each test-case run (in case of errors).

pxsalehi: @marcobebway I tried it out and the output seems to be ok. It looks like this (I changed the expected error to nil):
```
--- FAIL: TestReconciler_Reconcile (0.00s)
    --- PASS: TestReconciler_Reconcile/Return_nil_and_default_Result{}_when_there_is_no_error_from_the_reconciler_dependencies (0.00s)
    --- PASS: TestReconciler_Reconcile/Return_error_and_default_Result{}_when_backend_sync_returns_error (0.00s)
    --- PASS: TestReconciler_Reconcile/Return_error_and_default_Result{}_when_backend_delete_returns_error (0.00s)
    --- FAIL: TestReconciler_Reconcile/Return_error_and_default_Result{}_when_validator_returns_error (0.00s)



Expected :*errors.errorString(&errors.errorString{s:"invalid sink"})
Actual   :<nil>(<nil>)
```

pxsalehi: The difference seems to be that if you share the `t` then any failing sub-test would stop the parent test while if you use separate `t`s then after the failure above, the rest of the case would be run as well. I think, here, that difference wouldn't matter anyway.

marcobebway (approved):

PR #13787, opened 2022-03-31
Enhance skr-test to be run N times after some reconciliations

hanngos:
**Description**
Enhance fast-integration tests in a way, that they could be run on the same SKR cluster N times after some reconciliations.

- `skr-test-N.js` and `test-N.js` are based on `skr-test.js` and `test.js` respectively, but some changes needed to be introduced so that `skr-test` could be run n times on the same cluster.

- `OIDCE2ETest`
OIDC E2E Test starts with checking the initial configuration, then  
updates the cluster with the new OIDC config and admins. After every update test checks if everything was properly applied. 

To run this test N times OIDC config and admins must be restored to the initial configuration. That is why its cases with updating cluster with initial OIDC config and admins together with checking if it was correctly applied were introduced.
Additionally, in `gatherOptions` function in `helpers.js` file, `administrator0` was changed to array to proceed with the update of the initial config.

- `CommerceMock`
Regarding CommerceMock a slight change was introduced. In `gatherOptions` function in `helpers.js` file, the namespace was randomized so that in all iterations resources are created under different namespaces.

- in `package.json` a proper line was added to run the test with `npm run test-skr-N` command and `test-skr` was specified.

- new function wait for reconciliation was implemented – it takes the last reconciliation and compares `schedulingID` with the current one as well checks if the status of the last reconciliation is ready, if so next iteration of tests can be executed.

- new env parameter `N` indicating how many iterations of tests should be executed

- the previous function `unregisterKymaFromCompass` was split into two smaller ones:
`unregisterScenarioFromCompass` which unregisters test scenario from the compass is executed at the end of every iteration
`unregisterRuntimeFromCompass` which unregisters runtime from the compass is executed only once

**How the test looks?**

1. First, there is provisioning as it was before, but after that scenario is not added to the compass and runtime is not added to the scenario. 

2. This logic was placed in the loop before because we want to have a new scenario added to the compass in every iteration. That is also the reason for calling gatherOption every time because it generates a random namespace, application name, test name etc… But with `instanceID` of the already provisioned cluster. After that, actual tests are run – `oidcE2ETest` and `commerceMock`. In the end function responsible for waiting for the new ready reconciliation is being called.  At the end of the after, the scenario is unregistered from the compass.

3. Lastly, after N iterations, the cluster is deprovisioned and runtime is unregistered from the compass.

**How to run**
1. add N parameter to env file (the same as for `skr-test`)
2. please assure that `COMPASS_TENANT` is the same as `KEB_GLOBALACCOUNT_ID`

3. in `skr-test` subdirectory run:
```
export $(xargs < .env)
```

4. run the test scenario:

```bash
npm run test-skr-N
```

**Related issue(s)**


pbochynski:
Too old to be merged.

Issue #11368, opened 2021-05-27
Proposal for supporting verbose backend status information


pxsalehi:
**Description**

In the `Backend`-CRD:
Currently, if eventing component is down, there is only one boolean field to signal this w/o any error description. Extend this information with description about:
- when (date when the change happened)
- why (reason for the change)
- what (what is the state)

The implementation could use the common pattern of conditions to achieve this

**Acceptance**
- [ ] `Ready` status of the backend is in sync with the detailed `conditions`
- [ ] `conditions` work for different backends (not hard coded to eg NATS only)
- [ ] also create events to enable us to get a history for the backend

k15r:
* maybe in `conditions`
* or as an event attached to the crd ?

raypinto:
@k15r do we need to document this change?

k15r:
This will be changed with our planned rewrite of the backend handling

PR #13841, opened 2022-04-05
Make eventing backend status more verbose

raypinto:
**Description**

- Added conditions to the eventing-backend status
- Emit events on condition change

**Related issue(s)**

https://github.com/kyma-project/kyma/issues/11368


components/eventing-controller/controllers/backend/reconciler.go
```diff
@@ -300,6 +277,42 @@ func (r *Reconciler) reconcileBEBBackend(ctx context.Context, bebSecret *v1.Secr
 	return ctrl.Result{}, nil
 }
 
+func (r *Reconciler) syncOauth2ClientIDAndSecret(ctx context.Context, desiredStatus eventingv1alpha1.EventingBackendStatus, backend *eventingv1alpha1.EventingBackend) error {
```
mfaizanse, line 280: Should `desiredStatus` be passed by reference?  Since we are modifying it inside.
```
syncOauth2ClientIDAndSecret(ctx context.Context, desiredStatus *eventingv1alpha1.EventingBackendStatus, ...
```

components/eventing-controller/api/v1alpha1/condition.go
```diff
@@ -39,32 +42,53 @@ const (
 	ConditionReasonNATSSubscriptionActive     ConditionReason = "NATS Subscription active"
 	ConditionReasonNATSSubscriptionNotActive  ConditionReason = "NATS Subscription not active"
 	ConditionReasonWebhookCallStatus          ConditionReason = "BEB Subscription webhook call no errors status"
+
+	ConditionReasonSubscriptionControllerReady        ConditionReason = "Subscription manager and controller started"
+	ConditionReasonSubscriptionControllerNotReady     ConditionReason = "Subscription controller not started"
+	ConditionReasonPublisherDeploymentReady           ConditionReason = "Publisher proxy deployment ready"
+	ConditionReasonPublisherDeploymentNotReady        ConditionReason = "Publisher proxy deployment not ready"
+	ConditionReasonBackendCreationUpdateFailed        ConditionReason = "Backend CR creation or update failed"
+	ConditionReasonPublisherProxyCreationUpdateFailed ConditionReason = "Publisher Proxy deployment creation or update failed"
```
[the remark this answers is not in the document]
mfaizanse: (nit) Similar for other reasons, maybe its better to use `sync` instead of `create or update` and `get or update`.

components/eventing-controller/controllers/backend/reconciler.go
```diff
@@ -267,13 +236,17 @@ func (r *Reconciler) reconcileBEBBackend(ctx context.Context, bebSecret *v1.Secr
 	// Set environment with secrets for BEB subscription controller
 	err = setUpEnvironmentForBEBController(secretForPublisher)
 	if err != nil {
+		desiredStatus.SetSubscriptionControllerReadyCondition(false, eventingv1alpha1.ConditionReasonSubscriptionControllerNotReady, err.Error())
```
mfaizanse, line 239: (nit) `ConditionReasonSubscriptionControllerNotReady` does not seem appropriate here it's not ready because of an error. 

By "ConditionReasonSubscriptionControllerNotReady" message, I would understand that its in-progress w/o any errors.

components/eventing-controller/controllers/backend/reconciler.go
```diff
@@ -329,79 +342,49 @@ func setUpEnvironmentForBEBController(secret *v1.Secret) error {
...
-		}
-
 		r.namedLogger().Debug("update backend CR status when backend is nil")
-		if err := r.Client.Status().Update(ctx, desiredBackend); err != nil {
-			r.namedLogger().Errorw("update EventingBackend status failed", "error", err)
-			return err
-		}
-		return nil
+		// Do not change the value of backend type if it cannot be changed
+		desiredStatus.Backend = currentBackend.Status.Backend
```
mfaizanse, line 358: This comment `// Do not change the value of backend type if it cannot be changed` conflicts with the above comment `// When backend creation fails, marking the eventingbackend ready to false`

components/eventing-controller/controllers/backend/reconciler.go
```diff
@@ -470,14 +474,11 @@ func hasBackendTypeChanged(currentBackendStatus, desiredBackendStatus eventingv1
...
-		SubscriptionControllerReady: utils.BoolPtr(false),
-		PublisherProxyReady:         utils.BoolPtr(false),
-		EventingReady:               utils.BoolPtr(false),
-		BEBSecretName:               "",
-		BEBSecretNamespace:          "",
-	}
+func getDesiredBackendStatus() eventingv1alpha1.EventingBackendStatus {
+	defaultStatus := eventingv1alpha1.EventingBackendStatus{}
+	defaultStatus.InitializeBackendConditions()
+	defaultStatus.EventingReady = utils.BoolPtr(true)
```
mfaizanse, line 480: Why EventingReady is true by default? Should it be false by default?

Also (nit) the name `getDefaultBackendStatus()` seems better to me. `getDesiredBackendStatus` seems to me like it should return the final status with everything ready and all conditions ready (like a desire :p).

components/eventing-controller/controllers/subscription/beb/reconciler.go
```diff
@@ -633,7 +631,7 @@ func getSvcNsAndName(url string) (string, string, error) {
 // syncInitialStatus determines the desired initial status and updates it accordingly (if conditions changed)
 func (r *Reconciler) syncInitialStatus(subscription *eventingv1alpha1.Subscription) {
 	expectedStatus := eventingv1alpha1.SubscriptionStatus{}
-	expectedStatus.InitializeConditions()
+	expectedStatus.InitializeSubscriptionConditions()
```
mfaizanse, line 634: (nit) Why do we need to rename it? We are doing `Subscription.status.InitialiseConditions()` so according to GO recommentations we dont need to add `Subscription` word to `InitializeSubscriptionConditions`

components/eventing-controller/api/v1alpha1/condition.go
```diff
@@ -207,3 +240,108 @@ func (s *SubscriptionStatus) SetConditionAPIRuleStatus(ready bool) {
...
+}
+
+func (b *EventingBackendStatus) GetSubscriptionControllerReadyStatus() bool {
+	for _, condition := range b.Conditions {
+		if condition.Type == ConditionControllerReady {
+			return condition.Status == corev1.ConditionTrue
+		}
+	}
+	return false
+}
```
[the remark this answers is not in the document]
raypinto: Makes sense. Will do 👍

components/eventing-controller/api/v1alpha1/condition.go
```diff
@@ -207,3 +240,108 @@ func (s *SubscriptionStatus) SetConditionAPIRuleStatus(ready bool) {
...
+	}
+
+	// compile map of Conditions per ConditionType
+	existingMap := make(map[ConditionType]Condition, len(existing))
+	for _, value := range existing {
+		existingMap[value.Type] = value
+	}
+
+	for _, value := range expected {
+		if !ConditionEquals(existingMap[value.Type], value) {
```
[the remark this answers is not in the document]
raypinto: That's when the function `ConditionEquals` kicks in.

components/eventing-controller/api/v1alpha1/eventingbackend_types.go
```diff
@@ -27,27 +27,25 @@ type EventingBackendStatus struct {
 	// +optional
 	EventingReady *bool `json:"eventingReady"`
 
+	// Conditions defines the status conditions
 	// +optional
-	SubscriptionControllerReady *bool `json:"subscriptionControllerReady"`
-
-	// +optional
-	PublisherProxyReady *bool `json:"publisherProxyReady"`
```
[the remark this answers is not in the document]
raypinto: I thought about it, but it would be duplicated information right to have the Ready value both in the conditions and as a boolean? We can discuss this. 

Q: Does fast-integration tests use these individual booleans or just the `eventingReady` boolean?

components/eventing-controller/config/crd/bases/eventing.kyma-project.io_eventingbackends.yaml
```diff
@@ -67,12 +67,27 @@ spec:
                 description: The namespace of the secret containing BEB access tokens,
                   required only for BEB
                 type: string
+              conditions:
+                description: Conditions defines the status conditions
```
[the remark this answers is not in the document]
raypinto: This is an auto generated file. But I will see where I can override the description value 👍

components/eventing-controller/controllers/backend/reconciler.go
```diff
@@ -110,63 +111,49 @@ func (r *Reconciler) Reconcile(ctx context.Context, _ ctrl.Request) (ctrl.Result
...
 	// Default: NATS flow
-	return r.reconcileNATSBackend(ctx)
+	return r.reconcileNATSBackend(ctx, desiredStatus)
 }
 
-func (r *Reconciler) reconcileNATSBackend(ctx context.Context) (ctrl.Result, error) {
-	// CreateOrUpdate CR with NATS
+func (r *Reconciler) reconcileNATSBackend(ctx context.Context, desiredStatus eventingv1alpha1.EventingBackendStatus) (ctrl.Result, error) {
 	r.backendType = eventingv1alpha1.NatsBackendType
+	desiredStatus.Backend, desiredStatus.BEBSecretName, desiredStatus.BEBSecretNamespace = r.backendType, "", ""
```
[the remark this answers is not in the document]
raypinto: This is very backend specific right? Here we are setting the backendType. And the bebSecretName and Namespace. These values are not known when `getDesiredBackendStatus()` is called. We can discuss this.

components/eventing-controller/controllers/backend/reconciler.go
```diff
@@ -110,63 +111,49 @@ func (r *Reconciler) Reconcile(ctx context.Context, _ ctrl.Request) (ctrl.Result
...
 			return ctrl.Result{}, errors.Wrapf(err, "update status when create or update backend failed")
 		}
 		return ctrl.Result{}, errors.Wrapf(err, "create or update EventingBackend failed, type: %s", eventingv1alpha1.NatsBackendType)
 	}
 
 	// Stop the BEB subscription controller
 	if err := r.stopBEBController(); err != nil {
-		updateErr := r.UpdateBackendStatus(ctx, r.backendType, newBackend, nil, nil)
-		if updateErr != nil {
+		desiredStatus.SetSubscriptionControllerReadyCondition(false, eventingv1alpha1.ConditionReasonControllerStopStartFailed, err.Error())
```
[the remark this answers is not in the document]
raypinto: The Message would specify in detail what happened. But we can split this condition into two.

components/eventing-controller/controllers/backend/reconciler.go
raypinto: That's a fair point. I'll think of a new conditionReason then

components/eventing-controller/controllers/backend/reconciler.go
raypinto: Both statements are true. But I need to get rid of the second comment.

components/eventing-controller/controllers/backend/reconciler.go
raypinto: It's true since our desiredStatus would be to have eventingReady as `true`.

Actually that's the idea I propose: To desire the required status (where eventingReady is true), and if something doesn't match or breaks, update the status (to false) accordingly.

components/eventing-controller/controllers/backend/reconciler.go
raypinto: Yes that's right! Will change it

components/eventing-controller/controllers/subscription/beb/reconciler.go
raypinto: Its was done to differentiate between initializing `Subscription` and `Backend` conditions. But I see your point. I will change it. 👍

components/eventing-controller/api/v1alpha1/condition.go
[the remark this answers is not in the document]
mfaizanse: I mean, won't `existingMap[value.Type]` in `if !ConditionEquals(existingMap[value.Type], value)` return null in that case?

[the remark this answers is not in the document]
raypinto: Good catch. Will look into it.

components/eventing-controller/controllers/backend/reconciler.go
raypinto: As discussed,  we will reuse the `ConditionReasonSubscriptionControllerStartFailed` Reason.

components/eventing-controller/api/v1alpha1/condition.go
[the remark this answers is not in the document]
raypinto: Apparently, it's an empty key and value and does not return nil. I have added the testcase for it anyway.

components/eventing-controller/api/v1alpha1/condition.go
```diff
@@ -207,3 +241,108 @@ func (s *SubscriptionStatus) SetConditionAPIRuleStatus(ready bool) {
...
+func (b *EventingBackendStatus) IsSubscriptionControllerStatusReady() bool {
+	for _, condition := range b.Conditions {
+		if condition.Type == ConditionControllerReady {
+			return condition.Status == corev1.ConditionTrue
+		}
+	}
+	return false
+}
+
+func (b *EventingBackendStatus) GetPublisherReadyStatus() bool {
```
mfaizanse, line 305: (nit) same for this naming `IsPublisherStatusReady`?

components/eventing-controller/controllers/backend/reconciler.go
```diff
@@ -103,6 +103,7 @@ func NewReconciler(ctx context.Context, natsSubMgr subscriptionmanager.Manager,
 
 func (r *Reconciler) Reconcile(ctx context.Context, _ ctrl.Request) (ctrl.Result, error) {
 	var secretList v1.SecretList
+	desiredStatus := getDefaultBackendStatus()
```
mfaizanse, line 106: (nit) Can you add comments here explaining:
- getDefaultBackendStatus sets all the conditions and the eventingReady status to true.
- and if something breaks then we update the fields in status to false accordingly.

mfaizanse (approved):

PR #13867, opened 2022-04-06
Improve retry logic in telemetry-operator

chrkl:
**Description**

Errors from the Kubernetes API were not retried and could cause incomplete fluent-bit configurations. This change re-queues non-permanent errors and makes the reconciliation logic more robust to partial failures.

- Retry non-permanent Kubernetes API errors
- Fix linting errors

**Related issue(s)**


components/telemetry-operator/controllers/logpipeline_controller.go
```diff
@@ -206,6 +206,17 @@ func (r *LogPipelineReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 	return ctrl.Result{}, nil
 }
 
+// Indicate if an error from the kubernetes client should be retried. Errors caused by a bad request or configuration should not be retried.
+func requeueRequest(err error) bool {
```
lindnerby, line 210: Maybe `shouldRetryOn` for the func name?

components/telemetry-operator/controllers/logpipeline_controller.go
```diff
@@ -242,34 +254,40 @@ func (r *LogPipelineReconciler) syncSectionsConfigMap(ctx context.Context, logPi
...
+			controllerutil.AddFinalizer(logPipeline, sectionsConfigMapFinalizer)
+			changed = true
 		}
-		controllerutil.AddFinalizer(logPipeline, sectionsConfigMapFinalizer)
 	}
 
 	// Update ConfigMap
-	if err := r.Update(ctx, &cm); err != nil {
-		return false, err
+	if changed {
```
lindnerby, line 278: May sound picky, since we already have quite a lot of nesting going on (thankfully we have a refactoring issue planned :)
but if you reduce nesting here by `if !changed` just return `false, nil`
as well as in other sync methods

components/telemetry-operator/controllers/logpipeline_controller.go
```diff
@@ -242,34 +254,40 @@ func (r *LogPipelineReconciler) syncSectionsConfigMap(ctx context.Context, logPi
...
+		if err = r.Update(ctx, &cm); err != nil {
+			return changed, err
```
lindnerby, line 280: only sync method where `changed` is return instead of `false` in err case. is there are reason?

components/telemetry-operator/controllers/logpipeline_controller.go
```diff
@@ -414,34 +439,33 @@ func (r *LogPipelineReconciler) syncSecretRefs(ctx context.Context, logPipeline
 				if _, hasKey := secret.Data[k]; hasKey {
 					delete(secret.Data, k)
 					controllerutil.RemoveFinalizer(logPipeline, secretRefsFinalizer)
-					//nolint:ineffassign
```
lindnerby, line 417: :D thank you

lindnerby (changes requested):
Looks good so far, although review is really getting harder because of code readability. We should definitely start with refactoring after this one :)

PR #13876, opened 2022-04-07
Adapt PodPreset to k8s 1.22

piotrmiskiewicz:
**Description**

- change PodPreset CRD definition to version v1 and adapt Spec, add `x-kubernetes-preserve-unknown-fields: true` for necessary places

**Related issue(s)**


piotrmiskiewicz:
Tested with the following scenario:
1. Install Kyma with old CRD podpreset definition.
2. Create a deployment, serviceinstance, servicebinding and inject the binding into the deploy (SBU)
3. apply new PodPreset CRD
4. The env still injected
5. Remove SBU and create it again
6. Env is injected once again

PR #13890, opened 2022-04-11
Add JetStream retention policy workqueue option for stream storage

marcobebway:
####  Description

Add the JetStream retention policy workqueue option for the stream storage.

#### Notes

- The retention policy defined in the Eventing charts is not changed.


marcobebway:
Not needed anymore.

PR #13972, opened 2022-04-13
Fix circular dependency between Eventing and commerce mock in the fast-integration tests

marcobebway:
**Description**

Fix circular dependency between Eventing and commerce mock in the fast-integration tests.

**Examples**
- Sample run before fixing the circular dependency is [here](https://storage.googleapis.com/kyma-prow-logs/pr-logs/pull/kyma-project_kyma/13779/marcobebway_test_of_prowjob_pre-main-kyma-gardener-gcp-eventing/1513920647124750336/build-log.txt).
- Sample run after fixing the circular dependency is [here](https://storage.googleapis.com/kyma-prow-logs/pr-logs/pull/kyma-project_kyma/13972/pre-main-kyma-gardener-gcp-eventing/1514174122769256448/build-log.txt).

### DIFF
--- a/tests/fast-integration/eventing-test/common/common.js
+++ b/tests/fast-integration/eventing-test/common/common.js
@@ -1,0 +1,23 @@
+const fs = require('fs');
+const eventMeshSecretFilePath = process.env.EVENTMESH_SECRET_FILE || '';
+const eventMeshNamespace = getEventMeshNamespace();
+const natsBackend = 'nats';
+const bebBackend = 'beb';
+
+// returns the EventMesh namespace from the secret.
+function getEventMeshNamespace() {
+  try {
+    const eventMeshSecret = JSON.parse(fs.readFileSync(eventMeshSecretFilePath, {encoding: 'utf8'}));
+    return '/' + eventMeshSecret['namespace'];
+  } catch (e) {
+    console.error(e);
+    return undefined;
+  }
+}
+
+module.exports = {
+  eventMeshSecretFilePath,
+  eventMeshNamespace,
+  natsBackend,
+  bebBackend,
+};
--- a/tests/fast-integration/eventing-test/eventing-test-prep.js
+++ b/tests/fast-integration/eventing-test/eventing-test-prep.js
@@ -12,7 +12,6 @@ const {
   isSKR,
   backendK8sSecretName,
   backendK8sSecretNamespace,
-  eventMeshSecretFilePath,
   timeoutTime,
   slowTime,
   gardener,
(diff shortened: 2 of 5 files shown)


tests/fast-integration/eventing-test/common/common.js
```diff
@@ -0,0 +1,23 @@
+const fs = require('fs');
+const eventMeshSecretFilePath = process.env.EVENTMESH_SECRET_FILE || '';
+const eventMeshNamespace = getEventMeshNamespace();
```
mfaizanse, line 3: (nit) Can we only call `getEventMeshNamespace()` if `eventMeshSecretFilePath` is not empty? Otherwise, we would have error log in console every time we run tests for nats.

mfaizanse, line 3: (nit) Can we only call `getEventMeshNamespace()` if `eventMeshSecretFilePath` is not empty? Otherwise, we would have error log in console every time we run tests for nats.

tests/fast-integration/eventing-test/common/common.js
```diff
@@ -0,0 +1,23 @@
...
+const bebBackend = 'beb';
+
+// returns the EventMesh namespace from the secret.
+function getEventMeshNamespace() {
+  try {
+    const eventMeshSecret = JSON.parse(fs.readFileSync(eventMeshSecretFilePath, {encoding: 'utf8'}));
+    return '/' + eventMeshSecret['namespace'];
+  } catch (e) {
+    console.error(e);
+    return undefined;
```
mfaizanse, line 14: (nit) Should we throw the error, so we don't continue the tests if we fail here for beb?

mfaizanse, line 14: (nit) Should we throw the error, so we don't continue the tests if we fail here for beb?

tests/fast-integration/eventing-test/common/common.js
marcobebway: I just moved the code here to fix the circular dependency.
Also this situation is not going to happen because this condition here: https://github.com/marcobebway/kyma/blob/ea8923a743c02dde991baebb5c0757e1012e5044/tests/fast-integration/eventing-test/eventing-test.js#L127.

tests/fast-integration/eventing-test/common/common.js
marcobebway: Same as [here](https://github.com/kyma-project/kyma/pull/13972/files#r849374409).

mfaizanse (approved):

rakesh-garimella (approved):

Issue #13872, opened 2022-04-07
Flaky test case : "Getting Started Guide Tests"  


kwiatekus:
**Description**

kyma-integration-k3d test sometimes fails with 

```
  1) Executing Standard Testsuite:
       Getting Started Guide Tests:
         Order should be persisted and should survive pod restarts (redis storage):
     Error: Order not found: 762727210
      at findOrder (test/fixtures/getting-started-guide/index.js:98:9)
      at runMicrotasks (<anonymous>)
      at processTicksAndRejections (internal/process/task_queues.js:95:5)
      at async verifyOrderPersisted (test/fixtures/getting-started-guide/index.js:84:3)
      at async Context.<anonymous> (test/2-getting-started-guide.js:36:7)
```
I.e [here](https://storage.googleapis.com/kyma-prow-logs/logs/kyma-integration-k3d/1511842544034516992/build-log.txt) 

**Troubleshooting**

Let's make sure that the redis storage instance is properly bound to the function. ( i.e via additional assertion )
The symptoms show that the order is not found after function pod restart.
This could be the because of problems with service binding - therefore after  function pod restarts the inmemory storage is lost.
We had been experiencing issues with service bindings to function pods here as well
https://github.com/kyma-project/kyma/issues/13443

kwiatekus:
Some of the failing tests can be found [here](https://storage.googleapis.com/kyma-prow-logs/pr-logs/pull/kyma-project_control-plane/1521/pre-main-control-plane-reconciler-e2e/1511687021553258496/build-log.txt), [here](https://storage.googleapis.com/kyma-prow-logs/pr-logs/pull/kyma-project_control-plane/1477/pre-main-control-plane-reconciler-e2e/1509153873531506688/build-log.txt
), [here](https://storage.googleapis.com/kyma-prow-logs/pr-logs/pull/kyma-project_control-plane/1477/pre-main-control-plane-reconciler-e2e/1508791476417466368/build-log.txt) and [here](https://storage.googleapis.com/kyma-prow-logs/pr-logs/pull/kyma-project_control-plane/1467/pre-main-control-plane-reconciler-e2e/1504732331099820032/build-log.txt).

PR #13975, opened 2022-04-13
Reduce flakiness of getting started e2e tests

kwiatekus:
**Description**

- Improve verification if redis is actually bound to the orders-service in the getting-started test case

**Related issue(s)**
Fixes https://github.com/kyma-project/kyma/issues/13872

### DIFF
--- a/tests/fast-integration/test/fixtures/getting-started-guide/index.js
+++ b/tests/fast-integration/test/fixtures/getting-started-guide/index.js
@@ -81,10 +81,22 @@ async function verifyOrderPersisted() {
 
   await deleteAllK8sResources('/api/v1/namespaces/orders-service/pods', {labelSelector: `app=${orderService}`});
 
-  await retryPromise(() => findOrder(serviceDomain, order), 30, 2000);
+  await retryPromise(() => findOrder(serviceDomain, order), 50, 2000);
 
   // https://kyma-project.io/docs/root/getting-started/#getting-started-connect-an-external-application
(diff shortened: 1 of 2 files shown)


tests/fast-integration/test/fixtures/getting-started-guide/index.js
```diff
@@ -131,7 +132,14 @@ async function ensureGettingStartedTestFixture() {
   await waitForServiceBinding(orderService, orderService);
   await k8sApply([sbuObj], orderService);
   const sbu = await waitForServiceBindingUsage(orderService, orderService);
-  await waitForPodWithSbuToBeReady(sbu);
+  const pod = await waitForPodWithSbuToBeReady(sbu);
```
pbochynski, line 135: What if pod doesn't have the required env variables injected yet? I don't see retry here. Wouldn't it be better to return these env variables values directly from order-service? I see this `kubectl exec` is a little bit tricky.

kwiatekus: `waitForPodWithSbuToBeReady` checks if the pod has been labeled with sbu usage and if it's ready.
After that the envs has to be in.. no need to retry.. if they arent there its ok to fail the test.

kwiatekus: @pbochynski done. pls have a look

pbochynski (approved):

PR #14017, opened 2022-04-21
fast integration tests without sc - p2

tillknuesting:
use of noSC boolean as to be found to be unnecessary


wozniakjan (dismissed):
/lgtm
/approve

Issue #13718, opened 2022-03-24
Remove Application Registry and Connector Service from Kyma


janmedrek:
**Description**

Application Registry and Connector Service should be removed from Application Connector. There should be Compass and non-Compass versions of the Application Connector chart, it should also be simplified as much as possible.

In short, the default Kyma OS installation should no longer contain:
- Application Registry
- Connector Service
- Connection Token Handler
- Connectivity Validator (still present in the Compass/managed scenario)
- Istio mTLS gateway (still present in the Compass/managed scenario)
- Various jobs that set up the certificate flow (may still be needed in Compass/managed scenario)

The **default flow with Compass stays unchanged**, so we need to make sure that all the relevant components stay, thus **it should be possible to configure the whole Application Connector chart to either Compass or non-Compass mode**.

This change will force either:
- A simplified setup, where certificates are not needed in order to send events and the registration flow is replaced by a single YAML file containing Application CR (will be described in the latter part as a blogpost explaining that setup).
- A setup where Kyma and Compass are talking to each other and are covering the existing functionalities of certificate setup and registration flow.

The pipelines testing the flow should be obviously removed, **we need to keep those that do the Compass integration** as it will be the new way of working with Kyma and certificates.

**Reasons**

We are maintaining the solution for a very long time already. Since it was [deprecated back in mid-2020](https://kyma-project.io/blog/2020/5/29/release-notes-113#deprecated-connector-and-application-registry-ap-is) it brought near-zero value and only resulted in big maintenance effort regarding pipelines and security issues. We do have alternatives in form of Compass or other approaches to Kyma-External System communication. That's why we've decided to drop the flow and rely completely on Compass in this area.

This change will reduce the maintenance effort, simplify the whole design, and drop one of the more-complicated flows that are not really needed anymore in Kyma.

**Acceptance Criteria**
- Application Connector available in Compass and non-Compass version
- Non-Compass flavor does not support certificate and registration flows
- Flow with Compass stays untouched
- Unnecessary components and jobs removed
- Tests and pipelines adjusted and added/removed
- Whole Compass-related flow tested with an external system

janmedrek:
Here is a short visual representation of what needs to be removed. The components marked in red are to be simply deleted, while the orange ones have some use-case in the Compass scenario, and should only be present if Compass Runtime Agent and its configuration has been applied on the cluster.

[...]

PR #14073, opened 2022-04-22
Remove Application Registry and Connector Service from FIT

VOID404:
**Description**

Make Fast Integration Tests not depend on Application Registry and Connector Service.

**Related issue(s)**
- Part of #13718


VOID404:
- removed service registration
  - [ ] create Application CR by hand
  - [ ] create secrets by hand
  - [ ] make sure all URLs are correct (since they are made by hand now)
  - [ ] clean up resource creation (right now new resources are smooshed into existing YAMLs)
- removed eventing endpoint
  - [ ] expose eventing with unsecured API Rule (or remove external eventing)

Issue #12353, opened 2021-10-13
Application Connector reorganization and cleanup


janmedrek:
**Description**

There is a lot of legacy code and config that need to be cleaned up.

**Reasons**

We want to have a clear and easily understandable structure of code, charts, jobs,  and configuration files.

**Acceptance Criteria**

(issues to be created)

- Remove Rafter integration
- Remove AccessService creation from Application Registry (no longer used with Central App Gateway)
- Remove Application Operator (same as above)
- Remove Application Broker (ServiceCatalog integration going away)
- Rename 'central-' components as they are now default
- Extract Application Registry and Connector Service to a separate chart outside of Kyma
- Remove `kyma-integration` namespace
- Remove `compass-system` namespace

franpog859:
The `disableLegacyConnectivity` flag is not removed [in my PR](https://github.com/kyma-project/kyma/pull/13296/files), @janmedrek

franpog859:
@akgalwas , as the clean-up part at this point, can be done separately from the operator implementation, I created two separate issues - one for the clean-up process - https://github.com/kyma-project/kyma/issues/15824 - and the second one for the operator implementation - https://github.com/kyma-project/kyma/issues/15826. Therefore I'm closing this issue as a sort of duplicate. I copied all the "funnel process" and "kyma project" values to the newly created issues. If they should be changed, please, change them or help me do it. If this obsolete issue should be removed from some project, please, remove it or help me do it

PR #14095, opened 2022-04-26
Introduce post 2.4 cleanup script for orphaned resources

kwiatekus:
**Description**

- Cleanup CRD, deployments related to rafter, service catalog, application registry

**Related issue(s)**
https://github.com/kyma-project/kyma/issues/13422
https://github.com/kyma-project/kyma/issues/12353


mvshao (approved):
LGTM

janmedrek (changes requested):
We cannot delete the Application Gateways, this will kill all the flows that are set up using the old way, we need to give users time to migrate. Also, what about the Application Connector deployments? Those will be orphaned as well and I do not see them being removed anywhere.

We need to:
- Remove Application Operator
- Remove Application Broker
- Do not remove Application Gateways and their respective services to keep the "old flow" operational
- We are slowly finalizing the https://github.com/kyma-project/kyma/issues/13718, perhaps we can include resources affected by this task as well? I believe we should avoid having two different, separate cleanup scripts. Can we somehow handle all the cases (CMP and non-CMP) in a single one? (Assuming that in neither of those scenarios we want the Application Gateways gone). This would refer further to components, Istio resources (non-CMP), and some jobs (non-CMP).

docs/assets/2.2-2.3-cleanup-orphaned-svcat-resources.sh
```diff
@@ -0,0 +1,48 @@
...
+echo "Deleting Service Catalog"
+
+kubectl delete crd addonsconfigurations.addons.kyma-project.io
+kubectl delete crd clusteraddonsconfigurations.addons.kyma-project.io
+kubectl delete crd podpresets.settings.svcat.k8s.io
+kubectl delete crd clusterservicebrokers.servicecatalog.k8s.io
+kubectl delete crd clusterserviceclasses.servicecatalog.k8s.io
+kubectl delete crd clusterserviceplans.servicecatalog.k8s.io
+kubectl delete crd servicebindings.servicecatalog.k8s.io
+kubectl delete crd servicebindingusages.servicecatalog.kyma-project.io
```
PK85, line 43: I see sc-migration reconciler did not remove instances of SBU crd, we need to remove all finalizers from instances before removing SBU CRD

docs/assets/2.2-2.3-cleanup-orphaned-svcat-resources.sh
```diff
@@ -0,0 +1,48 @@
...
+kubectl -n kyma-system delete deployment rafter-ctrl-mngr
+kubectl -n kyma-system delete deployment rafter-front-matter-svc
+kubectl -n kyma-system delete deployment rafter-minio
+kubectl -n kyma-system delete deployment rafter-upload-svc
+
+echo "Deleting Service Catalog"
+
+kubectl delete crd addonsconfigurations.addons.kyma-project.io
+kubectl delete crd clusteraddonsconfigurations.addons.kyma-project.io
+kubectl delete crd podpresets.settings.svcat.k8s.io
```
akgalwas, line 38: This should not be removed.

akgalwas:
There was a separate [PR](https://github.com/kyma-project/kyma/pull/14413) for 2.3 created to handle clean up related to Application Registry and Connector Service removal (https://github.com/kyma-project/kyma/issues/13718).

docs/assets/2.2-2.3-cleanup-orphaned-svcat-resources.sh
wozniakjan: I think that was potentially some transient error, the SBUs eventually got removed and it's implemented as part of [scmigration reconciler](https://github.com/kyma-incubator/reconciler/blob/5ff701d373650dc494cabc588bd902277b187f6a/pkg/reconciler/instances/scmigration/sc_removal.go#L292-L296) along with [handling of finalizer and owner references](https://github.com/kyma-incubator/reconciler/blob/5ff701d373650dc494cabc588bd902277b187f6a/pkg/reconciler/instances/scmigration/sc_removal.go#L62)

wozniakjan:
hey @kwiatekus, @PK85 had an idea about dumping some stats in the beginning and/or the end of the script execution. For example, how many resources for a particular CRD exist before it's deleted (there should be usually 0 but in case of finalizers or some unexpected user action, some could still exist)

janmedrek:
To avoid old Application Gateways being undeployed, we need to make sure that Application Operator is removed *before* ServiceInstances are gone. If we do it another way, the customer workload may be influenced, resulting in the flow being broken.

docs/assets/2.3-2.4-cleanup-orphaned-svcat-resources.sh
```diff
@@ -0,0 +1,60 @@
...
+kubectl -n kyma-integration delete virtualservice application-operator
+kubectl -n kyma-integration delete destinationrule application-operator-health-rule
+kubectl delete podsecuritypolicy application-operator
+kubectl delete clusterrolebinding application-operator
+kubectl delete clusterrole application-operator
+kubectl -n kyma-integration delete serviceaccount application-operator
+
+kubectl delete serviceinstance --all-namespaces --all
+
+echo "Deleting rafter"
```
majakurcius, line 26:
```suggestion
echo "Deleting Rafter"
```

franpog859: done

majakurcius (approved):

docs/assets/2.3-2.4-cleanup-orphaned-svcat-resources.sh
```diff
@@ -0,0 +1,60 @@
...
+kubectl -n kyma-integration delete service application-operator-health
+kubectl -n kyma-integration delete service application-operator-service
+kubectl -n kyma-integration delete virtualservice application-operator
+kubectl -n kyma-integration delete destinationrule application-operator-health-rule
+kubectl delete podsecuritypolicy application-operator
+kubectl delete clusterrolebinding application-operator
+kubectl delete clusterrole application-operator
+kubectl -n kyma-integration delete serviceaccount application-operator
+
+kubectl delete serviceinstance --all-namespaces --all
```
janmedrek, line 24: Won't it affect the BTP ServiceInstances as well? The name is the same, but the group differs.

Anyways, the `serviceinstances.servicecatalog.k8s.io` CRD is removed at the end of the script, will it remove the CRs as well? If yes - we can skip that part, I am not sure we should directly influence what customer provisioned in their cluster.

janmedrek (changes requested):

docs/assets/2.3-2.4-cleanup-orphaned-svcat-resources.sh
franpog859: Hmm, you may be right. I'll delete this command as it's not a part of the Application Connector anyway

janmedrek (approved):

grego952 (approved):

PR #14118, opened 2022-04-28
Make slack FIRING alert only contains Firing alerts instead of both Firing and Resolved

vvxxvvxx:
**Description**

- Make slack alert FIRING notification only contains Firing alerts instead of including both Firing and Resolved alets.

**Related issue(s)**

Current slack firing alert contains both Firing and Resolved alerts which cause confusion.

### DIFF
--- a/resources/monitoring/values.yaml
+++ b/resources/monitoring/values.yaml
@@ -278,7 +278,7 @@ alertmanager:
        icon_emoji: ":ghost:"
        title: '{{`{{ template "__subject" . }}`}} ({{ .Values.global.domainName }})'
        title_link: 'https://console.{{ .Values.global.domainName }}'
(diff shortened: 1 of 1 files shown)


skhalash:
@vvxxvvxx Thanks for you contribution, will take a look

skhalash:
@vvxxvvxx Why would you need such a change? Currently, both Firing and Resolved alerts are getting pushed to Slack, which is the expected default behavior.

vvxxvvxx:
> @vvxxvvxx Why would you need such a change? Currently, both Firing and Resolved alerts are getting pushed to Slack, which is the expected default behavior.

@skhalash please refer to the issue https://github.tools.sap/kyma/backlog/issues/2495. now as `range .Alerts` is defined in the text, so it contains both Firing and Resolved alerts which will cause confusion..

skhalash:
@vvxxvvxx Ok, got it. But does it mean that resolved alerts are not pushed to Slack at all or they are not getting mixed up?

skhalash:
After a discussion with @vvxxvvxx it was decided for now that in this case a custom alertmanager config can be used. However, I agree that the default Slack template can be improved, will discuss it with the team.

Issue #13515, opened 2022-02-28
POC - Istio component test


strekm:
**Description**

Implement test in go listing istio pods. Test should assume that cluster is already provisioned and Istio installed. Test should be implemented in go lang assuming it will be integrated with cucumber and gherkin. 

**Reasons**

Reducing effort of manual testing.

**Attachments**

https://github.com/kyma-project/kyma/issues/13517

PR #14133, opened 2022-05-02
add basic istio integration test

piotrkpc:
**Description**

- add istio integration tests
- start integration with it using Gherkin files as inputs
- ...

**Related issue(s)**

https://github.com/kyma-project/kyma/issues/13515


piotrkpc:
Some experiences from playing with this:
- [godog](https://github.com/cucumber/godog) is currently the most popular tool for bdd style tests with Gherkin I've found. It's still before it's major release and until then: `The API is likely to change a few times before we reach 1.0.0`
- There are multiple ways to integrate with godog: the most simple ones does not fit well with golang stdlib testing framework, the more complex ways can let you run test without godog command installed. imo, in the followup task that would use the gherkin we should use the integration with the stdlib testing.
- Depending on scenario, we would need sometimes to re-trigger installation from test code, hence deploying cluster via makefile is sub-optimal and it would be nice to have Golang bindings for that. For this case, makefile is good enough.
- Flow would potentially look different if we would integrate Gherkin with KUTTL.
  - KUTTL-flow: Gherkin file input -> godog tests scans the kuttl directory structure and validates fields against Gherkin input -> after validation it would start the kuttl test suite using kuttl golang bindings or run selected steps of kuttl test.
  - standard flow: would require a lot of retries and better knowledge of client-go package.

barchw (dismissed):

barchw (approved):

cnvergence (approved):

cnvergence:
Great PoC for component tests 👍🏻

strekm:
> What is the value of this test? The test doesn't even check if istio works. We have a few other integration tests that test istio ingress gateway already in fast-integration folder. Please justify what problem you solve with this test.

@pbochynski we want to have quick, cheap isolated istio component test, which will help us catch obvious bugs during istio bumps. it is not completed yet and it will be continued with more PRs. with some time we want to introduce `istioctl analyse`. we can also implement a flow when fast int tests are not triggered when this test is failing

pbochynski (approved):

PR #14146, opened 2022-05-02
Make EmsSubscriptionStatus a pointer type

raypinto:
**Description**

Made EmsSubscriptionStatus a pointer type in the Subscription Status, so that `omitempty` works when the object is empty.
A kyma subscription that uses the NATS backend, does not need to have EmsSubscriptionStatus field in the status. 

**Related issue(s)**


marcobebway (approved):
/lgtm

PR #14153, opened 2022-05-03
Limit Fluent Bit plugin usage in LogPipeline

### LINKED ISSUE #13146: [Telemetry Operator] Limit the initial plugins to a minimal set
**Description**
The goal is to offer the most open and typical integration options. Anything specific should be covered by running a custom fluentd instance. Assure that the offered feature set is minimal for now and that the way of doing advanced integrations is documented.

**Reasons**
By keeping the options minimal in the beginning the operations will be simplified a lot and improved over time to support even more in future with a more mature setup.

**Criterias**
- Validate what elements are usually needed and can be really supported by the kyma team (for example for a typical ELK integration via HTTP output) and enable only these ones (to not get new features in accidently when updating)
  - filters
  - outputs
 - Either use whitelisting, strict schema for the CRD or bundle only supported plugins in the binary
 - Loki, fluentd via forward and ELK via fluentd HTTP scenario must be supported

chrkl:
**Description**

The unlimited use of Fluent Bit plugin can cause unexpected behaviour in terms of stability and resource usage. This PR allows a whitelisting of filter and output plugins in the telemetry-operator. This whitelist can be used to permit only approved plugin to be used in LogPipeline instances.

- Allow limiting the usable filter and output plugins in telemetry-operator
- Bump buildpack image for telemetry-operator to use Golang 1.18
- Limit the output plugins for the telemetry chart to `grafana-loki` and `http`

**Related issue(s)**

Resolves #13146


components/telemetry-operator/internal/fluentbit/sections.go
```diff
@@ -0,0 +1,24 @@
...
+)
+
+func parseSection(section string) (map[string]string, error) {
+	result := make(map[string]string)
+
+	for _, line := range strings.Split(section, "\n") {
+		line = strings.TrimSpace(line)
+		if len(line) == 0 || strings.HasPrefix(line, "#") {
+			continue
+		}
```
skhalash, line 15: Let's add a unit test for comments as well

skhalash (changes requested):

skhalash (approved):

Issue #13828, opened 2022-04-04
Improve Usability of Function CRD


kwiatekus:
**Description**

Provide a stable version of Function CRD incorporating the sub-type division between the inline and git-sourced types

**AC**
 - [x] Clean up spec fields that are spread on the root level (i.e source related) - [branch](https://github.com/m00g3n/kyma/tree/fn-crd-refactor)
 - [x] Clean up scaling related fields (and support scale subresource to make Function scalable by any scaler) - [PR](https://github.com/m00g3n/kyma/pull/3)
 - [x] Implement conversion webhooks so that old specs requested by non-migrated clients are supported - [Issue](https://github.com/kyma-project/kyma/issues/14593)
 - [x] Plan an issue for next serverless apiversion support in kyma CLI [issue](https://github.com/kyma-project/cli/issues/1314)
 - [x] Plan an issue for next serverless apiversion support in Kyma Dashboard
 - [x] Document CRD changes - [issue](https://github.com/kyma-project/kyma/issues/15002)
 
**Reasons**
 - cleaning up API of the Function CRD

**Attachments**
![serverless-architecture drawio](https://user-images.githubusercontent.com/6783567/180386324-71585751-3d8e-44d1-ace2-426cf772a58d.png)

[serverless-architecture.drawio.zip](https://github.com/kyma-project/kyma/files/8588716/serverless-architecture.drawio.zip)

Disper:
Work plan:
- [ ] generate the project 
- [ ] put it into the repository
- [ ] adapt function controller to support watching
- [ ] create new reconciler to handle it
- [ ] add tests to cover scenarios

kwiatekus:
We need to poc the following:

 - getting rid of GitRepo CRD !!
 - propose changes in Function CRD (eliminate git related pros ??)
 - design GitFunction CRD 
 - decision whether we need a dedicated controller (or have a reconciliation branch in the function controller)

m00g3n:
https://github.com/kyma-project/kyma/pull/14142

kwiatekus:
Currently blocked by missing conversion between versions
https://github.com/kyma-project/kyma/issues/14593

PR #14197, opened 2022-05-05
Use fetch instead of clone for reference update in function controller

moelsayed:
**Description**

- Use fetch instead of clone for reference update in function controller
- Update internal healthz check timeout
- Update chart probs configuration

**Related issue(s)**

https://github.com/kyma-project/kyma/issues/13828

### DIFF
--- a/components/function-controller/internal/git/go2git_test.go
+++ b/components/function-controller/internal/git/go2git_test.go
@@ -65,10 +65,10 @@ func TestNewGit2Go_LastCommit(t *testing.T) {
 		t.Run(testcase.name, func(t *testing.T) {
 			repoPath := prepareRepo(t)
 			defer deleteTmpRepo(t, repoPath)
-			cloner := &git2goClonerMock{repoPath: repoPath}
+			fetcher := &git2goFetcherMock{repoPath: repoPath}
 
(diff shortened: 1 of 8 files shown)


components/function-controller/internal/git/go2git_test.go
```diff
@@ -65,10 +65,10 @@ func TestNewGit2Go_LastCommit(t *testing.T) {
 		t.Run(testcase.name, func(t *testing.T) {
 			repoPath := prepareRepo(t)
 			defer deleteTmpRepo(t, repoPath)
-			cloner := &git2goClonerMock{repoPath: repoPath}
+			fetcher := &git2goFetcherMock{repoPath: repoPath}
 
-			opts := Options{Reference: testcase.refName}
-			client := Git2GoClient{cloner}
+			opts := Options{Reference: testcase.refName, URL: repoPath}
```
dbadura, line 70: URL as `repoPath`?

moelsayed: Well, the struct member is called `URL` and the value we are passing is a path. Internally git2go accepts paths, http(s) and ssh url values.

dbadura (approved):

Issue #14235, opened 2022-05-09
Investigate possible performance improvements


kwiatekus:
**Description**
Multiple git functions result in fetching sources frequently. In case the size of git repository is big, it uses all resources and crashes the controler 

**Symptoms**
 - controller is CPU throttled - revisit CPU resources (requested and limits )
 - 5 functions declared simultaneously - total readiness time of all of them is lower than from before refactor
 - controller is entering a `crashloopBackoff`

**To Investigate**
 - optimise mounting of function sources into the build job
 - workflow for multiple functions reconciliation
 - arent'we throttled by github ? (Network bandwidth )
 - skip code fetching in some reconciliation conditions (currently it done excessively)
 - local cache per repo - to fetch only deltas
 
 
 **Attachments**
<img width="1326" alt="Screenshot 2022-05-06 at 14 23 57" src="https://user-images.githubusercontent.com/6783567/167364814-e9eb0175-d672-4a2d-9fdc-7bfbd39c0b12.png">

kwiatekus:
I used a test runtime with kyma 2.2
I have setup multiple functions. ( many of those are git functions )
I upgraded kyma to main version

After update I see that serverless-ctrl-mngr pod doesnt restart and doesnt suffer from CPU throttling that was present before upgrade.
Thats good news 👏 

However I dont think I get how this label should work
https://github.com/kyma-project/kyma/pull/14251/files#diff-48c000138a07f94412f9f8d07d5eda1152ea8f8c0c4cf7ebcbeceb457d1bb9e7R87

I labeled one of the git functions with 
```yaml
  labels:
    serverless.kyma-project.io/continuousGitCheckout: 'true'
 ``` 
and I dont see that it reacts to git code changes more often than a function that is sourced from the same repo buut doesnt have the label.

Actually I observed quite  oposite  - labeled function is rebuilt much later than non labeled.
Do I do sth wrong?

moelsayed:
Consider the following scenario: 
- You are working on a git function with a large git repo like test-infra. Before the performance optimization changes, we checked the code for changes _every time_ the function is reconciled. So, if you apply the function, and _during it's initial_ reconcile, pushed another change _before_ the function is ready, the changes will be picked immediately, and the new updates should be be used.

With the performance changes, we only check the source once at the beginning of the reconcile loop, and don't check it again until the function is ready and a cool off period has passed. The label allows the user to go back to the previous behavior if needed.

[...]

PR #14251, opened 2022-05-10
Avoid repeated git function version check

moelsayed:
**Description**

- By default, skip git revision check if the last time it was done is less than `FunctionReadyRequeueDuration`
- Define a function label to disable this default if needed
- Add helper methods to the function status.

**Related issue(s)**

https://github.com/kyma-project/kyma/issues/14235

### DIFF
--- a/docs/03-tutorials/00-serverless/svls-02-create-git-function.md
+++ b/docs/03-tutorials/00-serverless/svls-02-create-git-function.md
@@ -84,7 +84,7 @@ Follow these steps:
         secretName: # "git-creds-basic" or "git-creds-key"
     ```
    
-    >**NOTE:** To avoid performance degradation caused by large git repositories and large monorepos, the function controller implements a configurable backoff period for source checkout based on `APP_FUNCTION_REQUEUE_DURATION`. This behavior can be disabled, allowing the controller to perform source checkout with every reconciliation loop by labeling the function CR with the label `serverless.kyma-project.io/continuousGitCheckout: true`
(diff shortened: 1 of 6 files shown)


docs/03-tutorials/00-serverless/svls-02-create-git-function.md
```diff
@@ -84,8 +84,7 @@ Follow these steps:
         secretName: # "git-creds-basic" or "git-creds-key"
     ```
    
-    >**NOTE:** It is recommended to use small repositories to source your git functions. Using large multi-repos can cause problems with function controller stability as it may run out of CPU and i/o resources.
-
+    >**NOTE:** To avoid performance degradation caused by large git repositories and large monorepos, the function controller implements a configurable backoff period for source checkout based on `APP_FUNCTION_REQUEUE_DURATION`. This behavior can be disabled, allowing the controller to perform source checkout with every reconciliation loop by labeling the function CR with the label `serverless.kyma-project.io/continuousGitCheckout: true`
```
grego952, line 87:
```suggestion
    >**NOTE:** To avoid performance degradation caused by large Git repositories and large monorepos, Function Controller implements a configurable backoff period for the source checkout based on `APP_FUNCTION_REQUEUE_DURATION`. This behavior can be disabled, allowing the controller to perform the source checkout with every reconciliation loop by labeling the Function CR with the label `serverless.kyma-project.io/continuousGitCheckout: true`
```

grego952 (dismissed):

pPrecel (approved):

grego952 (approved):

Issue #14245, opened 2022-05-10
Liveness/Readiness probes config for serverless controller need revisiting


kwiatekus:
**Description**
Internal health check is should verify if the controller is actually working and not stuck from some reason. This can be achieved using the internal controller-runtime implementations of readiness/liveness checks.
We should at look into those.
We should reduce our custom backoff configuration.

PR #14264, opened 2022-05-11
Use healthz instead of metrics and increase liveness probe

dbadura:
**Description**

- implement readyz endpoint
- fix readiness probe by increasing timeout as copying long repo can trigger liveness.

**Related issue(s)**
https://github.com/kyma-project/kyma/issues/14245

### DIFF
--- a/components/function-controller/cmd/manager/main.go
+++ b/components/function-controller/cmd/manager/main.go
@@ -4,8 +4,9 @@ import (
 	"errors"
 	"fmt"
 	"os"
-	"sigs.k8s.io/controller-runtime/pkg/healthz"
 	"time"
+
+	"sigs.k8s.io/controller-runtime/pkg/healthz"
 
 	ctrl "sigs.k8s.io/controller-runtime"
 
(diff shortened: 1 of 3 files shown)


moelsayed (commented):
Technically, the readiness probe is used to add the pod/remove endpoints in a service. Since our core operation is not dependent on exposing an HTTP endpoint, this doesn't have a major impact on our stability.

I think we should extend this PR to also use the [controller-runtime liveness probe](https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.12.0/pkg/healthz), which would directly impact the controller stability, as it's the probe responsible to triggering pod restarts.

dbadura:
The readiness is a cosmetic change.

`Controller-runtime` liveness probe is a basic http check and there is no difference between readiness and liveness.
The default liveness probe can't detect whether the reconcile loop hang. 
Additionally it run seperate http server within controller. https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/manager/internal.go#L340=

In our scenario the timeout is too small for cloning git repository and that's why the controller get restarted.

moelsayed (approved):

PR #14319, opened 2022-05-17
use alpine in fi eventing dockerfile

FriedrichWilken:
this PR switches from debian to alpine in `kyma/tests/fast-integration/eventing-test/image/Dockerfile` and adapts the commands for installing additional packages accordingly.


tests/fast-integration/image/Dockerfile
```diff
@@ -7,18 +7,12 @@ ENV FIT_MAKE_TARGET=ci-skr
...
-    curl \
+RUN apk --no-cache add \
+    build-base \
     git \
-    && apt-get clean \
-    && rm -rf /var/lib/apt/lists/*
-
-RUN curl -fLSs https://deb.nodesource.com/setup_14.x | bash -
-RUN apt-get -y --no-install-recommends install nodejs
+    nodejs=16.14.2-r0
```
nachtmaar, line 15: I think it is safer to use the same nodejs version on both Prow and Spinnaker 🤔

FriedrichWilken: might makes sense. On the other hand wouldn't it be better to bump both to the latest LTS version (16.x) ?

nachtmaar: Also fine for me :) If it works, then I don't see a reason for not doing it :)

nachtmaar:
We need to modify the script because it uses [bash in the shebang](e2e-fi-skr-test-dvsrd-qtj8x). Alpine comes with ash only.

nachtmaar:
Also we need to add the npm command: `apk add npm`

nachtmaar:
There was another problem with the image then running the skr-test:
```
ENOENT: no such file or directory, open '/KUBECONFIG/kubeconfig.yaml'
```

I did remove the env vars. They will be set by the k8s job instead.

nachtmaar (dismissed):

jeremyharisch (approved):

nachtmaar (approved):

jeremyharisch (approved):

PR #14335, opened 2022-05-18
move tests/fast-integration/eventing/image to tests/fast-integration/image

FriedrichWilken:
The files in `tests/fast-integration/eventing/image` are not specific to `eventing` but are used for all over `fast-integration` tests so this PR will move the them `tests/fast-integration/image`


VladislavPaskar:
Totally makes sense to not encapsulate this inside the eventing folder. It can also be a sign for other teams, that it is ready to use by anyone, since it's not only eventing-related.

VladislavPaskar (approved):
LGTM, Totally makes sense to not encapsulate this inside the eventing folder. It can also be a sign for other teams, that it is ready to use by anyone, since it's not only eventing-related.

jeremyharisch (approved):

PR #14382, opened 2022-05-24
refactor extracting of fluent bit dry run output

dennis-ge:
**Description**

- With the update of fluent-bit in the telemetry operator [Dockerfile](https://github.com/kyma-project/kyma/blob/main/components/telemetry-operator/Dockerfile#L22), the output of the `fluent-bit --dry-run` changed. - This functionality is used in the validation webhook to validate the fluent bit configuration
- This PR refactors the regexes for extracting the errors from the `fluent bit --dry-run` output to work with the current fluent bit version (1.9.3)

**Related issue(s)**

### DIFF
--- a/components/telemetry-operator/internal/fluentbit/config_validator.go
+++ b/components/telemetry-operator/internal/fluentbit/config_validator.go
@@ -81,23 +81,29 @@ func listPlugins(pluginPath string) ([]string, error) {
 
 // extractError extracts the error message from the output of fluent-bit
 // Thereby, it supports the following error patterns:
-// 1. Error <msg>\nError: Configuration file contains errors. Aborting
-// 2. Error: <msg>. Aborting
-// 3. [<time>] [  Error] File <filename>\n[<time>] [  Error] Error in line 4: <msg> Error: Configuration file contains errors. Aborting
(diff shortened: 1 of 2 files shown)


tests/fast-integration/telemetry-test/telemetry-test.js
```diff
@@ -80,7 +80,9 @@ describe('Telemetry Operator tests', function() {
       assert.fail('Should not be able to apply invalid LogPipeline');
     } catch (e) {
       assert.equal(e.statusCode, 403);
-      expect(e.body.message).to.have.string('denied the request', 'Invalid indentation level');
+      expect(e.body.message).to.have.string('denied the request');
+      const errMsg = 'section \'abc\' tried to instance a plugin name that don\'t exists';
```
rakesh-garimella, line 84: Will this expression fail if the error message fails next time ? Should we may be check the comparison to be exact ? Like `expect.().to.equal` ?

dennis-ge: With the current setup we only check if the substring is present in the complete error message (which is a little bit longer). We could also check for equality for the complete error message. But, IMO the substring mechanism should be fine, but also open to check for equality.

rakesh-garimella (approved):

PR #14410, opened 2022-05-26
Fast-Integration-test refactoring after removal of SC: Helm-Broker

tillknuesting:
helm-Broker

### DIFF
--- a/tests/fast-integration/skr-svcat-migration-test/deploy-sample-resources.js
+++ b/tests/fast-integration/skr-svcat-migration-test/deploy-sample-resources.js
@@ -59,9 +59,6 @@ async function installRedisExample(options) {
 
   options = options || {};
 
-  debug(`Waiting for pod with label app=helm-broker...`);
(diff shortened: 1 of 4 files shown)


tests/fast-integration/skr-svcat-migration-test/deploy-sample-resources.js
```diff
@@ -252,7 +249,6 @@ async function deploy() {
   await waitForPodWithLabel('app', 'service-broker-proxy-k8s', 'kyma-system');
 
   try {
-    await waitForClusterServiceBroker('helm-broker', 5 * 60 * 1000);
```
wozniakjan, line 255: nit: technically this whole test can be removed after svcat is removed from kyma but I think we can start iteratively and disable it by parts

wozniakjan (approved):
/lgtm
/approve

pbochynski (approved):

Issue #9585, opened 2020-09-30
Function controller should retry failed function builds 


tgorgol:
**Description**

At the moment if a function build fails for any reason, the function controller will not try rebuilding the function unless the source changes. This behavior saves us some resources needed for rebuilding the function in case of failure caused by an invalid function source. However, rarely build may fail for other reasons such as temporary infrastructure problems (unavailable Docker registry). When that happens it is expected that the function controller will try to rebuild the function.

**AC**
 - retry should be implemented as  retry-able back-off (each time retry time is extended exponentially ) - so that we don waste cpu time on non-recoverable errors

 
**Reasons**

Having automatic retries after failed builds would improve the resiliency and user experience because the user wouldn't have to make changes to the function source if it was valid.

**Workaround**
As a workaround user needs to delete the failed  build jobs manually ( for the functions that are stuck in the error state. )
```
kubectl get jobs -A -l serverless.kyma-project.io/managed-by=function-controller -ojson | jq -r '.items[] | select(.status.failed==1) | "kubectl delete job -n " + .metadata.namespace + " " + .metadata.name ' | xargs -I {} sh -c "{}"

```

kwiatekus:
Not a priority

kwiatekus:
This problem priority has increased as we identify the scenario which leads to a lot of failed ( and not re-scheduled ) function builds: 

If a kyma release introduces changes in function CRD, then functions are scheduled for rebuild after upgrade.
With 2.1 there were no major changes in Function CRD; hence no issues experienced as functions were not rebuilt by the controller.
With 2.2 we introduced a new [feature](https://github.com/kyma-project/kyma/issues/13555) that adds to the Function CRD.  After 2.2 upgrade ( or even during the update ) the re-build jobs want to schedule pods.

During kyma upgrade there is a higher demand on CPU:
![Screenshot 2022-05-25 at 11 40 26](https://media.github.tools.sap/user/14697/files/2ddf85d6-48f2-47c7-bafb-86b4fd79fb39)
If there are a lot of functions to be rebuilt, not all of them would get a re-built pod scheduled due to insufficient CPU.

Normally the function state should eventually resolve to normal state, but because of this bug the failed jobs are not re-scheduled.

PR #14412, opened 2022-05-26
Fix hanging jobs on upgrade

m00g3n:
**Description**
Fix hanging jobs on upgrade

- failed jobs are now retried 
- logging fixed
- function status change events are no longer triggering reconciliation
- job events no longer trigger reconciliation

**Related issue(s)**

See also: https://github.com/kyma-project/kyma/issues/9585

### DIFF
--- a/components/function-controller/internal/controllers/serverless/fsm.go
+++ b/components/function-controller/internal/controllers/serverless/fsm.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"log"
 	"strings"
 	"time"
 
@@ -66,6 +67,9 @@ loop:
 		}
 	}
 
+	log.Println("result: ", m.result.RequeueAfter, m.result.Requeue)
+	log.Println("err:", m.err)
+
 	return m.result, m.err
 }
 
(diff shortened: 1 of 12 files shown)


components/function-controller/internal/controllers/serverless/fsm.go
```diff
@@ -61,11 +64,15 @@ loop:
 			m.err = ctx.Err()
 			break loop
 		default:
+			log.Println("Next fn state:", runtime.FuncForPC(reflect.ValueOf(m.fn).Pointer()).Name())
```
dbadura, line 67: use m.log as we have configured the logger and use debug level.

components/function-controller/internal/controllers/serverless/fsm.go
```diff
@@ -61,11 +64,15 @@ loop:
...
 			m.fn = m.fn(ctx, m, &state)
 
 		}
 	}
 
+	log.Println("result", m.result.RequeueAfter, m.result.Requeue)
```
dbadura, line 73: We can write this log in one line

dbadura (approved):

PR #14423, opened 2022-05-26
Improve Makefile in function controller: add targets for k3d

dbadura:
**Description**

- add ability to build and update serverless manager and webhook in k3d in one step

**Related issue(s)**


pPrecel (commented):
Looks great 🚀 ! I've tested it and I really like this idea! One suggestion only. In the Makefile, we have sections created to categorize specific targets. What do you think about moving these two targets under the `######## function manager` and the `######## function webhook` sections?

PR #14442, opened 2022-05-27
Move webhook to webhooks from manager

dbadura:
**Description**

- move webhook to webhooks from manager

**Related issue(s)**


components/function-controller/cmd/webhook/main.go
```diff
@@ -95,6 +95,12 @@ func main() {
 		},
 	)
 
+	whs.Register("/defaulting/config-registry-secrets",
```
dbadura, line 98: registry-config-secrets?

components/function-controller/internal/webhook/resources/webhook_config.go
```diff
@@ -27,8 +27,11 @@ const (
 	serverlessAPIGroup   = "serverless.kyma-project.io"
 	serverlessAPIVersion = "v1alpha1"
 
-	DefaultingWebhookName = "defaulting.webhook.serverless.kyma-project.io"
-	ValidationWebhookName = "validation.webhook.serverless.kyma-project.io"
+	DefaultingWebhookName     = "defaulting.webhook.serverless.kyma-project.io"
+	SecretMutationWebhookName = "mutating.secret.serverless.k8s.io"
```
moelsayed, line 31: Should we use the same naming convention as the other webhooks?

components/function-controller/internal/webhook/resources/webhook_config.go
```diff
@@ -71,55 +74,102 @@ func ensureValidatingWebhookConfigFor(ctx context.Context, client ctlrclient.Cli
...
+			},
+		},
+		FailurePolicy:           &failurePolicy,
+		MatchPolicy:             &matchPolicy,
+		TimeoutSeconds:          pointer.Int32(30),
+		SideEffects:             &sideEffects,
+		AdmissionReviewVersions: []string{"v1beta1", "v1"},
+		ObjectSelector: &metav1.LabelSelector{
+			MatchLabels: map[string]string{
+				RemoteRegistryLabelKey: "config",
```
moelsayed, line 159: Would it be possible to define this in the function body? Technically is it's fine here, but it's not easy to find/read when placed here. And it's important.

moelsayed, line 159: Would it be possible to define this in the function body? Technically is it's fine here, but it's not easy to find/read when placed here. And it's important.

components/function-controller/internal/webhook/resources/webhook_config.go
```diff
@@ -71,55 +74,102 @@ func ensureValidatingWebhookConfigFor(ctx context.Context, client ctlrclient.Cli
...
+	sideEffects := admissionregistrationv1.SideEffectClassNone
+
+	return admissionregistrationv1.MutatingWebhook{
+		Name: SecretMutationWebhookName,
+		ClientConfig: admissionregistrationv1.WebhookClientConfig{
+			CABundle: config.CABundel,
+			Service: &admissionregistrationv1.ServiceReference{
+				Namespace: config.ServiceNamespace,
+				Name:      config.ServiceName,
+				Path:      pointer.String("/defaulting/registry-config-secrets"),
```
moelsayed, line 148: I think it makes sense to put this path in a const and use it everywhere.

moelsayed, line 148: I think it makes sense to put this path in a const and use it everywhere.

resources/serverless/templates/admission_webhook.yaml
```diff
@@ -1,51 +0,0 @@
...
-    app: {{ template "fullname" . }}
-    chart: "{{ .Chart.Name }}-{{ .Chart.Version }}"
-    release: "{{ .Release.Name }}"
-    heritage: "{{ .Release.Service }}"
-type: Opaque
-data:
-  tls.crt: {{ b64enc $cert.Cert }}
-  tls.key: {{ b64enc $cert.Key }}
----
-apiVersion: admissionregistration.k8s.io/v1
```
moelsayed, line 21: I think this should be moved the the webhook chart, and add the same labels used in the other webhook configurations.

moelsayed, line 21: I think this should be moved the the webhook chart.

moelsayed (changes requested):
Looks very good, thank you. Did you get a chance to test upgrade scenarios from older versions?

resources/serverless/templates/admission_webhook.yaml
dbadura: it's already in webhook chart:
https://github.com/kyma-project/kyma/blob/main/resources/serverless/charts/webhook/templates/webhooks.yaml

components/function-controller/internal/webhook/resources/webhook_config.go
dbadura: I moved it at the top of function

moelsayed (approved):

Issue #14256, opened 2022-05-11
[Istio] implement component test for Istio evaluation installation


strekm:
**Description**

Implement component test testing Istio installation using evaluation profile on k3d cluster. Test should assume that connection to a cluster is provided and Istio is installed. Test should check proper configuration for evaluation profile: if all necessary pods are running, HPA configuration. 

Expected result:

- [x] Istio pods should be running ( istiod, ingress-gateway)
- [x] check that there is not hpa in place (only one pod per mentioned one)
- [x] prepare gherkin file and integrate with test suite using godog.

**Attachments**

part of: https://github.com/kyma-project/kyma/issues/13517

PR #14464, opened 2022-06-01
add component test for istio eval profile with godog support

piotrkpc:
**Description**
Adds Istio component test for evaluation profile installation

Run the tests:
```bash
export KYMA_PROFILE=evaluation
make test-k3d # deply k3d, install kyma, run tests
or
make test # run tests (go test ./...)
```

**Related issue(s)**

https://github.com/kyma-project/kyma/issues/14256


barchw (approved):

piotrkpc:
Notes with feedback from Oleksandr:
- the feature file does not captures full the intent and it could be improved. Example
```
Scenario: Istio component is installed with a configured profile
Istio component is always installed in a Kyma cluster with a
profile configured during an installation
  Given a running Kyma cluster
  And Kyma profile setting is "evaluation"
  Then Istio component is installed
  And Istio's profile is "evaluation"


Scenario: Istio component has all required pods running
  Given a running Kyma cluster with "evaluation" profile
  When Istio component is installed
  Then there is 1 pod for Pilot
  And there is 1 pod for "Ingress gateway"
  And "HPA" is not deployed
  And Istio pods are available for at least 3 seconds
```
- godog integration does not integrate with `testing.T` and relies on errors, this makes writing tests less readable but it can be improved in the future.

mjakobczyk:
> Scenario: Istio component has **all required pods** running

What are **all required pods**? Should we define this term earlier or could it be more specific? From my point of view is too broad and should be narrowed down to for example: **internal components** or **control plane and date plane**.

>   Then there is 1 pod for Pilot
>   And there is 1 pod for "Ingress gateway"
>   And "HPA" is not deployed

Shouldn't we quote **Pilot** term in this scenario also? Or define some common vocabulary ahead to use them freely?


> And Istio pods are available for at least 3 seconds

Don't we specify startup availability such as after what time pods should be reachable?

piotrkpc:
> Notes with feedback from Oleksandr:
> 
> * the feature file does not captures full the intent and it could be improved. Example
> 
> ```
> Scenario: Istio component is installed with a configured profile
> Istio component is always installed in a Kyma cluster with a
> profile configured during an installation
>   Given a running Kyma cluster
>   And Kyma profile setting is "evaluation"
>   Then Istio component is installed
>   And Istio's profile is "evaluation"
> 
> 
> Scenario: Istio component has all required pods running
>   Given a running Kyma cluster with "evaluation" profile
>   When Istio component is installed
>   Then there is 1 pod for Pilot
>   And there is 1 pod for "Ingress gateway"
>   And "HPA" is not deployed
>   And Istio pods are available for at least 3 seconds
> ```
> 
> * godog integration does not integrate with `testing.T` and relies on errors, this makes writing tests less readable but it can be improved in the future.

Since the first scenario is not really applicable since we don't have the setting of profile other than during deployment command.

Best practices I've tried to follow:
- keep the same order for registering steps with `ctx.Step` as we have in feature file.
Comments are welcome.

tests/components/istio/istio_installed_test.go
```diff
@@ -0,0 +1,201 @@
...
+	}
+	k8sClient, err := kubernetes.NewForConfig(config)
+	if err != nil {
+		log.Fatalf(err.Error())
+	}
+	return k8sClient
+}
+
+func TestIstioInstalled(t *testing.T) {
+
```
cnvergence, line 66:
```suggestion

```

Issue #14309, opened 2022-05-16
[Telemetry Operator] Community Plugins


a-thaler:
**Description**
The operator supports flexible configuration and usage of any filter and output plugin. As the Kyma main contributors cannot test all the different available plugins on every update (only the most relevant), it needs to rely on the tests done by the fluent bit contributors.
With https://github.com/kyma-project/kyma/issues/13146 we introduced an option to allow only specific plugins to be used at runtime to reflect that support conflict. However, that way is limiting the usage too much and will make the feature a lot unusable. Instead of forbidding the usage for a lot of plugins and making a native integration into other systems harder, the conflict must be expressed directly to the user and the user should be able to accept the risk of having a not fully integration-tested feature.
Still, there are some filters which can be really harmful and have no real usage without entering that harmful path. It should be possible to forbid these filters fully.

**Criterias**
- Have an option to enable "community" plugins explicitly via the CR (disabled by default)
- If disabled, you can only use the "allowed" plugins and filters in the CR (maybe rename "allowed"
- Revise the list of "allowed" plugins (currently filters are missing)
- Make the info if "community" mode is on available in the CR list view and make it observable (custom metric?)
- Forbid the usage of specific filters (kubernetes, rewrite_tag, multiline)

TorstenD-SAP:
@a-thaler, @rakesh-garimella: Does it makes sense/is it possible to add some kind of warning somewhere that the usage of unsupported plugins can harm the system?

rakesh-garimella:
The unsupported plugins is more like (untested plugins), we can may be give a warning message when they enable the flag in the CR.

rakesh-garimella:
- `enableUnsupported` flag has been introduced. When set to `true` users can use unsupported/untested plugins
- When the above flag is enabled users are also presented with warning
- Printer column has been added to show the status of the flag
- A new metric `EnableUnsupportedPlugin` has been introduced to show the status of the flag

PR #14485, opened 2022-06-02
Allow unsupported plugins in fluentbit

rakesh-garimella:
**Description**

- Allow unsupported plugins in fluentbit via a flag in crd. If set user can configure unsupported plugins in the pipelines
- Have printer columns for unsupported plugins
- Expose if pipelines support unsupported plugins via metric
**Related issue(s)**

https://github.com/kyma-project/kyma/issues/14309


components/telemetry-operator/main.go
```diff
@@ -65,8 +65,10 @@ var (
 	logFormat                  string
 	logLevel                   string
 	certDir                    string
-	allowedFilterPlugins       string
-	allowedOutputPlugins       string
+	supportedFilterPlugins     string
+	supportedOutputPlugins     string
+	deniedFilterPlugins        string
```
chrkl, line 70: I don't think we need an explicit list of denied filter and output plugins. All plugins that are not explicitly allowed, are denied.

components/telemetry-operator/internal/sync/sync.go
```diff
@@ -71,6 +72,10 @@ func (s *LogPipelineSyncer) SyncAll(ctx context.Context, logPipeline *telemetryv
 		log.Error(err, "Failed to sync secret references")
 		return false, err
 	}
+	err = s.syncEnableUnsupportedPluginMetric(ctx)
+	if err != nil {
+		return false, err
```
chrkl, line 77: Let's log the error. It should indicate a problem with the communication to the API server

chrkl (changes requested):

components/telemetry-operator/api/v1alpha1/logpipeline_types.go
```diff
@@ -137,6 +138,7 @@ func filterOutCondition(conds []LogPipelineCondition, condType LogPipelineCondit
 //+kubebuilder:resource:scope=Cluster
 //+kubebuilder:subresource:status
 //+kubebuilder:printcolumn:name="Status",type=string,JSONPath=`.status.conditions[-1].type`
+//+kubebuilder:printcolumn:name="Unsupported-Plugins",type=boolean,JSONPath=`.spec.enableUnsupportedPlugins`
```
chrkl, line 141: Let's name the column "Unsupported-Plugin" since it contains only one value per row

resources/telemetry/charts/operator/values.yaml
```diff
@@ -82,8 +82,29 @@ serviceMonitor:
...
+    - rewrite_tag
+    - multiline
+  output: []
+
+supportedPlugins:
+  filter:
+    - aws
+    - expect
+    - grep
+    - lua
```
chrkl, line 97: Not sure if lua should be supported

resources/telemetry/charts/operator/values.yaml
```diff
@@ -82,8 +82,29 @@ serviceMonitor:
...
+    - record_modifier
+    - nightfall
+    - stdout
```
chrkl, line 100: stdout should be rather unsupported to avoid log explosion problems

resources/telemetry/charts/operator/values.yaml
```diff
@@ -82,8 +82,29 @@ serviceMonitor:
...
+    - checklist
+    - geoip2
+    - parser
+    - modify
+    - nest
+    - throttle
+    - tensorflow
   output:
```
chrkl, line 108: forward should also be suported

chrkl (changes requested):

chrkl (approved):

chrkl (approved):

Issue #14027, opened 2022-04-21
Replace logging lib


kwiatekus:
**Description**
In our current logger setup ( logr with zap) we are vulnerable to mistakes that cause controller to panic

Verify if we should replace it or introduce a wrapper on the existing logger 

**Reason**
Current logger library can panic very easily because of wrong number of arguments.
It's very hard to catch such bug and to detect wrong initialisation during code review

**Attachments**
Learn from investigation:
 - use kyma [commons](https://github.com/kyma-project/kyma/tree/main/common/logging) for [consistent logging](https://github.com/kyma-project/community/tree/0efcbad113df60b3a9ddb0fafd0ef92f1bd9abf4/concepts/observability-consistent-logging)

dbadura:
Example panic log:
```
1.6504539847547677e+09  DPANIC  controllers.function    odd number of arguments passed as key-value pairs for logging   {"kind": "Function", "name": "function-sleepy-karolina5", "namespace": "default", "version": 1, "ignored key": ""}
github.com/kyma-project/kyma/components/function-controller/internal/controllers/serverless.(*reconciler).reconcile
        /workspace/go/src/github.com/kyma-project/kyma/components/function-controller/internal/controllers/serverless/fsm.go:51
github.com/kyma-project/kyma/components/function-controller/internal/controllers/serverless.(*FunctionReconciler).Reconcile
        /workspace/go/src/github.com/kyma-project/kyma/components/function-controller/internal/controllers/serverless/function_reconcile.go:138
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Reconcile
        /workspace/go/src/github.com/kyma-project/kyma/components/function-controller/vendor/sigs.k8s.io/controller-runtime/pkg/internal/controller/controller.go:114
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).reconcileHandler
        /workspace/go/src/github.com/kyma-project/kyma/components/function-controller/vendor/sigs.k8s.io/controller-runtime/pkg/internal/controller/controller.go:311
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).processNextWorkItem
        /workspace/go/src/github.com/kyma-project/kyma/components/function-controller/vendor/sigs.k8s.io/controller-runtime/pkg/internal/controller/controller.go:266
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Start.func2.2
[...]

PR #14486, opened 2022-06-03
Configure logging to use Text output as default

Cortey:
**Description**
So far logging was outputed as a json and used timestamps in EPOCH which was hard to read, now they will be logged as Text with option to still get json output if needed.
Example text log:
`2022-06-03T12:33:43+02:00       INFO    controller-runtime.metrics      logr/logr.go:261        Metrics server is starting to listen    {"context": {"addr": ":8080"}}
`

**Related issue(s)**
https://github.com/kyma-project/kyma/issues/14027


components/function-controller/cmd/manager/main.go
```diff
@@ -70,25 +67,41 @@ func main() {
...
+		ctrl.SetLogger(ctrlzap.New())
+		setupLog.Error(err, "unable to set setup logger level")
+		os.Exit(3)
+	}
+
+	if err := logger.InitKlog(l, logLevel); err != nil {
+		setupLog.Error(err, "unable to set setup logger level")
+		os.Exit(5)
+	}
+	ctrl.SetLogger(zapr.NewLogger(l.WithContext().Desugar()))
```
moelsayed, line 94: I think we can should wrap the log setup in a function. I would probably be useful when we apply the same to the webhook as well.

components/function-controller/cmd/manager/main.go
```diff
@@ -2,34 +2,30 @@ package main
 
 import (
 	"fmt"
+	serverlessLogging "github.com/kyma-project/kyma/components/function-controller/internal/logging"
```
moelsayed, line 5: This shouldn't be up here.

components/function-controller/cmd/manager/main.go
```diff
@@ -69,26 +66,24 @@ func main() {
 		setupLog.Error(err, "unable to load config")
 		os.Exit(1)
 	}
-
-	logLevel, err := toZapLogLevel(config.LogLevel)
+	l, err := serverlessLogging.ConfigureLogger(config.LogLevel, config.LogFormat)
 	if err != nil {
 		ctrl.SetLogger(ctrlzap.New())
```
moelsayed, line 71: I don't thing we need this. We exit(1) anyway..

moelsayed (changes requested):
Looks good, couple of comments

moelsayed (approved):

Issue #14440, opened 2022-05-27
Function jobs shouldn't request resources by default


dbadura:
**Description**
In current implementation by default we enforce build profile.

**Expected result**
By default the build profiles are not applied.

**Actual result**
Having those values set as default, that:
- function deployment request 500m cpu
- function build job request 1000m cpu
let's consider following situation:
User creates 3 function on node with 2000cpu available:
- first job requires 1000m, the available left is 1000m
- job finished and then we deploy the function which requires 500m, the available cpu is 1500m
- user deploy second function and at the end available cpu is 1000m
- user deploy third function. The job can fit, because we have 1000m cpu available and at the end we deploy function and cluster have available 500m cpu.
- the user want to upgrade one of those three functions and the upgrade stuck, because the job doesn't have enough space.

By default we shouldn't apply any build profiles or at least not enforce minimum request for build jobs.
I would consider to also remove the default profile for functions if the user didn't specify.

**Steps to reproduce**

**Troubleshooting**
Settings for minimum requests: https://github.com/kyma-project/kyma/blob/main/components/function-controller/pkg/apis/serverless/v1alpha1/function_validation.go#L28=

kwiatekus:
function build job cannot start in an over-commited CPU environment.
The following shows the scenario: CPU utilisation is really low ( ~14%) but CPU commitment is ~100%

![Screenshot 2022-06-01 at 09 57 39](https://user-images.githubusercontent.com/6783567/171356532-3f2639ba-1403-4051-bdab-533404b9b2cb.png)

We should lower the overcommitment of CPU for kyma components

kwiatekus:
We should 
 - delete the min and max values for the requested CPU time for function builds
 - analyse the actual CPU utilisation (% of the requested CPU time ) for kyma components that sit on most of the requested CPU
 - propose a PR with adjusted requested CPU time for kyma components

kwiatekus:
Thanks
We have reduced the cpu overcommitment across kyma components - saving cpu cycles for build jobs.
Additionally, we removed the requested cpu and memory for the build jobs by default - allowing for more flexible job scheduling in scenarios where cpu is limited on the worker nodes

PR #14500, opened 2022-06-07
Adjust components resources requests

moelsayed:
**Description**
We have observed several cases where we think the Kyma upgrade process is stuck due to insufficient CPU resources on the node. To improve this, this PR tries to establish a base line for components CPU resource requests. It applies a reduction of approximately 10% of the requested CPU resources for Kyma.

- Before:
image

- After:
image

- Normalize CPU requests across all components.

**Related issue(s)**

[#14440](https://github.com/kyma-project/kyma/issues/14440)

### DIFF
--- a/resources/application-connector/charts/application-operator/values.yaml
+++ b/resources/application-connector/charts/application-operator/values.yaml
@@ -13,7 +13,7 @@ controller:
       cpu: 100m
       memory: 256Mi
     requests:
-      cpu: 100m
+      cpu: 10m
       memory: 128Mi
   livenessProbe:
     initialDelaySeconds: 50
--- a/resources/application-connector/charts/central-application-connectivity-validator/values.yaml
+++ b/resources/application-connector/charts/central-application-connectivity-validator/values.yaml
@@ -11,7 +11,7 @@ deployment:
       cpu: 100m
       memory: 128Mi
     requests:
-      cpu: 50m
+      cpu: 10m
       memory: 64Mi
 
 service:
--- a/resources/application-connector/charts/central-application-gateway/profile-production.yaml
+++ b/resources/application-connector/charts/central-application-gateway/profile-production.yaml
@@ -4,7 +4,7 @@ deployment:
       cpu: 100m
       memory: 128Mi
     requests:
-      cpu: 50m
+      cpu: 10m
       memory: 64Mi
 autoscaling:
   enable: true
--- a/resources/application-connector/charts/central-application-gateway/values.yaml
+++ b/resources/application-connector/charts/central-application-gateway/values.yaml
@@ -15,7 +15,7 @@ deployment:
       cpu: 100m
       memory: 128Mi
     requests:
-      cpu: 50m
+      cpu: 10m
       memory: 64Mi
 
 service:
--- a/resources/application-connector/profile-production.yaml
+++ b/resources/application-connector/profile-production.yaml
@@ -6,7 +6,7 @@ application-broker:
         cpu: 100m
       requests:
         memory: 48Mi
-        cpu: 60m
+        cpu: 10m
 application-operator:
   controller:
     resources:
@@ -15,7 +15,7 @@ application-operator:
         cpu: 100m
         memory: 256Mi
       requests:
-        cpu: 100m
+        cpu: 10m
         memory: 128Mi
 central-application-gateway:
   autoscaling:
@@ -29,7 +29,7 @@ central-application-gateway:
         cpu: 100m
         memory: 128Mi
       requests:
-        cpu: 50m
+        cpu: 10m
         memory: 64Mi
 central-application-connectivity-validator:
   deployment:
@@ -38,5 +38,5 @@ central-application-connectivity-validator:
         cpu: 100m
         memory: 128Mi
       requests:
-        cpu: 50m
+        cpu: 10m
         memory: 64Mi
--- a/resources/eventing/profile-production.yaml
+++ b/resources/eventing/profile-production.yaml
@@ -14,7 +14,7 @@ controller:
       cpu: 500m
       memory: 1Gi
     requests:
-      cpu: 50m
+      cpu: 10m
       memory: 512Mi
   publisherProxy:
     resources:
@@ -22,7 +22,7 @@ controller:
         cpu: 500m
         memory: 512Mi
       requests:
-        cpu: 50m
+        cpu: 10m
         memory: 256Mi
 
 nats:
@@ -43,7 +43,7 @@ nats:
         cpu: 500m
         memory: 1Gi
       requests:
-        cpu: 50m
+        cpu: 10m
         memory: 512Mi
     logging:
       debug: false
--- a/resources/eventing/profile-production.yaml
+++ b/resources/eventing/profile-production.yaml
@@ -14,7 +14,7 @@ controller:
       cpu: 500m
       memory: 1Gi
     requests:
-      cpu: 400m
+      cpu: 50m
       memory: 512Mi
   publisherProxy:
     resources:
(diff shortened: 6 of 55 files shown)


kwiatekus (approved):

kwiatekus:
I have installed kyma runtime from main branch using evaluation profile.
I deployed an application to put some load. My setup  uses api rules, eventing and serverless
 (bunch of functions connected together via eventing triggered via api rule)
I upgraded the runtime to PR-14500
the resource consumption ( checked at grafana) looks solid

Screenshot 2022-06-14 at 13 38 07

resources/eventing/profile-production.yaml
```diff
@@ -14,15 +14,15 @@ controller:
       cpu: 500m
       memory: 1Gi
     requests:
-      cpu: 400m
+      cpu: 10m
```
mfaizanse, line 17: Isn't `10m` too less for production profile?

I deployed this PR with production profile and ran some tests:

With 50 events/sec:
image

With 10 events/sec:
image

You can see that `10m` was not enough even for 10events/sec. 
IMO, the requested CPU should be at-least enough to handle minimum production workload.

I would say, that we use atleast `150m` for production profile, so it can support 50events/sec using the requested CPU.

resources/eventing/profile-production.yaml
```diff
@@ -14,15 +14,15 @@ controller:
...
       memory: 512Mi
   publisherProxy:
     resources:
       limits:
         cpu: 500m
         memory: 512Mi
       requests:
-        cpu: 400m
+        cpu: 10m
```
mfaizanse, line 25: Same comment as [this](https://github.com/kyma-project/kyma/pull/14500/files?file-filters%5B%5D=.yaml&owned-by%5B%5D=mfaizanse&show-viewed-files=true#r896752619)

resources/eventing/profile-production.yaml
kwiatekus: @mfaizanse 🤔 we are talking about requested cpu time - which should rather correspond to idle/startup state.
What we see is that an iddle kyma runtime suffers from huge overbooking of the CPU time across all components.

When eventing is under load it uses more  - it is seen in you screenshots, ie:
nats requests `0.01` from k8s scheduler at the beginning.
When the load starts - the actual  CPU usage rises up to `0.15`, that corresponds to 1450% of the initially requested amount.. but still.. its only 30% of the limit!
The pod is free to rise cpu consumption until its limits (the limits where not changed by the PR).
IMO there is no justification to overbook cpu time "just in case".

mfaizanse: Question: When a pod is scheduled by K8s on a node, does it also check if the pod will be able to reach it limits on that node or not? For example, if a node has only `50m` CPU free (or unreserved) and the `eventing-controller` pod is scheduled on that node, will the eventing-controller pod be able to scale further up 50m? or what will happen in this case?

kwiatekus: in my cluster I also apply load to eventing component
The actual load ( cpu usage ) is not affected by the reduced cpu request time.
It can freely go up until limit

Screenshot 2022-06-14 at 15 24 42

lindnerby (approved):

resources/application-connector/charts/central-application-connectivity-validator/values.yaml
```diff
@@ -11,7 +11,7 @@ deployment:
       cpu: 100m
       memory: 128Mi
     requests:
-      cpu: 50m
+      cpu: 10m
```
janmedrek, line 14: Please bring that up to at least `25m`.

resources/application-connector/charts/application-operator/values.yaml
```diff
@@ -13,7 +13,7 @@ controller:
       cpu: 100m
       memory: 256Mi
     requests:
-      cpu: 100m
+      cpu: 10m
```
janmedrek, line 16: This one should not be changed. This component may get resource hungry when there are multiple Service Instances, we already sometimes have problems with the limit here.

resources/application-connector/charts/central-application-gateway/profile-production.yaml
```diff
@@ -4,7 +4,7 @@ deployment:
       cpu: 100m
       memory: 128Mi
     requests:
-      cpu: 50m
+      cpu: 10m
```
janmedrek, line 7: This one should not be changed.

resources/application-connector/charts/central-application-gateway/values.yaml
```diff
@@ -15,7 +15,7 @@ deployment:
       cpu: 100m
       memory: 128Mi
     requests:
-      cpu: 50m
+      cpu: 10m
```
janmedrek, line 18: Please bring that up to at least `25m`.

resources/application-connector/profile-production.yaml
```diff
@@ -6,7 +6,7 @@ application-broker:
         cpu: 100m
       requests:
         memory: 48Mi
-        cpu: 60m
+        cpu: 10m
```
janmedrek, line 9: This one should not be changed.

resources/application-connector/profile-production.yaml
```diff
@@ -15,7 +15,7 @@ application-operator:
         cpu: 100m
         memory: 256Mi
       requests:
-        cpu: 100m
+        cpu: 10m
```
janmedrek, line 18: This one should not be changed.

resources/application-connector/profile-production.yaml
```diff
@@ -29,7 +29,7 @@ central-application-gateway:
         cpu: 100m
         memory: 128Mi
       requests:
-        cpu: 50m
+        cpu: 10m
```
janmedrek, line 32: This one should not be changed.

resources/application-connector/profile-production.yaml
```diff
@@ -38,5 +38,5 @@ central-application-connectivity-validator:
...
```
janmedrek, line 41: This one should not be changed.

janmedrek (changes requested):
I would rather not modify marked entries as the components can eat up way more CPU in the active state than in standby mode. We have tested that and it tends to spike when multiple applications are connected. The lower limits on those resources can create situations where the user workload is unable to reach connected systems due to Gateway/Validator pods downtime.

moelsayed:
Hi @janmedrek @mfaizanse..

Requests are the minimum guaranteed CPU shares allocated to the application. It's more efficient to try to minimize that, since it allows for better resources usage and will actually improve the reliability of the entire cluster, especially in upgrade/reboot cases. Limits are what actually controls how much CPU time an application can use, and in this PR, we don't touch those at all. 

Kubernetes uses resource requests for mainly for scheduling. The scheduler will ensure that pods are started on nodes that have enough capacity to satisfy the resources request. Resources requests are not overcommited. Kubernetes will only schedule pods on a node as long as the node has enough capacity to satisfy _all_ the requests at the same time. That's why it's recommended that requests are set based on _mean_ consumption and not based on usage spikes. Planning for spikes is a wast of resources in this case, those resources will remain unused almost all of the time. 

For example: if an application is configured with `100m` and only uses `5m` for 80% of the time, and spikes to `90m` for 20% if the time, that means that for 80% of the time, `95m` CPU are wasted. Lowering the the requests will reduce this waste and still allow applications to spike to their limits with no problem.

Some services are sensitive, databases for example, and probably need to be in the guaranteed QoS class (have the same limits, requests values). And I tried to not change those at all. 

As I mentioned, the idea here is to establish a baseline. Allocated resources should be revised often anyway because applications change and their requirements change as well. It's usually more effective to start low and increase the limits than start with a high number and try to reduce it later.

janmedrek (approved):

janmedrek:
@moelsayed you are right, I was convinced it is the limits that are changed, my bad - sorry for that. Dropped an approval.

hanngos (approved):

Issue #13977, opened 2022-04-14
Revisit Application Connector tests


janmedrek:
**Description**

We need to provide a clear suite of tests that will go through all the functionalities that Application Connector presents.
Right now we have multiple tests that run through the same scenario and test the same code.

The desired state would be to have:
- ~~a FIT scenario that goes through the whole flow available without Compass and testing available Application Gateway auth methods~~
  - ~~called either from lambda via internal eventing (presents some coupling to the Serverless and Eventing area)~~
  - ~~called from within a cluster with regular call~~
- ~~a FIT scenario that goes through Compass integration, tests the whole registration flow with Commerce Mock, sends an event through the mTLS gateway, and checks whether the external API of Commerce Mock can be called (make sure the cleanup is done properly in Compass)~~
- A test checking the functionalities of Application Gateway (proxying, various auth methods available - we could introduce mocks here for checking different methods)
- A test checking whether Applications are properly synchronized by the Compass Runtime Agent (we can use the static Compass tenant and just check if Compass Runtime Agent downloaded and applied the proper YAML)
- A test checking Connectivity Validator functionality (consider using a static TLS certificate and just sending a request there)

ad 1) we have multiple Octopus tests in Go that should be removed as they are no longer used (https://github.com/kyma-project/kyma/issues/13338)
ad 2) we need to clean up old prow pipelines (https://github.com/kyma-project/kyma/issues/11870)

**Reasons**

A clear test structure and responsibility division will reduce maintenance effort and provide better quality for the area. It will be transparent which components fail and which require some attention.

**Acceptance Criteria**
- [x] Mentioned test cases implemented
- [x] Unnecessary tests removed
- [x] Testing should take minimal possible execution time
- [x] Tests implemented should be easily maintainable

Disper:
[This testing strategy draft documentation](https://github.tools.sap/pandas/testing-strategy) might be somehow relevant.

janmedrek:
I have updated the test cases - with the modularization on the horizon we want to make sure that tests for the Application Connector module are separated.

VOID404:
Gateway Tests PR: [#1450](https://github.com/kyma-project/kyma/pull/14504)

akgalwas:
The following requirements were defined for the test project structure:
- It will be added here: https://github.com/kyma-project/kyma/tree/main/tests/components
- There will be separate subfolder for each test-suite
- There will be separate image built for each test-suite
[...]

PR #14504, opened 2022-06-07
Rewrite Central Gateway tests in Go

VOID404:
Needs some chart cleanup.
Test job will fail because it doesn't wait for test app to deploy.

**Description**

- ...
- ...
- ...

**Related issue(s)**
Part of #13977


franpog859:
I believe it's not needed anymore. Reopen it if I'm wrong

PR #14549, opened 2022-06-10
Fix helpers.tpl name definition for Application Gateway

janmedrek:
**Description**

- Fixing the name definition in `_helpers.tpl` file


koala7659 (approved):
LGTM

PR #14577, opened 2022-06-13
Disable istio sidecar injection in ory postgresql slaves

ebensom:
**Description**

Disable istio sidecar injection in ory postgresql slaves


werdes72 (approved):
✅

Issue #14532, opened 2022-06-09
Move serverless functions labels/continuousGitCheckout to annotations


pPrecel:
**Description**

Right now `function-controller` supports disabling git source checking for a specific function by adding a label to it. Problem is that such logic should be based on annotations, not labels. 

 - [x] Move the `serverless.kyma-project.io/continuousGitCheckout` label to annotations.

**Reasons**

Such logic should be based on annotations, not labels. 

**Attachments**

https://blog.getambassador.io/kubernetes-labels-vs-annotations-95fc47196b6d

Issue #12838, opened 2021-12-10
[POC] KEDA scalable functions


kwiatekus:
**Description**

Prove the following Concept:
"User may install KEDA and configure kyma serverless to stop using HPA based scaling and use KEDA API (Scaled Object CRD ) to delegate scaling function deployment based on arbitrary scaler. serverless uses the existing HPA strategy to replicate functions by default"

Please propose any necessary API changes (Function CRD).  Propose draft changes in the serverless tooling ( kyma CLI, Kyma Dashboard ) 

**Assumptions**:
 - We should not assume KEDA is part of serverless. we offer integration point towards KEDA. 
 - Using KEDA instead of HPA is a decision made by user and realised via configuration
 - Customer may use any arbitrary scaler that suits his usecase and matches the technologies used in his application (i.e various eventing mechanisms, various metrics, etc ...). 
 - We make no assumptions about Keda installation. It may be installed by user. It may be uninstalled at any time.
 - Mixing Keda with HPA is not allowed. A selected scaler strategy is global
 - User who is able to install KEDA should be able to change the scaling strategy (permission-wise)
 - Changing the scaling to KEDA should not be reverted by reconciler.
 
 
 **Questions**:

  - Could  [Prometheus scaler](https://keda.sh/docs/2.5/scalers/prometheus/) be a default scaler suggested to the users ( i.e in the ui ) as a "generic" scaler based on incoming requests? How to observe an incomming request outside of an istio side cart (scaling to zero case)
  -  Is it possible to change the scaling strategy in runtime (as opposed to installation time)
**Reasons**

We want to give  developers freedom to scale parts of their kyma hosted applications based on their own strategies.
We want to support  KEDA  integration  as it comes with a variety of official and community driven scalers.

**Attachments**

https://github.com/kyma-project/kyma/issues/11225
https://github.com/kyma-project/kyma/issues/11997

a-thaler:
Will be also interesting to see the impact of shipping a 3party component-> What is the estimated effort to keep all dependencies up-to-date.
Have you thought about the expectations to users? They might want to use it for regular deployments as well but based on other metrics or even other metrics sources as well? Can you support that and to which extend, what is the integration point?

PR #14591, opened 2022-06-15
Disable hpa annotation

pPrecel:
**Description**

- Add the annotation ( `serverless.kyma-project.io/disableHorizontalPodAutoscaler` ) to allow users to disable a hpa creation for the specific function

**Related issue(s)**

https://github.com/kyma-project/kyma/issues/12838
Depends on #14532

### DIFF
--- a/components/function-controller/internal/controllers/serverless/fsm.go
+++ b/components/function-controller/internal/controllers/serverless/fsm.go
@@ -52,6 +52,7 @@ type reconciler struct {
 
 const (
 	continuousGitCheckoutAnnotation = "serverless.kyma-project.io/continuousGitCheckout"
+	disableHorizontalPodAutoscaler  = "serverless.kyma-project.io/disableHorizontalPodAutoscaler"
 )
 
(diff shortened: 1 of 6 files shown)


components/function-controller/internal/controllers/serverless/fsm.go
```diff
@@ -51,7 +51,8 @@ type reconciler struct {
 }
 
 const (
-	continuousGitCheckoutLabel = "serverless.kyma-project.io/continuousGitCheckout"
+	continuousGitCheckoutLabel     = "serverless.kyma-project.io/continuousGitCheckout"
```
moelsayed, line 54: We should probably change the `Label` suffix

pPrecel: Yup, thanks. This change is committed in [this PR](https://github.com/kyma-project/kyma/pull/14671/files#diff-76e76f20345ca2b85f060c5e41f385358d07742aa541c9f6c70967b3063e59cbL54)

PR #14599, opened 2022-06-17
add a replace for kube-openapi in event-publisher-proxy

FriedrichWilken:
his pr adds a replace for `k8s.io/kube-openapi` to the latest commit to prevent security issues.


resources/eventing/values.yaml
```diff
@@ -5,11 +5,11 @@ global:
   images:
     eventing_controller:
       name: eventing-controller
-      version: PR-14580
+      version: PR-14599
```
nachtmaar, line 8: Is this a test ? :-D
Doesn't need to be updated but why not :-D

nachtmaar (approved):

resources/eventing/values.yaml
FriedrichWilken: uuups

nachtmaar (approved):

Issue #14558, opened 2022-06-10
[Telemetry Operator] Refactor secretRefs


a-thaler:
**Description**
As outlined in https://github.com/kyma-project/community/pull/675 the secretRefs attribute needs to be transformed into a "variables" attribute. Hereby, an explicit key mapping is required by design.

The minimal scope should cover that API:
```YAML
  variables: # env variables to be used in custom plugins
    - name: myEnv2
      valueFrom:
        secretKeyRef:
          name: my-elastic-credentials
          namespace: default
          key: ES_ENDPOINT
```

**Criterias**
- API is formulated as proposed
- User can provide secrets to a custom plugin config
- variables names must be global unique, if not, a validation fails with a nice error message pointing to the conflicting variables (incl. name of pipeline)
- examples are adjusted

rakesh-garimella:
- The api has been adjusted.
- There is warning. msg  fro admission hook if user references to a non existent secret. The logpipeline stays in pending state until the secret is available

a-thaler:
I tried out the global uniqueness feature and got that warning: `denied the request: variable with name 'icke2' has a been previously used in pipeline 'icke2'` Looks like there is a grammar mistake @NHingerl 

I deployed a pipeline referencing a secret with non existing key. It was nicely in pending state till I fixed the problem. However, it prints a warning to the kubectl output. I think that is not a best practice. Think of a gitops fashion of rolling out an application. By design you will apply all resources at the same time and they get eventual consistent at some point, no warning should be printed.

rakesh-garimella:
The secret rotation has been implemented via periodic reconciliation mechanism in reconciler, where the reconciler is checking for changes in all the watched resources periodically and reconciles if necessary. The current period is set to 1 hour by default which can be changed via overrides.

However if there is a need for forced rotation then one can achieve by adding a annotation: `kyma-project.com/timeStamp: <current-time>`

PR #14620, opened 2022-06-20
Refactor Logpipeline api for secret refs

rakesh-garimella:
**Description**

- Update the CRD as per latest specification
- If the secret referenced is not present then the pipeline stays in pending. While creating such pipeline user is presented with warning message.
- Included sync period to check watch all secrets periodically.

**Related issue(s)**

https://github.com/kyma-project/kyma/issues/14558


components/telemetry-operator/internal/sync/sync.go
```diff
@@ -257,56 +258,100 @@ func (s *LogPipelineSyncer) syncFilesConfigMap(ctx context.Context, logPipeline
...
+	newSecret.Data = make(map[string][]byte)
+
+	var logPipelines telemetryv1alpha1.LogPipelineList
+	err = s.List(ctx, &logPipelines)
+	if err != nil {
+		return false, err
+	}
+
+	if logPipeline.DeletionTimestamp != nil {
+		controllerutil.RemoveFinalizer(logPipeline, secretRefsFinalizer)
```
chrkl, line 296: I think we don't need that finalizer anymore since the secret is regenerated on any change. Also with your change it is not added anymore.

components/telemetry-operator/internal/sync/sync.go
```diff
@@ -257,56 +258,100 @@ func (s *LogPipelineSyncer) syncFilesConfigMap(ctx context.Context, logPipeline
...
+	}
+
+	for _, l := range logPipelines.Items {
+		for _, varRef := range l.Spec.Variables {
```
chrkl, line 300: We should skip LogPipelines that have a DeletionTimestamp. Would probably not make a big difference since they will finally disappear and not be added again in the next reconciliation round, but skipping them would immediately result in the desired state.

chrkl (changes requested):
Please also give more details in the PR description

components/telemetry-operator/internal/sync/sync.go
rakesh-garimella: I fetch the current list of logpipelines and here we would not have the deleting logpipeline. So thats why have moved the logic related to deleting log pipeline outside.

components/telemetry-operator/main.go
```diff
@@ -132,8 +133,10 @@ func main() {
 		setupLog.Error(err, "Invalid flag provided")
 		os.Exit(1)
 	}
+	var syncPeriod = 1 * time.Minute
```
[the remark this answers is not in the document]
rakesh-garimella: will do

components/telemetry-operator/internal/sync/sync.go
rakesh-garimella: removed.

components/telemetry-operator/internal/sync/sync.go
```diff
@@ -257,56 +258,100 @@ func (s *LogPipelineSyncer) syncFilesConfigMap(ctx context.Context, logPipeline
 	return changed, nil
 }
 
-// Copy referenced secrets to global Fluent Bit environment secret.
-func (s *LogPipelineSyncer) syncSecretRefs(ctx context.Context, logPipeline *telemetryv1alpha1.LogPipeline) (bool, error) {
+func isSecretRef(fromType telemetryv1alpha1.ValueFromType) bool {
```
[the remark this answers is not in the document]
rakesh-garimella: done.

components/telemetry-operator/internal/fluentbit/variable_validator.go
```diff
@@ -0,0 +1,72 @@
...
+	for _, variable := range logPipeline.Spec.Variables {
+		err := v.validateSecretKeysExist(context, variable.ValueFrom.SecretKey)
+		if err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func (v *variablesValidator) validateSecretKeysExist(ctx context.Context, secretKeyRef telemetryv1alpha1.SecretKeyRef) error {
```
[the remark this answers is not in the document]
rakesh-garimella: done.

components/telemetry-operator/internal/fluentbit/variable_validator.go
```diff
@@ -0,0 +1,63 @@
+package fluentbit
```
chrkl, line 1: This is not really related to fluentbit. How about moving all the validators to an own `validation` package?

components/telemetry-operator/internal/secret/secret_handling.go
```diff
@@ -0,0 +1,76 @@
...
+	"fmt"
+
+	telemetryv1alpha1 "github.com/kyma-project/kyma/components/telemetry-operator/api/v1alpha1"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	logf "sigs.k8s.io/controller-runtime/pkg/log"
+)
+
+type SecretValidator struct {
```
chrkl, line 15: Since this does not only validate secrets, let's find a better name. How about `SecretHelper`?

components/telemetry-operator/internal/fluentbit/variable_validator_test.go
```diff
@@ -0,0 +1,99 @@
...
+	"testing"
+
+	telemetryv1alpha1 "github.com/kyma-project/kyma/components/telemetry-operator/api/v1alpha1"
+	"github.com/kyma-project/kyma/components/telemetry-operator/internal/sync/mocks"
+	"github.com/stretchr/testify/require"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+func TestValidateSecretRefs(t *testing.T) {
+
```
chrkl, line 14: remove

components/telemetry-operator/internal/webhook/logpipeline_webhook.go
```diff
@@ -101,6 +106,17 @@ func (v *LogPipelineValidator) Handle(ctx context.Context, req admission.Request
 		}
 	}
 
+	secretsFound := v.validateSecrets(ctx, logPipeline)
+	if !secretsFound {
+		warnMsg := "One or more secrets do not exist, the log pipeline creation would not be completed"
```
chrkl, line 111: Let's be more precise about the behaviour in the warn message. I think it might be unclear if the user has to fix this, or it will resolve itself. Better write that applying the configuration will be pending until all secrets are available.

chrkl (approved):

chrkl (approved):

PR #14701, opened 2022-06-27
Limit number of log pipelines in telemetry-operator

### LINKED ISSUE #14459: [Telemetry Operator] Limit amount of pipelines
**Description**
The telemetry-operator is not managing yet the fluent-bit daemonset itself. With that the resource limits are hardcoded and a concept for scaling on base of amount of pipelines is still missing.
For now, the amount of pipelines which the daemonset can handle is limited (see https://github.com/kyma-project/kyma/issues/13144). So the operator should limit the amount to avoid instable setups.

**Criterias**
- define proper resource limits (have minimal settings for evaluation profile)
- derive max amount of pipelines and make it configurable (strive for 5 on prod and 3 on eval)
- if a user creates more pipelines then maximal allowed, pipeline validation must fail with a meaningful error message hinting the user in the right direction for self-helping

chrkl:
**Description**

Adding log pipelines results in creating a new emitter in the Fluent Bit configuration, which increases the memory usage through using an additional log buffer. This PR adds the option to define an overall limit for the number of log pipelines in the telemetry-operator to keep the resource usage of Fluent Bit under control.

- Add option to limit number of log pipelines to telemetry operator
- Make number of log pipelines configurable per override value
- Define defaults for evaluation and production profile

**Related issue(s)**

Resolves #14459


components/telemetry-operator/main.go
```diff
@@ -118,6 +119,7 @@ func main() {
 	flag.StringVar(&supportedOutputPlugins, "supported-output-plugins", "", "Comma separated list of supported output plugins. If empty, all output plugins are allowed.")
 	flag.StringVar(&deniedFilterPlugins, "denied-filter-plugins", "", "Comma separated list of denied filter plugins even if allowUnsupportedPlugins is enabled. If empty, all filter plugins are allowed.")
 	flag.StringVar(&deniedOutputPlugins, "denied-output-plugins", "", "Comma separated list of denied output plugins even if allowUnsupportedPlugins is enabled. If empty, all output plugins are allowed.")
+	flag.IntVar(&maxPipelines, "max-pipelines", 5, "Maximum number of log pipelines to be created. If 0, no limit is applied.")
```
dennis-ge, line 122: > If 0, no limit is applied

Is this covered by the current application?

dennis-ge (commented):
Looks great 💯 One small comment/question

components/telemetry-operator/main.go
chrkl: Good catch, covered that case in the validation.

dennis-ge (approved):

PR #14704, opened 2022-06-27
Implement scale sub-resource

pPrecel:
**Description**

- implement scale sub-resource to enable scaling functions using Kubernetes-based scalers (like HPA)

**Related issue(s)**

https://github.com/kyma-project/kyma/issues/12838

### DIFF
--- a/components/function-controller/internal/controllers/serverless/function_reconcile_test.go
+++ b/components/function-controller/internal/controllers/serverless/function_reconcile_test.go
@@ -21,6 +21,7 @@ import (
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/tools/record"
+	"k8s.io/utils/pointer"
 	ctrl "sigs.k8s.io/controller-runtime"
 
 	serverlessv1alpha2 "github.com/kyma-project/kyma/components/function-controller/pkg/apis/serverless/v1alpha2"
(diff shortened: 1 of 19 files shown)


pPrecel:
Merge this PR after this one:
https://github.com/kyma-project/kyma/pull/14976

components/function-controller/internal/controllers/serverless/system_state.go
```diff
@@ -481,6 +476,21 @@ func (s *systemState) buildDeployment(cfg buildDeploymentArgs) appsv1.Deployment
 	}
 }
 
+func (s *systemState) getReplicas(defaultVal int32) *int32 {
+	if s.instance.Spec.ScaleConfig != nil {
+		// TODO: question - should it looks like this?
+		// Maybe better way is to return s.deployments.Items[0].Spec.Replicas?
+		// this resource is managed by HPA and returning always minReplicas is confusing
```
pPrecel, line 483: What do you think about it? I think assigning always minReplicas to the deployment would cause problems when HPA scales up deployment (change deplo.spec.replicas to higher value) and at the same time fun-ctrl would always detect change on the deployment and at the same time scale down it (because of this template issue)

pPrecel:
To test this functionality create HPA for the function by applying HPA for example:

```yaml
apiVersion: autoscaling/v1
kind: HorizontalPodAutoscaler
metadata:
  name: <func_name>
spec:
  maxReplicas: 10
  minReplicas: 5
  scaleTargetRef:
    apiVersion: serverless.kyma-project.io/v1alpha2
    kind: Function
    name: demo
  targetCPUUtilizationPercentage: 50
```

or use kubectl to do that:

```bash
kubectl autoscale function <func_name> --cpu-percent=50 --min=5 --max=10
```

of course you can use KEDA ScaledObject or any Kubernetes based scaler

components/function-controller/internal/controllers/serverless/function_reconcile_test.go
```diff
@@ -35,6 +36,246 @@ const (
...
+		g.Expect(getConditionStatus(function.Status.Conditions, serverlessv1alpha2.ConditionConfigurationReady)).To(gomega.Equal(corev1.ConditionTrue))
+		g.Expect(getConditionStatus(function.Status.Conditions, serverlessv1alpha2.ConditionBuildReady)).To(gomega.Equal(corev1.ConditionTrue))
+		g.Expect(getConditionStatus(function.Status.Conditions, serverlessv1alpha2.ConditionRunning)).To(gomega.Equal(corev1.ConditionTrue))
+		g.Expect(getConditionReason(function.Status.Conditions, serverlessv1alpha2.ConditionRunning)).To(gomega.Equal(serverlessv1alpha2.ConditionReasonDeploymentReady))
+	})
+
+	t.Run("should propagate spec.replicas value to deployment", func(t *testing.T) {
+		//GIVEN
+		g := gomega.NewGomegaWithT(t)
+		inFunction := newFixFunction(testNamespace, "replicas", 1, 2)
```
dbadura, line 204: Why not to create function straight away with `spec.replicas` set to 3

components/function-controller/internal/controllers/serverless/scaling_test.go
```diff
@@ -284,43 +285,45 @@ func Test_isScalingEnabled(t *testing.T) {
 		maxReplicas int32
 	}
 	tests := []struct {
-		name string
-		args args
-		want bool
+		name        string
```
dbadura, line 288: Please remove `args` struct

components/function-controller/internal/controllers/serverless/service.go
```diff
@@ -34,7 +34,7 @@ func stateFnCheckService(ctx context.Context, r *reconciler, s *systemState) sta
 		return buildStateFnUpdateService(expectedSvc)
 	}
 
-	return stateFnCheckHPA
+	return initStateFnCheckScaling
```
dbadura, line 37: why `init`? I suggest `checkScaling`

components/function-controller/internal/controllers/serverless/system_state.go
dbadura: Return replicas from deployment from cluster

dbadura, line 483: https://github.com/kyma-project/kyma/pull/14704#discussion_r934352178

components/function-controller/pkg/apis/serverless/v1alpha2/function_defaults.go
```diff
@@ -89,6 +99,10 @@ func (spec *FunctionSpec) defaultReplicas(config *DefaultingConfig, fn *Function
 	}
 }
 
+func (spec *FunctionSpec) defaultSpecReplicas(replicas *int32) {
```
dbadura, line 102: Do we need that function?

components/function-controller/pkg/apis/serverless/v1alpha2/function_defaults.go
```diff
@@ -68,19 +68,29 @@ func (spec *FunctionSpec) defaultReplicas(config *DefaultingConfig, fn *Function
 	defaultingConfig := config.Function.Replicas
 	replicasPreset := mergeReplicasPreset(fn, defaultingConfig.Presets, defaultingConfig.DefaultPreset)
 
+	if spec.Replicas == nil && spec.ScaleConfig == nil && replicasPreset.Max == replicasPreset.Min {
```
dbadura, line 71: I would change it a little bit:
if ScaleConfig != nil -> defaultScaleConfig()
else
check and default if needed the replicas field.

We will end up in situation like this:
if nothing is set, we set the simplest field `replicas`

dbadura (changes requested):

components/function-controller/internal/controllers/serverless/function_reconcile_test.go
pPrecel: This test creates a function with min/max replicas, checks if everything is ok, and then changes the `scaleConfig` to nil & uses the `replicas` field. The reason is to test change from scaleConfig to replicas as well

components/function-controller/internal/controllers/serverless/system_state.go
pPrecel: This topic is more complicated than I expected. I moved it to the separate issue:
https://github.com/kyma-project/kyma/issues/15152

dbadura (dismissed):

dbadura (dismissed):

tobiscr (approved):

dbadura (approved):

Issue #14559, opened 2022-06-10
[Telemetry Operator] Validate and default filesystem buffer limits


a-thaler:
**Description**
After a recommended filesystem buffer size is determined (https://github.com/kyma-project/kyma/issues/13833) the operator need to assure that the output configs of the pipelines do not exceed the maximum buffer size.
The operator should generate a default setting (calculated out of the maximum) if no setting is provided and it should validate that a custom setting in combination with all other outputs is not overcommitting the buffer size.

In a first iteration, the operator could just take the max buffer size and generate a setting for all outputs, overwriting custom settings.

**Criterias**
- The filesystem buffer of the managed fluentbit instance will never grow above the agreed size
- A custom output cannot specify the buffer storage size. Otherwise validation fails
- The max buffer size for an output has a static value (will not change dependent on the amount of current pipelines)
- Ensure that all emitters use the storage type filesystem

a-thaler:
Message got fixed

PR #14705, opened 2022-06-28
validate and default buffer limits

dennis-ge:
**Description**

- Add flag to operator called `fluent-bit-max-filesystem-buffer-size` defaulted to `1G` and configurable via helm value.
- Set this flag in each output as `storage.total_limit_size`
- Validate in the webhook if the field `storage.total_limit_size` is set in the any output. Fail if this is the case
- Add `storage.type filesystem` to  tail plugin

**Related issue(s)**

Resolves https://github.com/kyma-project/kyma/issues/14559


components/telemetry-operator/controllers/logpipeline_controller_test.go
```diff
@@ -51,13 +51,15 @@ var _ = Describe("LogPipeline controller", func() {
     Emitter_Mem_Buf_Limit 10M
 
 [FILTER]
-    Name   grep
-    Match   *
-    Regex   $kubernetes['labels']['app'] my-deployment
+    match *
```
dennis-ge, line 54: This should be possible, right? 
But since this test doesn't valide the config with the webhook, it is unnoticed. What should we do?

components/telemetry-operator/controllers/logpipeline_controller.go
```diff
@@ -52,13 +52,14 @@ type LogPipelineReconciler struct {
 }
 
 // NewLogPipelineReconciler returns a new LogPipelineReconciler using the given FluentBit config arguments
-func NewLogPipelineReconciler(client client.Client, scheme *runtime.Scheme, daemonSetConfig sync.FluentBitDaemonSetConfig, emitterConfig fluentbit.EmitterConfig) *LogPipelineReconciler {
+func NewLogPipelineReconciler(client client.Client, scheme *runtime.Scheme, daemonSetConfig sync.FluentBitDaemonSetConfig, emitterConfig fluentbit.EmitterConfig, fluentBitMaxFSBufferSize string) *LogPipelineReconciler {
```
chrkl, line 55: Maybe let's make the EmitterConfig a more general "PipelineConfig" and have the max. filesystem buffer size in there. Then we would not grow this argument list to infinity ;-) Maybe you can find a better name than "PipelineConfig".

components/telemetry-operator/internal/webhook/logpipeline_webhook.go
```diff
@@ -236,3 +245,18 @@ func (v *LogPipelineValidator) validateSecrets(ctx context.Context, logPipeline
 	secVal := secret.NewSecretHelper(v.Client)
 	return secVal.ValidateSecretsExist(ctx, logPipeline)
 }
+
+func (v *LogPipelineValidator) validateOutput(logPipeline *telemetryv1alpha1.LogPipeline) error {
```
chrkl, line 249: Let's move this to a OutputValidator. Then also add unit tests for it and cover it in the logpipeline_webhook_test.

chrkl (changes requested):

components/telemetry-operator/controllers/logpipeline_controller_test.go
chrkl: What do you mean? The lower-case `match` or indentation? I guess both should work.

dennis-ge: I mean the `*` (match all) expression. It works since the validation webhook is not deployed in this test.

Issue #11113, opened 2021-04-13
Enhance Eventing dashboards


marcobebway:
**Description**

Enhance Eventing dashboards:

- Enable the Eventing NATS dashboard.
- Address the review comments from https://github.com/kyma-project/kyma/pull/10975.
  Items:
  - [x] https://github.com/kyma-project/kyma/pull/10975#discussion_r611953011
  - [x] https://github.com/kyma-project/kyma/pull/10975#discussion_r611945680
  - [ ] use templating or constants in a consistent way across all the usages of ports

PR #14764, opened 2022-07-06
Make the port usages consistent across the eventing templates

VladislavPaskar:
**Description**

This PR is trying to maximise the port usage consistency throughout the eventing templates.

**Related issue(s)**
https://github.com/kyma-project/kyma/issues/11113

### DIFF
--- a/resources/eventing/charts/controller/values.yaml
+++ b/resources/eventing/charts/controller/values.yaml
@@ -42,7 +42,7 @@ metrics:
     nameSuffix: "-metrics"
   config:
     port: 8080
-    portName: "-metrics"
+    portName: metrics
     nameSuffix: "-metrics"
     interval: 30s
 
--- a/resources/eventing/charts/controller/values.yaml
+++ b/resources/eventing/charts/controller/values.yaml
@@ -42,7 +42,7 @@ metrics:
     nameSuffix: "-metrics"
   config:
     port: 8080
-    portName: http-metrics
+    portName: "-metrics"
     nameSuffix: "-metrics"
     interval: 30s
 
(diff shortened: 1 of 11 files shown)


resources/eventing/charts/publisher-proxy/values.yaml
```diff
@@ -18,10 +18,11 @@ nats:
 metrics:
   config:
     port: 9090
-    portName: http-metrics
+    portName: "-metrics"
     nameSuffix: "-metrics"
```
VladislavPaskar, line 22: for reviewer:

We could potentially merge those two constants (`portName` and `nameSuffix`), as they are equal. But I didn't do this, because for me those still refer to different thins: one is name of the port and the second one is used in metadata. So I would keep them separated in case want want to change one of them in the future.

resources/eventing/values.yaml
```diff
@@ -34,10 +34,13 @@ global:
...
   domainName: "kyma.example.com"
 
+  ports:
+    httpPrefix: http
+
   istio:
     proxy:
       statusPort: 15020
-      portName: http-status
+      portName: "-status"
```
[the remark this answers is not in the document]
VladislavPaskar: I like the idea, done.

resources/eventing/charts/controller/values.yaml
```diff
@@ -42,7 +42,7 @@ metrics:
     nameSuffix: "-metrics"
```
raypinto, line 42: do we need this nameSuffix and the nameSuffix in line 46?

VladislavPaskar: They were pointing to the same, so I removed them. Also removed some over unused constants.

raypinto (approved):

Issue #14562, opened 2022-06-10
[Telemetry Operator] Match expression only for own pipeline


a-thaler:
**Description**
The operator supports custom filters and outputs with match expressions refering to other pipelines.
With https://github.com/kyma-project/community/pull/675/files that is obsolete and should be declined.

**Criterias**
- Custom filters and outputs cannot define a match expression, validation fails
- The match expression gets generated always with the expression for the respective pipeline the plugin is defined in
- Improved error message not being misleading
- Examples are adjusted and working

Teneroy:
Prohibited match conditions for options and filters, removed the function to search for user defined match expressions, adapted and refactored tests. Now, no user can define their match expressions in outputs and filters, those are being generated automatically.

PR #14767, opened 2022-07-06
Prohibited matches in filters and outputs in telemetry-operator

Teneroy:
**Description**

- Prohibited "Match" in filters and outputs
- Fixed validator tests
- Refactored tests to adapt to the new logic
- Changed examples

**Related issue(s)**
#14562


NHingerl (dismissed):

components/telemetry-operator/internal/validation/plugin_validator_test.go
```diff
@@ -39,6 +39,7 @@ func TestValidateForbiddenFilters(t *testing.T) {
 	err := sut.Validate(logPipeline, logPipelines)
 
 	require.Error(t, err)
+	assert.Contains(t, err.Error(), "plugin 'grep' contains match condition. Match conditions are forbidden")
```
chrkl, line 42: Looks like the test before did not what the name suggests. Let's split it to two tests. One to validate that a LogPipeline that has a `match` expression is rejected and one that tests for the forbidden `lua` filter.

components/telemetry-operator/internal/validation/plugin_validator_test.go
```diff
@@ -57,14 +58,15 @@ func TestValidateForbiddenOutputs(t *testing.T) {
 	err := sut.Validate(logPipeline, logPipelines)
 
 	require.Error(t, err)
+	assert.Contains(t, err.Error(), "plugin 'es' contains match condition. Match conditions are forbidden")
```
chrkl, line 61: Same here, let's also test that the `es` output is forbidden.

chrkl (changes requested):

chrkl (approved):

PR #14779, opened 2022-07-08
improve warn msgs

rakesh-garimella:
**Description**

- Fix validation error (https://github.com/kyma-project/kyma/issues/14559#issuecomment-1178936089)
- Improve the warn msg you get when user is using unsupported plugin in pipeline. https://github.com/kyma-project/kyma/issues/14558#issuecomment-1178916240
- Removed warning if the secret used in secretRefs is not existing. https://github.com/kyma-project/kyma/issues/14558#issuecomment-1178916240

**Related issue(s)**

https://github.com/kyma-project/kyma/issues/14558

### DIFF
--- a/components/telemetry-operator/internal/validation/variable_validator.go
+++ b/components/telemetry-operator/internal/validation/variable_validator.go
@@ -56,7 +56,7 @@ func validateMandatoryFieldsAreEmpty(vr telemetryv1alpha1.VariableReference) boo
 func findConflictingVariables(logPipeLine *telemetryv1alpha1.LogPipeline, vr telemetryv1alpha1.VariableReference, existingPipelineName string) error {
 	for _, v := range logPipeLine.Spec.Variables {
 		if v.Name == vr.Name {
-			return fmt.Errorf("variable '%s' has been previously used in pipeline '%s'", v.Name, existingPipelineName)
+			return fmt.Errorf("Variable name must be globally unique. Variable '%s' is used in pipeline '%s'", v.Name, existingPipelineName)
 		}
 	}
 	return nil
--- a/components/telemetry-operator/internal/validation/variable_validator.go
+++ b/components/telemetry-operator/internal/validation/variable_validator.go
@@ -56,7 +56,7 @@ func validateMandatoryFieldsAreEmpty(vr telemetryv1alpha1.VariableReference) boo
 func findConflictingVariables(logPipeLine *telemetryv1alpha1.LogPipeline, vr telemetryv1alpha1.VariableReference, existingPipelineName string) error {
 	for _, v := range logPipeLine.Spec.Variables {
 		if v.Name == vr.Name {
-			return fmt.Errorf("variable with name '%s' has been previously used in pipeline '%s'", v.Name, existingPipelineName)
+			return fmt.Errorf("variable '%s' has been previously used in pipeline '%s'", v.Name, existingPipelineName)
 		}
 	}
 	return nil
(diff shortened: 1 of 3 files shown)


components/telemetry-operator/internal/validation/variable_validator.go
```diff
@@ -56,7 +56,7 @@ func validateMandatoryFieldsAreEmpty(vr telemetryv1alpha1.VariableReference) boo
 func findConflictingVariables(logPipeLine *telemetryv1alpha1.LogPipeline, vr telemetryv1alpha1.VariableReference, existingPipelineName string) error {
 	for _, v := range logPipeLine.Spec.Variables {
 		if v.Name == vr.Name {
-			return fmt.Errorf("variable with name '%s' has a been previously used in pipeline '%s'", v.Name, existingPipelineName)
+			return fmt.Errorf("variable with name '%s' has been previously used in pipeline '%s'", v.Name, existingPipelineName)
```
NHingerl, line 59: Maybe we can shorten it a bit?

```suggestion
			return fmt.Errorf("variable '%s' has been previously used in pipeline '%s'", v.Name, existingPipelineName)
```

Is the problem that it's been used before (no matter whether it's still in use now), or is it the problem that it's used in two places at the same time?

```suggestion
			return fmt.Errorf("Cannot use variable name '%s', already used in pipeline '%s'", v.Name, existingPipelineName)
```
Not sure whether I may change the syntax like this; will it be weird in rendered version?

components/telemetry-operator/internal/webhook/logpipeline_webhook.go
```diff
@@ -117,12 +115,10 @@ func (v *LogPipelineValidator) Handle(ctx context.Context, req admission.Request
...
 	var warnMsg []string
 
-	secretsFound := v.validateSecrets(ctx, logPipeline)
-	if !secretsFound {
-		warnMsg = append(warnMsg, "One or more secrets do not exist. The LogPipeline stays in pending state until the secrets are available.")
-	}
 	if v.pluginValidator.ContainsCustomPlugin(logPipeline) {
-		warnMsg = append(warnMsg, "Caution: LogPipeline contains an unsupported custom filter or output. This means that you proceed without support!")
+		helpText := "https://kyma-project.io/docs/kyma/latest/01-overview/main-areas/observability/obsv-04-telemetry-in-kyma/"
+		msg := fmt.Sprintf("Logpipeline '%s' uses unsupported custom filters or outputs. Please consider changing the pipeline to use supported filters or output as described in the documentation: %s", logPipeline.Name, helpText)
```
NHingerl, line 120: Just a bit shorter ;) 

```suggestion
		msg := fmt.Sprintf("Logpipeline '%s' uses unsupported custom filters or outputs. We recommend changing the pipeline to use supported filters or output. See the documentation: %s", logPipeline.Name, helpText)
```

components/telemetry-operator/internal/validation/variable_validator.go
rakesh-garimella: not sure if second message is a bit hard tone. We want to have a bit softer tone if I understand correct. I will go with first one.

NHingerl (approved):

components/telemetry-operator/internal/validation/variable_validator.go
```diff
@@ -56,7 +56,7 @@ func validateMandatoryFieldsAreEmpty(vr telemetryv1alpha1.VariableReference) boo
 func findConflictingVariables(logPipeLine *telemetryv1alpha1.LogPipeline, vr telemetryv1alpha1.VariableReference, existingPipelineName string) error {
 	for _, v := range logPipeLine.Spec.Variables {
 		if v.Name == vr.Name {
-			return fmt.Errorf("variable with name '%s' has a been previously used in pipeline '%s'", v.Name, existingPipelineName)
+			return fmt.Errorf("variable '%s' has been previously used in pipeline '%s'", v.Name, existingPipelineName)
```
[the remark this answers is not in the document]
NHingerl: Good idea! 
Would this suggestion work in the rendered message?

```suggestion
			return fmt.Errorf(Variable name must be unique. "variable '%s' is used in pipeline '%s'", v.Name, existingPipelineName)
```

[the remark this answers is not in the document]
a-thaler: For me that global aspect is missing and important. It is not enough to have it unique per pipeline but across all pipelines

PR #14781, opened 2022-07-08
Update BEB subscription Status to show error messages in the conditions

raypinto:
**Description**

- Update the APIRule ready and Subscribed conditions to show error messages, if any
- Update sink validator message

**Related issue(s)**

### DIFF
--- a/components/eventing-controller/controllers/subscription/beb/reconciler.go
+++ b/components/eventing-controller/controllers/subscription/beb/reconciler.go
@@ -272,7 +272,7 @@ func (r *Reconciler) syncBEBSubscription(subscription *eventingv1alpha1.Subscrip
 	}
 
 	if _, err := r.Backend.SyncSubscription(subscription, r.eventTypeCleaner, apiRule); err != nil {
-		r.syncConditionSubscribed(subscription, false)
+		r.syncConditionSubscribed(subscription, err)
 		return false, err
 	}
 
@@ -284,7 +284,7 @@ func (r *Reconciler) syncBEBSubscription(subscription *eventingv1alpha1.Subscrip
 	}
 
 	// sync the condition: ConditionSubscribed
-	r.syncConditionSubscribed(subscription, true)
+	r.syncConditionSubscribed(subscription, nil)
 
 	// sync the condition: ConditionSubscriptionActive
 	r.syncConditionSubscriptionActive(subscription, isActive, logger)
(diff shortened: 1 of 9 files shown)


components/eventing-controller/controllers/subscription/beb/reconciler.go
```diff
@@ -311,7 +313,8 @@ func (r *Reconciler) syncConditionSubscriptionActive(subscription *eventingv1alp
 	condition := eventingv1alpha1.MakeCondition(eventingv1alpha1.ConditionSubscriptionActive, eventingv1alpha1.ConditionReasonSubscriptionActive, corev1.ConditionTrue, "")
 	if !isActive {
 		logger.Debugw("Waiting for subscription to be active", "name", subscription.Name, "status", subscription.Status.EmsSubscriptionStatus.SubscriptionStatus)
-		condition = eventingv1alpha1.MakeCondition(eventingv1alpha1.ConditionSubscriptionActive, eventingv1alpha1.ConditionReasonSubscriptionNotActive, corev1.ConditionFalse, "")
+		message := fmt.Sprintf("current subscription status: %s", subscription.Status.EmsSubscriptionStatus.SubscriptionStatus)
```
mfaizanse, line 316: (nit) Whats the benefit of this message? Won't the user see value `subscription.Status.EmsSubscriptionStatus.SubscriptionStatus` in the status? Can we write something like: "Waiting for subscription to be active.."

components/eventing-controller/controllers/subscription/beb/reconciler_test.go
```diff
@@ -271,7 +271,7 @@ var _ = Describe("Subscription Reconciliation Tests", func() {
 				eventingv1alpha1.ConditionAPIRuleStatus,
 				eventingv1alpha1.ConditionReasonAPIRuleStatusNotReady,
 				v1.ConditionFalse,
-				"",
+				"sink URL scheme should be 'http' or 'https'",
```
mfaizanse, line 274: Are we sure that sink url can be `https`?

components/eventing-controller/controllers/subscription/beb/reconciler_test.go
```diff
@@ -747,7 +747,8 @@ var _ = Describe("Subscription Reconciliation Tests", func() {
 			ensureAPIRuleStatusUpdatedWithStatusReady(ctx, &apiRuleCreated).Should(BeNil())
 
 			By("Setting a subscription not created condition")
-			subscriptionNotCreatedCondition := eventingv1alpha1.MakeCondition(eventingv1alpha1.ConditionSubscribed, eventingv1alpha1.ConditionReasonSubscriptionCreationFailed, v1.ConditionFalse, "")
+			message := "create subscription failed: 500; 500 Internal Server Error;{\"message\":\"sorry, but this mock does not let you create a BEB subscription\"}\n"
```
mfaizanse, line 750: (nit) This message looks more like log, would it be okay to show such formatted message in the conditions? Can we only extract the `error.message` part?

components/eventing-controller/controllers/subscription/beb/reconciler_test.go
raypinto: This is the same error message from [here](https://github.com/kyma-project/kyma/blob/cc5af9ca8fae62d9166896554bf83615272c0fe5/components/eventing-controller/pkg/handlers/sink/validator.go#L21). But I will see if I can use that variable directly instead :)

components/eventing-controller/controllers/subscription/beb/reconciler_test.go
raypinto: I will have a look if that's possible. For now I print the entire error as string as you can see [here](https://github.com/raypinto/kyma/blob/0547523da9401f518dad3d06d9c6fc2319813461/components/eventing-controller/api/v1alpha1/condition.go#L229)

PR #14784, opened 2022-07-11
Fluentbit filesystem buffer metrics

Teneroy:
**Description**

- Fixed merge conflicts with Ben's repo

**Related issue(s)**


components/telemetry-fluentbit-sidecar/main_test.go
```diff
@@ -0,0 +1,85 @@
+package main
```
chrkl, line 1: I would suggest to have some end-to-end test that also pulls the /metrics URI and validates that the metrics are updated if the directory size changes.

components/telemetry-fluentbit-sidecar/main.go
```diff
@@ -0,0 +1,126 @@
...
+
+	http.Handle("/metrics", promhttp.Handler())
+	err = http.ListenAndServe(":2021", nil)
+	if err != nil {
+		panic(err)
+	}
+}
+
+func recordMetrics(logPath string) {
+	ticker := time.NewTicker(10 * time.Second)
```
chrkl, line 65: Maybe align this with the scraping frequency of our Prometheus

components/telemetry-fluentbit-sidecar/main.go
```diff
@@ -0,0 +1,126 @@
...
+	fsbufferMax = promauto.NewGauge(prometheus.GaugeOpts{
+		Namespace:   "",
+		Subsystem:   "",
+		Name:        "telemetry_fsbuffer_limit",
+		Help:        "The maximum defined size of the fluentbit chunk buffer",
+		ConstLabels: nil,
+	})
+)
+
+func main() {
```
chrkl, line 46: Let's have some minimal logging that at least initially prints the watched directory and served port. In best case, you would also have an info log level that logs every iteration. Consider using [this](https://github.com/kyma-project/kyma/tree/main/common/logging/logger) logging library.

chrkl (changes requested):

Issue #14654, opened 2022-06-22
New Application Connectivity tests - Test suite for Application Gateway


mvshao:
**Description**
This task aims to implement separate Application Gateway tests into a separate minimal suite, written in Go. 
The general idea is described [here](https://github.com/kyma-project/kyma/issues/13977#issuecomment-1148410870).

**Acceptance criteria**
- [x] New project created under the following directory: https://github.com/kyma-project/kyma/tree/main/tests/components
- [x] All authentication methods listed [here](https://github.com/kyma-project/kyma/issues/13977#issuecomment-1148492062) covered.
- [x] All test-cases listed [here](https://github.com/kyma-project/kyma/issues/13977#issuecomment-1148462789) covered.
- [x] Test is deployable on Kubernetes.
- [x] Makefile compatible with pipeline

**Related issue**
Test's build pipeline: [test-infra#5563](https://github.com/kyma-project/test-infra/pull/5563)
#13977

VOID404:
# Workplan

- [ ]  Endpoints + AppCR entries ([docs](https://kyma-project.io/docs/kyma/main/03-tutorials/00-application-connectivity/ac-04-register-secured-api/)):
    - [x]  no authorisation
    - [x]  Basic Auth
    - [x]  OAuth protected with mTLS (please see #14401)
    - [ ]  OAuth (grant client credentials flow)
    - [ ]  mTLS (a.k.a CertGen)
    - [ ]  CSRF with:
        - [ ]  no authorisation
        - [ ]  Basic Auth
        - [ ]  OAuth (grant client credentials flow)
        - [ ]  OAuth protected with mTLS
        - [ ]  mTLS
    - [ ]  Expired scenarios
- [ ]  Logging
    - [ ]  in test-app
    - [ ]  in tests
- [ ]  Log formatting?
- [ ]  Deployments
    - [ ]  Dockerfile
    - [ ]  Helm Charts
    - [ ]  `Makefile`

akgalwas:
Plan
- [ ] Prepare initial PR with project structure. It should test BasicAuth with positive authorisation only. The goal is to deliver MVP and allow parallel work. https://github.com/kyma-project/kyma/pull/14743
- [ ] Prepare pipeline for building Gateway tests 
https://github.com/kyma-project/test-infra/pull/5563/files
- [ ] Prepare pipeline for executing Gateway tests 
- [ ] Enhance test with path related scenarios https://github.com/kyma-project/kyma/pull/14762
[...]

PR #14792, opened 2022-07-12
Add Basic with CSRF scenario to the App Gateway tests

franpog859:
**Description**

- Add Basic with CSRF scenario to the App Gateway tests

**Related issue(s)**

https://github.com/kyma-project/kyma/issues/14654


tests/components/application-connector/resources/charts/gateway-test/templates/basic-cases.yml
```diff
@@ -30,6 +30,48 @@ spec:
...
+      entries:
+        - type: API
+          targetUrl: "http://test-app.{{ .Values.namespace  }}.svc.cluster.local:8080/v1/api/csrf-basic/ok"
+          centralGatewayUrl: "http://central-application-gateway.kyma-system:8080/test-app/csrf-basic"
+          credentials:
+            secretName: basic-test
+            type: Basic
+            csrfInfo:
+              tokenEndpointURL: "http://test-app.{{ .Values.namespace  }}.svc.cluster.local:8080/v1/api/csrf/token"
+    - displayName: bad csrf token basic
```
mvshao, line 47: In my opinion, separating positive and negative cases into separate Application CRD will increase the readability of the tests.

franpog859: Hmm, I'm not sure as we probably will have only one or two positive cases per authorization method (so also per an Application) and the cases are going to be quite well described with the return codes. If you still think we should separate it, I'll do it after the rebase onto main with the PR that adds the multiple Application map to the test suite, @mvshao

tests/components/application-connector/internal/testkit/test-api/csrf.go
```diff
@@ -0,0 +1,93 @@
...
+				handleError(w, http.StatusForbidden, "CSRF token header missing")
+				return
+			}
+
+			ch.mutex.RLock()
+			_, found := ch.tokens[headerToken]
+			ch.mutex.RUnlock()
+
+			if !found {
+				handleError(w, http.StatusForbidden, "Invalid token from the header")
```
VOID404, line 71:
```suggestion
				handleError(w, http.StatusForbidden, "Invalid CSRF token from the header")
```

VOID404 (changes requested):

VOID404 (approved):

PR #14799, opened 2022-07-13
Add Application Gateway OAuth tests

franpog859:
**Description**

- Add Application Gateway OAuth tests

**Related issue(s)**

https://github.com/kyma-project/kyma/issues/14654


tests/components/application-connector/internal/testkit/test-api/apis.go
```diff
@@ -11,13 +11,20 @@ import (
...
 	router := mux.NewRouter()
 
 	router.HandleFunc("/v1/health", alwaysOk).Methods("GET")
 	api := router.PathPrefix("/v1/api").Subrouter()
 	api.Use(Logger(logOut, logger.DevLoggerType))
 
+	oauth := NewOAuth(oAuthCredentials.ClientID, oAuthCredentials.ClientSecret)
+
+	{
+		api.HandleFunc("/oauth/token", oauth.Token).Methods(http.MethodPost)
```
VOID404, line 24: there is already an `/oauth` block, starting from line 40

tests/components/application-connector/internal/testkit/test-api/oauth.go
```diff
@@ -0,0 +1,138 @@
...
+const (
+	clientIDKey     = "client_id"
+	clientSecretKey = "client_secret"
+	grantTypeKey    = "grant_type"
+)
+
+type OauthResponse struct {
+	AccessToken string `json:"access_token"`
+	TokenType   string `json:"token_type"`
+	ExpiresIn   int    `json:"expires_in"`
```
VOID404, line 28: This field [is not required](https://www.rfc-editor.org/rfc/rfc6749#section-4.2.2) and we do not handle it correctly, so maybe we should omit it

tests/components/application-connector/internal/testkit/test-api/oauth.go
```diff
@@ -0,0 +1,138 @@
...
+	Scope       string `json:"scope"`
```
VOID404, line 29: This field [is not required](https://www.rfc-editor.org/rfc/rfc6749#section-4.2.2) and we do not handle it correctly, so maybe we should omit it

tests/components/application-connector/internal/testkit/test-api/oauth.go
```diff
@@ -0,0 +1,138 @@
...
+			splitToken := strings.Split(authHeader, "Bearer")
+			if len(splitToken) != 2 {
+				handleError(w, http.StatusForbidden, "Bearer token missing")
+				return
+			}
+
+			token := strings.TrimSpace(splitToken[1])
+
+			oh.mutex.RLock()
+			_, found := oh.tokens[token]
```
VOID404, line 106: If we ignore values in the map, maybe we could have `interface{}` and `nil` inside, to signify the lack of meaning of the stored value?

VOID404 (changes requested):

tests/components/application-connector/internal/testkit/test-api/apis.go
franpog859: I wanted to separate the token endpoints and the _target url endpoints_ as the token ones will be used across also the CSRF test but the _target urls_ will be different. In this case, there will be shared token urls and separated _target urls_ for each test case

VOID404: We could consider naming endpoint blocks in the future 🤔

VOID404 (approved):

PR #14805, opened 2022-07-13
Reconcile method unit tests for eventing reconcilers

raypinto:
**Description**

- Added Reconcile method unit tests for BEB reconciler
- Added Reconcile method unit tests for JetStream reconciler
- Add Finalizer test as part of JetStream integration test

**Related issue(s)**


components/eventing-controller/controllers/subscription/jetstream/reconciler_internal_test.go
```diff
@@ -30,6 +33,125 @@ const (
...
+			},
+			wantReconcileResult: ctrl.Result{},
+			wantReconcileError:  cleanerErr,
+		},
+	}
+
+	for _, testCase := range testCases {
+		tc := testCase
+		t.Run(tc.name, func(t *testing.T) {
+			reconciler := testCase.givenReconcilerSetup()
```
marcobebway, line 143: Should we use `tc.givenReconcilerSetup()` instead?

components/eventing-controller/controllers/subscription/beb/reconciler_internal_test.go
```diff
@@ -14,17 +27,119 @@ import (
...
+			wantReconcileError:  backendSyncErr,
+		},
+		{
+			name:              "Return error and default Result{} when backend delete returns error",
+			givenSubscription: testSubUnderDeletion,
+			givenReconcilerSetup: func() *Reconciler {
+				te := setupTestEnvironment(t)
+				fakeClient := te.clientBuilder.WithObjects(testSubUnderDeletion).Build()
+				te.backend.On("Initialize", mock.Anything).Return(nil)
+				te.backend.On("DeleteSubscription", mock.Anything).Return(backendDeleteErr)
```
marcobebway, line 109: Why do we return a backend error when we delete a subscription?

components/eventing-controller/controllers/subscription/beb/reconciler_internal_test.go
```diff
@@ -14,17 +27,119 @@ import (
...
 	reconcilertesting "github.com/kyma-project/kyma/components/eventing-controller/testing"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
 const (
 	domain = "domain.com"
 )
 
+// Test the return values of the Reconcile() method of the reconciler. This is important, as it dictates whether the
```
marcobebway, line 40: Should start with `TestReconciler_Reconcile`.

components/eventing-controller/controllers/subscription/beb/reconciler_internal_test.go
raypinto: Actually that's the idea of the testcase: if the deletion handler returns an error of any sort, that error is propagated and returned as the reconciliation error by the reconciler.

marcobebway: > to check if the deletion handler returns an error of any sort

Exactly, the test should handle any error, and the `backendDeleteErr` seems too specific when you first read it.

(nit) How about returning something like `anyError` to make it more obvious?

components/eventing-controller/controllers/subscription/beb/reconciler_internal_test.go
```diff
@@ -14,17 +27,119 @@ import (
...
+		givenReconcilerSetup func() *Reconciler
+		wantReconcileResult  ctrl.Result
+		wantReconcileError   error
+	}{
+		{
+			name:              "Return nil and default Result{} when there is no error from the reconciler dependencies",
+			givenSubscription: testSub,
+			givenReconcilerSetup: func() *Reconciler {
+				te := setupTestEnvironment(t)
+				fakeClient := te.clientBuilder.WithObjects(testSub).Build()
```
marcobebway, line 81: (nit) Does it make sense to move this setup inside the `setupTestEnvironment` func?

```golang
func setupTestEnvironment(t *testing.T, objs ...client.Object) *testEnvironment
```

components/eventing-controller/controllers/subscription/beb/reconciler_internal_test.go
```diff
@@ -14,17 +27,119 @@ import (
...
+	var testCases = []struct {
+		name                 string
+		givenSubscription    *eventingv1alpha1.Subscription
+		givenReconcilerSetup func() *Reconciler
+		wantReconcileResult  ctrl.Result
+		wantReconcileError   error
+	}{
+		{
+			name:              "Return nil and default Result{} when there is no error from the reconciler dependencies",
+			givenSubscription: testSub,
```
marcobebway, line 78: We use the same instance of `testSub` in multiple test cases, and IMO it is not future proof, because reconciling the same instance from multiple test-cases could lead to unwanted/unforeseen changes.

I suggest to either to initialise a new instance every time a test case is executed, or use a deepCopy from the original instance.

marcobebway:
Thanks @raypinto for the PR. I believe we need to add a couple more test cases:
- Reconciliation request for a non-existing subscription.
- Reconciliation request which is requeued.

components/eventing-controller/controllers/subscription/beb/reconciler_internal_test.go
raypinto: This is nice! Done

components/eventing-controller/controllers/subscription/beb/reconciler_internal_test.go
raypinto: I verified this and I can confirm that since we add the subscription `testSub` to the fake clientBuilder each time during a testcase, the subscription is as good as new.

components/eventing-controller/controllers/subscription/beb/reconciler_internal_test.go
raypinto: As we discussed, `backendDeleteErr` is a generic error message that is mocked. It could be a real error or nil. As long as it is propagated and returned by the Reconcile method, we should be good.

marcobebway:
I checked the JetStream reconciler tests and did not find any test case which covers adding the finalizer to the Kyma subscription. Could you please add a test for it?

components/eventing-controller/controllers/subscription/jetstream/reconciler_test.go
```diff
@@ -98,11 +98,13 @@ func TestCreateSubscription(t *testing.T) {
 				reconcilertesting.WithFilter(reconcilertesting.EventSource, reconcilertesting.OrderCreatedEventTypeNotClean),
 				reconcilertesting.WithWebhookForNATS(),
 				reconcilertesting.WithSinkURLFromSvc(ens.SubscriberSvc),
+				reconcilertesting.WithFinalizers([]string{}),
```
marcobebway, line 101: +1

marcobebway (approved):

Issue #14521, opened 2022-06-08
Subscription duplicates Content-Type on Cloud Event retry


johgoe:
**Description**

We are using kyma 2.2 with a nats backend. Sometimes it happens that a event could not handled by the the subscription sink service which results in an http error e.g. status 500.

We saw that the event will be delivered automatically after a some time but on the second try the content-type of the event is `application/json,application/json`. On the 3rd try the content-type header contains `application/json,application/json,application/json` and (or multiple content-type headers not sure about that because we can't log it on the network level but only in our application)

**Expected result**

We expected the content-type header stays stable with value of `application/json` on every retry.

**Actual result**

content-type header is appendens an additional `application/json` on every retry 

**Steps to reproduce**

1) Subscribe an event to a function which return an error 500 on first call.
2) Publish an event
3) Watch the received event on the 2nd retry.

FriedrichWilken:
Thank you @johgoe we will look into this.

bayern01kahn:
Hi Guys.  I met the same issue here. below are some logs.

`2022-06-15 05:18:59.513 WARN 7 --- [nio-8021-exec-2] .w.s.m.s.DefaultHandlerExceptionResolver : Resolved [org.springframework.web.HttpMediaTypeNotSupportedException: Invalid mime type "application/json,application/json,application/json,application/json": Invalid token character ',' in token "json,application/json,application/json,application/json"]
2022-06-15 05:25:56.630 WARN 7 --- [nio-8021-exec-4] .w.s.m.s.DefaultHandlerExceptionResolver : Resolved [org.springframework.web.HttpMediaTypeNotSupportedException: Invalid mime type "application/json,application/json,application/json": Invalid token character ',' in token "json,application/json,application/json"]
2022-06-15 05:30:56.691 WARN 7 --- [nio-8021-exec-6] .w.s.m.s.DefaultHandlerExceptionResolver : Resolved [org.springframework.web.HttpMediaTypeNotSupportedException: Invalid mime type "application/json,application/json,application/json,application/json": Invalid token character ',' in token "json,application/json,application/json,application/json"]
[...]

PR #14835, opened 2022-07-15
Rework tests nats subscriber

friedrichwilken:
This is a rework of the nats tests in the nats handlers for cloudevents to test not only against event data but against the whole cloud event including it's meta data.

- check against CE instead of CE data only

**Related issue(s)**
https://github.com/kyma-project/kyma/issues/14521

### DIFF
--- a/components/eventing-controller/controllers/backend/reconciler_test.go
+++ b/components/eventing-controller/controllers/backend/reconciler_test.go
@@ -667,8 +667,8 @@ func bebSecretExists(ctx context.Context) bool {
 	return len(secretList.Items) > 0
 }
 
-// creates a secret containing the oauth2 credentials that is expected to be
-// created by the Hydra operator
+// createOAuth2Secret creates a secret containing the oauth2 credentials that is expected to be
+// created by the Hydra operator.
 func createOAuth2Secret(ctx context.Context, clientID, clientSecret []byte) {
 	sec := &corev1.Secret{
 		ObjectMeta: metav1.ObjectMeta{
--- a/components/eventing-controller/pkg/handlers/nats_test.go
+++ b/components/eventing-controller/pkg/handlers/nats_test.go
@@ -7,22 +7,67 @@ import (
 	"testing"
 	"time"
 
-	"github.com/kyma-project/kyma/components/eventing-controller/pkg/handlers/metrics"
-
 	"github.com/avast/retry-go/v3"
-	cev2event "github.com/cloudevents/sdk-go/v2/event"
+	cenats "github.com/cloudevents/sdk-go/protocol/nats/v2"
+	ce "github.com/cloudevents/sdk-go/v2"
+	cebinding "github.com/cloudevents/sdk-go/v2/binding"
+	ceevent "github.com/cloudevents/sdk-go/v2/event"
 	kymalogger "github.com/kyma-project/kyma/common/logging/logger"
-	"github.com/nats-io/nats.go"
-	. "github.com/onsi/gomega"
-
 	eventingv1alpha1 "github.com/kyma-project/kyma/components/eventing-controller/api/v1alpha1"
 	"github.com/kyma-project/kyma/components/eventing-controller/logger"
 	"github.com/kyma-project/kyma/components/eventing-controller/pkg/application/applicationtest"
 	"github.com/kyma-project/kyma/components/eventing-controller/pkg/application/fake"
 	"github.com/kyma-project/kyma/components/eventing-controller/pkg/env"
 	"github.com/kyma-project/kyma/components/eventing-controller/pkg/handlers/eventtype"
+	"github.com/kyma-project/kyma/components/eventing-controller/pkg/handlers/metrics"
 	eventingtesting "github.com/kyma-project/kyma/components/eventing-controller/testing"
+	"github.com/nats-io/nats.go"
+	. "github.com/onsi/gomega"
 )
+
+const (
+	Structured SendEncoding = 0
+	Binary     SendEncoding = 1
+)
+
+type SendEncoding uint8
+
+// SendCloudEventToNATS sends a CloudEvent to a NATS client using a defined encoding method.
+func SendCloudEventToNATS(natsClient *Nats, cloudEvent *ceevent.Event, encoding SendEncoding) error {
+	// create a NATS-sender
+	natsOpts := cenats.NatsOptions()
+	url := natsClient.config.URL
+	subject := cloudEvent.Subject()
+	sender, err := cenats.NewSender(url, subject, natsOpts)
+	if err != nil {
+		return fmt.Errorf("failed to create NATS protocol sender, %s", err.Error())
+	}
+
+	// create a CloudEvent client
+	ceClient, err := ce.NewClient(sender)
+	if err != nil {
+		return fmt.Errorf("failed to create an CloudEvent client, %s", err.Error())
+	}
+
+	// define the encoding for sending the event
+	ctx := contextWithSendEncoding(encoding)
+
+	if result := ceClient.Send(ctx, *cloudEvent); ce.IsUndelivered(result) {
+		return fmt.Errorf("failed to send CloudEvent, %s", result.Error())
+	}
+	return nil
+}
+
+func contextWithSendEncoding(encoding SendEncoding) context.Context {
+	ctx := context.Background()
+	switch encoding {
+	case Structured:
+		ctx = cebinding.WithForceStructured(ctx)
+	case Binary:
+		ctx = cebinding.WithForceBinary(ctx)
+	}
+	return ctx
+}
 
 var (
 	nextSinkPort = &portGenerator{port: 8088}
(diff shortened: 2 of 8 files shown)


components/eventing-controller/controllers/backend/reconciler_test.go
```diff
@@ -667,7 +667,7 @@ func bebSecretExists(ctx context.Context) bool {
 	return len(secretList.Items) > 0
 }
 
-// creates a secret containing the oauth2 credentials that is expected to be
+// createOAuth2Secret creates a secret containing the oauth2 credentials that is expected to be
```
VladislavPaskar, line 670: (nit)
in case you want to improve the comment, you could also add point to the end of the comment as per go guidelines.

components/eventing-controller/pkg/handlers/nats_test.go
```diff
@@ -4,26 +4,81 @@ import (
...
+	Structured SendEncoding = 0
+	Binary     SendEncoding = 1
 )
 
+type SendEncoding uint8
+
+// SendCloudEventToNATS sends a CloudEvent to a NATS client using a defined encoding method.
+func SendCloudEventToNATS(natsClient *Nats, cloudEvent *ceevent.Event, encoding SendEncoding) error {
+	// validate the cloudEvent
+	if err := cloudEvent.Validate(); err != nil {
```
VladislavPaskar, line 39: the `ceClient.Send()` from line 63 also validates the cloudEvent, so I guess you can omit this check.

components/eventing-controller/pkg/handlers/nats_test.go
```diff
@@ -4,26 +4,81 @@ import (
...
+	sender, err := cenats.NewSender(url, subject, natsOpts)
+	if err != nil {
+		log.Fatalf("Failed to create nats protocol sender, %s", err.Error())
+		return err
+	}
+
+	// create a CloudEvent client
+	ceClient, err := ce.NewClient(sender)
+	if err != nil {
+		log.Fatalf("Failed to create an cloudEvent client, %s", err.Error())
```
VladislavPaskar, line 56: The fatal include the `os.exit(1)`, in that case what's after doesn't matter. So I would better use `log.Printf`

components/eventing-controller/pkg/handlers/nats_test.go
```diff
@@ -4,26 +4,81 @@ import (
...
+		return err
+	}
+
+	// define the encoding for sending the event
+	ctx := contextWithSendEncoding(encoding)
+
+	if result := ceClient.Send(ctx, *cloudEvent); ce.IsUndelivered(result) {
+		err = fmt.Errorf("failed to send cloudEvent: %s", result.Error())
```
VladislavPaskar, line 64: why not just replace these 3 lines with this one-liner:
```
return fmt.Errorf("failed to send CloudEvent: %s", result.Error())
```

components/eventing-controller/testing/event/cehelper/converter.go
```diff
@@ -0,0 +1,36 @@
...
+//   datacontenttype: application/json
+// Data,
+//   "{\"foo\":\"bar\"}"
+func RequestToEventString(r *http.Request) (string, error) {
+	msg := cehttp.NewMessageFromHttpRequest(r)
+	event, err := cebinding.ToEvent(context.Background(), msg)
+	if err != nil {
+		err = fmt.Errorf("failed to build a CloudEvent: %v", err)
+		log.Fatal(err)
+		return "", err
```
VladislavPaskar, line 32: one this: better don't use fatal for the reasons explained here. I am also not sure, whether you need to log something here, since you log the error output in the both of the function calls.

second, you can do it in one-liner.

```suggestion
		return "", fmt.Errorf("failed to build a CloudEvent: %v", err)
```

components/eventing-controller/testing/event/cehelper/converter.go
```diff
@@ -0,0 +1,36 @@
...
+	}
+	s := event.String()
+	return s, nil
```
VladislavPaskar, line 35:
```suggestion
	return event.String(), nil
```

components/eventing-controller/testing/subscriber.go
```diff
@@ -73,6 +174,7 @@ func getServeMux() *http.ServeMux {
 		retries.Inc()
 		w.WriteHeader(http.StatusInternalServerError)
 	})
+	// this endpoint returns the number of attempted retries.
```
VladislavPaskar, line 177: (nit)
for consistency 
```suggestion
	// this Endpoint returns the number of attempted retries.
```

components/eventing-controller/testing/subscriber.go
```diff
@@ -26,20 +36,108 @@ type Subscriber struct {
...
 
-const (
-	maxNoOfData           = 10
-	maxAttempts           = 10
-	storeEndpoint         = "/store"
-	checkEndpoint         = "/check"
-	internalErrorEndpoint = "/return500"
-	checkRetriesEndpoint  = "/check_retries"
-)
+// NewSubscriber creates a simple test Subscriber with http endpoints to received, store end answer
```
VladislavPaskar, line 39: (nit)

Not sure why you change the sequence of the functions, I guess it really makes more sense to have the Constructor sooner in the code. My small problem is, that with such changes you hardly can see the code differences, if there were some.

components/eventing-controller/testing/subscriber.go
```diff
@@ -124,24 +198,31 @@ func (s *Subscriber) IsRunning() bool {
...
 	err := retry.Do(
 		func() error {
+			// check if a response was received and that it's code is in 2xx-range
 			resp, err := http.Get(s.checkURL)
 			if err != nil {
 				return pkgerrors.Wrapf(err, "get HTTP request failed")
 			}
 			if !is2XXStatusCode(resp.StatusCode) {
-				return fmt.Errorf("Response code is not 2xx, received Response code is: %d", resp.StatusCode)
+				return fmt.Errorf("expected resonse code 2xx, actual response code: %d", resp.StatusCode)
```
VladislavPaskar, line 213: (nit)
```suggestion
				return fmt.Errorf("expected response code 2xx, actual response code: %d", resp.StatusCode)
```

components/eventing-controller/testing/subscriber.go
```diff
@@ -124,24 +198,31 @@ func (s *Subscriber) IsRunning() bool {
...
+
+			// try to read the response body
 			defer func() { _ = resp.Body.Close() }()
 			body, err = ioutil.ReadAll(resp.Body)
 			if err != nil {
 				return pkgerrors.Wrapf(err, "read data failed")
 			}
-			if string(body) != expectedData {
+
+			// compare response body with expectations
```
VladislavPaskar, line 223: what is the benefit of doing it in two lines ?

components/eventing-controller/pkg/handlers/nats_test.go
```diff
@@ -957,20 +1012,28 @@ func TestIsValidSubscription(t *testing.T) {
...
+		MaxIdleConnsPerHost: 2,
+		IdleConnTimeout:     6 * time.Second,
 	}
 	defaultMaxInflight := 1
 	defaultSubsConfig := env.DefaultSubscriptionConfig{MaxInFlightMessages: defaultMaxInflight}
 	natsBackend := NewNats(natsConfig, env.DefaultSubscriptionConfig{MaxInFlightMessages: defaultMaxInflight}, metrics.NewCollector(), defaultLogger)
 	g.Expect(natsBackend.Initialize(nil)).Should(Succeed())
 
-	subscriber := eventingtesting.NewSubscriber()
+	subscriber := eventingtesting.NewSubscriber(
```
VladislavPaskar, line 1034: Why not make the `eventingtesting.WithCloudEventServeMux()` part of the `NewSubscriber` constructor, since we want the same logic every time ?

components/eventing-controller/pkg/handlers/nats_test.go
friedrichwilken: because then the error would not be printed but just returned to the expect() of the tests. or would it still be printed?

components/eventing-controller/pkg/handlers/nats_test.go
friedrichwilken: No, some of the old test still need the old serveMux. For that reason I added `WithCloudEventServeMux()` as an option only for the case that CloudEvents are explicitly needed.

components/eventing-controller/testing/event/cehelper/converter.go
```diff
@@ -28,9 +27,7 @@ func RequestToEventString(r *http.Request) (string, error) {
 	event, err := cebinding.ToEvent(context.Background(), msg)
 	if err != nil {
 		err = fmt.Errorf("failed to build a CloudEvent: %v", err)
```
VladislavPaskar, line 29: I think you want to remove this line also.

VladislavPaskar (changes requested):
Could you please squash the commits before the PR gets merge, so we get a clean commit message after the merge.

VladislavPaskar (approved):

Issue #14593, opened 2022-06-15
Migration from serverless.kyma-project.io/v1alpha1 to serverless.kyma-project.io/v1alpha2


kwiatekus:
**Description**

As part of https://github.com/kyma-project/kyma/issues/13828 we plan to implement a migration of customers' functions from serverless.kyma-project.io/v1alpha1 to serverless.kyma-project.io/v1
We assume that there is only one valid function CR version per kyma version.

 - Document the migration steps (for Kyma OSS users)
 - Migrate customer's function automatically via reconciler
 - Cleanup GitRepo CRs

**Reason**
Releasing of stable Function CRD version

moelsayed:
a quick update on the status of the conversion webhook task:
I restored the v1alpha1 types
I added the required gen comments to be able to regenerate the crd with both versions
I implemented a webhook and 95% of the conversion code.

The controller-runtime uses a very simple hub-spoke pattern to implement version conversion. You just implement a couple of interfaces with your types and write the conversion code. The problem with this is that it's very rigid and abstracted. Our conversion code actually needs to "merge" two CRDs (function and the gitrepo) into one. It's not possible to do that with the hub-spoke pattern (at least I couldn't do it).

I went in another direction and tried to implement the webhook server from scratch. At this point, everything seems to be working, but the API (kubernetes API!) is sending an incorrect request. Instead of getting a conversionReview request from the API, I am getting a AdmissionRewiew request .. I am still trying to figure out why. I think it's a configuration issue, but can't figure out where exactly..

for now, the changes are all here: https://github.com/m00g3n/kyma/compare/fn-crd-refactor...moelsayed:migration_wip2?expand=1 . This is just experimental code, please excuse the quality.
I will have some time today, so I might keep trying to get it working. If not, I will be available on slack to answer any questions, please do feel free to ping me at any time. If I am not on a plane, I will respond :D

kwiatekus:
Leftovers
 - [ ] Inject CA bundle cert into converting webhook definition @Cortey 
 - [ ] Implement conversion from `v1alpha1` into `v1` @pPrecel @anoipm

anoipm:
State for now.

Newest sources: 
https://github.com/m00g3n/kyma/tree/fn-crd-refactor-tests

TODO in webhook:
- [ ] function controller doesn't create services after conversion,
- [ ] "apply function" says the function name does not match,
- [ ] in the logs of pod we can see infinite attempts to convert v1alpha2->v1alpha1,
[...]

PR #14848, opened 2022-07-19
Initial two-way conversion implementation

moelsayed:
**Description**

- Implement v1alpha1 <-> v1alpha2 conversions
- Update charts
- Restore `GitRepository` CRD

**Related issue(s)**

https://github.com/kyma-project/kyma/issues/14593

### DIFF
--- a/components/function-controller/internal/webhook/conversion_webhook.go
+++ b/components/function-controller/internal/webhook/conversion_webhook.go
@@ -177,7 +177,8 @@ func (w *ConvertingWebhook) convertSpecV1Alpha1ToV1Alpha2(in *serverlessv1alpha1
 	out.Runtime = serverlessv1alpha2.Runtime(in.Runtime)
 	out.RuntimeImageOverride = in.RuntimeImageOverride
 
-	//TODO: out.ResourceConfiguration / profile
+	//TODO: out.Profile
+	//TODO out.CustomRuntimeCOnfiguration
 	out.ResourceConfiguration = serverlessv1alpha2.ResourceConfiguration{
 		Build: serverlessv1alpha2.ResourceRequirements{
 			Resources: in.BuildResources,
@@ -199,7 +200,6 @@ func (w *ConvertingWebhook) convertSpecV1Alpha1ToV1Alpha2(in *serverlessv1alpha1
 		if err != nil {
 			return err
 		}
-
 		out.Source = serverlessv1alpha2.Source{
 			GitRepository: &serverlessv1alpha2.GitRepositorySource{
 				URL: repo.Spec.URL,
@@ -273,6 +273,7 @@ func (w *ConvertingWebhook) convertSpecV1Alpha2ToV1Alpha1(in *serverlessv1alpha2
 		out.Source = in.Source.Inline.Source
 		out.Deps = in.Source.Inline.Dependencies
 	} else {
+		out.Type = serverlessv1alpha1.SourceTypeGit
 		// check repo name in the annotations,
 		if repoName == "" {
 			// create a random name git repo with the information. This is not supposed to happen, it's just a precaution.
--- a/components/function-controller/internal/webhook/conversion_webhook.go
+++ b/components/function-controller/internal/webhook/conversion_webhook.go
@@ -269,7 +269,7 @@ func (w *ConvertingWebhook) convertSpecV1Alpha2ToV1Alpha1(in *serverlessv1alpha2
 	out.Labels = in.Template.Labels
 
 	// TODO: clean check
-	if in.Source.GitRepository.URL == "" {
+	if in.Source.GitRepository == nil {
 		out.Source = in.Source.Inline.Source
 		out.Deps = in.Source.Inline.Dependencies
 	} else {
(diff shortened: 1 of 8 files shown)


dbadura (changes requested):

components/function-controller/internal/webhook/conversion_webhook_test.go
```diff
@@ -0,0 +1,389 @@
+package webhook
```
dbadura, line 1: Can you add use case when downgrading from v2 to v1 withouth repository name in annotation?

resources/serverless/charts/webhook/templates/cluster-role.yaml
```diff
@@ -45,3 +45,12 @@ rules:
     verbs:
       - get
       - update
+  - apiGroups:
+      - serverless.kyma-project.io
+    resources:
+      - gitrepositories
+    verbs:
+      - get
+      - list
```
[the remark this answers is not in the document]
moelsayed: Yes. It looks like the controller-runtime is using it internally to setup caches.

components/function-controller/internal/webhook/conversion_webhook_test.go
```diff
@@ -0,0 +1,508 @@
...
+	tests := []struct {
+		name          string
+		src           runtime.Object
+		wantDst       runtime.Object
+		wantVersion   string
+		wantErr       bool
+		checkRecreate bool
+	}{
+		{
+			name: "v1alpha1 to v1alpha2 inline function",
```
dbadura, line 52: I would add test case when resources are not filled or even consider to remove all resources in conversion scenarios to make them smaller and add special test case with resources.

components/function-controller/internal/webhook/conversion_webhook.go
```diff
@@ -0,0 +1,378 @@
...
+		return w.convertFunctionV1Alpha2ToV1Alpha1(src, dst)
+	default:
+		dstGVK := dst.GetObjectKind().GroupVersionKind()
+		return fmt.Errorf("unsupported convert source version %s ", dstGVK)
+	}
+}
+
+// v1alpha1 -> v1alpha2
+func (w *ConvertingWebhook) convertFunctionV1Alpha1ToV1Alpha2(src, dst runtime.Object) error {
+	in := src.(*serverlessv1alpha1.Function)
```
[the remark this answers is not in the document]
moelsayed: That would be nicer, I agree. I chose to go this way to make the caller more readable though: https://github.com/kyma-project/kyma/pull/14848/files#diff-e954c26e3ad25788289bae5d2f07044a4ce5183e84a1a28402b4bd7ffdf74005R126-R138

If I change this to use function types, I will have to move the type casting and validation into the switch and that will make it ugly and much less readable..

components/function-controller/internal/webhook/conversion_webhook.go
```diff
@@ -0,0 +1,378 @@
...
+	if err := w.client.Create(context.Background(), repo); err != nil {
+		return "", errors.Wrap(err, "failed to create GitRepository")
+	}
+
+	return repoName, nil
+
+}
+
+func recreatedRepoName(functionName string) string {
+	return fmt.Sprintf("%s-recreated-repo", functionName)
```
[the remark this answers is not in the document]
moelsayed: I am not trying to "mark" it as much as I am trying to avoid name collisions.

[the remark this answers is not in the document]
dbadura: This function can be deleted

components/function-controller/internal/webhook/conversion_webhook_test.go
```diff
@@ -0,0 +1,365 @@
...
+								SecretName: testGitRepo.Spec.Auth.SecretName,
+							},
+						},
+					},
+				},
+			},
+			wantVersion: serverlessv1alpha2.GroupVersion.String(),
+		},
+		{
+			name: "v1alpha2 to v1alpha1 Git function",
```
dbadura, line 281: I would add the case with status conversion from v1 to v2.

components/function-controller/internal/webhook/conversion_webhook.go
```diff
@@ -0,0 +1,348 @@
...
+	out.Spec.BuildResources = in.Spec.ResourceConfiguration.Build.Resources
+	out.Spec.Resources = in.Spec.ResourceConfiguration.Function.Resources
+
+	out.Spec.Labels = in.Spec.Template.Labels
+
+	return w.convertSourceV1Alpha2ToV1Alpha1(in, out)
+}
+
+func (w *ConvertingWebhook) convertSourceV1Alpha2ToV1Alpha1(in *serverlessv1alpha2.Function, out *serverlessv1alpha1.Function) error {
+	if in.Spec.Source.GitRepository == nil {
```
dbadura, line 294: I would check if `Inline` is not nil:
- code is clearer and you only work with inline, other types doesn't matter
- possible nil (this shouldn't happen)

dbadura (commented):
Almost ready to go :)

components/function-controller/internal/webhook/conversion_webhook_test.go
moelsayed: Good Call :+1:

dbadura (approved):

Issue #14557, opened 2022-06-10
[Telemetry Operator] Add http output with minimal attribute set


a-thaler:
**Description**
Introduce an output definition for pushing logs using the http output plugin, also see https://github.com/kyma-project/community/pull/675/files. Settings for a typical OpenSearch integration are needed:
- host - value and valueFrom support needed
- http_User - value and valueFrom support needed
- http_Passwd  - value and valueFrom support needed
- uri: value

All other values need meaningful defaults like having TLS switched on

For now the valueFrom must support the secretKeyRef notation only

```YAML
  output: #only one output, no output should fail validation
    http: # enables http output
      Host: # next 3 attributes should be configurable by static value and valueFrom incl. secret rotation
        value: "icke.com"
        valueFrom: ...
      HTTP_User:
        value: "icke"
        valueFrom:
          secretKeyRef:
            name: my-elastic-credentials
            namespace: default
            key: ES_USER
      HTTP_Password:
        value: "icke.com"
        valueFrom: ...
      URI: /customindex/kyma
```

**Criterias**
- defining an output of type "http" is possible
- users can inject values from secrets
- default setup is using TLS
- integration tests with in-cluster HTTP backend

chrkl:
HTTP output API has been implemented by https://github.com/kyma-project/kyma/pull/14863.

PR #14863, opened 2022-07-20
Add http output to LogPipeline

chrkl:
**Description**

This PR introduces an explicit http Fluent Bit output plugin to the LogPipeline CRD as described in the [configurable logging concept](https://github.com/kyma-project/community/blob/main/concepts/observability-configurable-logging/08-revised-api.md#revised-api-1).

- Extend LogPipeline CRD with http output
- Generate http output plugin configuration from new API by the telemetry-operator
- Handle referenced secret values from http output API

**Related issue(s)**

See also #14557


components/telemetry-operator/config/samples/telemetry_v1alpha1_logpipeline_http.yaml
```diff
@@ -0,0 +1,81 @@
+apiVersion: v1
```
rakesh-garimella, line 1: I think splitting the file would be nice here as one can directly apply only http logpipeline by few adjustments

components/telemetry-operator/internal/validation/plugin_validator.go
```diff
@@ -103,6 +124,26 @@ func checkIfPluginIsValid(content string, pipeline *telemetryv1alpha1.LogPipelin
 	return nil
 }
 
+func validateHTTPOutput(httpOutput telemetryv1alpha1.HTTPOutput) error {
+	if httpOutput.Host.Value != "" && !validHostname(httpOutput.Host.Value) {
```
rakesh-garimella, line 128: he default of URI is `/` . Here user would need to pass the `/` as URI may be we need to cover this in docs. ?

components/telemetry-operator/internal/validation/plugin_validator.go
```diff
@@ -103,6 +124,26 @@ func checkIfPluginIsValid(content string, pipeline *telemetryv1alpha1.LogPipelin
...
+		return fmt.Errorf("invalid hostname '%s'", httpOutput.Host.Value)
+	}
+	if httpOutput.URI != "" && !strings.HasPrefix(httpOutput.URI, "/") {
```
rakesh-garimella, line 131: ../../../resources/telemetry/values.yaml

rakesh-garimella (approved):

PR #14867, opened 2022-07-21
Implement verbs tests

VOID404:
**Description**
Test if gateway uses correct method and forwards body for various HTTP verbs

**Related issue(s)**

#14654


tests/components/application-connector/resources/charts/gateway-test/templates/applications/body-forwarding.yml
```diff
@@ -0,0 +1,38 @@
+apiVersion: applicationconnector.kyma-project.io/v1alpha1
+kind: Application
+metadata:
+  name: body-forwarding
```
franpog859, line 4: I think the methods are more key in this test, aren't they? Maybe something like `methods-with-body`

tests/components/application-connector/resources/charts/gateway-test/templates/applications/body-forwarding.yml
```diff
@@ -0,0 +1,38 @@
...
+  namespace: "{{ .Values.namespace }}"
+spec:
+  description: body forwarding with various methods
```
franpog859, line 7: Uppercase. I'd also keep in mind that we should somehow document the idea with the usage of `service.Description` as we use it as it is to create a request body and the method

tests/components/application-connector/test/application-gateway/runner_test.go
```diff
@@ -71,3 +73,32 @@ func (gs *GatewaySuite) TestResponseBody() {
...
+		gs.Run(service.Description, func() {
+			httpCli := NewHttpCli(gs.T())
+
+			for _, entry := range service.Entries {
+				if entry.Type != "API" {
+					gs.T().Log("Skipping event entry")
+					continue
+				}
+
+				req, err := http.NewRequest(service.Description, entry.CentralGatewayUrl, strings.NewReader(service.Description))
```
franpog859, line 90: I'd explicitly save these in some variables to make it more readable. You know, `method := service.Description` and `body := strings.NewReader(service.Description)`. It'd be more clear that we use this description as we do

franpog859 (commented):
Looks good! Please, bump the images, because the tests are not running yet :D

franpog859 (approved):
Much better! 🚀

franpog859 (approved):

PR #14875, opened 2022-07-21
fix the validation problem with parsers

rakesh-garimella:
**Description**

- Validate parsers by passing `-R` flag as normal dry-run is not able to catch faulty parser config.
- Validate current logparser
- Validate current logpipeline + all the parsers present
**Related issue(s)**


components/telemetry-operator/internal/validation/config_validator.go
```diff
@@ -99,11 +104,16 @@ func extractError(output string) string {
 		return r2Matches[2] // 0: complete output, 1: label, 2: description
 	}
 
-	r3 := regexp.MustCompile(`.*(?P<label>\[error]\s)(?P<description>.+)`)
+	r3 := regexp.MustCompile(`.*(?P<label>\[error]\s\[parser]\s)(?P<description>.+)(\sin.+)`)
```
chrkl, line 107: Update the godoc comment above the function

chrkl (changes requested):

Issue #14564, opened 2022-06-10
[Telemetry Operator] Add input section for application logs with filters for rewrite_tag


a-thaler:
**Description**
According to https://github.com/kyma-project/community/pull/675/files a new root section for defining a single input is needed. Here, you can select a type of input and configure the different types with more details. 

Furthermore the namespace and container selectors are required, which will result into a rule expression for the related rewrite_tag filter.

```YAML
  input: # singular, different inputs require different pipelines and output configs
    application: # maps to central tail pipeline, dealing with the actual application logs
      namespaces: [] # generates the rule for the rewrite_tag assigned to every pipeline
      excludeNamespaces: []
      containers: []
      excludeContainers: []
```

**Criterias**
- the input section is optional, by default "application" type is selected without any selector
- all input types can be configured simultaneous (no exclusive or) as the used type will be selected by the "type" selector
- namespaces and containers selected or excluded need to be evaluated already on the rewrite_tag so that unnecessary logs can be dropped here already
- the selectors do not support wildcards for now
- include and exclude expressions cannot be used in parallel, validation should fail with appropriate validation message
- the defaults should not collect any logs from Kyma and Kubernetes system components
- the defaults should not collect any logs from fluentbit containers

lindnerby:
Introduced an extra flag to `includeSystemNamespaces` within the application input to include system namespaces `kyma-system`, `kyma-integration`, `istio-system` and `kube-system` that are excluded by default when applying a logpipeline. Flag is overriden by explicitly defining exluded or included namespaces.

lindnerby:
Adapted the API and documentation accordingly after feedback to match the following structure:
```
input:
  application:
    namespaces:
      include:
      exclude:
      system:
    containers:
      include:
      exclude:
```

PR #14934, opened 2022-07-27
Telemetry input section

lindnerby:
**Description**

- Create input section definition for LogPipeline
- Create Rewrite Tag filter rules according to `spec.input.application` selectors
- Add default exclusion of system namespaces
- Refactor emitter config generation
- Add more tests

**Related issue(s)**
* https://github.com/kyma-project/kyma/issues/14564


components/telemetry-operator/internal/fluentbit/emitter.go
```diff
@@ -0,0 +1,85 @@
...
+	return sb
+}
+
+func (sb *SectionBuilder) ToString() string {
+	sb.builder.WriteByte('\n')
+	return sb.builder.String()
+}
+
+// CreateEmitter creates the Fluent Bit Rewrite Tag Filter section
+func CreateEmitter(config PipelineConfig, logPipeline *telemetryv1alpha1.LogPipeline) string {
```
skhalash, line 45: Not sure about the naming here. I like the idea of extracting this piece of logic into a separate file, but I'd rather call it `rewrite_tag` / `CreateRewriteTagSection` or something. Otherwise, it can be confused with rewrite tag's emitter fields: https://docs.fluentbit.io/manual/pipeline/filters/rewrite-tag.

components/telemetry-operator/internal/helper/defaults.go
```diff
@@ -0,0 +1,62 @@
+package helper
```
skhalash, line 1: I'd prefer to have a segment in the package name that indicates that this `helper` is not a generic helper, but a `logpipeline` helper. For example, here some good example: https://github.com/kubernetes/kubernetes/blob/master/pkg/controller/deployment/util/deployment_util.go.

components/telemetry-operator/internal/validation/input_validator_test.go
```diff
@@ -0,0 +1,64 @@
+package validation
```
skhalash, line 1: If we create a `logpipipeline` subdirectory as discussed above, it can be moved inside it since it's validating a log pipeline.

components/telemetry-operator/internal/helper/sync.go
```diff
@@ -1,4 +1,4 @@
-package sync
+package helper
```
skhalash, line 1: Is there any reason this package was merged/renamed to `helper`?

skhalash (changes requested):
The logic itself looks solid but there are some questions with regards to the module names/structure.

components/telemetry-operator/internal/helper/defaults.go
lindnerby: Agree as discussed, but let's adapt the structure of internal pkg in a separate PR. 👌

skhalash (approved):

skhalash (approved):

PR #14942, opened 2022-07-28
Add unit test for auth removal

moelsayed:
**Description**

- Add unit test for auth removal
- Remove dead code
- ...

**Related issue(s)**

https://github.com/kyma-project/kyma/issues/14593

### DIFF
--- a/components/function-controller/internal/webhook/conversion_webhook_test.go
+++ b/components/function-controller/internal/webhook/conversion_webhook_test.go
@@ -389,35 +389,11 @@ func TestConvertingWebhook_convertFunctionWithRemovedAuth(t *testing.T) {
 		},
 	}
 
-	v1alpha2FunctionWithAuth := &serverlessv1alpha2.Function{
-		ObjectMeta: metav1.ObjectMeta{
-			Name: "test", Namespace: "test",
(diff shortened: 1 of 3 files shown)


components/function-controller/internal/webhook/conversion_webhook_test.go
```diff
@@ -380,6 +380,82 @@ func TestConvertingWebhook_convertFunction(t *testing.T) {
...
+	tests := []struct {
+		name        string
+		src         runtime.Object
+		repo        serverlessv1alpha1.GitRepository
+		wantDst     runtime.Object
+		wantVersion string
+		wantErr     bool
+	}{
+		{
+			name: "v1alpha1 to v1alpha2 Git function - auth removed",
```
dbadura, line 407:
```suggestion
			name: "v1alpha1 to v1alpha2 Git function without auth",
```

dbadura (commented):
Why do we need seperate test for this?
Can we just add new test case?

moelsayed:
@dbadura it will just complicate the other test cases and the test body. I have already started with that as you can probably tell by the leftover code :smile:

dbadura (approved):

Issue #14737, opened 2022-07-04
Define and document API lifecycle management


varbanv:
**Description**

We do not have an established and documented process on what and how the API lifecycle is managed. 
We need to define and document the process and rules around the API lifecycle model.
Some questions to consider:
- Do we allow alpha/beta features on production clusters?
- How long should an API be in alpha/beta stage?
- What is the process around deprecating an API?
- What commitments do we want to make around API versions?

**Links**
[Kubernetes API changes](https://kubernetes.io/docs/concepts/overview/kubernetes-api/#api-changes)
[Kubernetes features](https://kubernetes.io/docs/reference/command-line-tools-reference/feature-gates/#feature-stages)

a-thaler:
I think there is nothing against applying the k8s policies to Kyma. Major implications where we need to improve are:
- alpha features must be disabled by default, maybe we can introduce a dedicated profile to enable them?
- beta features are enabled by default but can be disabled. Do we then really want to enable beta features by default for the production profile as you might to apply manual migration routines on upgrades

varbanv:
- At the moment, we cannot disable alpha features by default.
- We need to start planing the migration of alpha to stable.
- We would provide advance user notice for the changes - proposed timeline 9 months, or 3 releases?
- We strive to provide an out-of-the-box migration between versions - do we need to document the details for end-users?
-

PR #14975, opened 2022-07-29
Add API versioning and deprecation policy

varbanv:
**Description**

- Document API versioning and deprecation information

**Related issue(s)**
https://github.com/kyma-project/kyma/issues/14737


docs/05-technical-reference/00-api-overview/README.md
```diff
@@ -0,0 +1,90 @@
...
+**Rule #3: An API version in a given track may not be deprecated in favor of a less stable API version.**
+
+  * GA API versions can replace beta and alpha API versions.
+  * Beta API versions can replace earlier beta and alpha API versions, but *may not* replace GA API versions.
+  * Alpha API versions can replace earlier alpha API versions, but *may not* replace GA or beta API versions.
+
+**Rule #4a: minimum API lifetime is determined by the API stability level**
+
+   * **GA API versions may be marked as deprecated, but must not be removed within a major version of a Kyma module**
+   * **Beta API versions must be supported for 9 months or 3 releases (whichever is longer) after deprecation**
```
pbochynski, line 69:
```suggestion
   * **Beta API versions must be supported for 6 months or 3 releases (whichever is longer) after deprecation**
```

pbochynski (changes requested):
Majority of the content is copied from https://kubernetes.io/docs/reference/using-api/deprecation-policy
Please make it a quote, provide the source (or maybe just link it instead).

pbochynski (dismissed):

docs/05-technical-reference/00-api-overview/README.md
```diff
@@ -0,0 +1,92 @@
...
+
+### REST resources (aka API objects)
+
+Consider a hypothetical REST resource named Widget, which was present in API v1 in the above timeline, and which needs to be deprecated. We document and announce the deprecation in sync with release X+1. The Widget resource still exists in API version v1 (deprecated) but not in v2alpha1. The Widget resource continues to exist and function in releases up to and including X+8. Only in release X+9, when API v1 has aged out, does the Widget resource cease to exist, and the behavior get removed.
+
+### Fields of REST resources
+
+As with whole REST resources, an individual field which was present in API v1 must exist and function until API v1 is removed.  Unlike whole resources, the v2 APIs may choose a different representation for the field, as long as it can be round-tripped. For example a v1 field named "magnitude" which was deprecated might be named "deprecatedMagnitude" in API v2. When v1 is eventually removed, the deprecated field can be removed from v2.
```
IwonaLanger, line 84: The phrase "whole REST resources" is used a few times and I'd like to clarify if the meaning is all of REST resources meaning each of them, or each of them as a whole - shall we say - unit.

varbanv: It should be the second - so you can deprecate the whole REST resource, or just one or more fields.

docs/05-technical-reference/00-api-overview/README.md
```diff
@@ -0,0 +1,92 @@
...
+
+These rules are enforced between official releases, not between arbitrary commits to main or release branches.
+
+**Rule #1: API elements may only be removed by incrementing the version of the API group.**
+
+Once an API element has been added to an API group at a particular version, it cannot be removed from that version or have its behavior significantly changed, regardless of the track.
+
+**Rule #2: API objects must be able to round-trip between API versions in a given release without information loss, with the exception of whole REST resources that do not exist in some versions.**
+
+For example, an object can be written as v1 and then read back as v2 and converted to v1, and the resulting v1 resource will be identical to the original.  The representation in v2 might be different from v1, but the system knows how to convert between them in both directions.  Additionally, any new field added in v2 must be able to round-trip to v1 and back, which means v1 might have to add an equivalent field or represent it as an annotation.
```
IwonaLanger, line 60:
```suggestion
For example, an object can be written as v1 and then read back as v2 and converted to v1, and the resulting v1 resource is identical to the original.  The representation in v2 might be different from v1, but the system knows how to convert between them in both directions.  Additionally, any new field added in v2 must be able to round-trip to v1 and back, which means v1 might have to add an equivalent field or represent it as an annotation.
```

docs/05-technical-reference/00-api-overview/README.md
```diff
@@ -0,0 +1,92 @@
...
+  * Beta API versions can replace earlier beta and alpha API versions, but *may not* replace GA API versions.
+  * Alpha API versions can replace earlier alpha API versions, but *may not* replace GA or beta API versions.
+
+**Rule #4a: minimum API lifetime is determined by the API stability level.**
+
+   * **GA API versions may be marked as deprecated, but must not be removed within a major version of a Kyma module**
+   * **Beta API versions must be supported for 6 months or 3 releases (whichever is longer) after deprecation**
+   * **Alpha API versions may be removed in any release without prior deprecation notice**
+
+**Rule #4b: The `preferred` API version and the `storage version` for a given group may not advance until after a release has been made that supports both the new version and the previous version**
```
IwonaLanger, line 74:
```suggestion
**Rule #4b: The `preferred` API version and the `storage version` for a given group may not advance until after a release has been made that supports both the new version and the previous version.**
```

IwonaLanger (approved):

Issue #13882, opened 2022-04-07
[Telemetry Operator] Files can breaks pipeline isolation and lead to missing file content


skhalash:
**Description**

Adding and removing a pipeline which is sharing the same name of a file as an existing pipeline, should not bring the Fluent Bit in an error situation. Currently, it is not the case. Say, we add a `loki` pipeline, then add a `mockserver` pipeline and delete the `mockserver` pipeline. Both pipelines define a file named "labelmap.json".

**Expected result**

The `mockserver` pipeline is not applied anymore. The `loki` pipeline remains applied. Fluent Bit is running.

**Actual result**

Fluent Bit is crashing:
```
Fluent Bit v1.8.15
* Copyright (C) 2015-2021 The Fluent Bit Authors
* Fluent Bit is a CNCF sub-project under the umbrella of Fluentd
* https://fluentbit.io

[2022/04/07 14:42:14] [Warning] [config] I cannot open /fluent-bit/etc/parsers.conf file
[2022/04/07 14:42:14] [Warning] [config] I cannot open /fluent-bit/etc/dynamic-parsers/parsers.conf file
[2022/04/07 14:42:14] [ info] [engine] started (pid=1)
[2022/04/07 14:42:14] [ info] [storage] version=1.1.6, initializing...
[2022/04/07 14:42:14] [ info] [storage] in-memory
[2022/04/07 14:42:14] [ info] [storage] normal synchronization mode, checksum disabled, max_chunks_up=128
[2022/04/07 14:42:14] [ info] [cmetrics] version=0.2.2
[2022/04/07 14:42:14] [ info] [input:tail:tail.0] multiline core started
[2022/04/07 14:42:14] [ info] [filter:kubernetes:kubernetes.0] https=1 host=kubernetes.default.svc port=443
[2022/04/07 14:42:14] [ info] [filter:kubernetes:kubernetes.0] local POD info OK
[2022/04/07 14:42:14] [ info] [filter:kubernetes:kubernetes.0] testing connectivity with API server...
[2022/04/07 14:42:14] [ info] [filter:kubernetes:kubernetes.0] connectivity OK
level=error caller=out_grafana_loki.go:49 [flb-go]="failed to launch" error="failed to open LabelMap file: open /files/labelmap.json: no such file or directory"
[2022/04/07 14:42:14] [error] [go proxy]: plugin 'grafana-loki' failed to initialize
[2022/04/07 14:42:14] [error] [lib] backend failed
[...]

PR #14993, opened 2022-08-03
Validate file name

rakesh-garimella:
**Description**

- Allow only unique file names.
**Related issue(s)**

https://github.com/kyma-project/kyma/issues/13882


components/telemetry-operator/internal/webhook/logpipeline_webhook.go
```diff
@@ -73,7 +74,8 @@ func NewLogPipeLineValidator(
 	outputValidator validation.OutputValidator,
 	pipelineConfig fluentbit.PipelineConfig,
 	fsWrapper fs.Wrapper,
-	restartsTotal prometheus.Counter) *LogPipelineValidator {
+	restartsTotal prometheus.Counter,
+	fileValidator validation.FilesValidator) *LogPipelineValidator {
```
chrkl, line 78: Let's change the order or the arguments to have all the validators consecutively. Maybe it would also make sense to wrap all validators into some struct to shorten the argument list a bit.

chrkl (changes requested):

chrkl (approved):

PR #15024, opened 2022-08-05
Stub compas runetime agent tests

mvshao:
**Description**

- ...
- ...
- ...

**Related issue(s)**


mvshao:
There are two things we need to think about:
- what about target naming in the Makefile? For Gateway we have target `test-gateway` which is the shortened name of the component, but for Compass, the pipeline expects the full name of the component as a Makefile target e.g. `test-compass-runtime-agent`. It would be good for the naming convention to be standardised across all tests. 
- the script `check-pod-log.sh` is responsible for reading the logs from the pod and only works for one type of test (Gateway or Compass) at a time. We need to develop a solution that is universal for each test type. Maybe passing a parameter to the script that tells it from which pod to retrieve the logs would be a good idea.

akgalwas (approved):

PR #15054, opened 2022-08-09
make MergeParsersConfig permutation invariant

dennis-ge:
**Description**

- With the current implementation, deploying multiple Logparser CRs results in endless restarts of the telemetry fluent bit pod.
- The logic in the Logparser controller/reconciler loop is the reason:
    - Assuming that there are 5 CRs, the controller creates a ConfigMap entry with the content of 1 CR and then adds the content of the 4 other CRs to the ConfigMap entry
    - Since the status of each CR is still `Pending`, it is requeued and processed again by the controller. It is processed again until the status is `Running`.
    - Next, each CR creates a new parser config using the `MergeParsersConfig` function and reads the existing parser config from the ConfigMap entry. It then compares both for equality. If the parser configs are equal, the config didn't change, the CR status changes to `Running` and it is not processed again. If they are not equal, the CR is processed again.
    - Here comes the problem: The generated parser config of the `MergeParsersConfig` function always was different from the ConfigMap entry.  The `MergeParsersConfig` function retrieves  all Logparsers CRs, iterates over them and generates the config. However, the order of the CRs, i.e. `[CR1, CR2, CR3, CR4, CR5]` and `[CR2, CR1, CR3, CR4, CR5]`, resulted in different configs.
    - As a result, each CR recreated the ConfigMap entry, which is why the configs never reached a final state where they configs were the same for all CRs. This is the cause for the endless controller loop.
 - This PR changes the `MergeParsersConfig` function to be permutation invariant, i.e. `MergeParsersConfig([CR1, CR2, CR3, CR4, CR5]) == MergeParsersConfig([CR2, CR1, CR3, CR4, CR5])` for all possible permutations by ordering the CRs beforehand

**Related issue(s)**

### DIFF
--- a/components/telemetry-operator/internal/fluentbit/config_builder.go
+++ b/components/telemetry-operator/internal/fluentbit/config_builder.go
@@ -240,15 +240,25 @@ func generateMatchCondition(logPipelineName string) string {
 
 // MergeParsersConfig merges Fluent Bit parsers to a single Fluent Bit configuration.
 func MergeParsersConfig(logParsers *telemetryv1alpha1.LogParserList) string {
-	var sb strings.Builder
+	logParsersMap := make(map[string]string, len(logParsers.Items))
+	logParsersNames := make([]string, 0, len(logParsers.Items))
+
 	for _, logParser := range logParsers.Items {
-		var parser string
 		if logParser.DeletionTimestamp == nil {
-			name := fmt.Sprintf("Name %s", logParser.Name)
-			parser = fmt.Sprintf("%s\n%s", logParser.Spec.Parser, name)
-			sb.WriteString(BuildConfigSection(ParserConfigHeader, parser))
+			logParsersMap[logParser.Name] = logParser.Spec.Parser
+			logParsersNames = append(logParsersNames, logParser.Name)
 		}
 	}
+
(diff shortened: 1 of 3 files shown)


components/telemetry-operator/internal/fluentbit/config_builder.go
```diff
@@ -240,15 +240,25 @@ func generateMatchCondition(logPipelineName string) string {
...
 		if logParser.DeletionTimestamp == nil {
-			name := fmt.Sprintf("Name %s", logParser.Name)
-			parser = fmt.Sprintf("%s\n%s", logParser.Spec.Parser, name)
-			sb.WriteString(BuildConfigSection(ParserConfigHeader, parser))
+			logParsersMap[logParser.Name] = logParser.Spec.Parser
+			logParsersNames = append(logParsersNames, logParser.Name)
 		}
 	}
+
+	sort.Strings(logParsersNames)
```
chrkl, line 253: Maybe rather use sort.Slice on logParsers.Items.

dennis-ge: yes, makes the code way more readable :)

chrkl (approved):

PR #15055, opened 2022-08-09
Update nats charts to version 0.17.3

VladislavPaskar:
**Description**

the purpose of this PR is to stay up-to-date with the [upstream repository](https://github.com/nats-io/k8s/tree/main/helm/charts/nats) charts version 

- Changes in the ConfigMap:
   - new `lameDuckGracePeriod` and set it to the default, [reference](https://docs.nats.io/running-a-nats-service/configuration) 
- Changes in the StatefulSet:
   - `podManagementPolicy` is now set to `Parallel` in the upstream charts per default, [reference](https://kubernetes.io/docs/concepts/workloads/controllers/statefulset/#pod-management-policies). It used to be `OrderedReady` in the past, we still need to decide, which one to use.
   - new properties in the `livenessProbe`, which are for now set to the default upstream values
   - new startupProbe check
   - removed the readiness check, since it is covered by both liveness and startup checks, [reference](https://github.com/nats-io/k8s/blob/main/helm/charts/nats/values.yaml#L61)
- Changes in the `values.yaml`:
   - changed the `lameDuckDuration` and the `terminationGracePeriodSeconds` to the new default values
   - new `configChecksumAnnotation` value, which causes the StatefulSet to roll when the ConfigMap is updated, [reference](https://renehernandez.io/notes/rolling-pods-config-changes/)
   - new `uniqueTag` value, which prevents placing a stream in the same availability zone twice when set

### DIFF
--- a/resources/eventing/charts/nats/README.md
+++ b/resources/eventing/charts/nats/README.md
@@ -23,24 +23,12 @@ $ kubectl -n nats port-forward nats-1 4222
 
 ## Configuration
 
-### Server Image
-
-```yaml
-nats:
-  image: nats:2.8.4-alpine
-  pullPolicy: IfNotPresent
-```
-
 ### Limits
 
 ```yaml
 nats:
   # The number of connect attempts against discovered routes.
   connectRetries: 30
-
-  # How many seconds should pass before sending a PING
-  # to a client that has no activity.
-  pingInterval:
 
   # Server settings.
   limits:
@@ -81,12 +69,6 @@ https://docs.nats.io/nats-server/configuration/clustering#nats-server-clustering
 cluster:
   enabled: true
   replicas: 3
-```
-
-Example:
-
-```sh
-$ helm install nats nats/nats --set cluster.enabled=true
 ```
 
 ## JetStream
--- a/resources/eventing/charts/nats/README.md
+++ b/resources/eventing/charts/nats/README.md
@@ -23,6 +23,14 @@ $ kubectl -n nats port-forward nats-1 4222
 
 ## Configuration
 
+### Server Image
+
+```yaml
+nats:
+  image: nats:2.8.4-alpine
+  pullPolicy: IfNotPresent
+```
+
 ### Limits
 
 ```yaml
@@ -32,19 +40,19 @@ nats:
 
   # How many seconds should pass before sending a PING
   # to a client that has no activity.
-  pingInterval: 
+  pingInterval:
 
   # Server settings.
   limits:
-    maxConnections: 
-    maxSubscriptions: 
-    maxControlLine: 
-    maxPayload: 
+    maxConnections:
+    maxSubscriptions:
+    maxControlLine:
+    maxPayload:
 
-    writeDeadline: 
-    maxPending: 
-    maxPings: 
-    lameDuckDuration: 
+    writeDeadline:
+    maxPending:
+    maxPings:
+    lameDuckDuration:
 
   # Number of seconds to wait for client connections to end after the pod termination is requested
   terminationGracePeriodSeconds: 60
@@ -52,29 +60,40 @@ nats:
 
 ### Logging
 
-> **NOTE**: It is not recommended to enable trace or debug in production, since enabling it will significantly degrade performance.
+*Note*: It is not recommended to enable trace or debug in production since enabling it will significantly degrade performance.
 
 ```yaml
 nats:
   logging:
-    debug: 
-    trace: 
-    logtime: 
-    connectErrorReports: 
-    reconnectErrorReports: 
+    debug:
+    trace:
+    logtime:
+    connectErrorReports:
+    reconnectErrorReports:
 ```
-### Clustering
 
-If clustering is enabled, then a 3-node cluster will be set up. You can find more information in the [NATS documentation](https://docs.nats.io/running-a-nats-service/introduction/running/nats-kubernetes/helm-charts#clustering).
+## Clustering
+
+If clustering is enabled, then a 3-node cluster will be setup. More info at:
+https://docs.nats.io/nats-server/configuration/clustering#nats-server-clustering
 
 ```yaml
 cluster:
   enabled: true
   replicas: 3
 ```
-### JetStream
 
-To set up memory and file storage with JetStream, use this configuration: 
+Example:
+
+```sh
+$ helm install nats nats/nats --set cluster.enabled=true
+```
+
+## JetStream
+
+### Setting up Memory and File Storage
+
+File Storage is **always** recommended, since JetStream's RAFT Meta Group will be persisted to file storage.  The Storage Class used should be block storage.  NFS is not recommended.
 
 ```yaml
 nats:
--- a/resources/eventing/charts/nats/templates/configmap.yaml
+++ b/resources/eventing/charts/nats/templates/configmap.yaml
@@ -8,8 +8,22 @@ metadata:
     {{- include "nats.labels" . | nindent 4 }}
 data:
   nats.conf: |
+    # NATS Clients Port
+    port: {{ .Values.nats.ports.client }}
+
     # PID file shared with configuration reloader.
     pid_file: "/var/run/nats/nats.pid"
+
+    {{- if .Values.nats.config }}
+    ###########
+    #         #
+    # Imports #
+    #         #
+    ###########
+    {{- range .Values.nats.config }}
+    include ./{{ .name }}/{{ .name }}.conf
+    {{- end}}
+    {{- end }}
 
     ###############
     #             #
@@ -17,9 +31,17 @@ data:
     #             #
     ###############
     http: 8222
-    server_name: $POD_NAME
+    server_name: {{- if .Values.nats.serverNamePrefix  }}$SERVER_NAME{{- else }}$POD_NAME{{- end }}
 
-    {{- if .Values.global.jetstream.enabled }}
+    {{- if .Values.nats.serverTags }}
+    server_tags: [
+      {{- range .Values.nats.serverTags }}
+        "{{ . }}",
+      {{- end }}
+    ]
+    {{- end }}
+
+    {{- if .Values.global.jetstream.enabled  }}
     ###################################
     #                                 #
     # NATS JetStream                  #
@@ -52,6 +74,10 @@ data:
       {{- .Values.nats.jetstream.fileStorage.size }}
       {{- end }}
       {{- end }}
+
+      {{- if .Values.nats.jetstream.uniqueTag }}
+      unique_tag: {{ .Values.nats.jetstream.uniqueTag }}
+      {{- end }}
     }
     {{- end }}
 
@@ -78,6 +104,7 @@ data:
 
       routes = [
         {{ include "nats.clusterRoutes" . }}
+        {{ include "nats.extraRoutes" . }}
       ]
       cluster_advertise: $CLUSTER_ADVERTISE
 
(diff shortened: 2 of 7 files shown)


resources/eventing/charts/nats/README.md
```diff
@@ -23,6 +23,14 @@ $ kubectl -n nats port-forward nats-1 4222
 
 ## Configuration
 
+### Server Image
+
+```yaml
+nats:
+  image: nats:2.8.4-alpine
+  pullPolicy: IfNotPresent
+```
```
mfaizanse, line 32: Lets not keep this `Server Image` information in readme, otherwise we have to maintain it everytime we bump the image.

resources/eventing/charts/nats/README.md
```diff
@@ -32,49 +40,60 @@ nats:
...
   enabled: true
   replicas: 3
 ```
-### JetStream
 
-To set up memory and file storage with JetStream, use this configuration: 
+Example:
+
+```sh
+$ helm install nats nats/nats --set cluster.enabled=true
```
mfaizanse, line 89: have you tested this command with our helm chart? if its not working, then remove it from here.

resources/eventing/charts/nats/templates/configmap.yaml
```diff
@@ -141,6 +168,10 @@ data:
...
     {{- end }}
 
+    {{- with .Values.nats.limits.lameDuckGracePeriod }}
+    lame_duck_grace_period: {{ . }}
+    {{- end }}
+
     {{- with .Values.nats.limits.lameDuckDuration }}
     lame_duck_duration: {{ . }}
-    {{- end }}
+    {{- end }}
```
mfaizanse, line 177: put a newline!

resources/eventing/charts/nats/README.md
VladislavPaskar: I wanted to be stay in-sync with the original docs as much as possible, but you are also right. I removed it.

resources/eventing/charts/nats/README.md
VladislavPaskar: removed it.

resources/eventing/charts/nats/values.yaml
```diff
@@ -13,13 +13,83 @@ nats:
...
 
+  # The servers name prefix, must be used for example when we want a NATS cluster
+  # spanning multiple Kubernetes clusters.
+  serverNamePrefix: ""
+
+  # Server Tags
+  serverTags:
+  # - "foo"
+  # - "bar"
+
```
mfaizanse, line 24: If `serverNamePrefix` and `serverTags` are only useful when we want a NATS cluster spanning multiple Kubernetes clusters, then we can remove these configs because it not of any use in our scenario. Let's keep only the stuff in the helm chart which is useful for us.

resources/eventing/charts/nats/values.yaml
```diff
@@ -81,13 +158,35 @@ nats:
...
+  #
+  # e.g:
+  #
+  #  config:
+  #    - name: ssh-key
+  #      secret:
+  #        secretName: ssh-key
+  #    - name: config-vol
+  #      configMap:
+  #        name: log-config
```
mfaizanse, line 179: same, delete this whole section if its not used in out case.

resources/eventing/charts/nats/values.yaml
```diff
@@ -171,22 +271,48 @@ statefulSetPodLabels:
...
 
+# additionalContainers are the sidecar containers to add to the NATS StatefulSet
+additionalContainers: []
+
+# additionalVolumes are the additional volumes to add to the NATS StatefulSet
+additionalVolumes: []
+
+# additionalVolumeMounts are the additional volume mounts to add to the nats-server and nats-server-config-reloader containers
+additionalVolumeMounts: []
+
```
mfaizanse, line 282: same, remove these additionalContainers, additionalVolumes and additionalVolumeMounts, if not used.

resources/eventing/charts/nats/values.yaml
```diff
@@ -171,22 +271,48 @@ statefulSetPodLabels:
...
+# Define if NATS is using FQDN name for clustering (i.e. nats-0.nats.default.svc.cluster.local) or short name (i.e. nats-0.nats.default).
+useFQDN: true
+
 # Add labels to all the deployed resources
 commonLabels: {}
 
+
+# podManagementPolicy controls how pods are created during initial scale up,
+# when replacing pods on nodes, or when scaling down.
+podManagementPolicy: Parallel
```
mfaizanse, line 314: Is `Parallel` better for us or `OrderedReady`?

resources/eventing/charts/nats/templates/service.yaml
```diff
@@ -8,14 +8,13 @@ metadata:
     {{- include "nats.labels" . | nindent 4 }}
   {{- if .Values.serviceAnnotations}}
   annotations:
-  {{- range $key, $value := .Values.serviceAnnotations }}
-    {{ $key }}: {{ $value | quote }}
-  {{- end }}
+    {{- toYaml .Values.serviceAnnotations | nindent 4 }}
```
mfaizanse, line 11: Please verify if this will also generate the same set of annotations as we have right now.

resources/eventing/charts/nats/templates/statefulset.yaml
```diff
@@ -6,11 +6,9 @@ metadata:
   namespace: {{ .Release.Namespace | quote }}
   labels:
     {{- include "nats.labels" . | nindent 4 }}
-    {{- if .Values.statefulSetAnnotations}}
+  {{- if .Values.statefulSetAnnotations }}
   annotations:
-  {{- range $key, $value := .Values.statefulSetAnnotations }}
-    {{ $key }}: {{ $value | quote }}
-  {{- end }}
+    {{- toYaml .Values.statefulSetAnnotations | nindent 4 }}
```
mfaizanse, line 11: Same, Please verify if this will also generate the same set of annotations as we have right now.

resources/eventing/charts/nats/values.yaml
```diff
@@ -13,13 +13,83 @@ nats:
...
+
+      initialDelaySeconds: 10
+      timeoutSeconds: 5
+      periodSeconds: 10
+      successThreshold: 1
+      failureThreshold: 30
+
+  # Adds a hash of the ConfigMap as a pod annotation
+  # This will cause the StatefulSet to roll when the ConfigMap is updated
+  configChecksumAnnotation: true
```
mfaizanse, line 91: Do you what new behaviour will it introduce?

resources/eventing/charts/nats/templates/statefulset.yaml
```diff
@@ -22,48 +20,58 @@ spec:
...
+      {{- end }}
+      {{- with .Values.affinity }}
       affinity:
-{{- tpl (toYaml .) $ | nindent 8 }}
-{{- end }}
+        {{- tpl (toYaml .) $ | nindent 8 }}
+      {{- end }}
+      {{- if .Values.priorityClassName }}
+      priorityClassName: {{ .Values.priorityClassName | quote }}
+      {{- end }}
```
mfaizanse, line 56: I dont find any `priorityClassName` in the Values.yaml. If this is not used in our case, then lets remove it.

resources/eventing/charts/nats/templates/statefulset.yaml
```diff
@@ -206,34 +221,78 @@ spec:
...
+        {{- end }}
+        {{- end }}
+
+        {{- if .Values.nats.healthcheck.startup.enabled }}
+        startupProbe:
+          httpGet:
+            {{- $simpleVersion := .Values.global.images.nats.version | default "latest" | regexFind "\\d+(\\.\\d+)?(\\.\\d+)?" | default "2.7.1"  }}
+            {{- if and .Values.nats.healthcheck.enableHealthz (or (not .Values.nats.healthcheck.detectHealthz) (semverCompare ">=2.7.1" $simpleVersion)) }}
+            # for NATS server versions >=2.7.1, healthz will be enabled to allow for a grace period
+            # in case of JetStream enabled deployments to form quorum and streams to catch up.
```
mfaizanse, line 268: We know that we will be using versions > 2.7.1, therefore we can remove the else part where we cover the versions < 2.7.1.

resources/eventing/charts/nats/templates/statefulset.yaml
```diff
@@ -246,15 +305,11 @@ spec:
         name: {{ include "nats.fullname" . }}-js-pvc
         {{- if .Values.nats.jetstream.fileStorage.annotations }}
         annotations:
-        {{- range $key, $value := .Values.nats.jetstream.fileStorage.annotations }}
-          {{ $key }}: {{ $value | quote }}
-        {{- end }}
+          {{- toYaml .Values.nats.jetstream.fileStorage.annotations | nindent 10 }}
```
mfaizanse, line 308: Please check if the annotations remain the same.

resources/eventing/charts/nats/templates/statefulset.yaml
```diff
@@ -246,15 +305,11 @@ spec:
...
         {{- end }}
       spec:
         accessModes:
-        {{- range .Values.nats.jetstream.fileStorage.accessModes }}
-          - {{ . | quote }}
-        {{- end }}
+          {{- toYaml .Values.nats.jetstream.fileStorage.accessModes | nindent 10 }}
```
mfaizanse, line 312: Please check if the annotations remain the same.

resources/eventing/charts/nats/templates/statefulset.yaml
VladislavPaskar: it is generating the same output

resources/eventing/charts/nats/templates/service.yaml
VladislavPaskar: it is generating the same output

resources/eventing/charts/nats/templates/statefulset.yaml
VladislavPaskar: it is generating the same output

resources/eventing/charts/nats/values.yaml
VladislavPaskar: It is a good question, let's discuss with the team.

VladislavPaskar: let's discuss that with the team

resources/eventing/charts/nats/values.yaml
VladislavPaskar: this one causes the StatefulSet to roll when the ConfigMap is updated

mfaizanse: do we still need config-reloader sidecar then?

resources/eventing/charts/nats/templates/service.yaml
```diff
@@ -51,6 +49,3 @@ spec:
     {{- end }}
   - name: gateways
     port: {{ .Values.nats.ports.gateways }}
-    {{- if .Values.appProtocol.enabled }}
-    appProtocol: tcp
```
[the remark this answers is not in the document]
VladislavPaskar: we need it, it was just doubled :)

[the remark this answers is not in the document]
VladislavPaskar: we do :)

resources/eventing/charts/nats/values.yaml
VladislavPaskar: I think yes. Here is an answer from the NATS Team: "the reloader does not restart the process on updates whereas the annotation does. not everything can be reloaded so like leafnode remotes, so annotation instead can help"

resources/eventing/charts/nats/values.yaml
VladislavPaskar: After a discussion with the team, we decided to go with `Parallel` option

resources/eventing/charts/nats/values.yaml
VladislavPaskar: Let's keep it false right now, as we don't currently have any configMap changes, which will require a rescaling.

mfaizanse (approved):

majakurcius (approved):

Issue #13143, opened 2022-01-21
[Configurable Logging] Prepare release and verify upgrade procedure


a-thaler:
**Description**
When https://github.com/kyma-project/kyma/issues/11236 is ready to be released, several things need to be done in order to get the new component installed by default and the old functionality disabled

**Criterias**
- [x] The new component is enabled by default by all means of installation (CLI, pure reconciler, SAP offering) (PR #15057)
- [x] Feature toggle is disabled by default -> New fluent-bit is pushing logs to loki as loki-logPipeline is active
- [x] Old fluent-bit is not pushing logs to loki anymore (disable output plugin by re-using the disabled feature flag)
- [x] Docu is not describing anymore how to enable the feature (PR #15070)
- [x] Docu is not using "alpha" anymore. The CRD stays alpha for now but the feature itself is regular
- [x] Assure that an upgrade from standard kyma 2.5 cluster works as expected
  - [x] Open Source
  - [x] SKR
- [x] Have release notes ready
- [x] Ensure to have a cleanup script to delete the logging fluent-bit DaemonSet (PR #15092)
- [x] Migration guide for OSS users (PR https://github.com/kyma-project/kyma/pull/15122)

skhalash:
The following acceptance criterium will not be done:

* Old fluent-bit is not pushing logs to Loki anymore (disable output plugin by re-using the disabled feature flag)

In order to achieve this, we would have to replace the old fluent-bit output with null and then, in a subsequent version, disable the subchart rendering. This 2-step process seems to be too much. Anyway, the old fluent bit will push logs during a short time window after the upgrade until the cleanup script is executed, which is an acceptable tradeoff.

PR #15057, opened 2022-08-09
Enable telemetry component

chrkl:
**Description**

Enable the installation of the telemetry component.

- Enable telemetry chart to be installed by default (including Fluent Bit and operator)
- Disable Fluent Bit from logging chart
- Adapt Logging test to check access logs also from telemetry-fluent-bit

**Related issue(s)**

Related issue: #13143

### DIFF
--- a/tests/fast-integration/logging/loki.js
+++ b/tests/fast-integration/logging/loki.js
@@ -50,7 +50,7 @@ function parseJson(str) {
 }
 
 async function verifyIstioAccessLogFormat(startTimestamp) {
-  const query = '{container="istio-proxy",job="fluent-bit",namespace="kyma-system",pod="logging-loki-0"}';
+  const query = '{container="istio-proxy",namespace="kyma-system",pod="logging-loki-0"}';
 
   const responseBody = await queryLoki(query, startTimestamp);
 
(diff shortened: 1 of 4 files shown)


tests/fast-integration/logging/loki.js
```diff
@@ -50,7 +50,7 @@ function parseJson(str) {
 }
 
 async function verifyIstioAccessLogFormat(startTimestamp) {
-  const query = '{container="istio-proxy",job="fluent-bit",namespace="kyma-system",pod="logging-loki-0"}';
```
lindnerby, line 53: telemetry-tests should have even more queries with job="fluent-bit"

chrkl: The telemetry-tests explicitly query for job="telemetry-fluent-bit". Thought about adding the label here as well. Two reasons against that are:
1. We want to test logging in an end-to-end manner and not the telemetry-fluent-bit in this test
2. Having the telemetry-fluent-bit label here would break upgrade tests

lindnerby: Yes, was asking because telemetry component is default now if we even need the explicit query on job in telemetry tests anymore.

chrkl: Removed the job label from the telemetry test.

lindnerby (approved):

tobiscr (approved):

Issue #15063, opened 2022-08-09
[Telemetry operator] telemetry_fluentbit_restarts_total metric is not equal to the actual restart count


dennis-ge:
**Description**

When deploying X LogParser CR's (in my case 16), the Prometheus `telemetry_fluentbit_restarts_total` metric of the telemetry operator says that the Fluent Bit pod got restarted exactly X times. However, the pod only got restarted 2 times.

**Expected result**

The `telemetry_fluentbit_restarts_total` metric should be equal to the number of telemetry fluent bit pod restarts.

**Actual result**

Number of times the pod got restarted:
```shell
$ kubectl describe daemonset telemetry-fluent-bit
...
Events:
  Type    Reason            Age    From                  Message
  ----    ------            ----   ----                  -------
  Normal  SuccessfulCreate  7m34s  daemonset-controller  Created pod: telemetry-fluent-bit-8tq65
  Normal  SuccessfulDelete  7m14s  daemonset-controller  Deleted pod: telemetry-fluent-bit-8tq65
  Normal  SuccessfulCreate  6m57s  daemonset-controller  Created pod: telemetry-fluent-bit-7xq2b
```

Number of times the pod got restarted according to the metric:
```
$ kubectl port-forward svc/telemetry-operator-metrics 8080:8080 &
$ http http://localhost:8080/metrics | grep telemetry_fluentbit_restarts_total
# HELP telemetry_fluentbit_restarts_total Number of triggered Fluent Bit restarts
# TYPE telemetry_fluentbit_restarts_total counter
telemetry_fluentbit_restarts_total 16
```

**Steps to reproduce**

See above.

**Troubleshooting**

The counter is increased every time a new CR is deployed in the [utils.go](https://github.com/kyma-project/kyma/blob/main/components/telemetry-operator/internal/fluentbit/utils.go#L53), which is why the metric count is the number of CRs. However, I assume that as soon as the apiserver receives multiple pod restart requests, it will bundle them together and restart only the necessary amount of time.

dennis-ge:
The metric should not be equal to the number of total fluent bit restarts but should show the total number the telemetry operator triggered restarting fluent bit. Therefore, the metric is renamed to `telemetry_fluentbit_triggered_restarts_total`

PR #15066, opened 2022-08-10
rename fluent bit restarts total metric

dennis-ge:
**Description**

- The metric should not be equal to the number of total fluent bit restarts but should show the total number the telemetry operator triggered restarting fluent bit.
- Therefore the metric is renamed to `telemetry_fluentbit_triggered_restarts_total`

**Related issue(s)**
Fixes https://github.com/kyma-project/kyma/issues/15063


components/telemetry-operator/main.go
```diff
@@ -148,7 +148,7 @@ func main() {
 	}
 
 	restartsTotal := prometheus.NewCounter(prometheus.CounterOpts{
-		Name: "telemetry_fluentbit_restarts_total",
+		Name: "telemetry_fluentbit_triggered_restarts_total",
```
chrkl, line 151: There should also be a ServerMonitor that might need a change.

dennis-ge: Fixed :)

chrkl (approved):

chrkl (approved):

Issue #15039, opened 2022-08-08
[Telemetry Operator] Clean up dryrun package


skhalash:
**Description**

Having a meaningful encapsulation of the Fluent Bit dry-run functionality to allow further refactoring.

**Reasons**

**Attachments**

PR #15079, opened 2022-08-11
Telemetry Operator - Refactor dry-run pkg

lindnerby:
**Description**

- Move fluentbit utils for preparing dry-run workspace to the respective webhook dry-runs
- Refactor internals of preparing the workspace
- Fix parser dry-run since no parsers were fetched
- Establish first tests for webhook dry-runs

**Related issue(s)**
* https://github.com/kyma-project/kyma/issues/15039


components/telemetry-operator/internal/webhook/dryrun/file_writer.go
```diff
@@ -0,0 +1,164 @@
...
+type fileWriterImpl struct {
+	client client.Client
+	config *Config
+}
+
+func (f *fileWriterImpl) prepareParserDryRun(ctx context.Context, workDir string, parser *telemetryv1alpha1.LogParser) (func(), error) {
+	if err := makeDir(workDir); err != nil {
+		return nil, err
+	}
+	if err := f.writeConfig(ctx, workDir); err != nil {
```
dennis-ge, line 30: Is the config even necessary when validating the parsers files? Executing `fluent-bit --dry-run --parser=./custom_parsers.conf --parser=./dynamic-parsers/parsers.conf` shows me errors in the parsers files.

components/telemetry-operator/internal/webhook/dryrun/dryrun.go
```diff
@@ -2,122 +2,144 @@ package dryrun
...
-		return fmt.Errorf("error while validating Fluent Bit config: %v", err.Error())
+		return err
 	}
-	return nil
+	args = append(args, externalPluginArgs...)
+
+	return d.runCmd(ctx, args)
+}
+
+func (d *DryRunner) newWorkDirPath() string {
```
dennis-ge, line 76: could also be a free function, but not critical

components/telemetry-operator/internal/webhook/dryrun/dryrun.go
```diff
@@ -2,122 +2,144 @@ package dryrun
...
-	outBytes, err := cmd.CombinedOutput()
+func (d *DryRunner) runCmd(ctx context.Context, args []string) error {
+	outBytes, err := d.commandRunner.run(ctx, d.config.FluentBitBinPath, args...)
 	out := string(outBytes)
-	return out, err
+	if err != nil {
+		if strings.Contains(out, "error") || strings.Contains(out, "Error") {
+			return fmt.Errorf("failed to validate the supplied configuration: %s", extractError(out))
+		}
+		return fmt.Errorf("failed to execute dryrun: %v", err.Error())
```
dennis-ge, line 112: This error message is shown to the user right, so I don't know if mentioning `dryrun` is the best idea since we don't describe how we validate the config

components/telemetry-operator/internal/webhook/dryrun/file_writer_test.go
```diff
@@ -0,0 +1,94 @@
...
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/serializer"
+	"k8s.io/apimachinery/pkg/types"
+	clientgoscheme "k8s.io/client-go/kubernetes/scheme"
+	"sigs.k8s.io/controller-runtime/pkg/client/fake"
+)
+
+func mustLoadManifest[T runtime.Object](scheme *runtime.Scheme, filepath string) T {
```
dennis-ge, line 20: nice :D

components/telemetry-operator/internal/webhook/logparser/webhook.go
```diff
@@ -20,69 +20,35 @@ import (
...
-	"k8s.io/apimachinery/pkg/types"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	logf "sigs.k8s.io/controller-runtime/pkg/log"
 	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
 )
 
 //go:generate mockery --name DryRunner --filename dryrun.go
 type DryRunner interface {
-	RunParser(ctx context.Context, configFilePath string) error
+	DryRunParser(ctx context.Context, parser *telemetryv1alpha1.LogParser) error
```
dennis-ge, line 36: Unsure about the interface/method naming. It is in some way repetetive with the `DryRun` at the beginning. Thinking about something like `ValidateParser`/`ValidatePipeline` as method names because in the error messages we also say `could not validate ... `. This then resembles more to the the previous implementation, but it is just an idea, also am fine with the current naming.

dennis-ge (commented):
really nice work 🎉 definitely improves the code! Especially with the `file_writer` it is now visible how we map the given logpipeline/logparser config to the fb config 👍 Have some small comments, but nothing critical. Am ready to approve as soon as the tests pass

components/telemetry-operator/internal/webhook/dryrun/file_writer.go
lindnerby: Nope, got removed while you were reviewing 😁

components/telemetry-operator/internal/webhook/logparser/webhook.go
lindnerby: Could shorten it, since it's already the dryrunner that provides this `RunParser` and `RunPipeline` or something similar should suffice

components/telemetry-operator/internal/webhook/logparser/mocks/dryrun.go
```diff
@@ -1,25 +1,27 @@
...
 // DryRunner is an autogenerated mock type for the DryRunner type
 type DryRunner struct {
 	mock.Mock
 }
 
-// RunParser provides a mock function with given fields: ctx, configFilePath
-func (_m *DryRunner) RunParser(ctx context.Context, configFilePath string) error {
-	ret := _m.Called(ctx, configFilePath)
+// DryRunParser provides a mock function with given fields: ctx, parser
+func (_m *DryRunner) DryRunParser(ctx context.Context, parser *v1alpha1.LogParser) error {
```
chrkl: We have the same mock in the logparser and logpipeline package. Is that intended? I'm wondering if we should move it to dryrun/mocks.

components/telemetry-operator/internal/webhook/dryrun/dryrun.go
lindnerby: Give me something more creative, I'll use that :) Just wanted to convey that the error is not because of failed validation or errors from the binary, but rather the command execution itself is failing.

chrkl (changes requested):

components/telemetry-operator/internal/webhook/logparser/mocks/dryrun.go
lindnerby: Good catch 😅 forgot to run mockery. The interface has been split and is provided by the consumer now, in this case each webhook. Our DryRunner provides the implementations for both interfaces.

components/telemetry-operator/internal/webhook/dryrun/dryrun.go
dennis-ge: maybe `could not validate the supplied configuration: %v`

components/telemetry-operator/internal/webhook/logparser/mocks/dryrun.go
skhalash: Due to the fact that interfaces in go are implemented implicitly, usually, you would implement packages in a way that concrete structs are returned: https://github.com/golang/go/wiki/CodeReviewComments#interfaces. So interfaces are usually defined by consumers, not producers. In this case, for example, the `DryRunner` has 2 methods. But every webhook only needs one of them, so it can define an interface that only consists of what is needed. This approach allows you to write more loosely-coupled packages.

I believe, we should use the same approach in other packages as well.

components/telemetry-operator/internal/webhook/dryrun/dryrun.go
skhalash: @lindnerby Remember that constant we found weird? Maybe we should revert it back 🤣

skhalash: I just tested it. So if dryrun returns an error, it will print the first error: `failed to validate the supplied configuration: %s`. The second error is only printed if the fluent-bit binary returns a non-zero code and no error in the console output. It should not really happen, but there might be a bug in fluent-bit itself. I'd say, in this case we should not print the low-level error at all.

skhalash: @dennis-ge @lindnerby Please check. Also changed the wording to make it consistent with the rest of webhook errors.

dennis-ge (approved):

chrkl (approved):

Issue #14152, opened 2022-05-03
Kyma subscription and NATS infra are not in sync


raypinto:
**Description**

Kyma subscription and NATS infra are not in sync when we have a subscription that is not ready.

**Expected result**

Kyma subscription and NATS subscription are in sync

**Actual result**

Kyma subscription and NATS subscription are not in sync

**Steps to reproduce**

- Install kyma
- Create a subscription with 3 filters (3 subscriptions are created on NATS. You can verify it with consumers on JetStream using nats CLI)
- Edit the subscription - Delete 2 filters and have an invalid sink name
- We have only one cleanEventType but the subscription was not sync'd with the NATS infrastructure (as we still have 3 consumers instead of 1)

**Troubleshooting**

The reason for this is we do the sinkValidation before syncing the subscription with the NATS infra. We should do it regardless of an invalid sink.

raypinto:
Although [this bug](https://github.com/kyma-project/kyma/issues/14154) was resolved, the problem of syncing the subscription with the NATS infra is still a problem and needs to be addressed as part of this issue.

If we perform the `step to reproduce` as mentioned [here](https://github.com/kyma-project/kyma/issues/14154), when we change the filter value in the spec to an invalid event Type value or invalid prefix, we see that the cleanEventTypes are empty. But the subscriptions (or consumers) on the server are still present.

nachtmaar:
@raypinto I am a bit confused why you are talking about `cleanEventTypes` since they are defined as the `filter's event types after cleanup for use with the configured backend`, hence not important in our bug ? 
Isn't the only question whether the NATS subscriptions are created or not ? 

**From my understanding:**
- Each filter in the Subscription results in one NATS subscription
  - All NATS subscriptions share the sink from the eventing Subscription
 
**Case 1**: Should we delete all the 3 existing NATS subscriptions on the infra ? 
- Since the filters are defined on the eventing Subscription they should exist on NATS as well. However updating them with the invalid sink does not make sense I think 🤔 

**Case 2**: Should we leave the Subscription to contain the 3 valid subscriptions that are still present on the infra but not create the fourth invalid filter as a NATS subscription ? 
- I would say so. Creating a NATS subscription with an invalid sink does not make any sense 🤔

nachtmaar:
[...]

PR #15089, opened 2022-08-11
Jetstream Controller-Integration Test improvements

nachtmaar:
**Description**

Changes in the PR:

- This PR introduces the testing changes introduced in https://github.com/kyma-project/kyma/pull/14742 since the PR is on hold for now.

This PR also adds **several improvement**s to the controller integration **testing** in general:
- Make testing more **resilient** by using `Eventually`
- **Print a description** then a gingko matcher fails
- Add the support to **check for NATS subscriptions** with every possible event-type and not only hardcoded to order.created.v1 event type
- **Add logs** to the table tests to make it easier to understand in which step a matcher failed
- Fixes the problem `test executed panic(nil) or runtime.Goexit: subtest may have called FailNow on a parent test`

**Related issue(s)**

See Also https://github.com/kyma-project/kyma/issues/14152


components/eventing-controller/controllers/subscription/jetstream/reconciler_test.go
```diff
@@ -269,22 +271,44 @@ func TestCreateSubscription(t *testing.T) {
...
+			for eventType, matchers := range tc.want.NatsSubscriptions {
+				testSubscriptionOnNATS(ens, subscription, eventType, matchers...)
+			}
+
+			t.Log("testing the deletion of the subscription")
+			testSubscriptionDeletion(ens, subscription)
+
+			t.Log("testing the deletion of the NATS subscription(s)")
+			for _, filter := range subscription.Spec.Filter.Filters {
+				ensureNATSSubscriptionIsDeleted(ens, subscription, filter.EventType.Value)
```
VladislavPaskar, line 310: I see the `ensureNATSSubscriptionIsDeleted` function is always called after the `testSubscriptionDeletion` does it make sense to either merge them, or call the `ensureNATSSubscriptionIsDeleted` inside of `testSubscriptionDeletion` ?

components/eventing-controller/controllers/subscription/jetstream/reconciler_test.go
```diff
@@ -505,24 +587,53 @@ func TestChangeSubscription(t *testing.T) {
 
 	for _, tc := range testCases {
 		t.Run(tc.name, func(t *testing.T) {
+			// The new t instance needs to be used for making assertions. If ignored, errors like this will be printed:
```
VladislavPaskar, line 590: (nit)

do we need this pretty-long comment in every test ?

components/eventing-controller/controllers/subscription/testing/test_utils.go
```diff
@@ -182,17 +199,19 @@ func createSubscriberSvcInK8s(ens *TestEnsemble) {
 	if ens.SubscriberSvc.Namespace != "default " {
 		namespace := fixtureNamespace(ens.SubscriberSvc.Namespace)
 		if namespace.Name != "default" {
-			err := ens.K8sClient.Create(ens.Ctx, namespace)
-			if !k8serrors.IsAlreadyExists(err) {
-				fmt.Println(err)
-				g.Expect(err).ShouldNot(gomega.HaveOccurred())
-			}
+			g.Eventually(func() error {
```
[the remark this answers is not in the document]
nachtmaar: Sure :)

components/eventing-controller/controllers/subscription/jetstream/reconciler_test.go
nachtmaar: Good question. I think just by reading the test code it is not obvious why this code exists:

```
			g = gomega.NewGomegaWithT(t)
			ens.G = g
```

What about we remove the second line of the comment ? 

Another idea would be to add more details to the testing guides and then quickly reference it only

nachtmaar: So far we have 2 occurrences. I think let's leave it as it is until we have the problem in more tests. Then we can still reduce the comments or move it to the testing guidelines. What do you think ? :)

components/eventing-controller/controllers/subscription/testing/test_utils.go
```diff
@@ -182,17 +199,19 @@ func createSubscriberSvcInK8s(ens *TestEnsemble) {
...
+				if err := ens.K8sClient.Create(ens.Ctx, namespace); !k8serrors.IsAlreadyExists(err) {
+					return err
+				}
+				return nil
+			}, SmallTimeout, SmallPollingInterval).ShouldNot(gomega.HaveOccurred(), "unable to create the namespace for the subscriber")
```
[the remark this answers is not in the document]
nachtmaar: This is how it is rendered by gomega. I decided to go with the upper case version because `Expected` also uses upper case :)
Screenshot 2022-08-17 at 17 10 11

components/eventing-controller/controllers/subscription/jetstream/reconciler_test.go
VladislavPaskar: 2 so far are fine, let's think about it if it gets 2+ :)

VladislavPaskar (approved):

Issue #15037, opened 2022-08-08
New Application Connectivity tests - Test suite for Compass Runtime Agent


akgalwas:
**Description**

We need an integration test of Compass Runtime Agent. The test should be implemented **with a similar structure** to [the Application Gateway test](https://github.com/kyma-project/kyma/pull/14743). It should be fast, simple, and easy to maintain. If something breaks on some end-to-end pipeline or some environment, this test should clearly tell if there is a problem with the Validator or not

 **Acceptance criteria**
 
- [x] A test checking Compass Runtime Agent functionality is implemented 
- [x] The test takes minimal possible execution time
- [x] The test is considered as easy to maintain and has a similar structure to the Application Gateway tests
- [x] The test runs on the Prow pipeline
- [x] The test is runnable locally/ on a custom cluster

**Attachments**

epic issue - https://github.com/kyma-project/kyma/issues/13977

akgalwas:
Work plan:
- [ ] Prepare and merge initial PR with project structure. It should contain: Dockerfile, Makefile and minimal code changes that could be run on a Prow pipeline.
- [ ] Modify test pipeline to install Compass Runtime Agent 
- [ ] Implement testcase for creating Applications
- [ ] Implement testcase for updating Applications
- [ ] Implement testcase for deleting Applications

franpog859:
Please, remember to use the new Compass Director API in regard to the scenarios

franpog859:
Please, create a ticket for the additional follow-up that covers all the possible Compass Runtime Agent cases, that covers ADDITIONAL cases in regards to the manual tests. Certs renewal and stuff like this

PR #15101, opened 2022-08-12
Compass Runtime Agent tests - application CRD and Secret comparer

mvshao:
**Description**
Compass Runtime Agent tests - application CRD comparer

- create the tests and logic for comparing the Application CRD

**Related issue(s)**
#15037


tests/components/application-connector/test/compass-runtime-agent/application.go
```diff
@@ -0,0 +1,156 @@
...
+
+import (
+	"github.com/kyma-project/kyma/components/application-operator/pkg/apis/applicationconnector/v1alpha1"
+	"reflect"
+	"testing"
+)
+
+const actualSecretNamespace string = "kyma-integration"
+
+type SecretComparer interface {
```
koala7659, line 11: All datatypes and declarations not refering to Application (SecretComparer, Secret and Do) shuold be moved to specific file here: secret.go

tests/components/application-connector/test/compass-runtime-agent/secret.go
```diff
@@ -0,0 +1,116 @@
...
+	secret, err := c.getCredentials()
+	if err != nil {
+		return secretData{}, errors.Wrap(err, "Failed to save the secret from cluster")
+	}
+
+	return secret, nil
+}
+
+func SecretDataCompare(actual, expected string, t *testing.T) bool {
+	expectedSecretRepo, _ := initalizeSecretInterface("test") //TODO hardcoded, tried os.Getenv() didn't work
```
koala7659, line 93: We should not initialize interface for secret every time we want to compare secrets. Both interfaces can be initialized once and keep in memory for each use

tests/components/application-connector/test/compass-runtime-agent/secret.go
```diff
@@ -0,0 +1,116 @@
...
+	coreClientset, err := kubernetes.NewForConfig(k8sConfig)
+	if err != nil {
+		return nil, errors.Errorf("failed to create k8s core client, %s", err.Error())
+	}
+
+	return coreClientset.CoreV1().Secrets(namespace), nil
+}
+
+func initalizeSecretInterface(namespace string) (v1.SecretInterface, error) {
+	secretsRepo, err := newSecretsInterface(namespace)
```
koala7659, line 58: IMHO there is no need to have a separate function here. You can combine both into a single one.

koala7659 (changes requested):
Please see my code review comments.

tests/components/application-connector/test/compass-runtime-agent/secret.go
```diff
@@ -0,0 +1,107 @@
...
+	}
+
+	expectedSecret, err := secretExpectedClient.Get()
+	if err != nil {
+		s.t.Log(err, "Failed to get expected secret from cluster")
+		errors.Wrap(err, "Failed to get expected secret from cluster")
+		return false
+	}
+
+	return reflect.DeepEqual(actualSecret, expectedSecret)
```
koala7659, line 106: You don't need to use reflect.DeepEqual to compare simple structs with strings

Please check for example:
https://go.dev/play/p/k5JVVjHq1f0

tests/components/application-connector/test/compass-runtime-agent/secret.go
```diff
@@ -0,0 +1,107 @@
...
+
+type secretData struct {
+	password     string
+	username     string
+	crt          string
+	key          string
+	clientId     string
+	clientSecret string
+}
+
```
koala7659, line 25: You are missing mock type for this type 

//go:generate mockery -name=SecretClient

tests/components/application-connector/test/compass-runtime-agent/secret.go
```diff
@@ -0,0 +1,107 @@
...
+type SecretClient interface {
+	Compare(actual, expected string) bool
+	Get() (secretData, error)
```
koala7659, line 28: The better name instead of Get() would be GetSecretData()

tests/components/application-connector/test/compass-runtime-agent/secret.go
koala7659, line 25: You are missing unit tests for the SecretClient interface and Compare function. 

To implement it you can use fake client to make test secrets . "k8s.io/client-go/kubernetes/fake"

Example:

```
	import "k8s.io/client-go/kubernetes/fake"


         coreV1 := fake.NewSimpleClientset()
	secrets := coreV1.CoreV1().Secrets(namespace)

         createFakeCredentialsSecret(t, secrets, credentials)


func createFakeCredentialsSecret(t *testing.T, secrets core.SecretInterface, credentials credentials) {

	secret := &v1.Secret{
		ObjectMeta: meta.ObjectMeta{
			Name:      secretName,
			Namespace: namespace,
		},
		TypeMeta: meta.TypeMeta{
			Kind:       "Secret",
			APIVersion: "v1",
		},
		Data: map[string][]byte{
			"key1":       []byte("val1"),
			"key2":   []byte("val2"),
			"key3": []byte("val3"),
		},
	}

	_, err := secrets.Create(context.Background(), secret, meta.CreateOptions{})

	require.NoError(t, err)
}

```

koala7659 (changes requested):
Please see my comments

tests/components/application-connector/test/compass-runtime-agent/secret.go
mvshao: Yes, you have good point 👍🏻

Issue #14884, opened 2022-07-22
[Telemetry Operator] Support repeated keys in custom Fluent Bit plugins


a-thaler:
**Description**

Fluent Bit output or filter plugin configurations might expect repeated keys. For instance, the [`header`](https://docs.fluentbit.io/manual/pipeline/outputs/http) attribute of the HTTP output plugin can be set multiple times.

The telemetry-operator stores a configuration section internally in a map. When using repeated keys in a configuration section, oder the last occurrence of a key would be effective.

**Criteria**
* All occurrences of repeated configuration keys in custom plugin definitions need to be supported
* Attributes which are not repeatable should fail validation when multiple entries are provided

PR #15143, opened 2022-08-18
[Telemetry Operator] Support duplicate keys custom plugins

lindnerby:
**Description**

- When parsing custom filter or outputs allow duplicate keys
- Refactor output generator, modularize config building a bit
- Refactor section builder and add tests

**Related issue(s)**
* https://github.com/kyma-project/kyma/issues/14884


components/telemetry-operator/internal/configbuilder/output.go
```diff
@@ -0,0 +1,131 @@
...
+
+	if output.HTTP.Host.IsDefined() {
+		return generateHTTPOutput(&output.HTTP, config.FsBufferLimit, pipeline.Name)
+	}
+
+	if output.Loki.URL.IsDefined() {
+		return generateLokiOutput(&output.Loki, config.FsBufferLimit, pipeline.Name)
+	}
+
+	return ""
```
dennis-ge, line 26: I think we need to return an error here since we require an output. Before we had `fmt.Errorf("output plugin not defined")` in the old `config_builder`. Then `MergeSectionsConfig` also doesn't return `nil` all the time.

components/telemetry-operator/internal/configbuilder/config_builder.go
```diff
@@ -0,0 +1,82 @@
...
+func MergeSectionsConfig(pipeline *telemetryv1alpha1.LogPipeline, pipelineConfig PipelineConfig) (string, error) {
+	var sb strings.Builder
+	sb.WriteString(createRewriteTagFilterSection(pipeline, pipelineConfig))
+	sb.WriteString(createRecordModifierFilter(pipeline))
+	sb.WriteString(createCustomFilters(pipeline))
+	sb.WriteString(createKubernetesMetadataFilter(pipeline))
+	sb.WriteString(createLuaDedotFilter(pipeline))
+	sb.WriteString(createOutputSection(pipeline, pipelineConfig))
+
+	return sb.String(), nil
```
dennis-ge, line 28: This function always returns nil as error.

components/telemetry-operator/internal/secret/secret_helper.go
```diff
@@ -4,7 +4,6 @@ import (
 	"bytes"
 	"context"
 	"fmt"
-	"strings"
```
dennis-ge, line 7: All good with the changes in your PR, but on line 59 we log incorrect information: 
```
log.Error(err, "Failed reading secret '%s' from namespace '%s'", secretKey.Name, secretKey.Namespace)
```

`log.Error` expects an error, error message and then an arbitrary amount of key value pairs. Our current implementation assumes that the function formats the error message, which is not the case.

Just noticed this last time and thought it may be worth fixing it before the release.

components/telemetry-operator/internal/configbuilder/config_builder_test.go
```diff
@@ -0,0 +1,144 @@
...
+		InputTag:          "kube",
+		MemoryBufferLimit: "10M",
+		StorageType:       "filesystem",
+		FsBufferLimit:     "1G",
+	}
+
+	actual, err := MergeSectionsConfig(logPipeline, pipelineConfig)
+	require.NoError(t, err)
+	require.Equal(t, expected, actual)
+}
```
dennis-ge, line 144: There is no test case for `MergeParsersConfig` or `buildConfigSection`, there also was no test case before.

dennis-ge (commented):
Looks really nice, some small comments all concerning the refactoring. If you don't want to apply them within this PR, also fine for me :)

components/telemetry-operator/internal/configbuilder/section_builder.go
```diff
@@ -0,0 +1,75 @@
+package configbuilder
```
skhalash, line 1: I guess, everything here can be written lower-case, the `sectionBuilder` should not stick to the outside

components/telemetry-operator/internal/configbuilder/section_builder.go
```diff
@@ -0,0 +1,75 @@
...
+	}
+	return sb
+}
+
+func (sb *SectionBuilder) Build() string {
+	sort.Slice(sb.params, func(i, j int) bool {
+		if sb.params[i].key != sb.params[j].key {
+			return sb.params[i].key < sb.params[j].key
+		}
+		return sb.params[i].value < sb.params[j].value
```
skhalash, line 62: I think we should slightly adapt the sorting routine. As you can see for example [here](https://docs.fluentbit.io/manual/pipeline/outputs/http#configuration-file), usually you put name and match fields first and the rest of the fields are alphabetically sorted.

skhalash (changes requested):

components/telemetry-operator/internal/configbuilder/config_builder.go
lindnerby: Yes, was on my list and I forgot 😬  Just pushed an update on this, where validation on the custom multiline is done up front. We can decide on, where this has to be done though.

components/telemetry-operator/internal/configbuilder/output.go
lindnerby: See below

components/telemetry-operator/internal/configbuilder/config_builder.go
```diff
@@ -0,0 +1,82 @@
...
+
+type PipelineConfig struct {
+	InputTag          string
+	MemoryBufferLimit string
+	StorageType       string
+	FsBufferLimit     string
+}
+
+// MergeSectionsConfig merges Fluent Bit filters and outputs to a single Fluent Bit configuration.
+func MergeSectionsConfig(pipeline *telemetryv1alpha1.LogPipeline, pipelineConfig PipelineConfig) (string, error) {
```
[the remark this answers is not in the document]
lindnerby: I totally agree. Maybe we can introduce that in a later point in time since I just wanted to change the internals of this function. There is also a second function `MergeParsersConfig` We can introduce renaming or even the API of how to use them when we touch that function as well?

components/telemetry-operator/internal/configbuilder/config_builder_test.go
lindnerby: I agree, would love to do an iteration on `MergeParsersConfig` anyways but not here. Think it's too many changes already.

components/telemetry-operator/internal/configbuilder/config_builder.go
lindnerby: Please have look on this again @skhalash @dennis-ge

components/telemetry-operator/internal/configbuilder/config_builder.go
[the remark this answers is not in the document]
dennis-ge: yep, makes sense 👍

components/telemetry-operator/internal/configbuilder/section_builder.go
lindnerby: Question then: Should also just lowercase everything when we parse it? @dennis-ge @skhalash

components/telemetry-operator/internal/fluentbit/config/builder/output.go
```diff
@@ -0,0 +1,115 @@
...
+
+	if output.HTTP.Host.IsDefined() {
+		return generateHTTPOutput(&output.HTTP, config.FsBufferLimit, pipeline.Name)
+	}
+
+	if output.Loki.URL.IsDefined() {
+		return generateLokiOutput(&output.Loki, config.FsBufferLimit, pipeline.Name)
+	}
+
+	return ""
```
dennis-ge, line 26: What if no output is specified? With the current implementation no error is thrown right?

lindnerby: True. Re-introduced that.
Problem I found `OutputValidator` and the validation here in the configbuilder are separated and are validating different things.
Same with custom filter validation in the webhook and here. All these validation is distributed and we should have **one** validation logic for both.
Having the same validation in the configbuilder as in the webhook is important because we can not guarantee that the webhook is used.

lindnerby: Would like to solve that in a separate PR though.

skhalash (approved):

Issue #15199, opened 2022-08-22
Sometime telemetry operator fails to restart fluentbit pod


rakesh-garimella:
**Description**

Af the changes implemented [here](https://github.com/kyma-project/kyma/issues/15118) there is a corner case when fluentbit might not be restarted.

During the installation it can happen that webhook service might not be ready and if the logging chart has been installed before then the operator. In such scenario operator would update the fluentbit configmap but wont be able to add finalizer to pipeline CR (as the webhook service is not ready). In such a scenario the fluentbit restart is missed as currently we return the error when the pipeline CR update fails and requeue the request. In the next interation we only update the CR and not restart the fluentbit

To fix this we should collect the errors from pipeline update and fluentbit restart and then decide to return error and reque the request.

chrkl:
Will be solved with https://github.com/kyma-project/kyma/issues/15956. Closing this issue.

PR #15202, opened 2022-08-22
improve the logic for error collection

rakesh-garimella:
**Description**

-Aggregate errors from both CR update and fluent-bit restart and then requeue request if required.

**Related issue(s)**

https://github.com/kyma-project/kyma/issues/15199

### DIFF
--- a/components/telemetry-operator/controllers/telemetry/logparser_controller.go
+++ b/components/telemetry-operator/controllers/telemetry/logparser_controller.go
@@ -87,6 +87,7 @@ func (r *LogParserReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 		if err = r.updateLogParserStatus(ctx, req.NamespacedName, condition); err != nil {
 			return ctrl.Result{Requeue: shouldRetryOn(err)}, err
 		}
+		return ctrl.Result{RequeueAfter: requeueTime}, nil
 	}
 
 	if logParser.Status.GetCondition(telemetryv1alpha1.LogParserRunning) == nil {
--- a/components/telemetry-operator/controllers/telemetry/logparser_controller.go
+++ b/components/telemetry-operator/controllers/telemetry/logparser_controller.go
@@ -76,6 +76,11 @@ func (r *LogParserReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 
 	if result.LogParserChanged {
 		if err = r.Update(ctx, &logParser); err != nil {
+			// Return the error if we are only updating the CR
+			if !result.ConfigurationChanged {
+				log.Error(err, "Failed updating log parser")
+				return ctrl.Result{Requeue: shouldRetryOn(err)}, err
+			}
 			errMap["Failed updating log parser"] = err
 		}
 	}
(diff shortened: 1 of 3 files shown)


components/telemetry-operator/controllers/telemetry/logparser_controller.go
```diff
@@ -72,18 +72,26 @@ func (r *LogParserReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 		return ctrl.Result{Requeue: shouldRetryOn(err)}, nil
 	}
 
+	var errMap = make(map[string]error)
+
 	if result.LogParserChanged {
 		if err = r.Update(ctx, &logParser); err != nil {
-			log.Error(err, "Failed updating log parser")
-			return ctrl.Result{Requeue: shouldRetryOn(err)}, err
+			errMap["Failed updating log parser"] = err
```
chrkl, line 79: You never return this error if the condition in line 83 does not match

rakesh-garimella: good catch fixed that

components/telemetry-operator/controllers/telemetry/logparser_controller.go
```diff
@@ -122,6 +116,28 @@ func (r *LogParserReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 	return ctrl.Result{}, nil
 }
 
+func (r *LogParserReconciler) updateCR(ctx context.Context, updateCR bool, logParser telemetryv1alpha1.LogParser) error {
+	if !updateCR {
+		return nil
+	}
+	if err := r.Update(ctx, &logParser); err != nil {
```
chrkl, line 123: Just make a `return r.Update ...` (here and below)

components/telemetry-operator/controllers/telemetry/logpipeline_controller.go
```diff
@@ -186,6 +179,38 @@ func shouldRetryOn(err error) bool {
 		!errors.IsNotFound(err)
 }
 
+func getError(crUpdateError, fbRestartError error) error {
```
chrkl, line 182: Let's make this the following: `func getFirstError(err ...error) error`
The name describes better what is actually happening and you could just return the first list element (or nil). Maybe let's move this to a new `error_helpers` file together with `shouldRetryOn` to do not use code from a different controller.

components/telemetry-operator/controllers/telemetry/logparser_controller.go
```diff
@@ -122,6 +116,28 @@ func (r *LogParserReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 	return ctrl.Result{}, nil
 }
 
+func (r *LogParserReconciler) updateCR(ctx context.Context, updateCR bool, logParser telemetryv1alpha1.LogParser) error {
```
chrkl, line 119: Can we call it `updateLogParser` (and `updateLogPipeline` in the other controller)? The meaning of the shortcut CR is not very obvious.
Maybe we should also indicate that the behaviour is conditional: `updateLogParserIfChanged`

chrkl (changes requested):

chrkl:
We will not continue with the proposed changes since it does not cover all corner cases for the Fluent Bit restart properly.

PR #15209, opened 2022-08-22
Remove log that exposes the event internals

VladislavPaskar:
**Description**

- do not output the event internals in the logs


nachtmaar (approved):
Approving the PR so we can get rid of it asap.
However I wonder if a debug or info log with not the whole event but the event id is still useful 🤔

VladislavPaskar:
> Approving the PR so we can get rid of it asap. However I wonder if a debug or info log with not the whole event but the event id is still useful 🤔

But we have a log when we get the event: `h.namedLogger().Infof("CloudEvent received id:[%s]", event.ID())` in the `components/event-publisher-proxy/pkg/handler/nats/handler.go:217`

Issue #15099, opened 2022-08-12
Unify app label in serverless pods


kwiatekus:
**Description**
We have inconsistent labels across serverless pods:
 - Ctrl manager has: `app: serverless`
 - webhook has `app: serverless-webhook`

I think that we should unify them as together they are serverless application.
I would also look at internal container registry

**Reason**
Serverless is an app (module) that consists of multiple pods.
Using one label would help us with gathering logging from all parts of our system in test as we're using `kubectl logs -n kyma-system -l app=serverless --tail=-1`

dbadura:
Deployment `podSelector` field is not mutable (https://kubernetes.io/docs/concepts/workloads/controllers/deployment/#label-selector-updates) so we cannot simply change pod labels.
The only solution to fix it, is to delete and apply the deployment.
We can do it in several ways:
- write custom code in reconciler: https://github.com/kyma-incubator/reconciler/blob/67d6459b9b280ec7895d254d5df3f5330c03bb5b/pkg/reconciler/kubernetes/updatestrategyresolver.go
- wait for serverless operator and write the custom code
- set replicas to 0 to current deployment or delete (I don't know how reconciler will behave), create new deployment with new name and use proper labels.

dbadura:
Reconciler will delete deployment if we change the name of it

moelsayed:
@dbadura I am unable to reproduce this. If I changed the webhook deployment name, the old deployment is not deleted when I upgrade to the new version. Can you please describe exactly how you did this?

PR #15225, opened 2022-08-23
Unify labels in serverless components

moelsayed:
**Description**

This PR slightly changes the labels on the serverless components to align more with k8s best practices. However, it doesn't make any changes to the selector labels. The selector labels are immutable, and that would break upgrades.

- Align labels for server components with https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/.
- ...
- ...

**Related issue(s)**

[#15099](https://github.com/kyma-project/kyma/issues/15099)

### DIFF
--- a/resources/serverless/charts/docker-registry/templates/deployment.yaml
+++ b/resources/serverless/charts/docker-registry/templates/deployment.yaml
@@ -5,10 +5,12 @@ kind: Deployment
 metadata:
   name: {{ template "docker-registry.fullname" . }}
   labels:
-    app: {{ template "docker-registry.name" . }}
-    chart: {{ .Chart.Name }}-{{ .Chart.Version | replace "+" "_" }}
-    release: {{ .Release.Name }}
-    heritage: {{ .Release.Service }}
+    app: "{{ .Release.Name }}"
+    app.kubernetes.io/name: "{{ .Release.Name }}"
+    app.kubernetes.io/instance: "{{ .Release.Name }}"
+    app.kubernetes.io/component: '{{ template "docker-registry.name" . }}'
+    app.kubernetes.io/version: "{{ .Values.global.images.registry.version }}"
+
 spec:
   selector:
     matchLabels:
--- a/resources/serverless/charts/webhook/values.yaml
+++ b/resources/serverless/charts/webhook/values.yaml
@@ -8,10 +8,11 @@ image:
   pullPolicy: IfNotPresent
 
 commonLabels:
-  app: '{{ template "webhook.fullname" . }}'
-  app.kubernetes.io/name: '{{ template "webhook.fullname" . }}'
+  app: "{{ .Release.Name }}"
+  app.kubernetes.io/name: "{{ .Release.Name }}"
   app.kubernetes.io/instance: "{{ .Release.Name }}"
-  role: webhook
+  app.kubernetes.io/component: '{{ template "webhook.fullname" . }}'
+  app.kubernetes.io/version: "{{ .Values.global.images.function_webhook.version }}"
 
 deployment:
   replicas: 1
--- a/resources/serverless/values.yaml
+++ b/resources/serverless/values.yaml
@@ -61,9 +61,9 @@ global:
     app: '{{ template "name" . }}'
     version: "{{ .Values.global.images.function_controller.version }}"
     app.kubernetes.io/name: '{{ template "name" . }}'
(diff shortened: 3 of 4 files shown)


resources/serverless/values.yaml
```diff
@@ -58,12 +58,11 @@ tests:
 global:
   domainName: "kyma.example.com"
   commonLabels:
-    app: '{{ template "name" . }}'
-    version: "{{ .Values.global.images.function_controller.version }}"
-    app.kubernetes.io/name: '{{ template "name" . }}'
+    app: "{{ .Release.Name }}"
+    app.kubernetes.io/name: "{{ .Release.Name }}"
     app.kubernetes.io/instance: "{{ .Release.Name }}"
+    app.kubernetes.io/component: '{{ template "fullname" . }}-controller'
```
dbadura, line 64: What about just `controller`?

resources/serverless/charts/webhook/values.yaml
```diff
@@ -8,10 +8,11 @@ image:
   pullPolicy: IfNotPresent
 
 commonLabels:
-  app: '{{ template "webhook.fullname" . }}'
-  app.kubernetes.io/name: '{{ template "webhook.fullname" . }}'
+  app: "{{ .Release.Name }}"
+  app.kubernetes.io/name: "{{ .Release.Name }}"
   app.kubernetes.io/instance: "{{ .Release.Name }}"
-  role: webhook
+  app.kubernetes.io/component: '{{ template "webhook.fullname" . }}'
```
dbadura, line 14: Maybe `webhook` is enough?

dbadura (commented):
Do we need to have` instance` label` as we have one?

resources/serverless/charts/docker-registry/templates/deployment.yaml
```diff
@@ -5,10 +5,12 @@ kind: Deployment
...
   name: {{ template "docker-registry.fullname" . }}
   labels:
-    app: {{ template "docker-registry.name" . }}
-    chart: {{ .Chart.Name }}-{{ .Chart.Version | replace "+" "_" }}
-    release: {{ .Release.Name }}
-    heritage: {{ .Release.Service }}
+    app: "{{ .Release.Name }}"
+    app.kubernetes.io/name: "{{ .Release.Name }}"
+    app.kubernetes.io/instance: "{{ .Release.Name }}"
+    app.kubernetes.io/component: '{{ template "docker-registry.name" . }}'
```
dbadura, line 11: `docker-registry`

dbadura:
We can change the pod selector labels if we change the name of deployment to `serverless-webhook`.
The reconciler will delete the old deployment and apply new one.

resources/serverless/values.yaml
moelsayed: I think that might collide with other components that could be using the same pattern. This way it's more specific.

resources/serverless/charts/webhook/values.yaml
moelsayed: Same reasoning as above.

resources/serverless/charts/docker-registry/templates/deployment.yaml
moelsayed: also as above, but this one is mostly to stick to the pattern.

moelsayed:
@dbadura did you confirm this behavior ?

dbadura:
I tested it by `kyma` cli which uses reconciler and it worked.

resources/serverless/values.yaml
dbadura: I wouldn't say it collide.
If I want to get all k8s controllers, I am able to do that with that label.

dbadura: Look at the docs: https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/#labels
The component is named just: `database`

moelsayed: Controller deployment template has the `kyma-project.io/component: controller` which would suitable for matching controllers. Using the `app.kubernetes.io/component` label this way is more aligned with the kubernetes docs recommendation.

moelsayed:
@dbadura I am unable to reproduce this. If I changed the webhook deployment name, the old deployment is not deleted when I upgrade to the new version

moelsayed:
I am closing this PR for now, as this solution is not sufficient to achieve the acceptance criteria. We need to tackle this from the reconciler side.

Issue #15134, opened 2022-08-18
[Telemetry Operator] Bad default handling in LogPipeline resource


a-thaler:
**Description**
Creating a LogPipeline resource on the APIServer is defaulting several attributes to an empty map. That behaviour is not expected and not common.

For example, creating a resource like that:
```
apiVersion: telemetry.kyma-project.io/v1alpha1
kind: LogPipeline
metadata:
  name: icke1
spec:
  output:
    custom: |
      Name               stdout
```

and then calling `k get logpipeline icke1 -oyaml` will return:
```

apiVersion: telemetry.kyma-project.io/v1alpha1
kind: LogPipeline
metadata:
  XXX (not relevant here)
spec:
  output:
    custom: |
      Name               stdout
    grafana-loki:
      url:
        valueFrom:
          secretKeyRef: {}
    http:
      host:
        valueFrom:
          secretKeyRef: {}
      password:
        valueFrom:
          secretKeyRef: {}
      tls: {}
      user:
        valueFrom:
          secretKeyRef: {}
```

I expect that only the spec attributes are defined which I defined in the create statement

skhalash:
Made http,loki output fields and secret key reference nullable. In both cases we are dealing with discriminated union types where only one fields can be defined, and the rest should be nil:
1. If one output is defined, the rest must has to be nil.
2. If an HTTP user has a string value, the secret key reference has to be nil.

PR #15234, opened 2022-08-23
[Telemetry Operator] Make Loki and HTTP outputs nullable

skhalash:
**Description**

- Make Loki and HTTP output nullable to have a better user-experience
- Add a set of helper methods to improve output plugin checking
- Since I touched output plugin validation, I noticed that the output validation logic scatters over 2 types: `pluginValidator` and `outputValidator`. In this PR, I also moved all the output validation logic into the `outputValidator` and renamed the `pluginValidator` into `filterValidator`.

**Related issue(s)**
Resolves https://github.com/kyma-project/kyma/issues/15134


components/telemetry-operator/internal/webhook/logpipeline/validation/plugin_validator.go
```diff
@@ -76,28 +76,17 @@ func (pv *pluginValidator) validateOutput(pipeline *telemetryv1alpha1.LogPipelin
 	if err := validateHTTPOutput(pipeline.Spec.Output.HTTP); err != nil {
 		return err
 	}
-	if err := validateLokiOutPut(pipeline.Spec.Output.Loki); err != nil {
+	if err := validateLokiOutput(pipeline.Spec.Output.Loki); err != nil {
```
lindnerby, line 79: We have an explicit output_validator. Why is output validated here? Should we move it?

components/telemetry-operator/internal/fluentbit/config/builder/output.go
```diff
@@ -11,16 +11,16 @@ import (
 
 func createOutputSection(pipeline *telemetryv1alpha1.LogPipeline, config PipelineConfig) string {
 	output := &pipeline.Spec.Output
-	if len(output.Custom) > 0 {
+	if output.CustomDefined() {
```
lindnerby, line 14: What do you think about Is/Has prefixes here instead? IsCustom/IsHTTP etc. ?

components/telemetry-operator/apis/telemetry/v1alpha1/logpipeline_types.go
```diff
@@ -93,9 +93,43 @@ type TLSConfig struct {
 
 // Output describes a Fluent Bit output configuration section
 type Output struct {
-	Custom string     `json:"custom,omitempty"`
-	HTTP   HTTPOutput `json:"http,omitempty"`
-	Loki   LokiOutput `json:"grafana-loki,omitempty"`
+	Custom string      `json:"custom,omitempty"`
```
lindnerby, line 96: I guess since we intend to keep PR smaller, making this only about output fits, but we should also have a look at the rest of the CRD :)

components/telemetry-operator/internal/webhook/logpipeline/validation/plugin_validator.go
skhalash: Done, see the last bullet point in the PR description.

components/telemetry-operator/apis/telemetry/v1alpha1/logpipeline_types.go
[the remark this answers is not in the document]
lindnerby: Yes, but let's do that in a separate PR as discussed.

chrkl (approved):

lindnerby (approved):

PR #15236, opened 2022-08-24
Print SKR kyma version when provisioning SKR

mfaizanse:
**Description**

- Print SKR kyma version when provisioning SKR for spinnaker

**Related issue(s)**


szwedm (approved):
\lgtm

PR #15274, opened 2022-08-25
add a system-account user to nats-server

friedrichwilken:
This PR adds a user account `admin` with `system_account` privileges in eventing's **nats-server**. This allows sending commands like `nats server info` via the nats cli tool.

Here is a little overview of what this PR does in detail:
1. the [`Authorization `](https://github.com/nats-io/k8s/blob/81418cf9f9c0228a767941fac09123dc197389e5/helm/charts/nats/templates/configmap.yaml#L446) part was added to the `configmap`-template to allow configuration of NATS' `Authorization` properties. 
2. in the `statefulset`-template the mapping to mount a secret as the `resolver-volume` was added
3. in the `statefulset`-template the aforementioned `resolver-volume` is mounted (`volumeMounts`) into the nats-server pod
4. a secret was added that contains data to create a user dubbed `admin` with a random password that has `system-account` privileges
5. finally, in `values.yaml` everything gets connected by allowing authentication, setting the `resolver` to `memory` and mapping the aforementioned secret to it.

**How to acquire the user account data**:
The short way:
```bash
kubectl get secrets -n kyma-system eventing-nats-secret -oyaml | grep -e accountsJson | awk '{print $2}' | base64 -d | grep {user: | awk '{$1=$1};1' | awk '{print substr($0, 2, length($0) - 2)}'
```

this returns:
```bash
user: admin, password: WAIT3THIS4IS5NO7RANDOM8PASSWORD
```

The long way:
first get the secret:
```bash
kubectl get secrets -n kyma-system eventing-nats-secret -oyaml
```
this will return something like this:
```bash
apiVersion: v1
data:
  accounts.json: YWNjb3VudHM6IHsKICBTWVM6IHsKICAgIHVzZXJzOiBbCiAgICAgIHt1c2VyOiBhZG1pbiwgcGFzc3dvcmQ6IEN1bEN6WnpjemphWDMzWk5ObnNxaEQzR2N0b3hqUHFwfQogICAgweXQogIH0sCn0Kc3lzdGVtX2FjY291bnQ6IFNZUw==
kind: Secret
metadata:
  annotations:
    reconciler.kyma-project.io/managed-by-reconciler-disclaimer: |-
      DO NOT EDIT - This resource is managed by Kyma.
      Any modifications are discarded and the resource is reverted to the original state.
  creationTimestamp: "2022-08-25T14:39:38Z"
  labels:
    reconciler.kyma-project.io/managed-by: reconciler
    reconciler.kyma-project.io/origin-version: local
  name: eventing-nats-accounts
  namespace: kyma-system
  resourceVersion: "34881"
  uid: d652dcd8-e92c-4417-9330-8756570e7e68
type: Opaque
```
copy the value after `accounts.json:`. It is `base64` encoded so run
```bash
echo <paste here> | base64 -d
```
to decode it into human readable text:
```bash
accounts: {
  SYS: {
    users: [
      {user: admin, password: WAIT3THIS4IS5NO7RANDOM8PASSWORD}
    ]
  },
}
```
now you can run 
```bash
nats server info --user admin --password WAIT3THIS4IS5NO7RANDOM8PASSWORD
```


resources/eventing/charts/nats/templates/secret.yaml
```diff
@@ -0,0 +1,15 @@
...
+kind: Secret
+metadata:
+  name: "eventing-nats-accounts"
+type: Opaque
+stringData:
+  accounts.json: |-
+    accounts: {
+      SYS: {
+        users: [
+          {user: admin, password: {{ randAlphaNum 32 }}}
```
nachtmaar, line 11: As said today in the tech-talk, we need to make sure that the password is not changing with every reconciliation.

resources/eventing/charts/nats/templates/configmap.yaml
```diff
@@ -155,3 +155,123 @@ data:
...
+    #                #
+    # Authorization  #
+    #                #
+    ##################
+    {{- if .Values.auth.resolver }}
+      {{- if eq .Values.auth.resolver.type "memory" }}
+    resolver: MEMORY
+    include "accounts/{{ .Values.auth.resolver.configMap.key }}"
+      {{- end }}
+      {{- if eq .Values.auth.resolver.type "full" }}
```
nachtmaar, line 170: Do we need to support other resolver types (like `full`) atm ? At least for the use case of the issue the `memory` resolver is enough.

nachtmaar: This would reduce the amount of changes which we need to maintain a lot 🤔

nachtmaar:
Just some thoughts while checking the nats helm charts :)

- I just checked the nats helm chart and found the [`customConfigSecret`](https://github.com/nats-io/k8s/blob/81418cf9f9c0228a767941fac09123dc197389e5/helm/charts/nats/values.yaml#L162) option.
With this option, the whole nats config can be stored in a secret. This could make our current setup easier, because there is only one secret which holds the entire config instead of a configmap and the secret (which contains the username and password). But that also means that our whole config is base64 encoded 🙈 
  - The single secret is mounted then inside the nats pod, see [here](https://github.com/nats-io/k8s/blob/ae48f35e8fd92be30e915f980578a53d78cc0717/helm/charts/nats/templates/statefulset.yaml#L87).

- Ideally the nats charts would just use a secret instead of a [config map](https://github.com/nats-io/k8s/blob/81418cf9f9c0228a767941fac09123dc197389e5/helm/charts/nats/values.yaml#L162) or provide both options. Shall we think about getting your changes upstream instead ? 🤔 This would keep our charts in sync with theirs.

friedrichwilken:
@nachtmaar I reworked the secret so it will no longer will get deleted after reconciliation.

resources/eventing/charts/nats/values.yaml
```diff
@@ -194,6 +194,20 @@ nats:
...
+auth:
+  enabled: true
+  resolver:
+    ##############################
+    #                            #
+    # Memory resolver settings   #
+    #                            #
+    ##############################
+    type: memory
+    configMap:
```
mfaizanse, line 207:
```suggestion
    secret:
```
(nit) its a secret, not a configMap, let rename it for consistency?

resources/eventing/charts/nats/templates/statefulset.yaml
```diff
@@ -58,10 +58,16 @@ spec:
 {{- end }}
       # Common volumes for the containers.
       volumes:
+      ### the secret that holds account data ###
+      {{- if and .Values.auth.enabled .Values.auth.resolver }}
+      - name: resolver-volume
+        secret:
+          secretName: {{ .Values.auth.resolver.configMap.name }}
```
mfaizanse, line 65:
```suggestion
          secretName: {{ .Values.auth.resolver.secret.name }}
```

Related to: https://github.com/kyma-project/kyma/pull/15274/files#r959474476

resources/eventing/charts/nats/templates/statefulset.yaml
```diff
@@ -194,6 +205,14 @@ spec:
...
         {{- end }}
         volumeMounts:
+        ### the secret that holds account data ###
+          {{- if and .Values.auth.enabled .Values.auth.resolver }}
+          {{- if eq .Values.auth.resolver.type "memory" }}
+          - name: resolver-volume
+            mountPath: /etc/nats-config/accounts
+          {{- end }}
+          {{- end }}
+
```
mfaizanse, line 215: SInce we are updating the statefulset, have you tested the upgrade flow?

resources/eventing/charts/nats/templates/secret.yaml
```diff
@@ -0,0 +1,24 @@
+{{- $secret := (lookup "v1" "Secret" .Release.Namespace "eventing-nats-accounts") -}}
+apiVersion: v1
+kind: Secret
+metadata:
+  name: eventing-nats-accounts
```
mfaizanse, line 5: (nit) instead of hard coding name, can we follow the same approach as in [configmap.yaml](https://github.com/kyma-project/kyma/blob/f41c24f41f686ba420ea063d93f654e915e8cc1a/resources/eventing/charts/nats/templates/configmap.yaml)?

resources/eventing/charts/nats/templates/configmap.yaml
```diff
@@ -155,3 +155,17 @@ data:
...
+    {{- if .Values.auth.enabled }}
+    ##################
+    #                #
+    # Authorization  #
+    #                #
+    ##################
+    {{- if .Values.auth.resolver }}
+    {{- if eq .Values.auth.resolver.type "memory" }}
+    resolver: MEMORY
+    include "accounts/{{ .Values.auth.resolver.configMap.key }}"
```
mfaizanse, line 168:
```suggestion
    include "accounts/{{ .Values.auth.resolver.secret.key }}"
```

Related to: https://github.com/kyma-project/kyma/pull/15274/files#r959474476

resources/eventing/charts/nats/templates/statefulset.yaml
friedrichwilken: Yes, it seems to work if I upgrade from `main` to PR. The only problem so far is that IF you would actually change anything in the secret nats-server would not get updated. Usually this should not happen. A workaround is to add or remove some whitespace in the configmap which triggers the reloader.

mfaizanse (approved):

TorstenD-SAP (approved):
A Threat Modelling to identify potential threats and vulnerabilities was performed and necessary mitigations are identified.

PR #15318, opened 2022-08-29
[Telemetry Operator] Make ValueFrom and SecretRef nullable

skhalash:
**Description**

- Rename `ValueFromType` to `ValueFromSource` (similar to https://github.com/kubernetes/api/blob/master/core/v1/types.go#L2075)
- Make `ValueFrom` field nullable. If `Value` is defined, it must be set to null
- Make `SecretRef` optional (in the future we might want to add support for config maps) 

**Related issue(s)**
https://github.com/kyma-project/kyma/issues/15134


components/telemetry-operator/internal/utils/envvar/format.go
```diff
@@ -7,7 +7,7 @@ import (
 	"github.com/kyma-project/kyma/components/telemetry-operator/apis/telemetry/v1alpha1"
 )
 
-func GenerateName(logpipelineName string, secretKeyRef v1alpha1.SecretKeyRef) string {
+func GenerateName(logpipelineName string, secretKeyRef v1alpha1.SecretRef) string {
```
lindnerby, line 10: Rename input params as well please

components/telemetry-operator/apis/telemetry/v1alpha1/logpipeline_types.go
```diff
@@ -140,28 +140,32 @@ type FileMount struct {
...
 
-func (v *ValueFromType) IsSecretRef() bool {
-	return v.SecretKey.Name != "" && v.SecretKey.Key != ""
+type ValueFromSource struct {
+	SecretRef *SecretRef `json:"secretKeyRef,omitempty"`
 }
 
-type ValueFromType struct {
-	SecretKey SecretKeyRef `json:"secretKeyRef,omitempty"`
+func (v *ValueFromSource) IsSecretRef() bool {
```
lindnerby, line 164: Maybe this should be on the ValueType? Why ask the non-nil SecretRef if it "IsSecretRef"? Or rename it to something else if it's just for validation?

lindnerby (changes requested):

components/telemetry-operator/internal/utils/envvar/format.go
skhalash: Reverted back to `SecretKeyRef`

components/telemetry-operator/apis/telemetry/v1alpha1/logpipeline_types.go
skhalash: After a discussion, we decided to keep it like this so far

lindnerby (approved):

PR #15350, opened 2022-09-02
Delete commerce test from skr nightly

ukff:
Delete commerce test from skr nightly because it is performed in other place


tillknuesting (dismissed):

tillknuesting (approved):
Lgtm

Issue #15315, opened 2022-08-29
Generate alias for outputs on base of logpipeline name


a-thaler:
**Description**
I saw outputs in grafana named like http.1 and http.2. Please verify that all output definitions result in generated alias configs using the name of the pipeline, so all outputs of the fluent-bit config should have an alias like <pipelineName>-<outputName>

**Criterias**
- In grafana all outputs are named nicely

PR #15374, opened 2022-09-05
[Telemetry Operator] Set output plugin alias and rewrite tag emitter name

Teneroy:
**Description**

- Added logpipeline name to the name of input
- Added logpipeline name to the name of output
- Updated samples for telemetry-operator
- Adapted tests

The affected plugin fields are to be used by the fluent bit metrics as labels, which in turn make the Fluent Bit Grafana dashboard more descriptive. 

**Related issue(s)**
#15315


resources/telemetry/charts/fluent-bit/values.yaml
```diff
@@ -366,6 +366,7 @@ config:
   inputs: |
     [INPUT]
         Name tail
+        Alias tail-tele
```
chrkl, line 369: Make "tail" a suffix to have it consistent.

resources/telemetry/charts/fluent-bit/values.yaml
```diff
@@ -374,12 +375,12 @@ config:
...
         DB /data/flb_kube.db
         storage.type  filesystem
-
+ 
     [INPUT]
         Name tail
         Path /null.log
         Tag null.*
-        Alias dummy
+        Alias tail-dummy
```
a-thaler, line 383: Shouldn't it be called "null-dummy" as well as it is part of the "null" pipeline somehow

components/telemetry-operator/internal/fluentbit/config/builder/output.go
```diff
@@ -29,9 +29,20 @@ func createOutputSection(pipeline *telemetryv1alpha1.LogPipeline, defaults Pipel
 func generateCustomOutput(output *telemetryv1alpha1.Output, fsBufferLimit string, name string) string {
 	sb := NewOutputSectionBuilder()
 	customOutputParams := parseMultiline(output.Custom)
+	pipelineName := ""
```
chrkl, line 32: Use `var pipelineName string` (same below)

skhalash (changes requested):

PR #15375, opened 2022-09-06
[Telemetry operator] Add telemetry_all_logpipelines gauge

skhalash:
**Description**

- Add `telemetry_all_logpipelines ` gauge to count the current number of log pipelines
- Rename `telemetry_plugins_unsupported_total ` to `telemetry_unsupported_logpipelines `. The `_total` ending should be only used with counters: https://www.robustperception.io/on-the-naming-of-things/.
- Lift metric updating from the syncer to the controller itself
- Whitelist new metrics in the `servicemonitor`

**Related issue(s)**

### DIFF
--- a/components/telemetry-operator/controller/logpipeline/controller.go
+++ b/components/telemetry-operator/controller/logpipeline/controller.go
@@ -222,13 +222,13 @@ func (r *Reconciler) updateMetrics(allPipelines *telemetryv1alpha1.LogPipelineLi
 }
 
 // syncUnsupportedPluginsTotal checks if any LogPipeline defines a unsupported Filter or Output.
-func countUnsupportedPluginUsages(logPipelines *telemetryv1alpha1.LogPipelineList) int {
+func countUnsupportedPluginUsages(pipelines *telemetryv1alpha1.LogPipelineList) int {
 	count := 0
-	for _, l := range logPipelines.Items {
-		if !l.DeletionTimestamp.IsZero() {
+	for i := range pipelines.Items {
+		if !pipelines.Items[i].DeletionTimestamp.IsZero() {
 			continue
 		}
-		if usesUnsupportedPlugin(&l) {
+		if usesUnsupportedPlugin(&pipelines.Items[i]) {
 			count++
 		}
 	}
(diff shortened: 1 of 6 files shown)


components/telemetry-operator/controller/logpipeline/controller.go
```diff
@@ -204,3 +215,35 @@ func (r *Reconciler) updateLogPipelineStatus(ctx context.Context, name types.Nam
...
+		}
+		if usesUnsupportedPlugin(&pipelines.Items[i]) {
+			count++
+		}
+	}
+
+	return count
+}
+
+func usesUnsupportedPlugin(pipeline *telemetryv1alpha1.LogPipeline) bool {
```
lindnerby, line 239: Think you can use the method of LogPipeline `ContainsCustomPlugin`

components/telemetry-operator/controller/logpipeline/controller.go
```diff
@@ -44,10 +44,11 @@ type Config struct {
...
 	client.Client
-	config           Config
-	syncer           *syncer
-	unsupportedTotal prometheus.Gauge
-	daemonSetHelper  *kubernetes.DaemonSetHelper
+	config                  Config
+	syncer                  *syncer
+	daemonSetHelper         *kubernetes.DaemonSetHelper
+	allLogPipelines         prometheus.Gauge
+	unsupportedLogPipelines prometheus.Gauge
```
lindnerby, line 51: Just an idea: What do you think about (similar to the `syncer`) wrap metrics into a `metrics` or `controllerMetrics` struct, where the constructor function create the Gauges and possibly future metrics can be added.

Also calls to this would look like `metrics.Update(&allPipelines)`

components/telemetry-operator/controller/logpipeline/controller.go
skhalash: Agree, will do it

components/telemetry-operator/controller/logpipeline/controller.go
skhalash: I agree that it would make sense to collect all metrics under an overarching struct. However, having a single point of the metric update is hard because different metrics are updated in different phases of reconciliation.

lindnerby (approved):

lindnerby (approved):

PR #15390, opened 2022-09-07
Prepare code initializing Compass Runtime Agent before Compass Runtime Agent runs

akgalwas:
**Description**

- Creating formation
- Registering Runtime
- Unregistering Runtime
- Restarting Compass Runtime Agent

**Related issue(s)**
#15037


tests/components/application-connector/test/compass-runtime-agent/testkit/compassruntimeagentinit/compass.go
```diff
@@ -0,0 +1,77 @@
...
+	if err != nil {
+		var result *multierror.Error
+
+		multierror.Append(result, err)
+		{
+			err := rollBackFunc()
+			if err != nil {
+				multierror.Append(result, err)
+			}
+		}
```
franpog859, line 48: Why is `rollBackFunc` called in this place?

tests/components/application-connector/test/compass-runtime-agent/testkit/compassruntimeagentinit/compassconnection.go
```diff
@@ -0,0 +1,71 @@
...
+	if err != nil {
+		return nil, err
+	}
+	compassConnectionCR.ResourceVersion = ""
+
+	compassConnectionCRBackup := compassConnectionCR.DeepCopy()
+	compassConnectionCRBackup.ObjectMeta.Name = "compass-connection-backup"
+	_, err = cc.compassConnectionInterface.Create(context.TODO(), compassConnectionCRBackup, meta.CreateOptions{})
+	if err != nil {
+		return nil, err
```
franpog859: Just asking if I understand it correctly. We already deleted the `compass-connection` resource in this place but there is no rollback function because it'd be recreated by the Agent, is it right?

tests/components/application-connector/test/compass-runtime-agent/testkit/compassruntimeagentinit/deployment.go
```diff
@@ -0,0 +1,157 @@
...
+	err = waitForRollout(dc.deploymentName, deploymentInterface)
+	if err != nil {
+		var murliErr *multierror.Error
+
+		multierror.Append(murliErr, err)
+
+		err := rollbackDeploymentFunc()
+		if err != nil {
+			multierror.Append(murliErr, err)
+		}
```
franpog859, line 80: And also there. Why the rollback function is called there already? Shouldn't it be some higher where all the rollback functions would be called?

tests/components/application-connector/test/compass-runtime-agent/testkit/compassruntimeagentinit/deployment.go
```diff
@@ -0,0 +1,157 @@
...
+	}, retry.Attempts(RetryAttempts), retry.Delay(RetrySeconds*time.Second))
+}
+
+func waitForRollout(name string, deploymentInterface v13.DeploymentInterface) error {
+	return retry.Do(func() error {
+		deployment, err := deploymentInterface.Get(context.TODO(), name, v1.GetOptions{})
+		if err != nil {
+			return err
+		}
+		if deployment.Status.AvailableReplicas == 0 || deployment.Status.UnavailableReplicas != 0 {
```
franpog859, line 156: We should manually check if this `if` statement works correctly and if the deployment is indeed rolled out correctly at this place

franpog859 (commented):
Hey, a few comments from my side 👋

tests/components/application-connector/test/compass-runtime-agent/testkit/compassruntimeagentinit/compass.go
akgalwas: The idea was to clean up `Runtime` and `Formation` in case getting token failed. Now I see its redundant, as there is this piece of code in Compass Runtime Agent Configurator
```
compassRuntimeAgentConfig, compassConfiguratorRollbackFunc, err := crc.compassConfigurator.Do(runtimeName, formationName)
	if err != nil {
		return nil, crc.rollbackOnError(err,
			compassConfiguratorRollbackFunc)
	}
```

tests/components/application-connector/test/compass-runtime-agent/testkit/compassruntimeagentinit/deployment.go
akgalwas: There is a similar case as in Compass Configurator. I will remove this code.

tests/components/application-connector/test/compass-runtime-agent/testkit/compassruntimeagentinit/compassconnection.go
akgalwas: It is supposed to works as follows:
- in `Do` we delete original CompassConnection CR so that the CR will be recreated by the Compass Runtime Agent. Additionally we create a backup version of the CR.
- In rollback function backup CR is deleted ind CompassConnection CR restored

I think it needs some testing to make sure it is a correct approach. I didn't checked the case when Compass Runtime Agent is configured properly, and we run the test. The test should restore the state of the Compass Runtime Agent (so it should sync with Compass original configuration).

tests/components/application-connector/test/compass-runtime-agent/testkit/compassruntimeagentinit/deployment.go
akgalwas: Some initial testing shows that it is correct.

koala7659 (approved):
LGTM! Kudos for Go Generics!

PR #15400, opened 2022-09-08
Fix for cleanup in makefiles for application-connector component tests

koala7659:
After latest changes in makefiles the cleanup command was not called properly. 

This PR should fix this problem


franpog859 (commented):
What about the Compass Runtime Agent Tests?

franpog859 (approved):

PR #15423, opened 2022-09-09
Remove compass from skr-nightly-test

tillknuesting:
removing some left overs of compass in the test


wozniakjan (approved):
/lgtm
/approve

PR #15428, opened 2022-09-09
[Telemetry Operator] Secrets watcher

rakesh-garimella:
**Description**

- Add a mechanism that caches secrets referenced by all log pipelines. Also, watch all secret updates in the cluster - if the secret is cached, then trigger a reconciliation of the corresponding log pipelines
- Add an e2e test
- Decrease unnecessary log lines to the debug log level

**Related issue(s)**


components/telemetry-operator/controller/logpipeline/controller.go
```diff
@@ -81,9 +84,42 @@ func NewReconciler(
 func (r *Reconciler) SetupWithManager(mgr ctrl.Manager) error {
 	return ctrl.NewControllerManagedBy(mgr).
 		For(&telemetryv1alpha1.LogPipeline{}).
+		Watches(
+			&source.Kind{Type: &corev1.Secret{}},
+			handler.EnqueueRequestsFromMapFunc(r.enqueueRequests),
+			builder.WithPredicates(predicate.Funcs{
+				CreateFunc: func(event event.CreateEvent) bool {
```
rakesh-garimella, line 91: Predicate function is pretty intrusive, please check if with this change we can still create a new pipeline or not. In my tests I had some issues that creating a new pipeline was rejected. So I had to do type assertion to verify if incoming object is a secret

Also this change will not cover the case where a user creates a pipeline which is referencing a secret that does not exist yet and would be create later.

lindnerby: I think you were using the filter by predicates on the eventhandler registration level in `WithEventFilter` not on the `Watches` registration where events are already filtered by type Secret and then passed on to the predicates function. You are correct, it wasn't working with that setup so we needed to change it.
We tested it thoroughly with the current changes and they seem fine.

components/telemetry-operator/apis/telemetry/v1alpha1/logpipeline_types.go
```diff
@@ -273,6 +278,15 @@ func (l *LogPipeline) ContainsCustomPlugin() bool {
 	return l.Spec.Output.IsCustomDefined()
 }
 
+func ContainsCustomPlugin(l *LogPipeline) bool {
```
chrkl, line 281: Do we need this?

components/telemetry-operator/controller/logpipeline/controller.go
```diff
@@ -252,13 +307,13 @@ func (r *Reconciler) updateMetrics(allPipelines *telemetryv1alpha1.LogPipelineLi
 type keepFunc func(*telemetryv1alpha1.LogPipeline) bool
 
 func count(pipelines *telemetryv1alpha1.LogPipelineList, keep keepFunc) int {
-	count := 0
+	c := 0
```
chrkl, line 310: Was this change intentional?

components/telemetry-operator/controller/logpipeline/controller.go
```diff
@@ -131,16 +171,14 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 	}
 
 	if changed {
-		log.V(1).Info("Fluent Bit configuration was updated. Restarting the DaemonSet")
+		log.Info("Fluent Bit configuration was updated. Restarting the DaemonSet")
```
chrkl, line 174: This PR reaplces `log.Info` by `log.V(1).Info` for most of the occurrences (I guess for consistency?). Is there any reason to do it vice-versa in this line?

skhalash: Yeah, most of the info logs created too much noise, so they were demoted to debug. However, this one is good to have with the info level in order to see which reconciliation actually triggered a restart.

components/telemetry-operator/controller/logpipeline/controller.go
skhalash: I assume, GoLand was complaining since the variable name shadowed the func name, which was also count.

chrkl (approved):

PR #15450, opened 2022-09-12
Update GO Version in Function Controller

Cortey:
**Description**

Fixes CVE-2022-27664 & CVE-2022-29526


components/function-controller/go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/kyma-project/kyma/components/function-controller
 
-go 1.16
+go 1.19
```
pPrecel, line 3: Go version is also here:
https://github.com/kyma-project/kyma/blob/0306e0770dd5d00ff2794d301310752d43e5e4fe/components/function-controller/deploy/manager/Dockerfile#L1

If you would bump the builder version in the Dockerfile then the built binary on the image will have bumped (higher) version.

But it's great you bumped the version here as well. It re-organizes `go.mod` structure to a more readable one :)

pPrecel (commented):
Looks good but please read my one comment

pPrecel (approved):

PR #15458, opened 2022-09-12
remove unnecessary goroutine

k15r:
**Description**

the async callback is already put into a go routing somewhere inside the nats sdk. so we do not have to do it ourselves.

**Related issue(s)**


k15r:
Closing for now, but it will return ;)

PR #15481, opened 2022-09-14
Separate tests in `event-publisher-proxy`

friedrichwilken:
this pr 
- separates tests into `unit` and `integration` tests by adding a corresponding build-tags
- adds `make` targets to make `unit` or `integration` tests run only

it also separates the tests into files
- adds filename component `internal` for tests that need to exist in the same module (because they use private functions) as its system under test
- changes the module name by adding a `_test`-suffix to test that can be isolated from it's system under test.
- adds the filename component `integration` for integration tests


components/event-publisher-proxy/Makefile
```diff
@@ -66,3 +66,23 @@ $(cmds_images_push): %.push: %
...
+	@echo -n "Total coverage: "
+	@go tool cover -func=/tmp/artifacts/cover.out | grep total | awk '{print $$3}'
+ 
+test-local-unit: ## Run only unit tests
+	mkdir -p /tmp/artifacts
+	go test -coverprofile=/tmp/artifacts/cover.out --tags=unit ./...
+	@echo -n "Total coverage: "
+	@go tool cover -func=/tmp/artifacts/cover.out | grep total | awk '{print $$3}'
+ 
+test-local-integraion: ## Run only integration tests
```
marcobebway, line 82: typo `test-local-integraion` -> `test-local-integration`

components/event-publisher-proxy/Makefile
```diff
@@ -66,3 +66,23 @@ $(cmds_images_push): %.push: %
...
+	mkdir -p /tmp/artifacts
+	go test -coverprofile=/tmp/artifacts/cover.out --tags=integration ./...
+	@echo -n "Total coverage: "
+	@go tool cover -func=/tmp/artifacts/cover.out | grep total | awk '{print $$3}'
+
+test-local-int: test-local-integration
```
marcobebway, line 88: (nit) not needed IMO.

components/event-publisher-proxy/Makefile
```diff
@@ -66,3 +66,23 @@ $(cmds_images_push): %.push: %
 	@echo "+ Pushing container image $($@_img_name)"
 	docker image tag $($@_img_name) $(DOCKER_PUSH_REPOSITORY)$(DOCKER_PUSH_DIRECTORY)/$($@_img_name):$(TAG)
 	docker image push $(DOCKER_PUSH_REPOSITORY)$(DOCKER_PUSH_DIRECTORY)/$($@_img_name):$(TAG)
+
+test-local: ## Run all tests
```
marcobebway, line 70: For consistency, please let's move the comments on top of the targets.

components/event-publisher-proxy/Makefile
```diff
@@ -66,3 +66,23 @@ $(cmds_images_push): %.push: %
...
+	mkdir -p /tmp/artifacts
+	go test -coverprofile=/tmp/artifacts/cover.out --tags="unit integration" ./...
+	@echo -n "Total coverage: "
+	@go tool cover -func=/tmp/artifacts/cover.out | grep total | awk '{print $$3}'
+ 
+test-local-unit: ## Run only unit tests
```
marcobebway, line 76: Exec `make test-local-unit` failed:

```
pkg/handler/beb/handler_health_test.go:18:24: undefined: bigRequestSize
pkg/handler/beb/handler_health_test.go:19:24: undefined: defaultEventsEndpoint
```

components/event-publisher-proxy/pkg/application/clean_test.go
```diff
@@ -1,9 +1,13 @@
...
+//go:build unit
+// +build unit
+
+package application_test
 
 import (
 	"testing"
 
 	applicationv1alpha1 "github.com/kyma-project/kyma/components/application-operator/pkg/apis/applicationconnector/v1alpha1"
+	eppapplication "github.com/kyma-project/kyma/components/event-publisher-proxy/pkg/application"
```
marcobebway, line 10: The `eppapplication` gives an impression that this is the `Event Publisher Proxy Application`. I think `application` should be more convenient here.

components/event-publisher-proxy/Makefile
friedrichwilken: it makes actually sense, but I was not expecting this: if you define a const in a test file with the build tag `integration` you don't have access to it any more in a file in the same module but with the build tag `unit` if you run tests with the `unti` arg.

friedrichwilken: on a second glance I decided that all the tests are integration tests anyhow

components/event-publisher-proxy/pkg/application/clean_test.go
```diff
@@ -1,9 +1,13 @@
...
+//go:build unit
+// +build unit
+
+package application_test
 
 import (
 	"testing"
 
 	applicationv1alpha1 "github.com/kyma-project/kyma/components/application-operator/pkg/apis/applicationconnector/v1alpha1"
+	application "github.com/kyma-project/kyma/components/event-publisher-proxy/pkg/application"
```
marcobebway, line 10: (nit) Alias `application` can be omitted.

components/event-publisher-proxy/pkg/cloudevents/utils_test.go
```diff
@@ -12,6 +15,8 @@ import (
 
 	"github.com/kyma-project/kyma/components/event-publisher-proxy/internal"
 	"github.com/kyma-project/kyma/components/event-publisher-proxy/pkg/ems"
+
+	cloudevents "github.com/kyma-project/kyma/components/event-publisher-proxy/pkg/cloudevents"
```
marcobebway, line 19: (nit) Alias `cloudevents` can be omitted.

components/event-publisher-proxy/pkg/env/beb_config_test.go
```diff
@@ -1,8 +1,13 @@
...
+//go:build unit
+// +build unit
+
+package env_test
 
 import (
 	"net/http"
 	"testing"
+
+	env "github.com/kyma-project/kyma/components/event-publisher-proxy/pkg/env"
```
marcobebway, line 10: (nit) Alias `env` can be omitted.

components/event-publisher-proxy/pkg/handler/common_test.go
```diff
@@ -1,9 +1,14 @@
...
+// +build unit
+
+package handler_test
 
 import (
 	"testing"
 
 	"github.com/stretchr/testify/assert"
+
+	handler "github.com/kyma-project/kyma/components/event-publisher-proxy/pkg/handler"
```
marcobebway, line 11: (nit) Import alias can be omitted.

components/event-publisher-proxy/pkg/handler/health/health_integration_test.go
```diff
@@ -7,6 +10,8 @@ import (
 
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
+
+	health "github.com/kyma-project/kyma/components/event-publisher-proxy/pkg/handler/health"
```
marcobebway, line 14: (nit) Import alias can be omitted.

components/event-publisher-proxy/pkg/legacy-events/legacy_internal_test.go
```diff
@@ -0,0 +1,134 @@
...
+	"fmt"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	"github.com/kyma-project/kyma/components/event-publisher-proxy/internal"
+	legacyapi "github.com/kyma-project/kyma/components/event-publisher-proxy/pkg/legacy-events/api"
+	. "github.com/kyma-project/kyma/components/event-publisher-proxy/testing"
```
marcobebway, line 16: (nit) Please let's use an import alias.

components/event-publisher-proxy/pkg/oauth/client_integration_test.go
```diff
@@ -12,34 +15,41 @@ import (
 
 	"github.com/kyma-project/kyma/components/event-publisher-proxy/pkg/env"
 	testingutils "github.com/kyma-project/kyma/components/event-publisher-proxy/testing"
+
+	oauth "github.com/kyma-project/kyma/components/event-publisher-proxy/pkg/oauth"
```
marcobebway, line 19: (nit) Import alias can be omitted.

components/event-publisher-proxy/pkg/receiver/receiver_integration_test.go
```diff
@@ -9,7 +12,7 @@ import (
 
 	"github.com/kyma-project/kyma/components/eventing-controller/logger"
 
-	testingutils "github.com/kyma-project/kyma/components/event-publisher-proxy/testing"
+	receiver "github.com/kyma-project/kyma/components/event-publisher-proxy/pkg/receiver"
```
marcobebway, line 15: (nit) Import alias can be omitted.

components/event-publisher-proxy/pkg/sender/beb_integration_test.go
```diff
@@ -9,6 +12,8 @@ import (
 	"github.com/kyma-project/kyma/components/event-publisher-proxy/pkg/env"
 	"github.com/kyma-project/kyma/components/event-publisher-proxy/pkg/oauth"
 	testingutils "github.com/kyma-project/kyma/components/event-publisher-proxy/testing"
+
+	sender "github.com/kyma-project/kyma/components/event-publisher-proxy/pkg/sender"
```
marcobebway, line 16: (nit) Import alias can be omitted.

components/event-publisher-proxy/pkg/cloudevents/eventtype/clean_test.go
```diff
@@ -11,6 +14,8 @@ import (
 	"github.com/kyma-project/kyma/components/event-publisher-proxy/pkg/application"
 	"github.com/kyma-project/kyma/components/event-publisher-proxy/pkg/application/applicationtest"
 	"github.com/kyma-project/kyma/components/event-publisher-proxy/pkg/application/fake"
+
+	sut "github.com/kyma-project/kyma/components/event-publisher-proxy/pkg/cloudevents/eventtype"
```
marcobebway, line 18: Please let's use another alias.

friedrichwilken:
> (nit) Does it make sense to rename the test files to follow the unit/integration pattern introduced?
> 
> For example:
> 
> * handler.go
> * handler_unit_test.go
> * handler_integration_test.go

I was wondering about that too but wanted to avoid to do too much at once.
What about the `internal` section in the name? keep it?
eg `something_internal_unit_test.go`?

components/event-publisher-proxy/pkg/nats/connect_integration_test.go
```diff
@@ -7,7 +10,7 @@ import (
 	"github.com/nats-io/nats.go"
 	"github.com/stretchr/testify/assert"
 
-	pkgnats "github.com/kyma-project/kyma/components/event-publisher-proxy/pkg/nats"
+	eppnats "github.com/kyma-project/kyma/components/event-publisher-proxy/pkg/nats"
```
marcobebway, line 13: (nit) I think `pkgnats` makes more sense, also IMO `epp` is a redundant info.

components/event-publisher-proxy/pkg/subscribed/helpers_test.go
```diff
@@ -1,10 +1,14 @@
...
+// +build unit
+
+package subscribed_test
 
 import (
 	"reflect"
 	"testing"
 
-	eventingv1alpha1 "github.com/kyma-project/kyma/components/eventing-controller/api/v1alpha1"
+	subscribed "github.com/kyma-project/kyma/components/event-publisher-proxy/pkg/subscribed"
```
marcobebway, line 10: (nit) alias can be omitted.

marcobebway:
> > (nit) Does it make sense to rename the test files to follow the unit/integration pattern introduced?
> > For example:
> > 
> > * handler.go
> > * handler_unit_test.go
> > * handler_integration_test.go
> 
> @marcobebway I was wondering about that too but wanted to avoid to do too much at once. What about the `internal` section in the name? keep it? eg `something_internal_unit_test.go`?

IMO, the internal part is not need, so `something_unit_test.go` should be fine.

components/event-publisher-proxy/pkg/legacy-events/legacy_unit_test.go
```diff
@@ -103,9 +104,9 @@ func TestTransformLegacyRequestsToCE(t *testing.T) {
 			writer := httptest.NewRecorder()
 			app := applicationtest.NewApplication(tc.givenApplication, applicationTypeLabel(tc.givenTypeLabel))
 			appLister := fake.NewApplicationListerOrDie(ctx, app)
-			transformer := NewTransformer("test", tc.givenPrefix, appLister)
+			transformer := legacy.NewTransformer("test", tc.givenPrefix, appLister)
 			gotEvent, gotEventType := transformer.TransformLegacyRequestsToCE(writer, request)
-			wantEventType := formatEventType(tc.givenPrefix, tc.givenApplication, tc.givenEventName, tc.wantVersion)
+			wantEventType := fmt.Sprintf("%s.%s.%s.%s", tc.givenPrefix, tc.givenApplication, tc.givenEventName, tc.wantVersion)
```
marcobebway, line 109: I think we need to bring the func `formatEventType` back, because it also handles the empty prefix.

components/event-publisher-proxy/pkg/receiver/receiver_internal_unit_test.go
```diff
@@ -0,0 +1,29 @@
+//go:build unit
```
marcobebway, line 1: Please rename all unit test files from `...internal_unit_test.go` to `..._unit_test.go` as discussed here https://github.com/kyma-project/kyma/pull/15481#issuecomment-1248342968.

components/event-publisher-proxy/pkg/legacy-events/legacy_unit_test.go
friedrichwilken: `formatEventType` is the only reason we need to have the test in the same module; removing it allows to move the tests to a `_test` module. And then test table does not contain an empty-prefix case.

components/event-publisher-proxy/pkg/handler/beb/handler_integration_test.go
```diff
@@ -1,3 +1,6 @@
+//go:build integration
```
nachtmaar, line 1: I discussed this test with @k15r recently. I think this is a component test.
The reason why I think so is that this test basically tests the whole EPP component by:
- starting the EPP with a handler and message sender
- sending message to the EPP
- checking the return code of the EPP (there is even a real nats server behind the scenes).

I think this definition of a component test matches here:
> a component test is a test that takes the whole component, isolates it as much as possible from its dependencies (by replacing them with mocks and stubs if possible) and then tests how the component reacts to different stimuli in the test suite.

Why do you think this is only an integration test ? We could also discuss this with @ameteiko 🤔

Issue #15490, opened 2022-09-14
[Telemetry Operator] Exclusion of containers enables system namespace inclusion


a-thaler:
**Description**
Assume a pipeline configuration like that:
```yaml
apiVersion: telemetry.kyma-project.io/v1alpha1
kind: LogPipeline
metadata:
  name: cls-application-logs
spec:
  input:
    application:
      containers:
        exclude:
          - istio-proxy
...
```

then the expected behaviour according to documentation is that system namespaces are getting ignored (default behaviour for namespace selection stays untouched). In fact, that setting enables the streaming of logs for system namespaces.

lindnerby:
Fluent bit's rewrite_tag filter only evaluated rules until one matches, thus chaining rules for namespace and container selectors didn't work. We fixed this by piggybacking container selector rules in the rewrite_tag filter, if present and
for any namespace selectors we create an additional grep filter.

PR #15494, opened 2022-09-15
[Telemetry Operator] Add grep filter for namespace filtering in input

lindnerby:
**Description**
Apparently fluent bit's rewrite_tag filter only evaluates rules until one matches, thus chained rules for namespace and container selectors won't work. We now piggyback container selectors from application input in the rewrite_tag filter.
For any namespace selectors we create an additional grep filter.

- Extract namespace rule generation from rewrite_tag config builder
- Add grep filter builder function
- Adapt tests

**Related issue(s)**
* https://github.com/kyma-project/kyma/issues/15490


components/telemetry-operator/internal/fluentbit/config/builder/namespace_filter.go
```diff
@@ -0,0 +1,39 @@
...
+		return ""
+	}
+
+	var sectionBuilder = NewFilterSectionBuilder().
+		AddConfigParam("Name", "grep").
+		AddConfigParam("Match", fmt.Sprintf("%s.*", pipeline.Name))
+
+	if len(namespaces.Include) > 0 {
+		return sectionBuilder.
+			AddConfigParam("Regex", fmt.Sprintf("$kubernetes['namespace_name'] %s", strings.Join(namespaces.Include, "|"))).
```
chrkl, line 26: Is there a reason why we build the rule differently for the rewrite_tag filter and regex filter?

`"^(namespace1|namespace2)$"` vs `namespace1|namespace2`

lindnerby: Not really, just figured since grep filter has explicit config params for include (`regex`) and `exclude` the regex kann be simpler and doesn't need capturing group etc.
Probably rewrite_tag regex for containers could also be simplified but didn't test that out.
Good catch, we should have a look at the rewrite_tag regex creation as well and see if it can be simplified to save a few steps while matching

chrkl (approved):

chrkl (approved):

chrkl (approved):

skhalash (approved):

PR #15511, opened 2022-09-16
Update CPU requests for Istio pods with HPA

cnvergence:
**Description**

- Add more realistic CPU requests values for HPA


barchw (approved):
Approving at this is already merged in https://github.tools.sap/kyma/management-plane-config/pull/2099

PR #15558, opened 2022-09-19
Fix Istio authorization policy

skhalash:
**Description**

- Remove `logging-fluent-bit` leftover

**Related issue(s)**


chrkl (commented):
The flag to enable fluent bit in the logging chart is still working and we promised in the change logs to keep this for at least 2 releases. We should keep the policy until we remove the fluent-bit chart entirely.

skhalash:
@chrkl, Totally makes sense. Closing it then.

PR #15564, opened 2022-09-20
Set permissive mTLS policy for publisher-proxy metrics port

VladislavPaskar:
**Description**

- add flag to enable loosening of the mTLS mode for publisher proxy
- set the policy to `PERMISSIVE` for the publisher's metrics port

### DIFF
--- a/resources/eventing/charts/publisher-proxy/templates/peerauthentication.yaml
+++ b/resources/eventing/charts/publisher-proxy/templates/peerauthentication.yaml
@@ -9,4 +9,4 @@ spec:
     matchLabels: {{- include "publisher-proxy.peerauth.selectorLabels" . | nindent 6 }}
   portLevelMtls:
     {{ .Values.metrics.config.port }}:
-      mode: STRICT
+      mode: PERMISSIVE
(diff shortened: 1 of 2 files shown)


resources/eventing/charts/publisher-proxy/templates/peerauthentication.yaml
```diff
@@ -9,4 +9,4 @@ spec:
     matchLabels: {{- include "publisher-proxy.peerauth.selectorLabels" . | nindent 6 }}
   portLevelMtls:
     {{ .Values.metrics.config.port }}:
-      mode: STRICT
\ No newline at end of file
+      mode: PERMISSIVE
```
mfaizanse, line 12: Is there a way to make only the port `9090` as `PERMISSIVE`?

VladislavPaskar: This change results in:
```
portLevelMtls:
    "9090":
         mode: PERMISSIVE
```
according to [istio docs](https://istio.io/latest/docs/reference/config/security/peer_authentication/), the permissive option will only apply to the 9090 port, all the other ports will remain in the default(which is strict).

mfaizanse (approved):

Issue #15152, opened 2022-08-18
Function controller brings down replicas to the desired minimum for currently scaled up function deployments


pPrecel:
**Description**

Now template for functions deployment always use `minReplicas` as value for the `deployment.spec.replicas` field. It's a little bit confusing and may cause problems when existing deployment is scaled up to a value higher than the minimum. To avoid the potential risk of scaling problems we should think about a better solution and introduce them. Initial conversation and first iteration (first shape) of the solution for this problem were proposed [here](https://github.com/kyma-project/kyma/pull/14704#discussion_r934352178).

kwiatekus:
AFAIU the case is as follows:

Given that:
   deployment replicas is managed by HPA ( HPA managed by function controller ) 
When 
   function deployment's current state is compared with desired state by the control loop
AND
   deployment replicas has been scaled up by HPA
Then
   function controller sets the replicas down to the value of `minReplicas`

But it should do that. The current amount of replicas ( as managed by HPA ) should not be forcefully brought down by function controller

Shouldn't we simply leave that part for HPA to manage?

kwiatekus:
Closing.
There is a follow up bug to be fixed regarding corrupted v1alpha1->v1aloha2 conversion

Issue #14900, opened 2022-07-25
Fix defaulting and validation around the two modes for function scaling ( default vs. external scaler )


kwiatekus:
**Description**

Fix the validation, defaulting logic so that it conforms with the following case description.
Write  tests that validates the scenarios.

1) 
_when_ 
    user defines `spec.scaleConfig.minReplicas` and `spec.scaleConfig.maxReplicas` (higher than min)
_than_
    default ( serverless managed  Horizontal Pod Autoscaler kicks in to scale the function)

2)
_when_ 
    user defines `spec.scaleConfig.minReplicas` and `spec.scaleConfig.maxReplicas` (values equal or max lower than min)
_than_
    validation webhook rejects such configuration

3) 
_when_
  user leaves `spec.scaleConfig` empty 
_then_
  the default HPA scaler would not kick in

4) 
_when_
  user defines `spec.replicas` with X 
_then_
  the deployment has X replicas

5) 
_when_
  user leaves both `spec.replicas` and `spec.scaleConfig` empty 
_then_
  the defaulting webhook should fill in spec.replicas=1

**Reason**
Function CR support scale subresource that would enable scaling functions via external scalers.
Additionally it still offers the way to configure default scaling ( via serverless managed HPA ) based on cpu consumption
Those configuration modes should be tested

**Related**
https://github.com/kyma-project/kyma/issues/15291
https://github.com/kyma-project/kyma/issues/15292

kwiatekus:
Investigate whether  function controller managed HPA scaling the function would simplify the logic ( p 3 ).

dbadura:
I would add case when user to `max` or `min` replicas. 
In this case we should reject such configuration.

kwiatekus:
Works like described since https://github.com/kyma-project/kyma/pull/15576

PR #15576, opened 2022-09-21
Fix conflicting Serverless scaling options

moelsayed:
**Description**

- ...
- ...
- ...

**Related issue(s)**

#15152


moelsayed:
/hold
I am still working on a defaulting bug.

components/function-controller/pkg/apis/serverless/v1alpha2/function_defaults.go
```diff
@@ -65,6 +65,11 @@ func (fn *Function) Default(config *DefaultingConfig) {
 
 func (spec *FunctionSpec) defaultReplicas(config *DefaultingConfig, fn *Function) {
 	if spec.Replicas != nil {
+		if spec.ScaleConfig == nil {
```
dbadura, line 68: Why do we fill `scaleConfig`, if replicas is set?

components/function-controller/design/scaling_functions.md
```diff
@@ -0,0 +1,43 @@
+# Serverless Functions Scaling Modes
+
+## Summary
+Initially, the Serverless API supported the `spec.ScaleConfig` field only. The work flow was as follow:
+- Function resources were defaulted to min/max replicas = 1
+- If the user sets a different scaling config, the function controller will create an HPA resources with the user defined min/max values. 
+- The HPA resource target ref will be the Function runtime deployment.
```
dbadura, line 7: We can target function as it's scalable and update the deployment `replicas` using function `spec`

components/function-controller/pkg/apis/serverless/v1alpha2/function_defaults.go
moelsayed: 1- Show the user the setting. 
2- Validation will break if both are removed. So it should be there with a sane value when/if we remove spec.replicas to disable external scaling.

components/function-controller/design/scaling_functions.md
moelsayed: We actually discussed this. I didn't like initially, it makes sense because this why we avoid having a conflict with other user-created HPA resources to scale the function externally. Also mapping the work flow would be confusing: function.Spec.ScaleConfig.min/max -> Internal HPA resource - > function.Spec.Replicas - > runtime deployment. spec.Replicas :/

components/function-controller/pkg/apis/serverless/v1alpha2/function_defaults.go
dbadura: 1. Why? It will make the output much less readable. Use can see the structure by: reading documentation or by `kubectl explain`.
2. `replicas` is not removed. We can think about defaulting it to 1. 
Additionally this code can introduce bugs when we will decide to create HPA against function as it is mutating `scaleConfig`.

If `scaleConfig` is set to nil, it's very easy to decide to create hpa.

components/function-controller/design/scaling_functions.md
dbadura: This will hide user misconfiguration of scaling.

moelsayed: Can you give an example?

components/function-controller/pkg/apis/serverless/v1alpha2/function_defaults.go
moelsayed: 1. It's not ideal. The idea is to mix both features in the simplest possible way.
2. We don't have a flag to enable/disable external scaling. The simplest way I could find was to use `replicas` for that.
3. If you create an user-defined HPA against the function, it should work fine, since it will only target the `replicas` filed, as it doesn't use the `scalingConfig` field. However, if we change the controller to create internal HPAs with function targets, we will need to change this code as well. I think that's normal and expected..

components/function-controller/design/scaling_functions.md
dbadura: If user set `scaleConfig` he is aware that it enables scaling. In this situation the user shouldn't create any external scaler. As it can introduce bugs. 
Additionally our CR starting behaving strangely because we will set `scaleConfig` to the same values as `replicas` and will wipe user configuration.
The scaling topic should be well documented but it's in other task.

The same situation: user shouldn't create two external scalers as it can undefined behavior.

Adding such behaviour we hide user misconfiguration of scaling and changing the configuration. 

I think that the workflow you mentioned simplify code which is responsible for managing deployments.

kwiatekus (commented):
my 3 cents:

I see the benefits of the proposed idea.
I understand you tried to separate those scalability modes and distinguish which resource is scaled to avoid collisions.
But I think there are also problems.. for instance :
 - if the external scaler is used than it pollutes the scaleConfig (as replicas override the scaleConfig.. or did I get it wrong?)
 - the necessity to clear the replicas field in order to bring back serverless managed HPA. It would be more natural to me to clear the scaleConfig as in "I dont want you to scale it.. I will scale myself"
 - two sources of truth for scaling ( replicas and scaleconifg ) could be error prone and increase maintenance costs


I like to see the `spec.scaleConfig` a way to help user to configure a very basic scaler.
I believe this basic scaler (HPA) should target function.. this way function.spec.replicas would become the single source of truth for replication.
If the user enable other HPAs referencing the function indeed there might be collisions of two or more scalers trying to overrule others and set the number of replicas... but he can always do that, cant he? ( i.e by applying multiple scaledObjects). Its user's responsibility to model scalers that would not conflict with each other.

pPrecel (commented):
In general, this proposal looks ok. I have only one small suggestion for it - I think `mirroring` value of the `spec.replicas` in the `spec.scaleConfig` looks a little bit tricky. As a user, I would be really confused after reading the documentation, where we write that `scaleConfig` is used to define internal hpa, and then after applying an external one the `scaleConfig` has exactly the same values as replicas (?) and this config is following values from replicas? It would be really confusing.

I would suggest the other solution to our problem:

* `spec.replicas` is THE ONLY ONE source of truth for any functionality in the function-controller
*  `spec.scaleConfig` is ONLY used to create an internal scaler (and should be not used in any other way) - if `spec.scaleConfig` == nil, then do nothing
* Internal scalers should refer to function, not deployment

I think the huge advantage of this proposition is simplicity ( simple code == simple documentation == functionality easy to understand ). 

I remember a few discussions about `what if the user would use `scaleConfig` and external scaler at the same time?` and I think this question is totally not a case because if we would make `scaleConfig` optional and users would enable it intentionally then it's not our fault that it would not work :D the exact same case would be when the user will use two (or more) external scalers for one function at the same time.

What do you think about it @moelsayed? I don't want to decide which proposition is better I only want to open a discussion to talk about it because I think it's really important to decide on the such thing right now before we will make a mistake.

kwiatekus (commented):
Thank you for reflecting the discussion points into the latest version of the proposal.
👍

kwiatekus (commented):
Thank you for reflecting the discussion points into the latest version of the proposal.
👍

pPrecel (commented):
In general, the design file and the code look ok to me. I did not test this functionality

components/function-controller/pkg/apis/serverless/v1alpha2/function_defaults_test.go
```diff
@@ -110,6 +112,7 @@ func TestSetDefaults(t *testing.T) {
 						MinReplicas: &two,
 						MaxReplicas: &two,
 					},
+					Replicas: &two,
```
dbadura, line 115: Generally we can remove scaleConfig with two equal values in cases where we don't check it.

components/function-controller/internal/controllers/serverless/utils.go
```diff
@@ -265,7 +265,7 @@ func equalInt32Pointer(first *int32, second *int32) bool {
 	return *first == *second
 }
 
-func isScalingEnabled(instance *serverlessv1alpha2.Function) bool {
+func isScaleConfigEnabled(instance *serverlessv1alpha2.Function) bool {
```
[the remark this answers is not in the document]
moelsayed: This function doesn't really check if the scaling feature is enabled or not anymore. It checks only if `scaleConfig` is set/enabled and it's only used in that context. So, I think it makes sense to rename it and remove the abstraction.

components/function-controller/pkg/apis/serverless/v1alpha2/function_defaults.go
```diff
@@ -58,30 +58,34 @@ type DefaultingConfig struct {
...
+		// TODO: The presets structure and docs should be updated to reflect the new behavior.
+		spec.Replicas = &replicasPreset.Min
+	}
+
 	if spec.ScaleConfig == nil {
-		spec.ScaleConfig = &ScaleConfig{}
+		return
 	}
+
+	// spec.ScaleConfig is SET, but not fully configured, for sanity, we will default MinReplicas, we will also use it as a default spec.Replica
```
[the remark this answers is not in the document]
moelsayed: I am preserving as much as possible from the original behavior. Since I didn't really cover validation in the proposal doc. Since there is a [dedicated issue](https://github.com/kyma-project/kyma/issues/14900). I prefer that we address it there.

components/function-controller/pkg/apis/serverless/v1alpha2/function_defaults_test.go
moelsayed: I this is one also is related to the [defaulting and validation](https://github.com/kyma-project/kyma/issues/14900) issue.

components/function-controller/internal/controllers/serverless/function_reconcile_test.go
```diff
@@ -116,20 +116,20 @@ func TestFunctionReconciler_Reconcile_Scaling(t *testing.T) {
...
-		g.Expect(getConditionReason(function.Status.Conditions, serverlessv1alpha2.ConditionRunning)).To(gomega.Equal(serverlessv1alpha2.ConditionReasonDeploymentUpdated))
+		g.Expect(getConditionStatus(function.Status.Conditions, serverlessv1alpha2.ConditionRunning)).To(gomega.Equal(corev1.ConditionTrue))
+		g.Expect(getConditionReason(function.Status.Conditions, serverlessv1alpha2.ConditionRunning)).To(gomega.Equal(serverlessv1alpha2.ConditionReasonDeploymentReady))
 		deployments := &appsv1.DeploymentList{}
 		g.Expect(resourceClient.ListByLabel(context.TODO(), request.Namespace, fnLabels, deployments)).To(gomega.Succeed())
 		g.Expect(len(deployments.Items)).To(gomega.Equal(1))
 		deployment := &deployments.Items[0]
 		g.Expect(deployment).ToNot(gomega.BeNil())
-		g.Expect(deployment.Spec.Replicas).To(gomega.Equal(&two))
+		g.Expect(deployment.Spec.Replicas).To(gomega.Equal(&one))
```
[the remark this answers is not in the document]
moelsayed: You are correct. However, I am not sure if we should change something here at this point.

components/function-controller/pkg/apis/serverless/v1alpha2/function_defaults.go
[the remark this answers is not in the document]
dbadura: Added: https://github.com/kyma-project/kyma/issues/14900#issuecomment-1273095981

components/function-controller/internal/controllers/serverless/function_reconcile_test.go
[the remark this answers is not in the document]
dbadura: We should revert this change to previues values (2)  as this test case is checking artificial, non existing situation.

components/function-controller/internal/controllers/serverless/function_reconcile_test.go
```diff
@@ -151,10 +146,10 @@ func TestFunctionReconciler_Reconcile_Scaling(t *testing.T) {
 
 		t.Log("updating function to use scalable replicas number")
 		g.Expect(resourceClient.Get(context.TODO(), request.NamespacedName, function)).To(gomega.Succeed())
-		function.Spec.ScaleConfig.MaxReplicas = &four
+		function.Spec.Replicas = &four
```
dbadura, line 149: I would change the name of the test step to something like: `external scaler increase replicas amount`

components/function-controller/internal/controllers/serverless/utils.go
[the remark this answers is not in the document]
dbadura: Why doesn't it really check if scaling is enabled?

The field `scaleConfig` enables and disables scaling, so I would keep to the old name

dbadura (approved):

components/function-controller/design/scaling_functions.md
```diff
@@ -0,0 +1,40 @@
+# Serverless Functions Scaling Modes
+
+## Summary
+Initially, the Serverless API supported the `spec.ScaleConfig` field only. The work flow was as follow:
```
grego952, line 4:
```suggestion
Initially, the Serverless API supported the `spec.ScaleConfig` field only. The workflow was as follows:
```

components/function-controller/design/scaling_functions.md
```diff
@@ -0,0 +1,40 @@
...
+- Function resources were defaulted to min/max replicas = 1
+- If the user sets a different scaling config, the function controller will create an HPA resources with the user defined min/max values.
```
grego952, line 6:
```suggestion
- If you set a different scaling config, Function Controller created HPA resources with the user-defined min/max values. 
```

components/function-controller/design/scaling_functions.md
```diff
@@ -0,0 +1,40 @@
+# Serverless Functions Scaling Modes
+
+## Summary
+Initially, the Serverless API supported the `spec.ScaleConfig` field only. The work flow was as follow:
+- Function resources were defaulted to min/max replicas = 1
+- If the user sets a different scaling config, the function controller will create an HPA resources with the user defined min/max values. 
+- The HPA resource target ref will be the Function runtime deployment.
```
grego952, line 7:
```suggestion
- The HPA resource target ref was the Function runtime deployment.
```

components/function-controller/design/scaling_functions.md
```diff
@@ -0,0 +1,40 @@
...
+- The Function controller will not enforce the runtime deployment `spec.Replicas` any more and it will be handled by the HPA resources.
```
grego952, line 8:
```suggestion
- The Function Controller did not enforce the runtime deployment `spec.Replicas` anymore and it was handled by the HPA resources.
```

components/function-controller/design/scaling_functions.md
grego952, line 4: I changed lines 4-8 to past tense, because we're talking about the previous workflow, correct?

components/function-controller/design/scaling_functions.md
```diff
@@ -0,0 +1,40 @@
...
+
+
+Later, we extended the Serverless API to include a `Scale` subresource, which allows us to directly scale the function resources through the Kubernetes API.
```
grego952, line 11:
```suggestion
Later, the Serverless API was extended by the `Scale` subresource, which allows for direct scaling of the Function resources through the Kubernetes API.
```

components/function-controller/design/scaling_functions.md
```diff
@@ -0,0 +1,40 @@
...
+### Goals
+- Support Function scale subresource and `spec.ScaleConfig` without conflicts.
+- Provide frictionless UX for the feature.
+
+## Proposal
+Describe and implement two different scaling configuration. Both configurations already work to some extent. The point here is have an ergonomic UX and flow.
+
+### External scaling configuration
+This is managed and configured using `spec.Replicas`. It describes the scale subresource use case. It supports:
+- Manually scaling the Function up/down through the API.
```
grego952, line 24:
```suggestion
- Manual scaling of the Function up and down through the API.
```

components/function-controller/design/scaling_functions.md
```diff
@@ -0,0 +1,40 @@
...
+- Configuring an HPA resource with the Function resources as a target.
+- Using an external scaler like KEDA.
```
grego952, line 26:
```suggestion
- Using an external scaler like [KEDA](https://keda.sh/).
```

components/function-controller/design/scaling_functions.md
```diff
@@ -0,0 +1,40 @@
...
+
+### Built-in scaling configuration
+This is managed and enabled by setting `spec.ScaleConfig`. It should be configurable using Busola and it provide the most basic scaling configuration for the function.
```
grego952, line 29:
```suggestion
This is managed and enabled by setting `spec.ScaleConfig`. It is configurable using Busola and it provides the most basic scaling configuration for the Function.
```

components/function-controller/design/scaling_functions.md
```diff
@@ -0,0 +1,40 @@
...
+
+This configuration will be disabled by removing `spec.ScaleConfig`. Busola UI can be extended to allow users to add/remove `spec.ScaleConfig` to manage built-in scaling with minimal effort to the user.
```
grego952, line 31:
```suggestion
This configuration is disabled by removing `spec.ScaleConfig`. The Busola UI can be extended to allow you to add or remove `spec.ScaleConfig` to manage built-in scaling with minimal effort.
```

components/function-controller/design/scaling_functions.md
```diff
@@ -0,0 +1,40 @@
...
+
+## Implementation details
+
+- The controller should support and accept both `spec.Replicas` and `spec.ScaleConfig`. Current validation rule to block this will be removed.
+- `spec.Replcas` is the only source of truth for scaling the function/runtime deployment.
```
grego952, line 36:
```suggestion
- `spec.Replicas` is the only source of truth for scaling the Function/runtime deployment.
```

components/function-controller/design/scaling_functions.md
```diff
@@ -0,0 +1,40 @@
...
+- `spec.ScaleConfig` is only used to configure the controller internal HPA.
+- The internal HPA is removed if `spec.ScaleConfig == nil`
+- The HPA resource created by the controller will still target the Function resources.
```
grego952, line 39:
```suggestion
- The HPA resource created by the controller still targets the Function resources.
```

components/function-controller/design/scaling_functions.md
```diff
@@ -0,0 +1,40 @@
...
+- The current Function status update logic should be fixed to reflect the function current scale.
```
grego952, line 40:
```suggestion
- The current Function status update logic must be fixed to reflect the Function's current scale. 
```

grego952 (approved):

PR #15583, opened 2022-09-21
Add Subscription conversion logic v1alpha1 to v1alpha2 and backwards 

VladislavPaskar:
**Description**

- add following convertions:
   - v1alpha1 -> v1alpha2
   - v1alpha2 -> v1alpha1 
- add unit tests for it
-  restructure the v1alpha2 constants


components/eventing-controller/api/v1alpha2/constants.go
```diff
@@ -0,0 +1,23 @@
+package v1alpha2
+
+const (
+	STANDARD TypeMatching = "standard"
+	EXACT    TypeMatching = "exact"
```
nachtmaar, line 5: I propose to use a prefix like `TypeMatching`. This makes it easier for the different values to be discovered.

```suggestion
	TypeMatchingStandard TypeMatching = "standard"
	TypeMatchingExact     TypeMatching  = "exact"
```

<details>
  <summary>There are other examples (also in the golang stdlib) which follow this idea:</summary>

[source](https://www.sohamkamani.com/golang/enums/)

**net/http package**
- https://pkg.go.dev/net/http#pkg-constants
```go
const (
	MethodGet     = "GET"
	MethodHead    = "HEAD"
	MethodPost    = "POST"
	MethodPut     = "PUT"
	MethodPatch   = "PATCH" // [RFC 5789](https://rfc-editor.org/rfc/rfc5789.html)
	MethodDelete  = "DELETE"
	MethodConnect = "CONNECT"
	MethodOptions = "OPTIONS"
	MethodTrace   = "TRACE"
)
```

**Color package**
- https://pkg.go.dev/image/color#pkg-variables
This is an example where no prefix is required since the package name `color` is already meaningful enough. Example: `color.Black`

```go
var (
	Black       = [Gray16](https://pkg.go.dev/image/color#Gray16){0}
	White       = [Gray16](https://pkg.go.dev/image/color#Gray16){0xffff}
	Transparent = [Alpha16](https://pkg.go.dev/image/color#Alpha16){0}
	Opaque      = [Alpha16](https://pkg.go.dev/image/color#Alpha16){0xffff}
)
```

**Log package**:
```go
const (
	Ldate         = 1 << [iota](https://pkg.go.dev/builtin#iota)     // the date in the local time zone: 2009/01/23
	Ltime                         // the time in the local time zone: 01:23:23
	Lmicroseconds                 // microsecond resolution: 01:23:23.123123.  assumes Ltime.
	Llongfile                     // full file name and line number: /a/b/c/d.go:23
	Lshortfile                    // final file name element and line number: d.go:23. overrides Llongfile
	LUTC                          // if Ldate or Ltime is set, use UTC rather than the local time zone
	Lmsgprefix                    // move the "prefix" from the beginning of the line to before the message
	LstdFlags     = [Ldate](https://pkg.go.dev/log@go1.16.6#Ldate) | [Ltime](https://pkg.go.dev/log@go1.16.6#Ltime) // initial values for the standard logger
)
```

</details>

components/eventing-controller/api/v1alpha1/testing_helpers.go
```diff
@@ -0,0 +1,172 @@
...
+		newSub.Status.Config = nil
+	}
+
+	return newSub
+}
+
+func exemptHandshake(val bool) *bool {
+	exemptHandshake := val
+	return &exemptHandshake
+}
```
nachtmaar, line 60: nit: We do have `utils.Boolptr` already. Is there a reason to still keep this ?

components/eventing-controller/api/v1alpha1/subscription_conversion.go
```diff
@@ -1,19 +1,340 @@
...
 
+const (
+	hubVersionErrMsg     = "hub version is not the expected v1alpha2 version"
+	multipleSourceErrMsg = "subscription contains more than 1 eventSource"
+)
+
 // ConvertTo converts this Subscription to the Hub version (v2).
-func (s *Subscription) ConvertTo(rawObj conversion.Hub) error {
-	_ = rawObj.(*v1alpha2.Subscription)
+func (src *Subscription) ConvertTo(dstRaw conversion.Hub) error { // nolint
```
nachtmaar, line 19: Why do we need the `nolint` ? If we really need it, there should be a comment explaining why we need to exclude linting IMHO

components/eventing-controller/api/v1alpha1/subscription_conversion.go
```diff
@@ -1,19 +1,340 @@
...
+	"strings"
+
 	"github.com/kyma-project/kyma/components/eventing-controller/api/v1alpha2"
+	"github.com/pkg/errors"
 	"sigs.k8s.io/controller-runtime/pkg/conversion"
 )
 
+const (
+	hubVersionErrMsg     = "hub version is not the expected v1alpha2 version"
+	multipleSourceErrMsg = "subscription contains more than 1 eventSource"
```
nachtmaar, line 15: Should we prefix this ? To make errors easily discoverable to the API users.

```suggestion
	errorHubVersionErrMsg     = "hub version is not the expected v1alpha2 version"
	errorMultipleSourceErrMsg = "subscription contains more than 1 eventSource"
```

Alternative would be to provide our own error package.

components/eventing-controller/api/v1alpha1/subscription_conversion.go
```diff
@@ -1,19 +1,340 @@
...
+	// event types
+	src.setV2StatusTypes(dst)
+
+	// Backend-specific status
+
+	// BEB
+	src.bebBackendStatusToV2(dst)
+
+	// NATS
+	src.natsBackendStatusToV2(dst)
```
nachtmaar, line 71: Why do we need to copy the status ?  The status does not need to be copied as it will be calculated by the controller. Only the specification is important or am I missing something ?

components/eventing-controller/api/v1alpha1/subscription_conversion.go
nachtmaar: Same question for the other occurences :)

components/eventing-controller/api/v1alpha1/subscription_conversion.go
```diff
@@ -1,19 +1,340 @@
...
+
+	return nil
+}
+
+// ConvertFrom converts this Subscription from the Hub version (v2).
```
nachtmaar, line 76:
```suggestion
// ConvertFrom converts this Subscription from the Hub with version v2 to v1.
```

components/eventing-controller/api/v1alpha1/subscription_conversion.go
```diff
@@ -1,19 +1,340 @@
...
+	"github.com/pkg/errors"
 	"sigs.k8s.io/controller-runtime/pkg/conversion"
 )
 
+const (
+	hubVersionErrMsg     = "hub version is not the expected v1alpha2 version"
+	multipleSourceErrMsg = "subscription contains more than 1 eventSource"
+)
+
 // ConvertTo converts this Subscription to the Hub version (v2).
```
nachtmaar, line 18:
```suggestion
// ConvertTo converts this Subscription in version v1 to the Hub version v2.
```

components/eventing-controller/api/v1alpha1/subscription_conversion.go
```diff
@@ -1,19 +1,340 @@
...
+	if src.Spec.Config != nil {
+		dst.natsSpecConfigToV1(src)
+	}
+
+	// Conditions
+	for _, condition := range src.Status.Conditions {
+		dst.Status.Conditions = append(dst.Status.Conditions, ConditionV2ToV1(condition))
+	}
+
+	dst.Status.Ready = src.Status.Ready
```
nachtmaar, line 125: Same question as above regarding the copy of the status.

components/eventing-controller/api/v1alpha1/subscription_conversion.go
```diff
@@ -1,19 +1,340 @@
...
+	dst.setV1CleanEvenTypes(src)
+	dst.bebBackendStatusToV1(src)
+	dst.natsBackendStatusToV1(src)
 
 	return nil
 }
 
-func (s *Subscription) ConvertFrom(rawObj conversion.Hub) error {
-	_ = rawObj.(*v1alpha2.Subscription)
+// setV2TypeMatching sets the default typeMatching on the v1alpha2 Subscription version
```
nachtmaar, line 134: nit:
```suggestion
// setV2TypeMatching sets the default typeMatching on the v1alpha2 Subscription version.
```

components/eventing-controller/api/v1alpha1/subscription_conversion.go
```diff
@@ -1,19 +1,340 @@
...
+func (src *Subscription) setV2TypeMatching(dst *v1alpha2.Subscription) {
+	dst.Spec.TypeMatching = v1alpha2.EXACT
+}
+
+// setV2ProtocolFields converts the protocol-related fields from v1alpha1 to v1alpha2 Subscription version
```
nachtmaar, line 139: nit: Please use the godot linter (from golangci_lint) to find the other missing dots :)
```suggestion
// setV2ProtocolFields converts the protocol-related fields from v1alpha1 to v1alpha2 Subscription version.
```

components/eventing-controller/api/v1alpha1/subscription_conversion.go
```diff
@@ -1,19 +1,340 @@
...
+		dst.Spec.Types = append(dst.Spec.Types, filter.EventType.Value)
+	}
 	return nil
 }
+
+// natsSpecConfigToV2 converts the v1alpha2 Spec config to v1alpha1
+func (src *Subscription) natsSpecConfigToV1(dst *v1alpha2.Subscription) {
+	if maxInFlightMessages, ok := dst.Spec.Config[v1alpha2.MaxInFlightMessages]; ok {
+		intVal, err := strconv.Atoi(maxInFlightMessages)
+		if err == nil {
```
nachtmaar, line 227: What if there is an error ?

components/eventing-controller/api/v1alpha1/subscription_conversion_test.go
```diff
@@ -0,0 +1,202 @@
...
+						CleanType:    OrderDeletedEventType,
+					},
+				}),
+				v1alpha2.WithBEBStatusFields(),
+			),
+		},
+	}
+
+	for _, testCase := range testCases {
+		t.Run(testCase.name, func(t *testing.T) {
```
nachtmaar, line 130: nit: Since 2 methods are tested here I propose to use a **dedicated** `t.Run` with a **specific name** **for each method**.
This makes it easier in the case that the test fails to spot if there is a problem in converting from v1 -> v2 or the other way around :)

components/eventing-controller/api/v1alpha1/subscription_conversion_test.go
```diff
@@ -0,0 +1,202 @@
...
+			err := v1ToV2(testCase.alpha1Sub, convertedV1Alpha2)
+			if err != nil && testCase.wantErrMsg != "" {
+				require.Equal(t, err.Error(), testCase.wantErrMsg)
+			} else {
+				require.NoError(t, err)
+				v1ToV2Assertions(t, testCase.alpha2Sub, convertedV1Alpha2)
+			}
+
+			// test ConvertFrom
+			log.Print("Test v2 to v1 conversion")
```
nachtmaar, line 144: Why not use `t.Log` instead ?

components/eventing-controller/api/v1alpha1/subscription_conversion_test.go
nachtmaar: Another idea would be to move the definition of the test cases outside of the method and have 2 different go test functions :)

nachtmaar: I am wondering why we don't have a controller integration test or something like that where the check that the conversion logic is correctly integrated ?

components/eventing-controller/api/v1alpha2/testing_helpers.go
```diff
@@ -0,0 +1,128 @@
...
+			Type:   ConditionSubscriptionActive,
+			Status: "true",
+		},
+		{
+			Type:   ConditionSubscribed,
+			Status: "false",
+		}}
+)
+
+// +kubebuilder:object:generate=false
```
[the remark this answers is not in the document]
VladislavPaskar: this is required for the `controller-gen`, because it tries make a deep copy method for all the types declared in the api/v1alpha* packages, otherwise it throws you an error

components/eventing-controller/api/v1alpha2/testing_helpers.go
```diff
@@ -0,0 +1,128 @@
...
+	return func(sub *Subscription) {
+		if statusTypes == nil {
+			sub.Status.InitializeEventTypes()
+			return
+		}
+		sub.Status.Types = statusTypes
+	}
+}
+
+func WithWebhookAuthForBEB() SubscriptionOpt {
```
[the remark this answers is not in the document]
VladislavPaskar: The idea was to have non-shared testing methods to avoid the recursive imports in the both of the api packages. and when the v1alpha1 will be deprecated, we will get rid of it anyhow, so I think we could leave it as is.

components/eventing-controller/api/v1alpha2/testing_helpers.go
```diff
@@ -0,0 +1,128 @@
+package v1alpha2
```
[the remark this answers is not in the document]
VladislavPaskar: There was a problem with the recursive package imports, which took a lot of time, that's why I decided to go this way.

components/eventing-controller/api/v1alpha1/subscription_conversion.go
VladislavPaskar: The problem is the naming of the Subscription, on which we call this method. In the ConvertTo its' `src` and in the ConvertFrom it's the `dst`, which makes sense. But linter doesn't seem to like it.

components/eventing-controller/api/v1alpha1/subscription_conversion.go
VladislavPaskar: improved this, I propagate the error now to the main convert function.

components/eventing-controller/api/v1alpha1/subscription_conversion_test.go
VladislavPaskar: I added t.Run for every conversion. I am not sure if we need the integration tests for that, because in that case we would test the kubebuilder logic and not our. Our is already covered by this unit test IMO.

components/eventing-controller/api/v1alpha1/subscription_conversion.go
VladislavPaskar: Good point, but I the Status contains some configurations and I wanted to make sure, that if there is no reconciliation(there I am not sure) right after the conversion which would populate the status, that we would have the status there.

components/eventing-controller/api/v1alpha1/subscription_conversion.go
VladislavPaskar: https://github.com/kyma-project/kyma/pull/15583#discussion_r978445769

components/eventing-controller/api/v1alpha1/subscription_conversion_test.go
nachtmaar: I don't agree. We would test the integration of both to ensure that our code is correctly integrated into the kubebuilder logic. Our conversion logic could be fully correct but still it not used in the webhook.

components/eventing-controller/api/v1alpha1/subscription_conversion.go
nachtmaar: I don't think an empty status is a problem. This is also the case if we don't have a webhook and the controller is not running.
Do you have an example where the status contains configuration which is required to be copied ?

components/eventing-controller/api/v1alpha2/testing_helpers.go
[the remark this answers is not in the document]
nachtmaar:
```
$ make generate-local
controller-gen object:headerFile="hack/boilerplate.go.txt" paths="./..."
github.com/kyma-project/kyma/components/eventing-controller/api/v1alpha2:-: invalid type: func(subscription *github.com/kyma-project/kyma/components/eventing-controller/api/v1alpha2.Subscription)
```

components/eventing-controller/api/v1alpha1/subscription_conversion.go
nachtmaar:
```
$ revive api/v1alpha1/subscription_conversion.go
api/v1alpha1/subscription_conversion.go:77:1: receiver name dst should be consistent with previous receiver name src for Subscription
```

components/eventing-controller/api/v1alpha2/testing_helpers.go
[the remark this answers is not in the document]
nachtmaar: For now we could also move the testing code to the unit test itself. This would fix the problem at least for now until the code needs to be reused.

components/eventing-controller/api/v1alpha2/testing_helpers.go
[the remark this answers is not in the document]
nachtmaar: Let's check your original idea :)

components/eventing-controller/api/v1alpha1/subscription_conversion.go
```diff
@@ -1,19 +1,346 @@
...
+		intVal, err := strconv.Atoi(maxInFlightMessages)
+		if err != nil {
+			return err
+		}
+		if err == nil {
+			src.Spec.Config = &SubscriptionConfig{
+				MaxInFlightMessages: intVal,
+			}
+		}
+	}
```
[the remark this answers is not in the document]
VladislavPaskar: good point

components/eventing-controller/api/v1alpha2/testing_helpers.go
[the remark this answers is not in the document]
nachtmaar: We found a solution together. The changes are:
- Don't expose testing only fixtures.
- Reuse existing functional option functions from testing pkg
- Remove unused testing functions
- Convert the conversion test to a black box test
- Ignoring controller-gen for the SubscriptionOpt is no longer required (as it resides in testing package)
- Moving the new v2 test functions to the generic testing package (for reuse)

components/eventing-controller/api/v1alpha2/testing_helpers.go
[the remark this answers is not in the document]
nachtmaar: Fixed, see https://github.com/kyma-project/kyma/pull/15583#discussion_r981059655

nachtmaar (approved):
Thank you for your contribution 👍

Issue #15402, opened 2022-09-08
Implement data type to be used for advanced templating of function managed workloads


kwiatekus:
**Description**

Implement a reduced PodSpecTemplate-like type to be used as
runtime Template and build Template.

Propose the structure in the spec, i.e: `spec.buildTemplate` vs `spec.template.build`

**Reasons**

Give serverless users more flexibility to configure buildtime and runtime pods of the function.

PR #15592, opened 2022-09-22
Expand function api by adding templates

dbadura:
**Description**

- Add api for templating

**Related issue(s)**
https://github.com/kyma-project/kyma/issues/15402


components/function-controller/config/crd/.gitignore
```diff
@@ -0,0 +1 @@
+crd-serverless.yaml
```
dbadura, line 1: This file is not used anywhere and it's copied to the `installation` directory. It doesn't make sens to keep it in git as is transient file.

components/function-controller/pkg/apis/serverless/v1alpha2/function_types.go
```diff
@@ -118,6 +120,67 @@ type ResourceConfiguration struct {
...
+	Spec *PodSpecTemplate `json:"spec,omitempty"`
+}
+
+type Templates struct {
+	// Standard object's metadata.
+	// This metadata is applied on functionPod and buildPod.
+	// Can be overwritten by specific pod metadata
+	// More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#metadata
+	// +optional
+	Metadata *MetadataTemplate `json:"metadata,omitempty"`
```
kwiatekus, line 169: what is the purpose of metadata on this level ?

components/function-controller/pkg/apis/serverless/v1alpha2/function_types.go
```diff
@@ -118,6 +120,67 @@ type ResourceConfiguration struct {
 	Function *ResourceRequirements `json:"function,omitempty"`
 }
 
+type PodSpecTemplate struct {
+	// Compute Resources required by container created by function controller.
+	// +optional
+	Resources *ResourceRequirements `json:"resources,omitempty"`
```
kwiatekus, line 126: I believe on this level there should not be any "our" abstraction(i.e named presets "S", "L") but only k8s specs for  workload resources (v1.ResourceRequirements) that gets defaulted from the high level spec fields and/or can be fine tuned (customised ) here on this level

kwiatekus (changes requested):

components/function-controller/pkg/apis/serverless/v1alpha2/function_types.go
dbadura: To not repeat annotations and labels for build's pod and function's pods

components/function-controller/pkg/apis/serverless/v1alpha2/function_types.go
dbadura: We can add on root level field:
```
resourceProfiles:
  build: L
  runtime: S
```
and then default this template using profile.

components/function-controller/pkg/apis/serverless/v1alpha2/function_types.go
pPrecel: but we should repeat them, right? the main goal of creating the `templates`  field is to support config separation of build and function pods. Or maybe I misunderstood something...

kwiatekus: ok, so for shared ones?

components/function-controller/pkg/apis/serverless/v1alpha2/function_types.go
```diff
@@ -118,6 +120,67 @@ type ResourceConfiguration struct {
...
+	// +optional
+	BuildJobTemplate *PodTemplate `json:"buildJobTemplate,omitempty"`
+
+	// Additional specification for function pod
+	// +optional
+	FunctionPodTemplate *PodTemplate `json:"functionPodTemplate,omitempty"`
+
+	// Configuration of available volumes for function pod and buildjob pod
+	// +optional
+	Volumes []v1.Volume `json:"volumes,omitempty"`
```
dbadura, line 181: I would move `Volumes` to build and function to not mix them here.

components/function-controller/pkg/apis/serverless/v1alpha2/function_types.go
kwiatekus: I dont think we need to add anything on root level but reuse those profiles that already exist , ie:
```yaml
spec:
  resourceConfiguration:
    function:
      profile: S
```

components/function-controller/pkg/apis/serverless/v1alpha2/function_types.go
dbadura: Yes, it's used for shared ones to not create very complicated structure and duplicate data. 
If user want to set specific labels for function's pod or build's pod, the user is able to customize it.

components/function-controller/pkg/apis/serverless/v1alpha2/function_types.go
dbadura: I think that the structure is too big to what we want to achieve.
But we can leave it as it's if we don't want to change it.

components/function-controller/pkg/apis/serverless/v1alpha2/function_types.go
pPrecel: I would suggest moving this field to the `PodTemplate` struct. If we don't do this, we will lead to a possible situation when the volume mounted to one container would be unused in the second one

pPrecel: Same topic: https://github.com/kyma-project/kyma/pull/15592/files#r985779218

dbadura: Moved.

components/function-controller/pkg/apis/serverless/v1alpha2/function_types.go
pPrecel: > I dont think we need to add anything on root level but reuse those profiles that already exist , ie:
> 
> ```yaml
> spec:
>   resourceConfiguration:
>     function:
>       profile: S
> ```

this structure will work only until the new API version is released. I think @dbadura is right we should add another structure that would represent the t-shirt functionality after removing the `resourceConfiguration` struct. It may look like:

User creates function:

```
...
spec:
  resourceConfiguration:
    function:
      profile: S
```

Webhook moves the `resourceConfiguration` to the new place and translates it to the right resources:

```
...
spec:
  resourceConfiguration:
    function:
      profile: S
  resourceProfiles:
    function: S
  templates:
    functionPodTemplate:
      spec:
        resources:
        [...]
```

The new API version (v1/v1beta1/v1alpha3) would look like this (after conversion):
```
spec:
  resourceProfiles:
    function: S
  templates:
    functionPodTemplate:
      spec:
        resources:
        [...]
```

pPrecel: Here is another question. What do you think about NOT keeping old (high-level) fields after soft conversion? (soft means around one API spec version). 

Maybe we should accept high-level input from users and remove these fields after copying them into new ones. Then the example after moving and translating profiles by webhook would look like:

```
...
spec:
  resourceProfiles:
    function: S
  templates:
    functionPodTemplate:
      spec:
        resources:
        [...]
```

It looks exactly the same as the target spec we will introduce in the new version.

components/function-controller/pkg/apis/serverless/v1alpha2/function_types.go
```diff
@@ -118,6 +120,67 @@ type ResourceConfiguration struct {
...
+	// +optional
+	Metadata *MetadataTemplate `json:"metadata,omitempty"`
+
+	// Additional specification for build job pod
+	// +optional
+	BuildJobTemplate *PodTemplate `json:"buildJobTemplate,omitempty"`
+
+	// Additional specification for function pod
+	// +optional
+	FunctionPodTemplate *PodTemplate `json:"functionPodTemplate,omitempty"`
```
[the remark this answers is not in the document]
dbadura: It's good advice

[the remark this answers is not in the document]
dbadura: Done

components/function-controller/pkg/apis/serverless/v1alpha2/function_types.go
dbadura: I will remove this field.

moelsayed: @dbadura why? It made sense.

moelsayed:
Looks very good :+1:

components/function-controller/pkg/apis/serverless/v1alpha2/function_types.go
anoipm: @moelsayed 
IMO we have complicated logic for calculating some fields - merging, defaulting, ... It complicate our code and do it harder for maintain. 

This field complicate our code because we should merge it with `metadata` inside templates: `buildJobTemplate` and `functionPodTemplate`. Remove it in the future needs change API version. Maybe users don't need this functionality now.

When we don't add it now, code is simpler. We can add this in every moment.

components/function-controller/design/2022-10-v1alpha2-templating.md
```diff
@@ -0,0 +1,67 @@
+# Function templating
```
anoipm, line 1: Could we move this document to `adr` directory instead of `design`. ADR (sometimes ADL) is well known short. 

More about ADR: https://github.com/joelparkerhenderson/architecture-decision-record

kwiatekus (dismissed):

anoipm (approved):

anoipm (approved):

anoipm (approved):

Issue #15473, opened 2022-09-13
Bucket marks are set too small in histogram metric `event_publish_to_messaging_server_latency_bucket`


vvxxvvxx:
**Description**

The bucket marks now set in the histogram metric `event_publish_to_messaging_server_latency_bucket` are too small as below. As it's milliseconds (see https://github.com/friedrichwilken/kyma/blob/67786a940a8af6a82110f9be36a16c52171671f2/components/event-publisher-proxy/pkg/metrics/collector.go#L98), so 0.005ms, 0.01ms, 0.025ms, 0.05ms, 0.1ms, 0.25ms are too small and it doesn't make sense to collect them. (We could define the buckets using prometheus.DefBuckets .)
```
event_publish_to_messaging_server_latency_bucket{destination_service="nats://eventing-nats.kyma-system.svc.cluster.local:4222",response_code="204",le="0.005"} 0
event_publish_to_messaging_server_latency_bucket{destination_service="nats://eventing-nats.kyma-system.svc.cluster.local:4222",response_code="204",le="0.01"} 0
event_publish_to_messaging_server_latency_bucket{destination_service="nats://eventing-nats.kyma-system.svc.cluster.local:4222",response_code="204",le="0.025"} 0
event_publish_to_messaging_server_latency_bucket{destination_service="nats://eventing-nats.kyma-system.svc.cluster.local:4222",response_code="204",le="0.05"} 0
event_publish_to_messaging_server_latency_bucket{destination_service="nats://eventing-nats.kyma-system.svc.cluster.local:4222",response_code="204",le="0.1"} 0
event_publish_to_messaging_server_latency_bucket{destination_service="nats://eventing-nats.kyma-system.svc.cluster.local:4222",response_code="204",le="0.25"} 0
event_publish_to_messaging_server_latency_bucket{destination_service="nats://eventing-nats.kyma-system.svc.cluster.local:4222",response_code="204",le="0.5"} 0
event_publish_to_messaging_server_latency_bucket{destination_service="nats://eventing-nats.kyma-system.svc.cluster.local:4222",response_code="204",le="1"} 21
event_publish_to_messaging_server_latency_bucket{destination_service="nats://eventing-nats.kyma-system.svc.cluster.local:4222",response_code="204",le="2.5"} 57
event_publish_to_messaging_server_latency_bucket{destination_service="nats://eventing-nats.kyma-system.svc.cluster.local:4222",response_code="204",le="5"} 58
[...]

PR #15643, opened 2022-09-26
Add exponential buckets for the Event publisher proxy latency histogram

marcobebway:
**Description**

Add exponential buckets for the Event publisher proxy latency histogram.

**Related issue(s)**

- https://github.com/kyma-project/kyma/issues/15473.


components/event-publisher-proxy/pkg/metrics/collector_test.go
```diff
@@ -0,0 +1,36 @@
...
+	assert.NotNil(t, collector.errors)
+	assert.NotNil(t, collector.errors.MetricVec)
+	assert.NotNil(t, collector.latency)
+	assert.NotNil(t, collector.latency.MetricVec)
+	assert.NotNil(t, collector.eventType)
+	assert.NotNil(t, collector.eventType.MetricVec)
+	assert.NotNil(t, collector.requests)
+	assert.NotNil(t, collector.requests.MetricVec)
+
+	// then
```
k15r, line 32: nit:
```suggestion
```

components/event-publisher-proxy/Makefile
```diff
@@ -23,5 +23,8 @@ clean: resolve_clean
 resolve_clean:
 	rm -rf vendor
 
-build-local: test-local 
+build-local: test-local
 
+.PHONY: generate
+generate:
```
k15r, line 29: would it make sense to have the generate target as a dependency on other targets such as `test`?

k15r: we agreed that this will part of a bigger ticket for recreating a makefile without dependency on `generic-make-go.mk`

components/event-publisher-proxy/pkg/metrics/collector_test.go
marcobebway: Done

k15r (approved):

Issue #15363, opened 2022-09-05
migrate NATS/Jetstream controller to use new CRD version


k15r:
Acceptance
- [ ] use new crd in controller and tests

raypinto:
Current status:
Backend/Handler PR: https://github.com/kyma-project/kyma/pull/15647
Reconciler is in progress. Should be ready for review next week.

raypinto:
The reconciler implementation with unit tests are part of [this PR](https://github.com/kyma-project/kyma/pull/15722).
The integration tests will be added as part of [this issue](https://github.com/kyma-project/kyma/issues/15706).

PR #15647, opened 2022-09-26
JetStream handler version v1alpha2

raypinto:
**Description**

- Used the v1alpha2 CRD version for the JetStream handler
- Implemented a new Cleaner for JetStream

**Related issue(s)**

#15363


components/eventing-controller/pkg/backend/jetstreamv2/utils.go
```diff
@@ -0,0 +1,222 @@
...
+// It uses the crypto/md5 lib to return a string of 32 characters as recommended by the JetStream
+// documentation https://docs.nats.io/running-a-nats-service/nats_admin/jetstream_admin/naming.
+func computeConsumerName(subscription *eventingv1alpha2.Subscription, subject string) string {
+	cn := subscription.Namespace + separator + subscription.Name + separator + subject
+	h := md5.Sum([]byte(cn)) // #nosec
+	return hex.EncodeToString(h[:])
+}
+
+// computeNamespacedSubjectName returns Kubernetes namespaced name of the given subscription along with the subject.
+func computeNamespacedSubjectName(subscription *eventingv1alpha2.Subscription, subject string) string {
```
mfaizanse, line 190: Do we have unit tests for each individual util methods: NewSubscriptionSubjectIdentifier, computeConsumerName, NamespacedName, createKeyPrefix, getStreamConfig etc.?

raypinto: Most of them are in place. Will add if anything is missing!

raypinto:
Blocked by: https://github.com/kyma-project/kyma/issues/15655

components/eventing-controller/api/v1alpha2/subscription_types.go
```diff
@@ -79,6 +80,14 @@ func (s Subscription) MarshalJSON() ([]byte, error) {
 	return json.Marshal(a)
 }
 
+func (s *Subscription) GetMaxInFlightMessages() int {
+	val, err := strconv.Atoi(s.Spec.Config[MaxInFlightMessages])
+	if err != nil {
+		return defaultMaxInFlightMessages
```
mfaizanse, line 86: If the user has provided incorrect value then throw error instead of using default value. Otherwise, it would not be consistent with the information visible in the Sub CR.

components/eventing-controller/pkg/backend/cleaner/jetstream.go
```diff
@@ -1,6 +1,16 @@
...
+import (
+	"regexp"
+
+	"github.com/kyma-project/kyma/components/eventing-controller/logger"
+)
+
+var (
+	invalidSourceSegment = regexp.MustCompile(`[\s.>*]`)
+
+	invalidEventTypeSegment = regexp.MustCompile(`[\s>*]`)
```
mfaizanse, line 12: (nit) To me, segments means a whole block of characters. Maybe the name `invalidSourceCharacters` and `invalidEventTypeCharacters`?

components/eventing-controller/pkg/backend/jetstreamv2/utils.go
```diff
@@ -0,0 +1,222 @@
...
+
+	return unique
+}
+
+// getCleanEventTypes returns a list of clean eventTypes from the unique types in the subscription.
+func getCleanEventTypes(sub *eventingv1alpha2.Subscription, cleaner cleaner.Cleaner) ([]eventingv1alpha2.EventType, error) {
+	// TODO: Put this in the validation webhook
+	if sub.Spec.Types == nil || sub.Spec.Source == "" {
+		return []eventingv1alpha2.EventType{}, errors.New("event source and types must be provided")
+	}
```
mfaizanse, line 131: (nit) source is not used in this method, then let's only check for types?
```
if sub.Spec.Types == nil {
	return []eventingv1alpha2.EventType{}, errors.New("event types must be provided")
}
```

components/eventing-controller/pkg/backend/jetstreamv2/utils.go
```diff
@@ -0,0 +1,222 @@
...
+// computeNamespacedSubjectName returns Kubernetes namespaced name of the given subscription along with the subject.
+func computeNamespacedSubjectName(subscription *eventingv1alpha2.Subscription, subject string) string {
+	return subscription.Namespace + separator + subscription.Name + separator + subject
+}
+
+//----------------------------------------
+// CloudEvent utils
+//----------------------------------------
+
+func convertMsgToCE(msg *nats.Msg) (*cev2event.Event, error) {
```
mfaizanse, line 198: Instead of replicating this method, lets extract it our in some common pkg and re-use in both JetStreamV1 and JetStreamV2 ?

components/eventing-controller/pkg/backend/jetstreamv2/utils.go
```diff
@@ -0,0 +1,222 @@
...
+		return nil, err
+	}
+	return &event, nil
+}
+
+//----------------------------------------
+// Testing utils
+//----------------------------------------
+
+func getCleanEventType(eventType string, cleaner cleaner.Cleaner) (string, error) {
```
mfaizanse, line 214: If this is only used in tests, then lets move it to testing pkg or in test file?

components/eventing-controller/pkg/backend/jetstreamv2/utils_test.go
```diff
@@ -0,0 +1,517 @@
...
+	evtestingv2 "github.com/kyma-project/kyma/components/eventing-controller/testing/v2"
+	"github.com/stretchr/testify/require"
+)
+
+// maxJetStreamConsumerNameLength is the maximum preferred length for the JetStream consumer names
+// as per https://docs.nats.io/running-a-nats-service/nats_admin/jetstream_admin/naming
+const (
+	subName                        = "subName"
+	subNamespace                   = "subNamespace"
+	maxJetStreamConsumerNameLength = 32
```
mfaizanse, line 24: Shouldn't `maxJetStreamConsumerNameLength` be defined in `types.go` and reused in both code and test? 
Also a nit, would const defined in _test.go file be visible/accessible in whole pkg?

components/eventing-controller/pkg/backend/jetstreamv2/test_helpers.go
```diff
@@ -0,0 +1,94 @@
...
+	} else {
+		ctx = binding.WithForceStructured(ctx)
+	}
+	if err := client.Send(ctx, *event); err != nil {
+		return err
+	}
+	return nil
+}
+
+func AddJSCleanEventTypesToStatus(sub *v1alpha2.Subscription, cleaner cleaner.Cleaner) error {
```
mfaizanse, line 87: Should we move this file to testing pkg? The methods seems useful for reconciler testing as well!

components/eventing-controller/api/v1alpha2/constants.go
```diff
@@ -20,4 +20,6 @@ const (
 	WebhookAuthClientSecret = "clientSecret"
 	WebhookAuthTokenURL     = "tokenUrl"
 	WebhookAuthScope        = "scope"
+
+	defaultMaxInFlightMessages = 10
```
mfaizanse, line 24: check if we still need it after changes. IMO, not a good idea to replicate configs, common configs could be in separate pkg so it can be reused in multiple places.

components/eventing-controller/pkg/backend/nats/jetstream/jetstream_test.go
```diff
@@ -43,59 +48,77 @@ type jetStreamClient struct {
 // created when no stream exists in JetStream.
 func TestJetStreamInitialize_NoStreamExists(t *testing.T) {
```
mfaizanse, line 49: should we move these common tests into a separate file? so that later when we depreciate code we do not delete these as well.

components/eventing-controller/api/v1alpha2/subscription_types.go
raypinto: yeah you're right! changed

components/eventing-controller/pkg/backend/cleaner/jetstream.go
raypinto: done!

components/eventing-controller/pkg/backend/jetstreamv2/utils.go
raypinto: agreed! good catch

components/eventing-controller/pkg/backend/jetstreamv2/utils.go
raypinto: done! moved to utils

components/eventing-controller/pkg/backend/jetstreamv2/utils.go
raypinto: Its actually used in the same file! I additionally wrote tests for it

components/eventing-controller/pkg/backend/jetstreamv2/utils_test.go
raypinto: Actually this is just used for tests to confirm that the md5 32bit hash actually generates 32 characters. So I left it here

components/eventing-controller/pkg/backend/jetstreamv2/test_helpers.go
raypinto: Actually this was inside the testing file. I just exported it since I needed it in the shared tests for v1 and v2. I don't think its needed in the reconciler tests. If I do, I'll move it later

components/eventing-controller/api/v1alpha2/constants.go
raypinto: agreed! removed!

components/eventing-controller/pkg/backend/nats/jetstream/jetstream_test.go
raypinto: Actually all the tests are being reworked on. I would say its ok for now. I will add a comment if needed

raypinto: Thinking about it, I think its a good idea to do it! Done

mfaizanse (approved):

PR #15651, opened 2022-09-26
Restructure test files ec

friedrichwilken:
This PR renames and restructures the `_test.go` files in eventing controller.

1. Tests that do not use private properties of the system under test (sut) are moved into the `_test` package.
2. Files with tests that need private properties of the sut are renamed from `_test.go` to `_internal_test.go`.
3. Files that contain 
    - unit tests are renamed from `_test,go` to `_unit_test.go`.
    - integration tests are renamed from `_test.go` to `_integration_test.go`.
    - component tests are renamed from `_test.go` to `_component_test.go`.
 
This is a preparation to introduce build tags.


components/eventing-controller/api/v1alpha1/subscription_types_unit_test.go
```diff
@@ -1,4 +1,5 @@
-package v1alpha1
+// todo unit
```
VladislavPaskar, line 1: what about this todo ?

components/eventing-controller/api/v1alpha1/condition_unit_test.go
```diff
@@ -0,0 +1,474 @@
...
+		name            string
+		givenConditions []v1alpha1.Condition
+	}{
+		{
+			name:            "Conditions empty",
+			givenConditions: v1alpha1.MakeSubscriptionConditions(),
+		},
+		{
+			name: "Conditions partially initialized",
+			givenConditions: func() []v1alpha1.Condition {
```
VladislavPaskar, line 27: since you remove the callback from the previous test, I think you can improve it here also:
```suggestion
			givenConditions: []v1alpha1.Condition{
				{
					Type:               v1alpha1.ConditionSubscribed,
					LastTransitionTime: metav1.Now(),
					Status:             corev1.ConditionUnknown,
				},
			},
```

components/eventing-controller/api/v1alpha1/subscription_types_unit_test.go
```diff
@@ -7,6 +8,8 @@ import (
 	"github.com/stretchr/testify/require"
 
 	"github.com/kyma-project/kyma/components/eventing-controller/pkg/env"
+
+	"github.com/kyma-project/kyma/components/eventing-controller/api/v1alpha1"
 )
 
 const (
```
VladislavPaskar, line 15: the `orderCreatedEventType` is redeclared in the same package, can you please remove it. 

Ideally you could move some common files into the new `fixtures_test` file.

components/eventing-controller/api/v1alpha2/condition_unit_test.go
```diff
@@ -0,0 +1,450 @@
+// todo unit
```
VladislavPaskar, line 1: same thing, what is this todo about ?

components/eventing-controller/api/v1alpha2/condition_unit_test.go
```diff
@@ -0,0 +1,450 @@
...
+		name            string
+		givenConditions []v1alpha2.Condition
+	}{
+		{
+			name:            "Conditions empty",
+			givenConditions: v1alpha2.MakeSubscriptionConditions(),
+		},
+		{
+			name: "Conditions partially initialized",
+			givenConditions: func() []v1alpha2.Condition {
```
VladislavPaskar, line 27: let's remove the callback here:
```suggestion
			givenConditions: []v1alpha2.Condition{
				{
					Type:               v1alpha2.ConditionSubscribed,
					LastTransitionTime: metav1.Now(),
					Status:             corev1.ConditionUnknown,
				},
			},
```

components/eventing-controller/controllers/backend/reconciler_internal_unit_test.go
```diff
@@ -0,0 +1,93 @@
+// todo unit
```
VladislavPaskar, line 1: what about this todo ? 😄

components/eventing-controller/api/v1alpha1/subscription_types_unit_test.go
friedrichwilken: In the next step, after all the test files are restructured, we will a build tags like:
```
// +build integration
```
so we can run unit, component and integration separately.
The todo helps to define what tag there should be.

The original plan was to do it all in one big pr at the same time but that makes debugging too hard and messy.

components/eventing-controller/api/v1alpha1/subscription_types_unit_test.go
friedrichwilken: working on the tests in parallel is terrible.

VladislavPaskar (approved):

PR #15704, opened 2022-10-04
rename remaining test files in `event-publisher-proxy`

friedrichwilken:
renamed all remaining test files from `_test.go` to `_unit_test.go` or `integration_test.go` accordingly. All test files in `event-publisher-proxy` should follow this naming pattern now.


raypinto (commented):
Why is the `handler_health_test.go` renamed as unit_test for beb and integration test for nats? They pretty much do the same thing from first glance. They look like unit tests to me

raypinto:
whats the difference between integration test and component tests?

raypinto (approved):

friedrichwilken:
This is blocked for now, since #15672 might do the same things anyhow.

Issue #14717, opened 2022-06-30
Application Gateway handling redirects


janmedrek:
**Description**

We need to set up and describe the policy of handling redirect responses in Application Gateway.

**Reasons**

There are cases in which external systems return redirects to Application Gateway. There were some reports that those redirects break the request flow - redirected requests no longer pass through the Application Gateway, and thus - do not contain authorization header / other credentials and fail. We need to check the exact behavior, decide how the redirects should be handled, and implement/document the solution.

The idea is to have that configurable per Application so that the user can choose whether they want Gateway to follow redirects or not. It can be included as a parameter in Application CRD, something like `followRedirects` (by default set to true).

**Acceptance Criteria**

- By default redirects should be followed
- Application CR parameter added
- Application Gateway uses the CR parameter to determine whether redirects should be followed or not
- Documentation updated (both parameter description in CR docs and the way of handling redirects in the Application Gateway)

janmedrek:
A quick idea on how to set up a testing environment:
- Create an Application pointed to the httpbin
- Use the httpbin's redirect endpoint: https://httpbin.org/#/Redirects/get_redirect_to
- Redirect the call to [Postman's Echo Service](https://www.postman.com/postman/workspace/published-postman-templates/documentation/631643-f695cab7-6878-eb55-7943-ad88e1ccfd65?ctx=documentation)
- Response should contain all the request-related information

VishnAndr:
+1 report here about this issue. The current behaviour breaks the connection, for example, to SAP Field Service Management API (connected either manually or via SAP BTP). Checked 2.4.0 and 2.3.0 versions - same behaviour.

The flow is like this: in a kyma function, axios is used to do the requests. By default, axios client follows 5 redirects by itself. 
The explicit call from the function is done via central application gateway URL. CAG obtains OAuth token just fine, but while hitting targetUrl it gets back 308 response. So it passes back to the function and axios does the redirected call now directly to the location provided in 308 response. But apparently without authorization header. As can be expected, axios gets 401 from the new location.

If I set maxRedirects to 0 in axios client, I'm simply getting 308 response in one go.

The expected behaviour would be that CAG (go native http client?) handles redirects itself since it has authorization information and all.
[...]

PR #15710, opened 2022-10-04
Gateway redirects component tests

VOID404:
**Description**

~~One of the redirect tests redirects to `http://example.com`, which is questionable. It might be better to use some in-cluster endpoint. That would require an in-cluster endpoint that doesn't go through the gateway and has a constant URL (maybe gateway health check?), or making a `manual` test.~~

- Correct `Makefile` whitespace
- Add redirect endpoints to `mock-app`
- Add redirect test scenarios

**Related issue(s)**

- #14717


franpog859 (commented):
Looking good! Do we want to have some negative test cases regarding the redirection functionality, @VOID404 ?

franpog859 (approved):
Good!

Issue #14842, opened 2022-07-18
Align status codes for errors in Application Gateway


akgalwas:
**Description**
As a general rule REST service should return `5xx` for errors that occurred on the server side and `4xx` for errors on the client side. Error code should be as informative as possible. There is a number of cases that needs to be corrected.  

**Expected result**
There are the following error cases that needs to be fixed to adhere to HTTP errors conventions:
- when application specified in path doesn't exist Application Gateway must return `404 Not Found` 
- when application, service or entry for Compass use case is not specified in the path, Application Gateway must return `400 Bad Request `
- when call to target API timeouts Application Gateway must return `504 Gateway Timeout` 

**Actual result**
There are the following error cases that doesn't adhere to common HTTP error conventions:
- when call to target API timeouts Application Gateway returns `503 Service Unavailable` 
- when application specified in path doesn't exist Application Gateway returns '500 Internal Server Error' 
- when application, service or entry for Compass use case is not specified in the path, Application Gateway returns `500 Internal Server error `

**Steps to reproduce**

The above cases are tested in Application Gateway component integration tests (see [missing resources errors handling](https://github.com/kyma-project/kyma/blob/main/tests/components/application-connector/resources/charts/gateway-test/templates/applications/missing-resources-error-handling.yml) and [path related errors handling](https://github.com/kyma-project/kyma/blob/main/tests/components/application-connector/resources/charts/gateway-test/templates/applications/path-related-error-handling.yml))

mvshao:
**Workplan**

- [ ] align status code when the application specified in the path doesn't exist 
  - [ ] make appropriate corrections to the Application Gateway test code
  
- [ ] align status code when application, service or entry for Compass use case is not specified in the path
  - [ ] make appropriate corrections to the Application Gateway test code
  
- [ ] align status code when call to target API timeouts
  - [ ] make appropriate corrections to the Application Gateway test code

mvshao:
It's done. I'm closing this issue.

PR #15717, opened 2022-10-05
Align status codes for errors in Application Gateway

mvshao:
**Description**
REST service should return 5xx for errors that occurred on the server side and 4xx for errors on the client side. Error code should be as informative as possible. 

- `404 Not Found` status code when the application specified in the path doesn't exist
- `400 Bad Request` status code when application, service or entry for the Compass use case is not specified in the path
- `504 Gateway Timeout` when call to target API timeouts Application Gateway
- deletion of the `WriteTimeout` from the Internal Server and Internal Compass Server due to inconsistencies in timeouts. It was done to handle correctly timeout code rewriting.

**Related issue(s)**
#14842

### DIFF
--- a/components/central-application-gateway/internal/proxy/reverseproxy.go
+++ b/components/central-application-gateway/internal/proxy/reverseproxy.go
@@ -6,14 +6,13 @@ import (
 	"net/url"
 	"strings"
 
-	log "github.com/sirupsen/logrus"
-
 	"github.com/kyma-project/kyma/components/central-application-gateway/internal/csrf"
 	"github.com/kyma-project/kyma/components/central-application-gateway/pkg/apperrors"
 	"github.com/kyma-project/kyma/components/central-application-gateway/pkg/authorization"
 	"github.com/kyma-project/kyma/components/central-application-gateway/pkg/authorization/clientcert"
 	"github.com/kyma-project/kyma/components/central-application-gateway/pkg/httpconsts"
 	"github.com/kyma-project/kyma/components/central-application-gateway/pkg/httptools"
+	log "github.com/sirupsen/logrus"
 )
 
 func makeProxy(targetURL string, requestParameters *authorization.RequestParameters, serviceName string, skipTLSVerify bool, authorizationStrategy authorization.Strategy, csrfTokenStrategy csrf.TokenStrategy, clientCertificate clientcert.ClientCertificate, timeout int) (*httputil.ReverseProxy, apperrors.AppError) {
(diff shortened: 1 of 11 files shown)


VOID404 (changes requested):

components/central-application-gateway/internal/proxy/reverseproxy.go
```diff
@@ -54,7 +55,10 @@ func newProxy(targetURL string, requestParameters *authorization.RequestParamete
 
 		log.Infof("Modified request url : '%s', schema : '%s', path : '%s'", req.URL.String(), req.URL.Scheme, req.URL.Path)
 	}
-	return &httputil.ReverseProxy{Director: director, Transport: transport}, nil
+	errorHandler := func(rw http.ResponseWriter, req *http.Request, err error) {
+		codeRewriter(rw, err)
+	}
```
VOID404, line 60: Aren't we missing logging now?

mvshao: I will add necessary logging 👍🏻

components/central-application-gateway/internal/proxy/reverseproxy.go
```diff
@@ -1,19 +1,20 @@
...
 
-	log "github.com/sirupsen/logrus"
-
 	"github.com/kyma-project/kyma/components/central-application-gateway/internal/csrf"
 	"github.com/kyma-project/kyma/components/central-application-gateway/pkg/apperrors"
 	"github.com/kyma-project/kyma/components/central-application-gateway/pkg/authorization"
 	"github.com/kyma-project/kyma/components/central-application-gateway/pkg/authorization/clientcert"
 	"github.com/kyma-project/kyma/components/central-application-gateway/pkg/httpconsts"
 	"github.com/kyma-project/kyma/components/central-application-gateway/pkg/httptools"
+	log "github.com/sirupsen/logrus"
```
VOID404, line 17: Unsorting imports is a weird change to include :P

components/central-application-gateway/internal/proxy/reverseproxy.go
VOID404: And what about logging other errors?
Also we *could* take into account `ErrorLog` field of the `ReverseProxy`, 
but I don't feel like it's really necessary if we're not configuring that field.

VOID404 (changes requested):

components/central-application-gateway/internal/proxy/reverseproxy.go
mvshao: Other errors are handled by a default error handler in the reverse proxy package.

VOID404 (approved):

PR #15720, opened 2022-10-05
Add unit tests for backend.Initialize

nachtmaar:
**Description**

This PR adds unit tests for the jetstreamv2 Backend Initialize() method. Details:

- Moves the validation of the `env.NatsConfig` to the main entry point (fail-fast) instead of delaying it to the jetstream backend.
- Use jetstream client instead of jetstreambackend in SendEventToJetStream to send events

For the reviewer:
- There are some comments in the test files to help identify the system under test etc. As soon as your are through the PR I can remove them (they helped at least me).

**Related issue(s)**


components/eventing-controller/pkg/backend/jetstreamv2/mocks/ConnectionInterface.go
```diff
@@ -0,0 +1,81 @@
+// Code generated by mockery v2.14.0. DO NOT EDIT.
```
k15r, line 1: how does one generate this file?
Please add some `go:generate` comment on the source

components/eventing-controller/pkg/backend/jetstreamv2/jetstream.go
```diff
@@ -39,32 +41,178 @@ const (
...
+// To keep the tests green, for now I leave it as it is.
+func (js *JetStream) IsConnected() bool {
+	return js.connection.IsConnected()
+}
+
+// TODO(nils): I don't see a reason why the integration tests need to get the config from the backend. Since the tests
+// start the backend, they should already be aware of the config.
+// To keep the tests green, for now I leave it as it is.
+// Another solution might be to use a singleton for the config and don't access it from the JetStream backend.
+func (js *JetStream) GetConfig() *backendnats.Config {
```
k15r, line 69: why do we return a reference here? in all testcases where this is used, the reference is immediately dereferenced. just return the object by value and the code would be easier to read

components/eventing-controller/pkg/backend/jetstreamv2/jetstream_internal_unit_test.go
```diff
@@ -3,26 +3,431 @@
...
+	connection := connectionStub{isConnected: false, jetStreamContext: jsCtx}
+	jetStream := JetStream{
+		connectionBuilder: &connectionBuilderStub{conn: connection},
+		config:            config,
+		ceClientFactory:   ceClientFactoryStub{},
+	}
+
+	// when
+	err := jetStream.Initialize(connClosedHandler)
+
```
k15r, line 163: nit: //then

Issue #15093, opened 2022-08-12
Create a NATS Jetstream Consumer programatically in the eventing-controller


nachtmaar:
**TL;DR**
Do not let the JetStream SDK create the NATS Subscription and Consumer, instead do it in **two separate steps**.

**Description**
As a first step in the implementation of https://github.com/kyma-project/kyma/issues/15087, the eventing team decided to **create the NATS Jetstream Consumer independent** of the NATS Subscription.
Currently in the Jetstream Reconciler, the NATS Subscription and the NATS Consumer are created [in one step](https://github.com/kyma-project/kyma/blob/317ed95ca93e83297bd4491f7e877838064b9fa8/components/eventing-controller/pkg/handlers/jetstream.go#L403).
The idea of this ticket is to create the consumer and the subscriptions in two steps and not let the Jetstream SDK handle this for us.

**Reasons**

The advantage of this is that when deleting a NATS Subscription with [jsSubscription.Unsubscribe()](https://pkg.go.dev/github.com/nats-io/nats.go#Subscription.Unsubscribe), the [NATS JS Consumer](https://docs.nats.io/nats-concepts/jetstream/consumers) will not be automatically get deleted. We do use the [JS Interest Policy](https://docs.nats.io/using-nats/developer/develop_jetstream/model_deep_dive), this means that the events will not be deleted as there is active `interest` in them.
This will allow us to **pause a NATS Subscription** without losing events (`jsSubscription.Unsubscribe()`).

**Attachments**

nachtmaar:
Adding a consume should be possible with [js.jsCtx.AddConsumer()](https://github.com/nats-io/nats.go/blob/v1.16.0/jsm.go#L55)

VladislavPaskar:
since that change was introduced in this [PR](https://github.com/kyma-project/kyma/pull/15762), we can close this issue.

PR #15447, opened 2022-09-12
Add a Requeue after no NATS subscriptions synced

VladislavPaskar:
**Description**

Changes proposed in this pull request:

- ...
- ...
- ...

**Related issue(s)**

PR #15762, opened 2022-10-10
Improve consumer logic, add unit tests

VladislavPaskar:
**Description**

- pack all the consumer login into one function splitted into blocks
- add unit test using mocks
- add errors package

**Related issue(s)**
https://github.com/kyma-project/kyma/issues/15093


components/eventing-controller/pkg/backend/jetstreamv2/jetstream_internal_unit_test.go
```diff
@@ -0,0 +1,468 @@
...
+				jsSubKey := NewSubscriptionSubjectIdentifier(testCase.givenSubscription, jsSubject)
+
+				tc.givenAssertions(testCase.givenSubscription, testEnv.jsBackend, testEnv.jsCtxMock, jsSubKey)
+			}
+		})
+	}
+}
+
+// Test_CheckNATSSubscriptionsCount test the behaviour of the checkNATSSubscriptionsCount function.
+func Test_CheckNATSSubscriptionsCount(t *testing.T) {
```
nachtmaar, line 327: I tried to simplify the test. The details are in this [PR](https://github.com/VladislavPaskar/kyma/pull/426). Let me know what you think or we can discuss it in a call as well 😃

nachtmaar (commented):
Did not finish the review yet. Will continue tomorrow

components/eventing-controller/api/v1alpha2/subscription_types.go
```diff
@@ -86,12 +89,17 @@ func (s Subscription) MarshalJSON() ([]byte, error) {
...
-func (s *Subscription) GetMaxInFlightMessages() (*int, error) {
+// GetMaxInFlightMessages todo desc and test
+func (s *Subscription) GetMaxInFlightMessages(log *zap.SugaredLogger) int {
+	if s.Spec.Config == nil {
+		return env.DefaultMaxInFlight
+	}
 	val, err := strconv.Atoi(s.Spec.Config[MaxInFlightMessages])
 	if err != nil {
-		return nil, err
+		log.With("subscription-name", s.Name).Errorw("Failed to convert the maxInFlight config value", "error", err)
```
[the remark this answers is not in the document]
VladislavPaskar: The intention of the this function now is: when you cannot parse the sub's maxInFlight, you just use the fallback value and report to the user(by logging the message), that the config maxInFlight value is not parsable. By doing so, I make it easy for the caller, so they shouldn't handle any error.

components/eventing-controller/pkg/backend/jetstreamv2/jetstream_internal_unit_test.go
VladislavPaskar: I like it, thanks for contribution

components/eventing-controller/pkg/backend/jetstreamv2/jetstream.go
```diff
@@ -387,81 +389,112 @@ func (js *JetStream) deleteConsumerFromJetStream(name string) error {
 	return nil
 }
 
-// bindConsumersForInvalidNATSSubscriptions attempts to bind an existing consumer to a new NATS subscription,
-// when the previous subscription that the consumer was associated with becomes invalid. If binding fails,
-// we will delete the subscription from our internal subscriptions map.
-func (js *JetStream) bindConsumersForInvalidNATSSubscriptions(subscription *eventingv1alpha2.Subscription, asyncCallback func(m *nats.Msg), log *zap.SugaredLogger) {
+// syncNATSConsumers makes sure there are consumers and subscriptions created on the NATS Backend.
+func (js *JetStream) syncNATSConsumers(subscription *eventingv1alpha2.Subscription, asyncCallback func(m *nats.Msg)) error {
```
nachtmaar, line 393: nit: The name is misleading if it creates consumers and subscriptions 🤔

components/eventing-controller/pkg/backend/jetstreamv2/jetstream.go
```diff
@@ -387,81 +389,112 @@ func (js *JetStream) deleteConsumerFromJetStream(name string) error {
...
-		} else {
-			// save recreated JetStream subscription in storage
-			js.subscriptions[jsSubKey] = &Subscription{Subscription: jsSubscription}
-			log.Debugw("Recreated subscription on JetStream")
 		}
+
+		natsSubscription, subExists := js.subscriptions[jsSubKey]
+
+		// try to create a NATS Subscription if it doesn't exist
+		if !subExists && !consumerInfo.PushBound {
```
nachtmaar, line 419: Can't `consumerInfo` be nil here ? 

It is only checked for nilness later in the code:
```
		if consumerInfo != nil {
			if err := js.checkSubscriptionConfig(subscription, *consumerInfo, log); err != nil {
				return err
			}
		}
```

components/eventing-controller/pkg/backend/jetstreamv2/jetstream.go
```diff
@@ -387,81 +389,112 @@ func (js *JetStream) deleteConsumerFromJetStream(name string) error {
...
+	return nil
+}
 
-		// skip the subject, in case there is a bound consumer on NATS
-		if consumerInfo != nil && consumerInfo.PushBound {
-			continue
-		}
+// checkSubscriptionConfig checks that the latest Subscription Config changes are propagated to the consumer.
+// In our case config contains only the "maxInFlightMessages" property, which is the maxAckPending on the consumer side.
+func (js *JetStream) checkSubscriptionConfig(subscription *eventingv1alpha2.Subscription, consumerInfo nats.ConsumerInfo, log *zap.SugaredLogger) error {
```
nachtmaar, line 484: nit: IMHO we are updating the consumer config and not only checking it. Another name could be `syncConsumerConfig`.

```suggestion
func (js *JetStream) updateConsumerConfig(subscription *eventingv1alpha2.Subscription, consumerInfo nats.ConsumerInfo, log *zap.SugaredLogger) error {
```

components/eventing-controller/pkg/backend/jetstreamv2/jetstream.go
```diff
@@ -387,81 +389,112 @@ func (js *JetStream) deleteConsumerFromJetStream(name string) error {
...
+			}
+		}
+
+		if consumerInfo != nil {
+			if err := js.checkSubscriptionConfig(subscription, *consumerInfo, log); err != nil {
+				return err
+			}
+		}
+	}
+	if err := js.checkNATSSubscriptionsCount(subscription); err != nil {
```
nachtmaar, line 438: Why do we need to check this ? Isn't is the responsibility of this method to create them ? 🤔

components/eventing-controller/pkg/backend/jetstreamv2/jetstream_internal_unit_test.go
```diff
@@ -406,6 +208,99 @@ func Test_SyncNATSConsumers(t *testing.T) {
...
+			name: "ConsumerInfo's not found error should be ignored",
+			jetStreamContext: &jetStreamContextStub{
+				consumerInfoError: nats.ErrConsumerNotFound,
+				consumerInfo:      nil,
+
+				addConsumer: &nats.ConsumerInfo{Config: nats.ConsumerConfig{MaxAckPending: DefaultMaxInFlights}},
+
+				subscribe: &nats.Subscription{},
+			},
+			wantError: nil,
```
nachtmaar, line 235: Shouldn't we test here that if the consumer is not found it was created ? Therefore that addConsumer was called ?

components/eventing-controller/pkg/backend/jetstreamv2/jetstream_internal_unit_test.go
```diff
@@ -0,0 +1,425 @@
...
+			name:              "Adding a consumer/subscribing should be successful",
+			givenSubscription: subWithOneType,
+			givenMocks: func(sub *v1alpha2.Subscription, jsBackend *JetStream, jsCtx *mocks.JetStreamContext, jsSubKey SubscriptionSubjectIdentifier) {
+				for _, eventType := range sub.Status.Types {
+					_, jsSubKey, newConsumer, natsSub := generateConsumerInfra(jsBackend, sub, eventType)
+					subscribeArgs := getSubOptsAsVariadicSlice(jsBackend, sub, eventType)
+
+					jsCtx.On("ConsumerInfo", jsBackend.Config.JSStreamName, jsSubKey.ConsumerName()).Return(nil, nats.ErrConsumerNotFound)
+					jsCtx.On("AddConsumer", jsBackend.Config.JSStreamName, mock.AnythingOfType("*nats.ConsumerConfig")).Return(newConsumer, nil)
+					jsCtx.On("Subscribe", subscribeArgs...).Return(natsSub, nil)
```
nachtmaar, line 61: For every On() call you can just use `Test_SyncNATSConsumers` once in the table test body.

nachtmaar: Maybe we can get completely rid of `givenAssertions` then 🤔

components/eventing-controller/pkg/backend/jetstreamv2/jetstream_internal_unit_test.go
```diff
@@ -0,0 +1,425 @@
...
+	assert.NoError(t, err)
+}
+
+// Test_SyncNATSConsumers tests creation/binding of the consumer and NATS Subscriptions.
+func Test_SyncNATSConsumers(t *testing.T) {
+	testEnv := getTestEnvironmentWithMock(t)
+	callback := func(m *nats.Msg) {}
+	subWithOneType := subtesting.NewSubscriptionWithOneType()
+	_, jsSubKey, defaultConsumerInfo, _ := generateConsumerInfra(testEnv.jsBackend, subWithOneType, subWithOneType.Status.Types[0])
+	subscriberMock := &mocks.Subscriber{}
```
nachtmaar, line 41: It would be easier (I think) to understand for the test reader if we can simply construct these objects manually. Maybe we can refactor this as we did with `Test_CheckNATSSubscriptionsCount` 🤔

components/eventing-controller/pkg/backend/jetstreamv2/jetstream_internal_unit_test.go
```diff
@@ -0,0 +1,425 @@
...
+		jsCtx:            jsCtx,
+		subscriptions:    make(map[SubscriptionSubjectIdentifier]Subscriber),
+		sinks:            sync.Map{},
+		logger:           defaultLogger,
+		metricsCollector: metricsCollector,
+		cleaner:          jsCleaner,
+	}
+
+	return &TestEnvironment{
+		jsBackend:  jsBackend,
```
nachtmaar, line 388: Isn't is enough to just return the `JetStream` object since this is the `System under Test` ?

components/eventing-controller/pkg/backend/jetstreamv2/jetstream_internal_unit_test.go
```diff
@@ -0,0 +1,425 @@
...
+
+			// when
+			err := testEnv.jsBackend.syncNATSConsumers(testCase.givenSubscription, callback)
+
+			// then
+			// check if error is expected
+			if testCase.wantErr == nil {
+				require.NoError(t, err)
+			} else {
+				require.Error(t, err)
```
nachtmaar, line 195: This call is not required as there is require.ErrorIs() in the next line

components/eventing-controller/pkg/backend/jetstreamv2/jetstream_internal_unit_test.go
```diff
@@ -0,0 +1,425 @@
...
+				for _, eventType := range sub.Status.Types {
+					jsSubject := jsBackend.getJetStreamSubject(sub.Spec.Source, eventType.CleanType, sub.Spec.TypeMatching)
+					jsSubKey := NewSubscriptionSubjectIdentifier(sub, jsSubject)
+					subscriberMock.On("IsValid").Return(true)
+
+					subWithOneType.Spec.Config[v1alpha2.MaxInFlightMessages] = "20"
+					jsCtx.On("ConsumerInfo", jsBackend.Config.JSStreamName, jsSubKey.ConsumerName()).Return(defaultConsumerInfo, nil)
+					jsCtx.On("UpdateConsumer", jsBackend.Config.JSStreamName, mock.AnythingOfType("*nats.ConsumerConfig")).Return(defaultConsumerInfo, nil)
+				}
+			},
```
nachtmaar, line 154: I see that the for loop creates quite some complexity for the test. It is in `givenMocks`, `givenAssertions` and the table-test body.

I have 2 ideas in my mind to work around this:
- 1. We have the subscription types under our control in the test. So why do we even need to pass multiple types ? 
- 2. What if we refactor `syncNATSConsumers` in such a way that there is only the for-loop and then an inner function call (say `processNATSSubject` e.g.,). We could then test both pieces of code in isolation. 
  - case 1: ensure that for each event type the the function (processNATSSubject) is called
  - case 2: ensure the correctness of the inner function (by only providing a single subject)

components/eventing-controller/pkg/backend/jetstreamv2/jetstream_internal_unit_test.go
```diff
@@ -0,0 +1,425 @@
...
+
+					jsCtx.AssertCalled(t, "ConsumerInfo", jsBackend.Config.JSStreamName, jsSubKey.ConsumerName())
+					jsCtx.AssertNotCalled(t, "AddConsumer", jsBackend.Config.JSStreamName, mock.AnythingOfType("*nats.ConsumerConfig"))
+					jsCtx.AssertCalled(t, "Subscribe", jsSubject, mock.AnythingOfType("nats.MsgHandler"), mock.AnythingOfType("nats.subOptFn"))
+
+				}
+			},
+		},
+		{
+			name:              "Bind invalid NATS Subscription should result into error",
```
nachtmaar, line 108: I believe we don't need a mock here, but instead a stub. We are testing that a value is returned here, not that an interaction happened. So we can use a stub for both `sub` and `jsCtx`.

nachtmaar: By not mixing value tests and interaction tests we can also get completely rid of `wantErr` (this is the only value based test which is left in this table test) and the test related code in the table-test body.

components/eventing-controller/pkg/backend/jetstreamv2/jetstream.go
```diff
@@ -387,81 +387,113 @@ func (js *JetStream) deleteConsumerFromJetStream(name string) error {
...
+					js.getConsumerConfig(subscription, jsSubKey, jsSubject),
+				)
+				if err != nil {
+					return fmt.Errorf("%w: %v", ErrConsumerAdd, err)
+				}
+				log.Debug("Created consumer on JetStream")
+			} else {
+				return fmt.Errorf("%w: %v", ErrConsumerNotFound, err)
+			}
+		}
```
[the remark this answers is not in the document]
VladislavPaskar: good idea

components/eventing-controller/pkg/backend/jetstreamv2/jetstream_internal_unit_test.go
VladislavPaskar: That's a very-very good point !

components/eventing-controller/testing/v2/test_helpers.go
```diff
@@ -244,6 +245,26 @@ func NewSubscription(name, namespace string, opts ...SubscriptionOpt) *eventingv
...
+		WithTypeMatchingStandard(),
+		WithMaxInFlight(env.DefaultMaxInFlight),
+		WithStatusTypes([]eventingv1alpha2.EventType{
+			{
+				OriginalType: CloudEventType,
+				CleanType:    CloudEventType,
+			},
+		}),
+	)
+}
```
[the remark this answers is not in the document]
VladislavPaskar: moved them into my test fiel

components/eventing-controller/pkg/backend/jetstreamv2/jetstream_internal_unit_test.go
```diff
@@ -372,7 +346,7 @@ func TestVlad_CheckNATSSubscriptionsCount(t *testing.T) {
 			name:                 "unexpected empty subscriptions map should result into an error",
 			givenSubscription:    subWithType,
 			givenSubscriptionMap: map[SubscriptionSubjectIdentifier]Subscriber{},
-			wantErr:              &ErrMissingNATSSubscription{},
+			wantErr:              &backenderrors.ErrMissingNATSSubscription{},
```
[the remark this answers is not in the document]
VladislavPaskar: I think it's good to have an errors package, since it's easier to distinct them and structure. We followed the same pattern for skip.go error. I refactored them a bit though.

components/eventing-controller/pkg/backend/jetstreamv2/errors/failedtosubscribe.go
```diff
@@ -0,0 +1,27 @@
...
+)
+
+type ErrFailedToSubscribeOnNATS = FailedToSubscribeOnNATSError
+
+type FailedToSubscribeOnNATSError struct {
+	OriginalError error
+}
+
+func NewFailedToSubscribeOnNATSError(reason error) error {
+	return xerrors.Errorf("%w", &FailedToSubscribeOnNATSError{OriginalError: reason})
```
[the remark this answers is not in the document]
VladislavPaskar: check now

components/eventing-controller/pkg/backend/jetstreamv2/errors/missingnatssubscription.go
```diff
@@ -0,0 +1,28 @@
...
+)
+
+type ErrMissingNATSSubscription = MissingNATSSubscriptionError
+
+type MissingNATSSubscriptionError struct {
+	Subject eventingv1alpha2.EventType
+}
+
+func NewMissingNatsSubscriptionError(subject eventingv1alpha2.EventType) error {
+	return xerrors.Errorf("%w", &ErrMissingNATSSubscription{Subject: subject})
```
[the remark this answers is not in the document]
VladislavPaskar: done

components/eventing-controller/pkg/backend/jetstreamv2/jetstream.go
VladislavPaskar: I refactored the code, check it now

components/eventing-controller/pkg/backend/jetstreamv2/jetstream.go
VladislavPaskar: renamed to `syncConsumerMaxInFlight` since it's the most suitable one.

components/eventing-controller/pkg/backend/jetstreamv2/jetstream.go
VladislavPaskar: that's coming from https://github.com/kyma-project/kyma/pull/15447

components/eventing-controller/pkg/backend/jetstreamv2/jetstream_internal_unit_test.go
VladislavPaskar: done

components/eventing-controller/pkg/backend/jetstreamv2/jetstream_internal_unit_test.go
VladislavPaskar: I refactored the tests

components/eventing-controller/pkg/backend/jetstreamv2/jetstream.go
nachtmaar: I see it was refactored in the new version. Thanks for this 👍

components/eventing-controller/pkg/backend/jetstreamv2/stubs_test.go
```diff
@@ -0,0 +1,222 @@
+package jetstreamv2 //nolint:nolintlint,testpackage
```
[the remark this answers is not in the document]
VladislavPaskar: the problem is, that we the pull-lint jobs still fails due to this

components/eventing-controller/pkg/backend/jetstreamv2/jetstream_internal_unit_test.go
```diff
@@ -0,0 +1,445 @@
...
+		})
+	}
+}
+
+// Test_SyncConsumersAndSubscriptions_ForBindInvalidSubscriptions tests the
+// binding behaviour in the syncConsumersAndSubscriptions function.
+func Test_SyncConsumersAndSubscriptions_ForBindInvalidSubscriptions(t *testing.T) {
+	// pre-requisites
+	sub := NewSubscriptionWithOneType()
+	validSubMock := &mocks.Subscriber{}
```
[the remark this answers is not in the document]
VladislavPaskar: good point 👍

components/eventing-controller/pkg/backend/mocks/JetStreamContext.go
```diff
@@ -0,0 +1,1061 @@
+// Code generated by mockery v2.14.0. DO NOT EDIT.
```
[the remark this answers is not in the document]
VladislavPaskar: I really don't see a benefit for it right now for different reasons:
- I don't see a problem in exposing the autogenerated mocks
- having all the mocks in one file could cause difficulties
- we already have a lot of files in the source code package
- we already had other mocks in the same mocks package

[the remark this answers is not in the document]
nachtmaar: With the change I proposed we would still have several files. For `JetStreamContext.go` it would be `JetStreamContext_test.go`. The benefit is that the mock can never be used in production code since it would be only available in a `test_.go` file.

components/eventing-controller/pkg/backend/jetstreamv2/stubs_test.go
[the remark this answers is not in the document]
nachtmaar: I actually **like** the **recommendation** from `nolintlint` to give a **reason** why a linter is ignored because otherwise it is impossible to judge by looking at the code without running the linter:

```go
pkg/backend/jetstreamv2/jetstream_internal_unit_test.go:1:21: directive `//nolint:testpackage` should provide explanation such as `//nolint:testpackage // this is why` (nolintlint)
package jetstreamv2 //nolint:testpackage
pkg/backend/jetstreamv2/stubs_test.go:1:21: directive `//nolint:testpackage` should provide explanation such as `//nolint:testpackage // this is why` (nolintlint)
```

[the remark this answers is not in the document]
VladislavPaskar: This is a recommendation from linter in the IDE.

components/eventing-controller/pkg/backend/mocks/JetStreamContext.go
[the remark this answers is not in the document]
VladislavPaskar: I honestly don't see a big problem of having the mocks in our "production" code.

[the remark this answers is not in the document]
nachtmaar: Okay, let's leave it as it is. I will have face the same question in my [PR](https://github.com/kyma-project/kyma/pull/15720/files?diff=unified&w=0) and try to follow the approach. I will see if it causes any problems and report back to the team 😃

nachtmaar (approved):

Issue #15737, opened 2022-10-06
☠ Cleanup function-managed workload resources configuration


kwiatekus:
**Description**
Function spec introduces new `functionPod` and `buildPod` fields to allow defining templates for those workloads.
Those templates should contain `resources` objects ( consistent with PodTemplate Spec known to the kubernetes users )

**AC**
 - When user provides no data concerning workload resources a default preset is used and is propagated to the template resources
 - User should be able to use named presets (i.e `S`,`M`, `L`, `XL`) and the template resources should be filled in by defaulting webhook.
 - User should be able to use those new fields directly in order to precisely define workload resources for function runtime pod and build-time job. In this case user must clear the named preset field
 - Named preset takes precedence over resource values.

**Reasons**

Clean up Function spec API as we progress towards a stable v1 version. Make the function spec consistent with other known workload specs (i.e Job or Deployment) and include templates for function and build pods. Expose more control over the pod features.

**Attachments**

[proposal](https://github.com/kyma-project/kyma/blob/7ccc257701685054f05ea4859b81b6a53407035a/components/function-controller/design/2022-10-v1alpha2-templating.md#resources)

anoipm:
Steps:
- [ ] (defaulting) remove merging resource presets and custom settings (see use cases in [proposal](https://github.com/kyma-project/kyma/blob/7ccc257701685054f05ea4859b81b6a53407035a/components/function-controller/design/2022-10-v1alpha2-templating.md#resources)),
- [ ] (doc) info about changes in merging and priorities of presets and custom settings,
- [ ] (defaulting) propagate settings function resources to template,
- [ ] (defaulting) propagate settings build resources to template,
- [ ] (controller) use function resources from template,
- [ ] (controller) use build resources from template,
- [ ] (doc) info about new fields and deprecated old ones,
- [ ] (defaulting) remove defaulting old fields with function resources,
- [ ] (defaulting) remove defaulting old fields with build resources,

kwiatekus:
Waits for decision https://github.com/kyma-project/kyma/pull/15862

PR #15783, opened 2022-10-12
add conversion from and to v1alpha2 using Templates..Resources

anoipm:
**Description**

Add conversion from and to v1alpha2 using Templates..Resources

**Related issue(s)**
See also #15737 (this PR is only part of this issue)

### DIFF
--- a/components/function-controller/internal/webhook/conversion_webhook.go
+++ b/components/function-controller/internal/webhook/conversion_webhook.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	v1 "k8s.io/api/core/v1"
 	"net/http"
 
 	"go.uber.org/zap"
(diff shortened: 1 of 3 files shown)


components/function-controller/internal/webhook/conversion_webhook.go
```diff
@@ -216,10 +218,14 @@ func convertResourcesV1Alpha1ToV1Alpha2(in *serverlessv1alpha1.Function, out *se
 	}
 
 	if buildResourcesExists {
+		//TODO: remove ResourceConfiguration in next step
 		out.Spec.ResourceConfiguration.Build.Resources = &in.Spec.BuildResources
+		setV1Alpha2BuildJobResources(in.Spec.BuildResources, out)
 	}
 	if functionResourcesExists {
+		//TODO: remove ResourceConfiguration in next step
```
pPrecel, line 226: Is this PR ready to review? 🤔

anoipm: Yes, this part will be removed after changes in controller which are part of the same issue.

pPrecel (approved):

Issue #14998, opened 2022-08-03
Enable volume mounted secret bindings in Kyma Functions


gecko17:
**Description**

In order to use BTP service instances natively with @sap/xsenv, you need to create a "volume" from the secret (automatically created by the ServiceBinding resource) in the Pod template spec and then mount the volume to a specific path via a "volumeMount" in the container spec. This is described here: https://blogs.sap.com/2022/07/12/the-new-way-to-consume-service-bindings-on-kyma-runtime/.

In order to use BTP service instances also in Kyma serverless Functions, the Function CR would need to support the same: 
* support a volume spec
* support a volumeMount spec

Technically a function also creates a Pod and a Container, so it should be possible to allow volumes and volumeMounts. Under the hood, the function even uses a "tmp-dir" volume mounted to the path "tmp". So it should be a quite simple addition to allow custom volumes and volumeMounts in the Function spec?

Serverless function from the nature should be stateless... and not needing persistance.
Limit the volume mounts to secrets and config maps (Read only ).

**Related **
 - depends on the spec changes - https://github.com/kyma-project/kyma/issues/15233

**Reasons**

This is required in order to use BTP Service Instances (which are supported on Kyma) in serverless Functions.

**Attachments**
Proposal for *secretMounts*
https://github.com/kyma-project/kyma/blob/main/components/function-controller/design/extend_function_spec_proposal.md#final-version---the-compromise

kwiatekus:
Waits for decision https://github.com/kyma-project/kyma/pull/15862

anoipm:
Scenario from manual test: 
[test-14998-202211222150.txt](https://github.com/kyma-project/kyma/files/10071626/test-14998-202211222150.txt)

anoipm:
Task parts:
 - [x] function api + building pod + detecting changes by controller' s reconciliator - [PR-16108](https://github.com/kyma-project/kyma/pull/16108)
 - [ ]

Issue #14877, opened 2022-07-21
Allow to define annotations and labels for function managed workloads


kwiatekus:
**Description**

As a function developer I want to define metadata (labels and annotations) for function managed deployment.
 
 **AC**
 - [x]  User should be able to manage function runtime pod's labels and annotations - https://github.com/kyma-project/kyma/issues/16232
  ~User should be able to manage function build-time job's labels and annotations~  EDIT: Will be added when users will request that.  
 - [x]  Validation webhook shouldn't allow to add labels from `serverless.kyma-project.io` group

**Reason**
Various k8s tools are using annotations to control their features against the workload (i.e istio, fluentbit). User need to have control over metadata managed by function.
Labels are used to target workloads in kubernetes. User should have the possibility to add custom labels to function managed workloads to look them up, display logs or target them for any other reason. 

For Example:
Custom fluent bit log parser is enabled for a given pod by an annotation:
`fluentbit.io/parser: my-regex-parser`

**Attachments**
See the agreed [proposal](https://github.com/kyma-project/kyma/blob/main/components/function-controller/design/extend_function_spec_proposal.md#labels-and-annotations) for the Function CRD that includes labels and annotation management

kwiatekus:
Function spec already has a field to define pod annotations:
```
spec.template.annotations
```

However, we plan to rework this  `spec.template.` field and allow for more templating of function runtime pod, built-time job separately

kwiatekus:
Blocked by https://github.com/kyma-project/kyma/issues/15233

kwiatekus:
Waits for decision https://github.com/kyma-project/kyma/pull/15862

PR #15862, opened 2022-10-20
Propose changes in the Function CRD

kwiatekus:
**Description**

- Proposal for extending Function CRD

**Related issue(s)**
#14877 
#14998 
#15737


components/function-controller/design/extend_function_spec_proposal.md
```diff
@@ -0,0 +1,187 @@
...
+    app.kubernetes.io/name: my-function
+spec:
+  runtime: nodejs16
+  source:
+    ...
+  serviceBindings: 
+  - source: my-secret
+    mountPath: "/foo" # optional.. read from SERVICE_BINDING_ROOT ENV 
+
+  profile: S / M / L / XL / ... / Custom #optional
```
pPrecel, line 154: I would suggest thinking about the name of this field. I think that the `.spec.profile` name is too generic and does not introduce any context. What do you think about `resourcesProfile`?

components/function-controller/design/extend_function_spec_proposal.md
```diff
@@ -0,0 +1,233 @@
...
+
+  build:
+    resourcesProfile: S / M / L / XL / ... / Custom
+    resources: # optional... required if spec.build.profile==Custom
+      limits: ... #if_custom
+      requests: ... #if_custom
+
+```
+
+### Option 3
```
dbadura, line 226: I reserved it :)

components/function-controller/design/extend_function_spec_proposal.md
```diff
@@ -0,0 +1,233 @@
...
+### Option 4
```
anoipm, line 227: reserved

components/function-controller/design/extend_function_spec_proposal.md
```diff
@@ -0,0 +1,392 @@
...
+
+- Add more flexibility to the Serverless API - enable volume mounted secrets and pod annotations.
+- Organise spec attributes belonging to runtime and build-time configuration (?)
+- Provide a rollout plan to implement this functionality in small increments and avoid the need to roll out a new API version
+
+### Non-Goals
+- Define a new API version specification
+
+## Discussion points
+
```
moelsayed, line 27: Should we look into adding metadata to the build job pod?

components/function-controller/design/extend_function_spec_proposal.md
```diff
@@ -0,0 +1,392 @@
...
+metadata:
+  name: my-function
+  namespace: default
+  labels:
+    app.kubernetes.io/name: my-function
+spec:
+  runtime: nodejs16
+  source:
+    ...
+  serviceBindings:
```
moelsayed, line 150: This is an interesting name. I think it makes sense.

components/function-controller/design/extend_function_spec_proposal.md
```diff
@@ -0,0 +1,392 @@
...
+    profile: S / M / L / XL / ... / Custom
+    resources: # optional... required if spec.build.profile==Custom
+      limits: ... #if_custom
+      requests: ... #if_custom
+​
+```
+### Option 2
+
+- almost the same as `Option 1` (same pros)
+- dif1: move some fields from `.spec` to the new struct `.spec.runtimeSpec` to clearly distinguish fields desired to be used in the `build` and `running` phases. For example in `Option 1` users may have questions after seeing `.spec.env` and `.spec.build.env` fields for example "is .spec.env dedicated for the running function's pod? Would the field be merged with .spec.build.env for the building job?"
```
moelsayed, line 181: Also agree with this. The split is confusing especially considering our current API.

components/function-controller/design/extend_function_spec_proposal.md
```diff
@@ -0,0 +1,392 @@
...
+- Configure labels and annotations for the  Function runtime pod.
+
+### Goals
+
+- Add more flexibility to the Serverless API - enable volume mounted secrets and pod annotations.
+- Organise spec attributes belonging to runtime and build-time configuration (?)
+- Provide a rollout plan to implement this functionality in small increments and avoid the need to roll out a new API version
+
+### Non-Goals
+- Define a new API version specification
```
moelsayed, line 24: I don't think this would be possible. All the proposals have big changes to the current API, not just additions. I think if we are going to keep this non-goal in mind, we should include a high-level implementation plan on how to achieve that.

components/function-controller/design/extend_function_spec_proposal.md
```diff
@@ -0,0 +1,392 @@
...
+
+Give Serverless users the ability to:
+- Configure volume mounted secrets (and config maps) for Function's subresources.
+- Configure labels and annotations for the  Function runtime pod.
+
+### Goals
+
+- Add more flexibility to the Serverless API - enable volume mounted secrets and pod annotations.
+- Organise spec attributes belonging to runtime and build-time configuration (?)
+- Provide a rollout plan to implement this functionality in small increments and avoid the need to roll out a new API version
```
moelsayed, line 21: This is very important, yet not included.

components/function-controller/design/extend_function_spec_proposal.md
```diff
@@ -0,0 +1,392 @@
...
+  # resources
+  # envs
+  # metadata
+  buildSpec:
+    # resources
+    # envs (?)
+    # metadata (?)
+```
+
+### Mounts - own structure or k8s inherited
```
moelsayed, line 63: I prefer providing the user with more flexibility and exposing the full volume API, but if we can't, the serviceBinding option sound good.

moelsayed (commented):
The proposed options cover most possible permutations, so I am not going to add another one. I just added a few comments.

components/function-controller/design/extend_function_spec_proposal.md
```diff
@@ -0,0 +1,392 @@
...
+
+Under the hood the secret mount becomes a volume mount in the runtime pod.
+We could :
+ - expose the k8s volume mount spec  in function spec
+ - cover k8s volume mount spec via oor own [specialised](https://servicebinding.io/application-developer/) spec
```
pbochynski, line 68:
```suggestion
 - cover k8s volume mount spec via our own [specialized](https://servicebinding.io/application-developer/) spec
```

pbochynski, line 68: Do you mean to reuse service binding from BTP service operator or create something new? First will be too limiting, and second too confusing.

kwiatekus: I imagine a simple way to define a reference to a secret and optional mounting point (`SERVICE_BINDING_ROOT` env could be used by default ) to support volume mounted secrets as recommended [here](https://servicebinding.io/)
Something like 
```yaml
spec:
  serviceBindings:
  - source: my-redis-secret
  - mountPath: "/redis-secret" # optional
```

kwiatekus: discussed. and the outcome reflected in the PR

components/function-controller/design/extend_function_spec_proposal.md
kwiatekus: I discussed it with @pbochynski 

we could separate those cases.
We definitely need a convenient way to enable service bindings.
Those use natively  k8s secrets ( i.e created via BTP service binding CRs ) and leverage dedicated SDKs compliant with  service binding recommendation

volume mounts is more generic but would be less convinient for devs focused on the main use case ( service bindings )

Volume mounts could be a separate feature in our spec ( see last "compromise" option )

components/function-controller/design/extend_function_spec_proposal.md
kwiatekus: included in the last "compromise option"

components/function-controller/design/extend_function_spec_proposal.md
kwiatekus: the final name would have to be `secretBindings` though as `service binding` would clash with https://github.com/SAP/sap-btp-service-operator#service-binding

components/function-controller/design/extend_function_spec_proposal.md
kwiatekus: included in the last "compromise" option

pbochynski (dismissed):

components/function-controller/design/extend_function_spec_proposal.md
```diff
@@ -0,0 +1,496 @@
...
+    app: my-app
+  annotations: #?
+    fluentbit.io/parser: my-regex-parser
+    istio-injection: enabled
+
+  serviceBindings: 
+  - secretName: my-secret
+    mountPath: "/foo" # optional.. read from SERVICE_BINDING_ROOT ENV 
+
+  profile: S / M / L / XL / ... / Custom #optional
```
dbadura, line 449: I wouldn't use `Custom` to enable manual resoruces. The simplier to use option is to leave it empty.

components/function-controller/design/extend_function_spec_proposal.md
```diff
@@ -0,0 +1,496 @@
...
+
+  labels: #?
+    app: my-app
+  annotations: #?
+    fluentbit.io/parser: my-regex-parser
+    istio-injection: enabled
+
+  serviceBindings: 
+  - secretName: my-secret
+    mountPath: "/foo" # optional.. read from SERVICE_BINDING_ROOT ENV
```
dbadura, line 447:
```suggestion
    mountPath: "/foo" # optional.. if not set read from SERVICE_BINDING_ROOT ENV 
```

components/function-controller/design/extend_function_spec_proposal.md
```diff
@@ -0,0 +1,496 @@
...
+
+```yaml
+apiVersion: serverless.kyma-project.io/v1alpha2
+kind: Function
+metadata:
+  name: my-function
+  namespace: default
+  labels:
+    app.kubernetes.io/name: my-function
+spec:
```
dbadura, line 430:
```suggestion
spec: #Contains spec of Function and run stage (deployments, hpa)
```

components/function-controller/design/extend_function_spec_proposal.md
```diff
@@ -0,0 +1,496 @@
...
+  # - name: config
+  #   configmap:
+  #     name: function-configuration
+  #     items:
+  #       - key: config
+  #         path: config.yaml
+​​
+
+  # optional 
+  build:
```
dbadura, line 481:
```suggestion
  build: #Contains spec of build stage
```

components/function-controller/design/extend_function_spec_proposal.md
a-thaler: and fail validation if both are specified

components/function-controller/design/extend_function_spec_proposal.md
a-thaler: And SERVICE_BINDING_ROOT env is set by default?

components/function-controller/design/extend_function_spec_proposal.md
```diff
@@ -0,0 +1,496 @@
...
+    min: 1
+    max: 2
+
+  labels: #?
+    app: my-app
+  annotations: #?
+    fluentbit.io/parser: my-regex-parser
+    istio-injection: enabled
+
+  serviceBindings:
```
a-thaler, line 445: I'm wondering if that explicit relation to the ServiceBinding concept is good. That will introduce a dependency to it and you would need to explicitly test that (have an app using the specific libraries to interpret the metadata of a binding, to assure that actual serviceBindings can be consumed)
In contrast, if someone just wants to consume a 3party API using secret mounts he will be confused and wonder if that mechanism can be used for that.
So what are the actual parts which are specific to ServiceBindings? As far as I can see the defaulting of the mountPath to the SERVICE_BINDING_ROOT only, right?
We could either drop it fully and except a little loss in convenience

```yaml
secretMounts
  - secretName: icke
    mountPath: "myMountPath"
```
or we could add it as a capability to the binding itself?

```yaml
secretMounts
  - secretName: icke
     isServiceBinding: true
```

or more generic but coupled:
```yaml
secretMounts
  - fromSecret:
       name: icke
     mountPath: "/icke"
  - fromServiceBinding:
       name: "icke"
```

kwiatekus: Thank you @a-thaler 
In order to avoid implicit association with service bindings (https://github.com/SAP/sap-btp-service-operator#service-binding or https://servicebinding.io/) and user confusion I think the best option would be to simply enable the secret mount.

If user wants to use specialised libraries we would need to extra set the env variable, for example:

```yaml
secretMounts
  - secretName: icke
    mountPath: "/foo" #required.. no assumptions/validations towards SERVICE_BINDING_ROOT env value
    
 envs:
  - name: SERVICE_BINDING_ROOT
     value: "/foo"  #set explicitely if user wants to use a specialised library that expects it      
```

kwiatekus: the fact that `mountPath` and SERVICE_BINDING_ROOT value match indicates that this secret mount will be consumed in an "Service binding"- way via specialised library. 
But this is soft indication, not done by our API.
I think its sufficiently convenient

components/function-controller/design/extend_function_spec_proposal.md
kwiatekus: included in the PR

components/function-controller/design/extend_function_spec_proposal.md
kwiatekus: `mountPath` is required.
`SERVICE_BINDING_ROOT` ENV  is optional and used by users who are willing to use specialised libraries to consume mounted secrets as service bindings

components/function-controller/design/extend_function_spec_proposal.md
kwiatekus: in the end the implicit associations with service bindings (https://github.com/SAP/sap-btp-service-operator#service-binding or https://servicebinding.io/) turn out to be unwanted. 
To keep the API not clean from this domain the proposal was changed to `secretMounts`

components/function-controller/design/extend_function_spec_proposal.md
```diff
@@ -0,0 +1,499 @@
...
+    resources: # optional... required if spec.profile is empty
+      limits: ... #if profile empty
+      requests: ... #if profile empty
+
+
+```
+
+### Precedence, defaulting and validation
+
+`Profile` takes precedence over `resources`. If profile field is not set to "Custom" there should be no defaulting happening for the resources. Controller should fill the pod template resources accoriding to the selected profile preset.
```
dbadura, line 497:
```suggestion
`profile` takes precedence over `resources`. If profile field is not set there should be no defaulting happening for the resources. Controller should fill the pod template resources according to the selected profile preset.
```

components/function-controller/design/extend_function_spec_proposal.md
```diff
@@ -0,0 +1,499 @@
...
+spec: #Contains spec of Function and run stage (deployments, hpa)
+  runtime: nodejs16
+  source:
+    ...
+  replicas: 1
+  scalingConfig:
+    min: 1
+    max: 2
+
+  labels: #?
```
dbadura, line 439:
```suggestion
  labels:
```

components/function-controller/design/extend_function_spec_proposal.md
```diff
@@ -0,0 +1,499 @@
...
+    app: my-app
+  annotations: #?
```
dbadura, line 441:
```suggestion
  annotations:
```

dbadura:
Looks ok

components/function-controller/design/extend_function_spec_proposal.md
```diff
@@ -0,0 +1,529 @@
+# Extending CRD spec for Serverless v1alpha2
+
+## Summary
+
+The current Serverless API allows for limited configuration of the generated Function's deployment. 
+Currently users can only use ENVs to pass 3rd party service credentials but it doesn't allow for volume mounted secrets (which becomes the industry standard for [service bindings](https://servicebinding.io/application-developer/)).
```
grego952, line 6:
```suggestion
Currently, users can only use ENVs to pass 3rd party service credentials but it doesn't allow for volume-mounted Secrets (which become the industry standard for [service bindings](https://servicebinding.io/application-developer/)).
```

components/function-controller/design/extend_function_spec_proposal.md
```diff
@@ -0,0 +1,529 @@
...
+Moreover users have no control over the annotations that are applied on the function runtime pod. This excludes functions pods from features enabled by annotations (i.e custom log parsers via `fluentbit.io/parser: my-regex-parser`).
```
grego952, line 7:
```suggestion
Moreover, users have no control over the annotations applied on the function runtime pod. This excludes Function's Pods from features enabled by annotations (i.e., custom log parsers via `fluentbit.io/parser: my-regex-parser`).
```

components/function-controller/design/extend_function_spec_proposal.md
```diff
@@ -0,0 +1,529 @@
...
+
+
+
+## Motivation
+
+Give Serverless users the ability to:
+- Configure volume mounted secrets for Function's subresources.
```
grego952, line 14:
```suggestion
- Configure volume-mounted Secrets for Function's subresources.
```

components/function-controller/design/extend_function_spec_proposal.md
```diff
@@ -0,0 +1,529 @@
...
+- Configure labels and annotations for the Function runtime pod.
```
grego952, line 15:
```suggestion
- Configure labels and annotations for the Function's runtime Pod.
```

components/function-controller/design/extend_function_spec_proposal.md
```diff
@@ -0,0 +1,529 @@
...
+
+### Goals
+
+- Add more flexibility to the Serverless API - enable volume mounted secrets and pod annotations.
```
grego952, line 19:
```suggestion
- Add more flexibility to the Serverless API - enable volume-mounted Secrets and Pod annotations.
```

components/function-controller/design/extend_function_spec_proposal.md
```diff
@@ -0,0 +1,529 @@
...
+- Organise spec attributes belonging to runtime and build-time configuration (?)
+- Propose sample Function CR visualising different variants
+
+## Discussion points
+
+### Runtime, build-time separation
+
+Since Function CR is managing two workloads ( Deployment for runtime and Job for build-time ) we need to separate them in the spec in order to make it clear where do the mounts and annotations belong.
```
grego952, line 27:
```suggestion
Since Function CR is managing two workloads (Deployment for runtime and Job for build-time) we need to separate them in the spec in order to make it clear where the mounts and annotations belong. 
```

components/function-controller/design/extend_function_spec_proposal.md
```diff
@@ -0,0 +1,529 @@
...
+- Organise spec attributes belonging to runtime and build-time configuration (?)
+- Propose sample Function CR visualising different variants
+
+## Discussion points
+
+### Runtime, build-time separation
+
+Since Function CR is managing two workloads ( Deployment for runtime and Job for build-time ) we need to separate them in the spec in order to make it clear where do the mounts and annotations belong. 
+
+We can either make a clear separation, i.e:
```
grego952, line 29:
```suggestion
We can either make a clear separation, i.e.:
```

components/function-controller/design/extend_function_spec_proposal.md
```diff
@@ -0,0 +1,529 @@
...
+    # resources
+    # envs
+    # metadata
+  buildSpec:
+    # resources
+    # envs (?)
+    # metadata (?)
+```
+
+Alternatively we could promote runtime fields to the root (as the belong to kind: Function) and extract only the build-time fields
```
grego952, line 45:
```suggestion
Alternatively, we could promote runtime fields to the root (as they belong to kind: Function) and extract only the build-time fields 
```

components/function-controller/design/extend_function_spec_proposal.md
```diff
@@ -0,0 +1,529 @@
...
+  # metadata
+  buildSpec:
+    # resources
+    # envs (?)
+    # metadata (?)
+```
+
+### Mounts - own structure or k8s inherited
+
+Under the hood the secret mount becomes a volume mount in the runtime pod.
```
grego952, line 62:
```suggestion
Under the hood, the Secret mount becomes a volume mount in the runtime Pod.
```

components/function-controller/design/extend_function_spec_proposal.md
```diff
@@ -0,0 +1,529 @@
...
+We could :
```
grego952, line 63:
```suggestion
We could:
```

components/function-controller/design/extend_function_spec_proposal.md
```diff
@@ -0,0 +1,529 @@
...
+
+A) expose the k8s volume mount spec in function spec
```
grego952, line 65:
```suggestion
A) expose the k8s volume mount spec in the Function spec
```

components/function-controller/design/extend_function_spec_proposal.md
```diff
@@ -0,0 +1,529 @@
...
+
+Pros: 
+ - Generic solution - Allows mounting anything : secrets, config maps, any volumes
```
grego952, line 68:
```suggestion
 - Generic solution - Allows mounting anything: Secrets, ConfigMaps, any volumes
```

components/function-controller/design/extend_function_spec_proposal.md
```diff
@@ -0,0 +1,529 @@
...
+ - very easy to achieve (rewriting from function spec to pod teplate spec)
```
grego952, line 69:
```suggestion
 - very easy to achieve (rewriting from the Function spec to Pod template spec)
```

components/function-controller/design/extend_function_spec_proposal.md
```diff
@@ -0,0 +1,529 @@
...
+
+Cons: 
+ - Does not represent the service binging use case. User needs to translate service bindngs into volume mounts by themselves
```
grego952, line 72:
```suggestion
 - Does not represent the service binding use case. User needs to translate service bindings into volume mounts by themselves
```

components/function-controller/design/extend_function_spec_proposal.md
```diff
@@ -0,0 +1,529 @@
...
+ - Less compact ( elegant ) to configure requested service binding use case
```
grego952, line 73:
```suggestion
 - Less compact (elegant) to configure requested service binding use case
```

components/function-controller/design/extend_function_spec_proposal.md
```diff
@@ -0,0 +1,529 @@
...
+ - Noone requested it yet
```
grego952, line 74:
```suggestion
 - Noone has requested it yet
```

components/function-controller/design/extend_function_spec_proposal.md
```diff
@@ -0,0 +1,529 @@
...
+  - source: my-secret
+    mountPath: /bar # optional mount path
+  env:
+  - name: SERVICE_BINDING_ROOT # default mount path for service bindings
+    value: /foo
+```
+ Pros: 
+ - Purpose focused  
+ - Compact configuration - easy to adopt
+ - Less confusing (as volume mounts cause confusion as serverless functions are considered stateless and should not claim any persistance volumes )
```
grego952, line 105:
```suggestion
 - Less confusing (as volume mounts confuse Serverless Functions are considered stateless and should not claim any persistence volumes )
```

components/function-controller/design/extend_function_spec_proposal.md
```diff
@@ -0,0 +1,529 @@
...
+ - Enables using service binding natively with dedicated SDKs (i.e @sap/xsenv) - [related read](https://blogs.sap.com/2022/07/12/the-new-way-to-consume-service-bindings-on-kyma-runtime/)
+
+Cons: 
+ - Not allows to mount anything besides secrets
```
grego952, line 109:
```suggestion
 - Not allows mounting anything besides Secrets
```

components/function-controller/design/extend_function_spec_proposal.md
```diff
@@ -0,0 +1,529 @@
...
+
+A) and B) are not exclusive
+
+We could separate those cases. (See last 'compromise' option)
+
+
+### Metadata for function pod
```
grego952, line 116:
```suggestion
### Metadata for Function's Pod
```

components/function-controller/design/extend_function_spec_proposal.md
```diff
@@ -0,0 +1,529 @@
...
+      limits: ... #if_custom
+      requests: ... #if_custom
+    labels: #optional
+    annotations: #optional
+```
+
+### Option 2
+
+- almost the same as `Option 1` (same pros)
+- dif1: move some fields from `.spec` to the new struct `.spec.runtimeSpec` to clearly distinguish fields desired to be used in the `build` and `running` phases. For example in `Option 1` users may have questions after seeing `.spec.env` and `.spec.build.env` fields for example "is .spec.env dedicated for the running function's pod? Would the field be merged with .spec.build.env for the building job?"
```
grego952, line 203:
```suggestion
- dif1: move some fields from `.spec` to the new struct `.spec.runtimeSpec` to clearly distinguish fields desired to be used in the `build` and `running` phases. For example in `Option 1` users may have questions after seeing `.spec.env` and `.spec.build.env` fields for example "is .spec.env dedicated for the running Function's Pod? Would the field be merged with .spec.build.env for the building job?"
```

components/function-controller/design/extend_function_spec_proposal.md
```diff
@@ -0,0 +1,529 @@
...
+- dif2: rename the `.spec.profile` to the `.spec.resourcesProfile` to make this field more intuitive
+- dif3: this solution is simple but not as simple as `Option 1`
+
+>NOTE: the main idea is to close a specific configuration in a field that represents the specific phase of the function's lifecycle. It would be intuitive and easy to understand for a user that the `.spec.build` field contains configuration for the building phase and `.spec.runtimeSpec` for the running phase.
```
grego952, line 207:
```suggestion
>NOTE: the main idea is to close a specific configuration in a field that represents the specific phase of the Function's lifecycle. It would be intuitive and easy to understand for a user that the `.spec.build` field contains configuration for the building phase and `.spec.runtimeSpec` for the running phase.
```

components/function-controller/design/extend_function_spec_proposal.md
```diff
@@ -0,0 +1,529 @@
...
+    resourcesProfile: S / M / L / XL / ... / Custom
+    resources: # optional... required if spec.build.profile==Custom
+      limits: ... #if_custom
+      requests: ... #if_custom
+
+```
+
+### Option 3
+
+This option expose runtime pod configuration over build pod because runtime pod is final result, the build is transient phase.
```
grego952, line 252:
```suggestion
This option exposes runtime pod configuration over the build Pod because the runtime Pod is the final result, and the build is a transient phase.
```

components/function-controller/design/extend_function_spec_proposal.md
```diff
@@ -0,0 +1,529 @@
...
+        env: ...
+        volumeMounts: ...
+      volumes: ...
+```
+
+
+### Final version - the compromise
+
+ - clearly separate configuration of the build stage in the spec (treating `build` stage as second class citizen and keeping the main spec dedicated to the more important runtime stage).
+ - add convinient way to mount secrets (w/o polluting function API with dependencies to service bindings)
```
grego952, line 414:
```suggestion
 - add convenient way to mount Secrets (w/o polluting function API with dependencies to service bindings)
```

components/function-controller/design/extend_function_spec_proposal.md
```diff
@@ -0,0 +1,529 @@
...
+    resources: # optional... required if spec.profile is empty
+      limits: ... #if profile empty
+      requests: ... #if profile empty
+
+
+```
+
+### Precedence, defaulting and validation
+
+`profile` takes precedence over `resources`. If profile field is not set there should be no defaulting happening for the resources. Controller should fill the pod template resources according to the selected profile preset.
```
grego952, line 493:
```suggestion
`profile` takes precedence over `resources`. If the profile field is not set there should be no defaulting happening for the resources. The controller should fill the pod template resources according to the selected profile preset.
```

components/function-controller/design/extend_function_spec_proposal.md
```diff
@@ -0,0 +1,529 @@
...
+If profile field is set to "Custom", the user must then ( and only then ) set the values for resources manualy.
```
grego952, line 494:
```suggestion
If the profile field is set to "Custom", the user must then (and only then) set the values for resources manually.
```

grego952 (approved):

Issue #15401, opened 2022-09-08
[Telemetry Operator] Handshake errors


a-thaler:
**Description**
In the telemetry operator logs of one cluster I found several times such error blocks in the logs.

```
{"level":"INFO","timestamp":"2022-09-08T05:44:12Z","logger":"controller-runtime.certwatcher","caller":"logr@v1.2.0/logr.go:249","message":"Updated current TLS certificate","context":{}}
2022/09/08 05:50:57 http: TLS handshake error from xxx: remote error: tls: bad certificate
2022/09/08 05:50:58 http: TLS handshake error from xxx: remote error: tls: bad certificate
2022/09/08 05:51:00 http: TLS handshake error from xxx: remote error: tls: bad certificate
2022/09/08 05:51:04 http: TLS handshake error from xxx: remote error: tls: bad certificate
2022/09/08 05:51:13 http: TLS handshake error from xxx: remote error: tls: bad certificate
{"level":"INFO","timestamp":"2022-09-08T05:51:14Z","logger":"controller-runtime.certwatcher","caller":"logr@v1.2.0/logr.go:249","message":"Updated current TLS certificate","context":{}}
```

The error message is not in the proper json format, furthermore it should be clarified if that is just a random connectivity problem or an actual problem

chrkl:
PR https://github.com/kyma-project/kyma/pull/15873 ensures that messages from third party libraries are wrapped in JSON.

PR #15873, opened 2022-10-21
Fixing incorrect tls handshake error logging

Teneroy:
**Description**

- Copying the stream of standard logs to the klog. Since we expect every standard log to be properly logged, we are logging those using ERROR level;

**Related issue(s)**

#15401


components/telemetry-operator/go.mod
```diff
@@ -18,6 +18,7 @@ require (
 	k8s.io/api v0.25.3
 	k8s.io/apimachinery v0.25.3
 	k8s.io/client-go v0.25.3
+	k8s.io/klog/v2 v2.80.0
```
lindnerby, line 21: What's the difference between this and glog?

components/telemetry-operator/internal/logger/logger.go
```diff
@@ -29,5 +30,7 @@ func New(format, level string) (*Logger, error) {
 		return nil, err
 	}
 
+	klog.CopyStandardLogTo("ERROR")
```
lindnerby, line 33: Can you leave a comment what it does and why we need it please? 😁

lindnerby (changes requested):

components/telemetry-operator/go.mod
chrkl, line 21: Should be still an indirect dependency since it's only used by the logger package, right?

Teneroy: Not entirely sure. I can move it to the indirect manually, it won't really change anything. However, when I do `go mod tidy`, it automatically moves it from indirect block back to the direct one

components/telemetry-operator/internal/logger/logger.go
Teneroy: Sometimes logs from controller-runtime are being pushed not to the klog, but to the standard logs, which is happening when some component logs use regular "log" Go package. Because of that, we receive logs in incorrect format. Specifically, it happens when TLS handshake errors occur. Therefore, we need to redirect logs those are being written using standard logging mechanism to our klog. That is basically, why we need the CopyStandardLogTo method. However, we also need to decide to which level we push those logs. Since we consider all logs those are not being logged using klog as unexpected results, we can choose an ERROR as our level. That is why we use "ERROR" as methods' parameter

chrkl: I think the question was about having a comment in the code ;-)

lindnerby: Yes, short comment in code please :)

lindnerby (changes requested):

lindnerby (approved):

Issue #15660, opened 2022-09-28
Enable Java as Alpha Runtime  in Kyma Serverless


kwiatekus:
**Description**

Define configuration for java runtime

 - define config map that introduces config for java runtime - (draft in progress - https://github.com/kyma-project/kyma/pull/15910)
 - allow java runtime in the validation webhook
 - prove that Kyma serverless would be capable to build, package and serve a java application as a function.
 

**AC**
 - java runtime shoud be allowed only if feature flag is enabled. If FF enabled:
   - User should be able to set "java-alpha" in the `spec.runtime` field.
    - User should be able to provide java source code and dependencies
    - Function should build and be scheduled as a pod
 - Include java in the integration tests

kwiatekus:
For now we will revert the change:
 - Image used for runtime has 130+ vulnerabilities that needs to be addressed first
 - we will investigate a modified image based on sap machine ( SAP Java )

PR #15910, opened 2022-10-26
Enable Java function as alpha

dbadura:
**Description**

- Enable support for Java Runtime

**Related issue(s)**
https://github.com/kyma-project/kyma/issues/15660

### DIFF
--- a/components/function-controller/cmd/webhook/main.go
+++ b/components/function-controller/cmd/webhook/main.go
@@ -25,15 +25,6 @@ import (
 	ctrlwebhook "sigs.k8s.io/controller-runtime/pkg/webhook"
 )
 
-type config struct {
-	SystemNamespace     string               `envconfig:"default=kyma-system"`
-	WebhookServiceName  string               `envconfig:"default=serverless-webhook"`
-	WebhookSecretName   string               `envconfig:"default=serverless-webhook"`
-	WebhookPort         int                  `envconfig:"default=8443"`
-	WebhookConfigPath   string               `envconfig:"default=/appdata/config.yaml"`
-	WebhookFeatureFlags webhook.FeatureFlags `envconfig:""`
-}
-
 var (
 	scheme = runtime.NewScheme()
 )
@@ -51,12 +42,12 @@ func main() {
 	setupLog := ctrlzap.New().WithName("setup")
 
 	setupLog.Info("reading configuration")
-	cfg := &config{}
-	if err := envconfig.Init(cfg); err != nil {
+	cfg := &webhook.Config{}
+	if err := envconfig.InitWithPrefix(cfg, "WEBHOOK"); err != nil {
 		panic(errors.Wrap(err, "while reading env variables"))
 	}
 
-	logCfg, err := fileconfig.Load(cfg.WebhookConfigPath)
+	logCfg, err := fileconfig.Load(cfg.ConfigPath)
 	if err != nil {
 		setupLog.Error(err, "unable to load configuration file")
 		os.Exit(1)
@@ -80,7 +71,7 @@ func main() {
 	defer cancel()
 
 	logWithCtx := log.WithContext()
-	go logging.ReconfigureOnConfigChange(ctx, logWithCtx.Named("notifier"), atomic, cfg.WebhookConfigPath)
+	go logging.ReconfigureOnConfigChange(ctx, logWithCtx.Named("notifier"), atomic, cfg.ConfigPath)
 
 	logrZap := zapr.NewLogger(logWithCtx.Desugar())
 	ctrl.SetLogger(logrZap)
@@ -95,7 +86,7 @@ func main() {
 
 	mgr, err := manager.New(ctrl.GetConfigOrDie(), manager.Options{
 		Scheme:             scheme,
-		Port:               cfg.WebhookPort,
+		Port:               cfg.Port,
 		MetricsBindAddress: ":9090",
 		Logger:             logrZap,
 	})
@@ -109,9 +100,9 @@ func main() {
 	// because the webhook server needs to read it from disk to start.
 	if err := resources.SetupCertificates(
 		context.Background(),
-		cfg.WebhookSecretName,
+		cfg.SecretName,
 		cfg.SystemNamespace,
-		cfg.WebhookServiceName,
+		cfg.ServiceName,
 		logWithCtx.Named("setup-certificates")); err != nil {
 		logWithCtx.Error(err, "failed to setup certificates and webhook secret")
 		os.Exit(1)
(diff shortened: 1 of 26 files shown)


components/function-controller/cmd/webhook/main.go
```diff
@@ -26,11 +26,12 @@ import (
 )
 
 type config struct {
-	SystemNamespace    string `envconfig:"default=kyma-system"`
-	WebhookServiceName string `envconfig:"default=serverless-webhook"`
-	WebhookSecretName  string `envconfig:"default=serverless-webhook"`
-	WebhookPort        int    `envconfig:"default=8443"`
-	WebhookConfigPath  string `envconfig:"default=/appdata/config.yaml"`
+	SystemNamespace     string               `envconfig:"default=kyma-system"`
```
dbadura, line 29: I can move it all to package `webhook` and init config with prefix `WEBHOOK`

components/function-controller/internal/webhook/validating_webhook.go
```diff
@@ -16,17 +16,19 @@ import (
...
-func NewValidatingHook(configv1alpha1 *serverlessv1alpha1.ValidationConfig, configv1alpha2 *serverlessv1alpha2.ValidationConfig, client ctrlclient.Client) *ValidatingWebHook {
+func NewValidatingWebHook(configv1alpha1 *serverlessv1alpha1.ValidationConfig, configv1alpha2 *serverlessv1alpha2.ValidationConfig, featureFlags FeatureFlags, client ctrlclient.Client) *ValidatingWebHook {
 	return &ValidatingWebHook{
-		configv1alpha1: configv1alpha1,
-		configv1alpha2: configv1alpha2,
-		client:         client,
+		configv1alpha1:       configv1alpha1,
+		configv1alpha2:       configv1alpha2,
+		client:               client,
+		rejectJava17AlphaJVM: !featureFlags.Java17JvmAlphaEnabled,
```
moelsayed, line 31: Why are we using a negated field, and not the feature flag value directly?

components/function-controller/internal/webhook/validating_webhook.go
```diff
@@ -65,9 +67,15 @@ func (w *ValidatingWebHook) handleFunctionValidation(req admission.Request) admi
 			if err := w.decoder.Decode(req, fn); err != nil {
 				return admission.Errored(http.StatusBadRequest, err)
 			}
+
+			if w.rejectJava17AlphaJVM && fn.Spec.Runtime == serverlessv1alpha2.Java17JvmAlpha {
```
moelsayed, line 71: Nit: I think it's easier to read if we check the runtime first, then check it's enabled/rejected.

resources/serverless/charts/webhook/values.yaml
```diff
@@ -95,6 +95,8 @@ initContainer:
     readOnlyRootFilesystem: true # Mandatory
 
 values:
+  featureFlags:
+    java17AlphaEnabled: false
```
moelsayed, line 99: Do you think it makes sense to move `featureFlags` to it's own top level section? I know it's technically webhook configuration, but I feel it's important enough to have it's own top level section.

resources/serverless/templates/runtimes.yaml
```diff
@@ -73,3 +73,34 @@ data:
     RUN rm -rf /etc/pip.conf
 
     USER 1000
+---
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: dockerfile-java17-jvm-alpha
+  namespace: kyma-system
```
moelsayed, line 81: The namespace should be configurable.

resources/serverless/templates/runtimes.yaml
```diff
@@ -73,3 +73,34 @@ data:
...
+  labels:
+    serverless.kyma-project.io/config: runtime
+    serverless.kyma-project.io/runtime: java17-jvm-alpha
+data:
+  Dockerfile: |-
+    FROM {{ include "imageurl" (dict "reg" .Values.global.containerRegistry "img" .Values.global.images.function_runtime_java17_jvm_alpha) }} as builder
+
+    ARG BUILD_DIR=/build
+    ARG SOURCE_DIR=/src
+    ARG DEPS_DIR=/src
```
moelsayed, line 91: Do we need both? The have the same value are used once each.

resources/serverless/templates/runtimes.yaml
```diff
@@ -73,3 +73,34 @@ data:
...
+    WORKDIR $BUILD_DIR
+
+    COPY $DEPS_DIR/pom.xml $BUILD_DIR/handler-pom.xml
+
+    COPY $SOURCE_DIR/Handler.java $BUILD_DIR/src/main/java/io/project/kyma/serverless/handler/Handler.java
+    RUN mvn dependency:go-offline -f handler-pom.xml
+
+    RUN mvn clean && mvn package -f pom.xml
+
+    FROM eclipse-temurin:17-jre-alpine
```
moelsayed, line 101: Should we host this image as well?

moelsayed (commented):
Looks great, thank you! Just a few questions and nit comments.

components/function-controller/cmd/webhook/main.go
moelsayed: That sounds like a great idea :+1:

resources/serverless/templates/runtimes.yaml
dbadura: Unfortunately yes. I've done it to be able to use the same image for local build and kaniko build.

resources/serverless/templates/runtimes.yaml
dbadura: Yes, I will add in another PR.

resources/serverless/charts/webhook/values.yaml
dbadura: I have a feeling that our values are distribution in the whole `values.yaml`. 
I think that it should be refactored in the future.

components/function-controller/internal/webhook/validating_webhook.go
dbadura: In context of webhook, the runtime can be rejected. 
I've done it for readability.

resources/serverless/templates/runtimes.yaml
moelsayed: Probably a good idea to add a comment about that to avoid "fixing it" in the future ;)

moelsayed (approved):

resources/serverless/templates/runtimes.yaml
dbadura: Added :)

moelsayed (approved):
LGTM

Issue #15831, opened 2022-10-17
Updated TracePipeline updates the otel-collector configuration


a-thaler:
**Description**

The user of a TracePipeline expects that a update of the resource will result in an effective otel-collector config. That is not the case currently and you would need to re-create the pod/TracePipeline?.
A nice idea might be to just annotate the Deploment of the otel-collector with a hash representing the config value so that kubernetes will take care of the restart.

**Criterias**
- A changed TracePipeline will lead to otel-collector setup reflecting that change

**Reasons**

**Attachments**

chrkl:
With PR #15941, the telemetry-operator adds an annotation with a hash over the ConfigMap and Secret data to the Otel Collector Pod template (same way how it's usually done in Helm charts). With this, Kubernetes automatically recreates the pod if the configuration has changed.

PR #15941, opened 2022-10-28
Enfore trace collector restart if configuration changes

chrkl:
**Description**

The OpenTelemetry Collector does not automatically load changed configurations and requires a restart after a config change. This PR adds a hash of the relevant ConfigMap and Secret to the collector Pod template to enforce a restart when a new config is applied.

- Enfore trace collector restart if configuration changes

**Related issue(s)**

#15831

### DIFF
--- a/components/telemetry-operator/controller/tracepipeline/controller.go
+++ b/components/telemetry-operator/controller/tracepipeline/controller.go
@@ -117,7 +117,7 @@ func (r *Reconciler) installOrUpgradeOtelCollector(ctx context.Context, tracing 
 		return fmt.Errorf("failed to create otel collector configmap: %w", err)
 	}
 
-	configHash := NewConfigHash([]corev1.ConfigMap{*configMap}, []corev1.Secret{*secret}).Build()
+	configHash := CreateConfigHash([]corev1.ConfigMap{*configMap}, []corev1.Secret{*secret})
 	deployment := makeDeployment(r.config, configHash)
 	if err = controllerutil.SetControllerReference(tracing, deployment, r.Scheme); err != nil {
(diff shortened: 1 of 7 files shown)


components/telemetry-operator/controller/tracepipeline/resources.go
```diff
@@ -17,6 +17,7 @@ import (
 const (
 	basicAuthHeaderVariable = "BASIC_AUTH_HEADER"
 	otlpEndpointVariable    = "OTLP_ENDPOINT"
+	configHashAnnotationKey = "checksum/config"
```
lindnerby, line 20: What do you think about a maybe convention driven key for the annotation inspired by the  `kubectl.kubernetes.io/last-applied-configuration`, in our case something like `kyma-project.io/current-config-sha`?

components/telemetry-operator/controller/tracepipeline/controller.go
```diff
@@ -117,7 +117,8 @@ func (r *Reconciler) installOrUpgradeOtelCollector(ctx context.Context, tracing
 		return fmt.Errorf("failed to create otel collector configmap: %w", err)
 	}
 
-	deployment := makeDeployment(r.config)
+	configHash := NewConfigHash().AddByteMap(secretData).AddStringMap(configMap.Data).Build()
```
lindnerby, line 120: Since order of `AddByteMap` and `AddStringMap` matters and we want to also make this internal of a function? 
Also with a signature like `CreateConfigHash(secrets []v1.Secret, cms []configsMaps)`

lindnerby (approved):

PR #15952, opened 2022-10-31
Update go dependencies for application connector and compass-runtime-agent: net, crypto and text

koala7659:
For modules:

- central-application-connectivity-validator 
- central-application-gateway
- compass-runtime-agent
- application-operator


franpog859 (approved):
LGTM. I think after this PR we'll need to bump also the `github.com/kyma-project/kyma/components/application-operator` dependency in the components that use it

Issue #15916, opened 2022-10-27
Remove PSP and set security context for istio, ory and api-gateway components


strekm:
**Description**

Since PSP will be removed with k83 1.25 remove it from following kyma components:
- istio -> S -> @dariusztutaj 
- ory -> M -> @barchw 
- api-gateway -> XS -> @triffer 

Make sure proper security context is applied for pods in mentioned components.

**Reasons**

Applying POD Security Standards

**Attachments**

https://github.com/kyma-project/kyma/issues/15897

dariusztutaj:
ecause of using Istio CNI plugin, it's not needed to change anything. Istio is already compliant.
The CNI handling the netns setup replaces the current Istio approach using a NET_ADMIN privileged initContainers container, istio-init, injected in the pods along with istio-proxy sidecars. This removes the need for a privileged, NET_ADMIN container in the Istio users' application pods. https://github.com/istio/istio/tree/master/cni

dariusztutaj:
Installing Istio with CNI plugin deploys _istio-cni-node_ DaemonSet.  Install container, which is part of this daemonset, requires to be run with aid and gid 0, because of using host path directories.
**Related links**:
https://istio.io/latest/docs/setup/additional-setup/cni/#install-istio-with-cni-plugin
https://github.com/istio/istio/issues/35902

PR #15965, opened 2022-11-02
Remove PSP from api-gatway

triffer:
**Description**

- Remove PSP
- Add restricted level security context configuration

**Related issue(s)**

- https://github.com/kyma-project/kyma/issues/15916


resources/api-gateway/templates/webhook-certificates.yaml
```diff
@@ -8,12 +8,17 @@ metadata:
...
-  replicas: 1
   template:
     metadata:
       annotations:
         sidecar.istio.io/inject: "false"
     spec:
+      securityContext:
+        runAsUser: 65534
+        runAsGroup: 65534
+        runAsNonRoot: true
```
cnvergence, line 19: Shall we also include `readOnlyRootFilesystem: true`?

resources/api-gateway/values.yaml
```diff
@@ -16,10 +16,11 @@ deployment:
     capabilities:
       drop:
       - ALL
-    privileged: false
     runAsGroup: 65534
     runAsNonRoot: true
     runAsUser: 65534
+    seccompProfile:
+      type: RuntimeDefault
```
cnvergence, line 23: Shall we also include `readOnlyRootFilesystem: true?`

resources/api-gateway/templates/webhook-certificates.yaml
triffer: Thanks, good finding! It's not relevant for Pod Security Admission, but still we should have it.

cnvergence (approved):
lgtm

Issue #15897, opened 2022-10-24
Check activation of Pod Security Standards


TorstenD-SAP:
**Description**

We have to check, whether our Pod specifications meat the requirements of the Pod Security Standards or not. I created a [document describing the problem and listing a couple of steps to perform](https://github.com/kyma-project/community/tree/main/concepts/psp-replacement) in order to find out if we are able to activate Pod Security Standards in which level. The overall goal should be to enforce the Pod Security Standard _restricted_ on the Kyma namespace.

The [steps](https://github.com/kyma-project/community/tree/main/concepts/psp-replacement#compliance-with-pod-security-standards) listed in the document mentioned above must be performed and the results must be added to this issue.

**Reasons**

Pod Security Policies is no longer part of Kubernetes in version 1.25. As a replacement, Kubernetes introduced the so-called [Pod Security Standards](https://kubernetes.io/docs/concepts/security/pod-security-standards) and an admission controller ([Pod Security Admission](https://kubernetes.io/docs/concepts/security/pod-security-admission/)) to enforce the Pod Security Standards.

To not wait until Pod Security Policies are finally removed, this should be done as part of the upcoming Kubernetes update.
See #15732 Kubernetes upgrade from ver 1.23 to 1.24

**Attachments**

a-thaler:
We will have a problem with Daemonsets which typically require access to the node, see also my [PR](https://github.com/kyma-project/kyma/pull/15744) for the observability area which has two daemonsets

pPrecel:
The output from Otters' investigation is described [here](https://github.com/kyma-project/kyma/issues/15245#issuecomment-1293215929).

anoipm:
Function-controller (serverless) in kyma was adapted to be compatible with PSS. Now we have some incompabilities described [here](https://github.com/kyma-project/kyma/issues/15923#issuecomment-1310140557)

dariusztutaj:
Istio with CNI plugin enabled has one container which needs to be run with uid 0. This container is not directly attached to the workloads. More info [here](https://github.com/kyma-project/kyma/issues/15916#issuecomment-1316511519)

marcobebway:
Here is the applied security contexts config for:
- [Eventing](https://github.com/kyma-project/kyma/issues/15919#issuecomment-1412352214).
- [KMC](https://github.com/kyma-project/kyma/issues/15919#issuecomment-1412354351).

a-thaler:
[...]

Issue #15677, opened 2022-09-29
Migrate deprecated PodSecurityPolicy to PodSecurityAdmission in Compass Runtime Agent chart


franpog859:
**Description**

PodSecurityPolicy is deprecated and going to be removed in Kubernetes 1.25. We need to migrate these resources in order to function well on this k8s version. I found one in [the Compass Runtime Agent chart](https://github.com/kyma-project/kyma/blob/main/resources/compass-runtime-agent/templates/pod-security-policy.yaml). Please, migrate it in the same way, it's done [here](https://github.com/kyma-project/kyma/issues/15245). Check also the [Application Connector charts](https://github.com/kyma-project/kyma/tree/main/resources/application-connector). I didn't find any PSP there but I might have missed something

**Attachments**

Here is [the migration guide](https://kubernetes.io/docs/tasks/configure-pod-container/migrate-from-psp/).
Here is the [recommendation](https://github.com/kyma-project/community/pull/727/files) for Kyma project.

franpog859:
@TorstenD-SAP , thanks for looking at this issue! Just let us know when there is something to be done on our side

franpog859:
It's done in this PR - https://github.com/kyma-project/kyma/pull/15980 - Please, reopen it if needed

Issue #15245, opened 2022-08-24
Migrate deprecated PodSecurity Policy to PodSecurity Admission in serverless charts


grego952:
**Description**

Deprecated PSP is used in serverless charts and mentioned in documentation

 - [x] Investigate the required security standard levels for functions (as customer workloads in their namespaces)
 - [x] Investigate the required security standard levels for serverless workloads in kyma system namespace
 - [x] Remove PSPs `severless-build` and `serverless-function` and their service accounts and roles - https://github.com/kyma-project/kyma/issues/15921
 - [x] Apply security context for `baseline` level to the function build job and runtime pod - https://github.com/kyma-project/kyma/issues/15923
 - [x] Apply `restricted` security context in serverless charts for serverless controller and webhook - https://github.com/kyma-project/kyma/pull/16104
 - [x] Update documentation:
    - [x] update [Security considerations](https://kyma-project.io/docs/kyma/main/01-overview/main-areas/serverless/svls-06-security-considerations) document 
    - [x] document maximum restriction level ( applied on namespace ) that allows to run and build functions
    
    
**Reasons**
 PodSecurity Policy were deprecated in kubernetes 1.21 and will be removed in 1.25

**Attachments**
[migration guide](https://kubernetes.io/docs/tasks/configure-pod-container/migrate-from-psp/)

kwiatekus:
@dbadura pls document the findings about removing PSP in serverless charts

dbadura:
I removed PSP for function pod and add PSA label: `pod-security.kubernetes.io/warn: restricted`. It worked.

kwiatekus:
@dbadura should'nt we use enforce mode to really test the PSP removal ?
`pod-security.kubernetes.io/enforce: restricted`
https://kubernetes.io/docs/concepts/security/pod-security-admission/

kwiatekus:
We need to find out if build jobs can run with baseline- level  PSA enforced

kwiatekus:
With PSA the restrictions are enforced on namespace level. Its fundamentally different usability concept comparing to PSPs.
If we need to enforce restrictions for function-managed workloads ( runtime pod and build job ) we'd need to control PSA labels in user namespaces as function workloads dont live in kyma-system.

My standpoint is that it's up to the user where ( in which namespaces ) the PSAs are enforced.

[...]

PR #15980, opened 2022-11-02
Remove PSP from CRA

franpog859:
**Description**

- Remove PodSecurityPolicy from the Compass Runtime Agent chart

**Related issue(s)**

Some sort of an epic, I'd say - https://github.com/kyma-project/kyma/issues/15897
The migration instructional - https://kubernetes.io/docs/tasks/configure-pod-container/migrate-from-psp/
The proposal - https://github.com/kyma-project/community/tree/main/concepts/psp-replacement
The Compass Runtime Agent issue - https://github.com/kyma-project/kyma/issues/15677

Related to the PSP migration issue in the Serverless area - https://github.com/kyma-project/kyma/issues/15245
PR with the serverless PSPs removal -  https://github.com/kyma-project/kyma/pull/15946
PR with the global PSPs removal - https://github.com/kyma-project/kyma/pull/15929


resources/compass-runtime-agent/templates/deployment.yaml
```diff
@@ -19,15 +19,36 @@ spec:
...
         - name: {{ .Chart.Name }}
+          securityContext:
+            privileged: false
+            allowPrivilegeEscalation: false
+            runAsNonRoot: true
+            capabilities:
+              drop:
+                - ALL
+              add:
+                - NET_BIND_SERVICE
```
a-thaler, line 44: Is that capability really needed? If not the "ad" section should be removed

resources/compass-runtime-agent/templates/deployment.yaml
```diff
@@ -19,15 +19,36 @@ spec:
...
+            seLinuxOptions:
```
a-thaler, line 45: I think there is no need to explicitly set this seLinux setting. @TorstenD-SAP

resources/compass-runtime-agent/templates/deployment.yaml
```diff
@@ -19,15 +19,36 @@ spec:
...
+              type: container_t
+            procMount: default
```
a-thaler, line 47: I think this is also not needed as the default value is already good enough. @TorstenD-SAP

resources/compass-runtime-agent/templates/deployment.yaml
```diff
@@ -19,15 +19,36 @@ spec:
     metadata:
       annotations:
         sidecar.istio.io/inject: "true"
+        container.apparmor.security.beta.kubernetes.io: runtime/default
```
a-thaler, line 22: In the documentation is states that by default the runtime/default profile is active, see https://kubernetes.io/docs/tutorials/security/apparmor/#profile-reference

Adding the annotation using this profile does not bring any value, it just makes it more transparent. However, we need to maintain that annotations.. also when the feature becomes stable, it will be moved into the container spec which requires migration.

I propose to not add it to every single container as a mandatory action just to have it more transparent that we use the recommended default. @TorstenD-SAP

TorstenD-SAP: @a-thaler You are right, adding the annotation in that way makes no sense.

Therefor I agree, that we should NOT add it to the deployment spec. I already changed the provided templates (https://github.com/kyma-project/community/pull/738)

resources/compass-runtime-agent/templates/deployment.yaml
TorstenD-SAP: @franpog859 The capability `NET_BIND_SERVICE` is the only one which is allowed for restricted Pod Security Standard. Adding this is optional and only necessary if you really need this capability. Otherwise please remove the whole `add` section.

resources/compass-runtime-agent/templates/deployment.yaml
TorstenD-SAP: I don't see a huge maintenance effort for this because there is nothing to decide or change. Just add it to the spec as it is. Therefor I vote for leaving it in.

TorstenD-SAP (commented):
See my comments inline.

resources/compass-runtime-agent/templates/deployment.yaml
franpog859: Thanks!

akgalwas (approved):
LGTM

Issue #15978, opened 2022-11-02
Unsubscribing NATS subscriptions does not delete consumers


raypinto:
**Description**

When we delete kyma subscriptions or modify kyma subscription filters, the underlying NATS subscriptions are unsubscribed which eventually deletes the consumers. 

But if by any chance the NATS subscriptions becomes invalid, then we create a new NATS subscription and **bind** an existing consumer to this subscription. On deletion of this NATS subscription, it does not delete the underlying consumer since we manually bind the consumer and hence it results in dangling consumers.

**Expected result**

Deletion or modification of existing subscription filters should delete the old consumers.

**Actual result**

Deletion or modification of existing subscription filters does not delete the old consumers and we end up with dangling consumers.

**Steps to reproduce**

- Create a kyma subscription with one filter
- The number of NATS consumers should be one
- Restart the NATS server, make sure the reconnects were exhausted and check if [this line](https://github.com/kyma-project/kyma/blob/5cda76fc199c066e71ac43a4ae7f36e8fcb7263f/components/eventing-controller/pkg/backend/nats/jetstream/jetstream.go#L443) of code was called
- The number of NATS consumers should still be one
- Now modify the filter and change the value
- This should create a new consumer and delete the old one
- But there will be two NATS consumers instead of one.

**Tasks**

- [x] Reproduce the issue
- [ ] Fix the bug and write tests
- [ ] Create another function in the controller that will check all the existing consumers and see if they have a corresponding kyma filter(limited to our working stream). If not, delete it
- [ ] Cherry pick this change to 2.8 release

muralov:
There two cases dangling consumers are produced:
* Change subscription CR filter(s) while EC is down
   `Solution`: delete subscription consumers not associated with any of the filters. Consumer should be invalid as it has no subscription

* Change subscription CR filter(s) while NATS JetStream is down
  `Solution`: compare EC subscriptions cache to actual JetStream subscriptions while reconciliation

PR #16038, opened 2022-11-08
Delete Consumer with Unsubscription

muralov:
**Description**

Delete the NAT JetStream consumer always after unsubscription explicitly

- Delete consumer always after unsubscribing
- Adapt filter change/delete integration tests

**Related issue(s)**
[#15978](https://github.com/kyma-project/kyma/issues/15978)

### DIFF
--- a/components/eventing-controller/pkg/backend/nats/jetstream/jetstream.go
+++ b/components/eventing-controller/pkg/backend/nats/jetstream/jetstream.go
@@ -599,7 +599,6 @@ func (js *JetStream) deleteSubscriptionFromJetStream(jsSub backendnats.Subscribe
 		}
 	}
 
-	// if JS sub is not valid, then we need to delete the consumer on JetStream
 	if err := js.deleteConsumerFromJetStream(jsSubKey.ConsumerName()); err != nil &&
 		!errors.Is(err, nats.ErrConsumerNotFound) {
 		return fmt.Errorf("failed to delete consumer %s from JetStream: %w", jsSubKey.ConsumerName(), err)
--- a/components/eventing-controller/pkg/backend/nats/jetstream/jetstream.go
+++ b/components/eventing-controller/pkg/backend/nats/jetstream/jetstream.go
@@ -597,11 +597,12 @@ func (js *JetStream) deleteSubscriptionFromJetStream(jsSub backendnats.Subscribe
 		if err := jsSub.Unsubscribe(); err != nil {
 			return xerrors.Errorf("failed to unsubscribe subscription %v from JetStream: %v", jsSub, err)
 		}
-	} else {
-		// if JS sub is not valid, then we need to delete the consumer on JetStream
-		if err := js.deleteConsumerFromJetStream(jsSubKey.ConsumerName()); err != nil {
-			return err
-		}
+	}
+
+	// if JS sub is not valid, then we need to delete the consumer on JetStream
+	if err := js.deleteConsumerFromJetStream(jsSubKey.ConsumerName()); err != nil &&
+		!errors.Is(err, nats.ErrConsumerNotFound) {
+		return fmt.Errorf("failed to delete consumer %s from JetStream: %w", jsSubKey.ConsumerName(), err)
 	}
 
 	delete(js.subscriptions, jsSubKey)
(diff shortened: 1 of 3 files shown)


components/eventing-controller/pkg/backend/nats/jetstream/jetstream_integration_test.go
```diff
@@ -531,15 +547,19 @@ func TestJetStreamSubAfterSync_FilterRemoved(t *testing.T) {
...
-	require.True(t, jsSub.IsValid())
+	firstJsSubject := jsBackend.GetJetStreamSubject(firstSubject)
+	firstJsSubKey := NewSubscriptionSubjectIdentifier(sub, firstJsSubject)
+	firstJsSub := jsBackend.subscriptions[firstJsSubKey]
+	require.NotNil(t, firstJsSub)
+	require.True(t, firstJsSub.IsValid())
+	// make sure old filter doesn't have any JetStream consumer
+	secondCon, err := secondJsSub.ConsumerInfo()
+	require.Nil(t, secondCon)
+	require.ErrorIs(t, err, nats.ErrConsumerNotFound)
```
mfaizanse, line 558: I ran the tests `make test-local` with and without the bug fix, and in both cases the tests were passing. Can we make the test fail when the bug fix is not there?

components/eventing-controller/pkg/backend/nats/jetstream/jetstream.go
```diff
@@ -597,11 +597,12 @@ func (js *JetStream) deleteSubscriptionFromJetStream(jsSub backendnats.Subscribe
...
 			return xerrors.Errorf("failed to unsubscribe subscription %v from JetStream: %v", jsSub, err)
 		}
-	} else {
-		// if JS sub is not valid, then we need to delete the consumer on JetStream
-		if err := js.deleteConsumerFromJetStream(jsSubKey.ConsumerName()); err != nil {
-			return err
-		}
+	}
+
+	// if JS sub is not valid, then we need to delete the consumer on JetStream
```
mfaizanse, line 602: (nit) the comment needs to be updated.

components/eventing-controller/pkg/backend/nats/jetstream/jetstream_integration_test.go
```diff
@@ -1398,3 +1418,124 @@ func TestJetStream_CheckConsumerConfig(t *testing.T) {
...
+		return info != nil && err == nil
+	}, 60*time.Second, 5*time.Second)
+
+	// SyncSubscription binds the existing subscription to JetStream created one
+	err = jsBackend.SyncSubscription(sub)
+	// then
+	require.NoError(t, err)
+
+	// given
+	// Now, change the filter in subscription
```
mfaizanse, line 1509: (nit) why dont we delete the kyma subscription instead of changing the filter?

components/eventing-controller/pkg/backend/nats/jetstream/jetstream_integration_test.go
```diff
@@ -504,6 +515,11 @@ func TestJetStreamSubAfterSync_FilterRemoved(t *testing.T) {
 	secondSubject, err := backendnats.GetCleanSubject(sub.Spec.Filter.Filters[1], testEnvironment.cleaner)
 	require.NoError(t, err)
 	require.NotEmpty(t, secondSubject)
+	secondJsSubject := jsBackend.GetJetStreamSubject(secondSubject)
+	secondJsSubKey := NewSubscriptionSubjectIdentifier(sub, secondJsSubject)
+	secondJsSub := jsBackend.subscriptions[secondJsSubKey]
+	require.NotNil(t, secondJsSub)
+	require.True(t, secondJsSub.IsValid())
```
mfaizanse, line 522: (nit) Do we still need these changes for other tests? As we now have a new separate test. Also, these changes are meaningless w.r.t. the bug fix.

muralov: They might be meaning less w.r.t the bug fix, but they are additional checks, which make sure JS consumer added/deleted for subscription subject filters add/change/delete operations.

components/eventing-controller/pkg/backend/nats/jetstream/jetstream_integration_test.go
muralov: yes we could also delete it. I just used change as I was testing with change. I think we can leave it so.

mfaizanse (approved):

mfaizanse (approved):

Issue #15994, opened 2022-11-03
kyma-upgrade-k3d-kyma2-to-main fails on Tracing tests


Tomasz-Smelcerz-SAP:
**Description**
The pipeline fails from Nov 01.

The error from the logs:

```
   1) Upgrade test tests
       Tracing Tests:
         with Nats backend
           in-cluster structured event should have correct tracing spans:
      AssertionError: expected false to be true      
      actual expected
      
      falsetrue
      
      at checkTrace (test/fixtures/commerce-mock/index.js:344:25) 
```

**Expected result**
Tests are passing

**Actual result**

Tests are failing

**Steps to reproduce**
See the following links:

https://status.build.kyma-project.io/view/gs/kyma-prow-logs/logs/kyma-upgrade-k3d-kyma2-to-main/1587867151023214592
https://status.build.kyma-project.io/view/gs/kyma-prow-logs/logs/kyma-upgrade-k3d-kyma2-to-main/1587836951820505088
**Troubleshooting**

n/a

Tomasz-Smelcerz-SAP:
Same error happened in another pipeline: `kyma-upgrade-gardener-kyma2-to-main-reconciler-main`: https://status.build.kyma-project.io/view/gs/kyma-prow-logs/logs/kyma-upgrade-gardener-kyma2-to-main-reconciler-main/1587867150926745600

Tomasz-Smelcerz-SAP:
Another pipeline with the same error: `pre-main-control-plane-reconciler-upgrade-kyma2-latest-to-main-gardener`
https://status.build.kyma-project.io/view/gs/kyma-prow-logs/pr-logs/pull/kyma-project_control-plane/2212/pre-main-control-plane-reconciler-upgrade-kyma2-latest-to-main-gardener/1587861536104255488

mfaizanse:
The issue is also happening for upgrade from OSS Kyma `v2.8.1` to `v2.8.2`.

**Before Upgrade (on OSS kyma `v2.8.1`)**
![image](https://user-images.githubusercontent.com/9897945/200832871-1b170717-4019-4f9b-bad8-48ddf2c50a12.png)

**After Upgrade (on OSS kyma `v2.8.2`)**
![image](https://user-images.githubusercontent.com/9897945/200832912-04725147-e526-4074-b0c1-90571e0d575b.png)

The tracing is getting broken after the upgrade.

Sawthis:
[...]

PR #16052, opened 2022-11-09
Skip OTLP collector endpoint availability check in functions

kwiatekus:
**Description**

This PR removes the availability check of the OTLP endpoint in serverless functions.
Sometimes it happens (i.e whenever kyma upgrades include new istio versions ) that functions and jaeger gets restarted simultaneously.
At the function start time jaeger ( or its side-car ) might be still warming up. In such case the function not being able to reach it at start time would be configured to not send traces (until it is restarted again at the time when jaeger is available).

**Related issue(s)**
https://github.com/kyma-project/kyma/issues/15994


pPrecel (approved):
LGTM

PR #16061, opened 2022-11-10
Apply workaround for python CVE

dbadura:
**Description**

- Apply workaround from python: https://github.com/python/cpython/issues/97514 to mitigate CVE-2022-42919

**Related issue(s)**


moelsayed (approved):
LGTM

PR #16115, opened 2022-11-17
Change http response code based on backend error

k15r:
**Description**

- if a stream is full (jetstream) we report now 507
- if a stream does not exists (jetstream) we report 502
- ...

**Related issue(s)**

### DIFF
--- a/components/event-publisher-proxy/pkg/handler/handler_test.go
+++ b/components/event-publisher-proxy/pkg/handler/handler_test.go
@@ -329,6 +329,25 @@ func TestHandler_publishCloudEvents(t *testing.T) {
 				eventing_epp_errors_total 1
 			`,
 		},
+		{
+			name: "Publish binary CloudEvent but backend is full",
+			fields: fields{
+				Sender: &GenericSenderStub{
+					Err: fmt.Errorf("oh no, i cannot send: %w", sender.ErrInsufficientStorage),
+				},
+				collector:        metrics.NewCollector(latency),
+				eventTypeCleaner: &eventtypetest.CleanerStub{},
+			},
+			args: args{
+				request: CreateValidBinaryRequest(t),
+			},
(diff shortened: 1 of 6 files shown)


components/event-publisher-proxy/pkg/handler/handler_test.go
```diff
@@ -429,6 +448,50 @@ func TestHandler_publishLegacyEventsAsCE(t *testing.T) {
 					eventing_epp_requests_total{destination_service="FOO",response_code="204"} 1
 				`,
 		},
+		{
+			name: "Send valid legacy event but cannot send to backend due target not found (eg stream missing)",
```
friedrichwilken, line 452:
```suggestion
			name: "Send valid legacy event but cannot send to backend due to target not found (e.g. stream missing)",
```

components/event-publisher-proxy/pkg/sender/jetstream/jetstream.go
```diff
@@ -26,6 +26,7 @@ const (
 	natsBackend           = "nats"
 	jestreamHandlerName   = "jetstream-handler"
 	noSpaceLeftErrMessage = "no space left on device"
+	natsMaxBytesExceeded  = "nats: maximum bytes exceeded"
```
VladislavPaskar, line 29: can we use the `ErrMaxBytes` from nats-server library and use the n`errors.Is` instead of checking for error text ?

friedrichwilken (changes requested):

components/event-publisher-proxy/pkg/sender/jetstream/jetstream.go
k15r: No we cannot, but i improved the code to evaluate the errorcode returned by the api

friedrichwilken (approved):
/lgtm

PR #13809, opened 2022-04-01
Fix Ory port names in services to comply with Istio convention

cnvergence:
**Description**

We should follow [official Istio documentation about port naming convention](https://istio.io/latest/docs/reference/config/analysis/ist0118/). 

Istioctl analyze output:
```bash
Info [IST0118] (Service kyma-system/ory-hydra-admin) Port name status-port (port: 15020, targetPort: 15020) doesn't follow the naming convention of Istio port.
Info [IST0118] (Service kyma-system/ory-hydra-maester-metrics) Port name metrics (port: 8080, targetPort: 8080) doesn't follow the naming convention of Istio port.
Info [IST0118] (Service kyma-system/ory-oathkeeper-maester-metrics) Port name metrics (port: 8080, targetPort: 8080) doesn't follow the naming convention of Istio port.
Info [IST0118] (Service kyma-system/ory-oathkeeper-proxy) Port name status-port (port: 15020, targetPort: 15020) doesn't follow the naming convention of Istio port.
```

**Related issue(s)**

See: 
- https://github.com/kyma-project/kyma/issues/11686

PR #16120, opened 2022-11-17
Fix port names for Istio analyzer

cnvergence:
**Related issue(s)**

See also:
- https://github.com/kyma-project/kyma/issues/11686


skhalash:
@cnvergence Could you test an upgrade from the `main` to your PR? If I remember correctly, a port name is an immutable field in a `service`. In this case, an upgrade would fail.

cnvergence:
@skhalash should not be causing problems with service, but I can test it out.
I did several times before - example: https://github.com/kyma-project/kyma/pull/13811, https://github.com/kyma-project/kyma/pull/13809

cnvergence:
@skhalash confirmed - upgrade works

skhalash (approved):

PR #16131, opened 2022-11-18
Add stream re-creation check in eventing FI tests

raypinto:
**Description**

- Get the stream creation timestamp during the preparation phase, and store the value in a configmap
- The value is obtained by exposing the `eventing-nats` svc on port `8222` via an APIRule.
- During the eventing  upgrade test, we check if these values match
- We skip this test if either the creation timestamp is not exposed by the service (before NATS 2.9.7) or if the configmap is not found (the prep stage is run from the FI test of the previous release and hence the cm is not created).

Note: This test will only be effective (not skipped) after kyma 2.10 release.

**Related issue(s)**

### DIFF
--- a/tests/fast-integration/eventing-test/eventing-test.js
+++ b/tests/fast-integration/eventing-test/eventing-test.js
@@ -32,6 +32,7 @@ const {
   k8sApply,
   deleteK8sPod,
   eventingSubscription, waitForEndpoint, waitForPodStatusWithLabel, waitForPodWithLabelAndCondition,
+  createAPIRuleForService, getConfigMap, deleteApiRule,
 } = require('../utils');
 const {
   eventingMonitoringTest,
@@ -48,6 +49,11 @@ const {
   getNatsPods,
   getStreamConfigForJetStream,
   skipAtLeastOnceDeliveryTest,
+  skipStreamReCreationTest,
+  getJetStreamStreamData,
+  streamDataConfigMapName,
+  eventingNatsSvcName,
+  eventingNatsApiRuleAName,
   subscriptionNames,
 } = require('./utils');
 const {
@@ -104,6 +110,7 @@ describe('Eventing tests', function() {
     }
 
     if (backend === natsBackend) {
+      testStreamNotReCreated();
       testJetStreamAtLeastOnceDelivery();
     }
   }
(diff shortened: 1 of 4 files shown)


k15r:
@pbochynski : this test is part of the eventing specific upgrade test. its purpose is to guarantee, that during an upgrade the stream on jetstream is silently regenerated as this would mean data loss on nats. It cannot be moved down the pyramid, as it requires the full upgrade flow to be executed.

pbochynski:
@k15r thanks for the explanations

tests/fast-integration/eventing-test/utils.js
```diff
@@ -121,11 +130,26 @@ async function getStreamConfigForJetStream() {
...
+    streamCreationTime: streamCreationTime,
+  };
+}
+
 function skipAtLeastOnceDeliveryTest() {
   return !(streamConfig['retention_policy'] === 'limits' &&
       streamConfig['consumer_deliver_policy'] === 'all');
 }
 
+function skipStreamReCreationTest(streamData) {
```
friedrichwilken, line 149: This name to me implies that this function is skipping a test. Can you please rename this in a way that implies that it will return a bool? Something like `function isStreamRecreationTestSkippable(streamData)` or, because this is more what this function does, `isStreamCreationTimeMissing`. This will also create `if`s that are nicer to read.

tests/fast-integration/eventing-test/eventing-test.js
```diff
@@ -123,6 +130,45 @@ describe('Eventing tests', function() {
...
+      before('check if stream creation timestamp is available', async function() {
+        try {
+          const cm = await getConfigMap(streamDataConfigMapName);
+          wantStreamData = cm.data;
+        } catch (err) {
+          console.log('Skipping the stream recreation check due to missing configmap!');
+          this.skip();
+        }
+        if (skipStreamReCreationTest(wantStreamData)) {
+          console.log('Skipping the stream recreation check due to missing info!');
```
friedrichwilken, line 146: Can you please add what info, the creation timestamp, is missing? Think of the poor soul who has to debug with this message.

tests/fast-integration/eventing-test/eventing-test.js
```diff
@@ -123,6 +130,45 @@ describe('Eventing tests', function() {
     });
   }
 
+  function testStreamNotReCreated() {
```
friedrichwilken, line 133: Can you add a simple line that explains how we check if the stream was recreated? Old vs current timestamp.

tests/fast-integration/utils/index.js
```diff
@@ -1451,6 +1451,106 @@ function deleteEventingBackendK8sSecret(name, namespace='default') {
...
 }
 
+/**
+ * Creates apirule for the service specified
+ * @param {string} name - name of the configmap
+ * @param {string} namespace - namespace where to create the configmap
+ * @param {string} svcName - service to expose as apirule
+ * @param {int} port - port of the service to expose as apirule
+ */
+async function createAPIRuleForService(name, namespace='default', svcName, port) {
```
friedrichwilken, line 1461: can you please rename to `...Api...` to stay in style with the existing function naming conventions?

friedrichwilken, line 1461:
```suggestion
async function createApiRuleForService(name, namespace='default', svcName, port) {
```

tests/fast-integration/eventing-test/eventing-test.js
```diff
@@ -123,6 +130,45 @@ describe('Eventing tests', function() {
...
+        gotStreamData = await getJetStreamStreamData(vsHost);
+      });
+
+      it('Compare the stream creation timestamp', async function() {
+        assert.equal(gotStreamData.streamName, wantStreamData.streamName);
+        assert.equal(gotStreamData.streamCreationTime, wantStreamData.streamCreationTime);
+      });
+
+
+      it('Delete the APIRule created', async function() {
```
friedrichwilken, line 166:
```suggestion
      it('Delete the created APIRule', async function() {
```
Or do I get this wrong?

friedrichwilken:
@pbochynski I approved. Can you too, please? :)

lindnerby (approved):

pbochynski (approved):

Issue #16142, opened 2022-11-21
Compass Runtime Agent must not fetch spec data from Compass


akgalwas:
**Description**

Compass Runtime Agent fetches spec data from Compass. This data is no longer needed as code using it has been removed.

**Expected result**

Specs should not be fetched as a part of GraphQL queries to Compass Director.

PR #16144, opened 2022-11-21
Removed fetching specs from Compass

akgalwas:
**Description**

- Remove fetching specs for APIDefinitions and EventDefinitions
- Remove fetching documents

As a result of applying this changes less data will be fetched from Compass database what will improve overall performance.

**Related issue(s)**
#16142


components/compass-runtime-agent/internal/compass/director/query.go
```diff
@@ -73,25 +73,17 @@ func bundlesData() string {
 		instanceAuthRequestInputSchema
 		apiDefinitions {%s}
 		eventDefinitions {%s}
-		documents {%s}
 		defaultInstanceAuth {%s}
```
koala7659, line 76: If we don't query for API documents here we can remove its processing code from tests.

components/compass-runtime-agent/internal/compass/director/converter_test.go
```diff
@@ -339,85 +335,36 @@ func fixInternalAPIBundleEmptySpecs(suffix string) kymaModel.APIBundle {
...
 		ID:          baseAPIId + suffix,
+		Name:        baseAPIName + suffix,
 		Description: baseAPIDesc + suffix,
-		Title:       baseDocTitle + suffix,
-		DisplayName: baseDocDisplayName + suffix,
-		Format:      kymaModel.DocumentFormatMarkdown,
-		Kind:        &kind,
-		Data:        data,
 	}
 }
```
koala7659, line 367: Please remove Documents from function below member since we are not processing it anymore

components/compass-runtime-agent/internal/compass/director/converter_test.go
```diff
@@ -463,7 +406,7 @@ func fixCompassBundleExt(suffix string) *graphql.BundleExt {
 func fixCompassBundleExtWithEmptySpecs(suffix string) *graphql.BundleExt {
 	return &graphql.BundleExt{
 		Bundle:           fixCompassBundle(suffix, nil),
-		APIDefinitions:   fixAPIDefinitionPageExtWithEmptyApiSpecs(),
+		APIDefinitions:   fixAPIDefinitionPageExt(),
 		EventDefinitions: fixEventAPIDefinitionPageExtWithEmptySpecs(),
 		Documents:        fixDocumentPageExtWithEmptyDocs(),
```
koala7659, line 411: We are not fetching Documents anymore (same thing for the function above)

components/compass-runtime-agent/internal/compass/director/converter_test.go
```diff
@@ -643,26 +561,6 @@ func fixCompassDocument(suffix string, data *graphql.CLOB) *graphql.Document {
...
-	data := graphql.CLOB(`Open API spec`)
-
-	return &graphql.APISpec{
-		Data:   &data,
-		Type:   graphql.APISpecTypeOpenAPI,
-		Format: graphql.SpecFormatJSON,
-	}
-}
-
 func fixCompassAsyncAPISpec() *graphql.EventSpec {
```
koala7659, line 564: I believe functions fixCompassAsyncAPISpec and  fixCompassDocContent should also be removed. We don't fetch this data from Compass anymore.

components/compass-runtime-agent/internal/kyma/model/types.go
```diff
@@ -57,35 +37,14 @@ type APIDefinition struct {
...
-
-// APISpec contains API spec BLOB and its type
-type APISpec struct {
-	Data   []byte
-	Type   APISpecType
-	Format SpecFormat
+	ID          string
+	Name        string
+	Description string
 }
```
koala7659, line 49: Can we delete this type? We don't query for EventAPISpec and there is no need to process it later.

components/compass-runtime-agent/internal/kyma/model/types.go
```diff
@@ -4,32 +4,13 @@ type Labels map[string]interface{}
 
 type SpecFormat string
 
-const (
-	SpecFormatYAML SpecFormat = "YAML"
-	SpecFormatJSON SpecFormat = "JSON"
-	SpecFormatXML  SpecFormat = "XML"
-)
-
 type EventAPISpecType string
```
koala7659, line 7: Both APISpecType and EventAPISpecType types are not needed anymore

koala7659 (changes requested):
Please check my comments and remove the obsolete code and datatypes

components/compass-runtime-agent/internal/kyma/model/types.go
akgalwas: Probably we could do that, but I would be rather defensive here. We don't fetch spec, but we modify Application CR to store information that the Application can react on event. Should not break anything, but let's not enhance the impact of the fix.

koala7659 (approved):
LGTM

PR #16147, opened 2022-11-22
Add missing CRD_NAME env for api-gateway cert webhook

triffer:
**Description**

- Add CRD_NAME env to the api-gateway cert webhook so that a newer version of api-gateway relying on this configuration will not fail. The CRD_NAME env is required because of https://github.com/kyma-project/api-gateway/pull/55.

**Related issue(s)**


cnvergence (approved):
LGTM

PR #16172, opened 2022-11-24
Fix retry on conflict in Serverless integration tests

moelsayed:
**Description**
Fixes a bug in conflict handling during function updates. This should make the test less flaky.

- Fix retry on conflict in Serverless integration tests
- ...
- ...

**Related issue(s)**

### DIFF
--- a/tests/function-controller/pkg/resource/resource.go
+++ b/tests/function-controller/pkg/resource/resource.go
@@ -171,13 +171,10 @@ func (r *Resource) Update(res interface{}) (*unstructured.Unstructured, error) {
 	}
 
 	var result *unstructured.Unstructured
-	err = retry.RetryOnConflict(retry.DefaultBackoff, func() error {
-		var errUpdate error
-		result, errUpdate = r.ResCli.Update(context.Background(), unstructuredObj, metav1.UpdateOptions{})
-		return errUpdate
-	}, r.log)
+
+	result, err = r.ResCli.Update(context.Background(), unstructuredObj, metav1.UpdateOptions{})
 	if err != nil {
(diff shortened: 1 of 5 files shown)


tests/function-controller/pkg/resource/resource.go
```diff
@@ -171,13 +171,10 @@ func (r *Resource) Update(res interface{}) (*unstructured.Unstructured, error) {
...
-	err = retry.RetryOnConflict(retry.DefaultBackoff, func() error {
-		var errUpdate error
-		result, errUpdate = r.ResCli.Update(context.Background(), unstructuredObj, metav1.UpdateOptions{})
-		return errUpdate
-	}, r.log)
+
+	result, err = r.ResCli.Update(context.Background(), unstructuredObj, metav1.UpdateOptions{})
 	if err != nil {
-		return nil, errors.Wrapf(err, "while updating resource %s '%s'", r.kind, unstructuredObj.GetName())
+		return nil, err // upstream caller RetryOnConflict doesn't work with wrapped errors
```
dbadura, line 177: Can you give us the reason why?

moelsayed: I didn't read it's actual code, but I think they don't bother unwrapping the error. The recommendation is documented there: https://github.com/kubernetes/client-go/blob/9927afa2880713c4332723b7f0865adee5e63a7b/util/retry/util.go#L89-L93

dbadura: Please add the link to those comments, it would make easier to understand why it behaves like that after some time.

dbadura (approved):

PR #16177, opened 2022-11-24
[Telemetry Operator] Unify error handling in controllers

skhalash:
**Description**

- Always return an error from the `Reconcile` function. For now, we will retry on each error, so no retry timeout has to be explicitly set
- Status update is deferred to make sure it's always executed (even if in case of an early return)
- Trigger reconciliation for referenced secret creation, not only updates

**Related issue(s)**


components/telemetry-operator/controller/logpipeline/setup.go
```diff
@@ -17,32 +17,42 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/source"
 )
 
+var (
+	createOrUpdate = predicate.Funcs{
```
lindnerby, line 21: Let's share these as internal pkg between all three controllers

components/telemetry-operator/controller/logpipeline/controller.go
```diff
@@ -76,49 +75,56 @@ func NewReconciler(client client.Client, config Config, prober DaemonSetProber,
...
+				err = fmt.Errorf("failed while updating status: %v: %v", statusErr, err)
+			} else {
+				err = fmt.Errorf("failed to update status: %v", statusErr)
+			}
 		}
 	}()
 
-	if err := r.ensureFinalizers(ctx, &pipeline); err != nil {
-		return ctrl.Result{Requeue: controller.ShouldRetryOn(err)}, nil
+	if err = r.ensureFinalizers(ctx, pipeline); err != nil {
```
lindnerby, line 106: Can you align `ensureFinalizers` with the logparser version of this? (method vs func)

lindnerby (changes requested):

components/telemetry-operator/controller/logpipeline/controller.go
skhalash: Sure

lindnerby (approved):

components/telemetry-operator/controller/logpipeline/setup.go
skhalash: Sure

Issue #16201, opened 2022-11-28
Simple deployment status for TracePipeline


a-thaler:
**Description**
A simple status is required reflecting the state of the pipeline by reflecting the status of the related otel-collector deployment.

**Criteria**
- We have new states Pending, Running
- The status reflects 1:1 the status of the deployment, similar to the LogPipeline

PR #16221, opened 2022-11-30
TracePipeline simple status

hisarbalik:
**Description**

- Simple status attributed added
- Status currently support only `Pending` and `Running`
- Status depend on OpenTelemetry Deployment instance

**Related issue(s)**
#16201


components/telemetry-operator/controller/tracepipeline/setup.go
```diff
@@ -29,6 +29,11 @@ func (r *Reconciler) SetupWithManager(mgr ctrl.Manager) error {
 			&source.Kind{Type: &corev1.Secret{}},
 			handler.EnqueueRequestsFromMapFunc(r.mapSecret),
 			builder.WithPredicates(onlyUpdate()),
+		).
+		Watches(
```
chrkl, line 33: This should be unnecessary. The owner reference should already trigger a reconciliation if the Deployment changes.

hisarbalik: this code slices removed.

components/telemetry-operator/apis/telemetry/v1alpha1/tracepipeline_types.go
```diff
@@ -56,8 +56,73 @@ func (b *BasicAuthOptions) IsDefined() bool {
...
 
+type TracePipelineConditionType string
+
+// These are the valid statuses of LogPipeline.
+const (
+	TracePipelinePending TracePipelineConditionType = "Pending"
+	TracePipelineRunning TracePipelineConditionType = "Running"
+)
+
+const (
```
chrkl, line 67: In the LogPipeline status we also have a condition for a missing secret that is referenced. We should add this here as well since that might be an important information for the user. You can find the implementation for the LogPipeline [here](https://github.com/kyma-project/kyma/blob/main/components/telemetry-operator/controller/logpipeline/status.go#L59).

hisarbalik: secret reference check added

Issue #11300, opened 2021-05-18
Prometheus annotation support


a-thaler:
**Description**
The prometheus annotations are quite popular nowadays and are an alternative to the ServiceMonitor definition style. So instead of defining a concrete scrape configuration via a ServiceMonitor, there is a scrape config based on the prometheus service discovery which dynamically selects the endpoints to scrape.

That concept is interesting in regards to providing the feature out-of-the-box for kyma application developers. Furthermore it might be an interesting substitute in case the prometheus-operator is not a mandatory element of the kyma stack anymore (prometheus might be replaced with the openTelemetryCollector)

Goal: Figure out if additional support of prometheus metrics annotationsis feasible and valuable. You can deploy an application with annotations and metrics are getting scraped. Furthermore see if the serviceMonitors for the system components can be replaced by them so that the base kyma setup is not based on ServiceMonitors at all.

For that:
- Understand the service discovery mechanism of prometheus and come up with a scrqape config supporting annotations
- Check istio mergeMetrics option, decide if it must be switched on.. can istio scraping be separated from application metric scraping? The istio-prometheus might scrape the istio ports directly while the main prometheus is based on annotations? 
- Check for all podmonitors we have, can they be substituted
- What is with advanced ServiceMonitor options like metricRelabelling? Can we live without it, what is the replacement?
- Do we have components with annotations in place which might collide with existing scrape approach?
- Check prometheus stability

Major question to answer: Should we support annotations already additionally? Can it be a full replacement long-term thinking of not having the prometheus-operator in the cluster?

**Reasons**
Prometheus-operator might not be used in future, popular charts comes with annotations only

**Attachments**

a-thaler:
The feature should be the main scenario for the future metrics support as described in https://github.com/kyma-project/kyma/issues/13079.
However, we should not add it as a feature to the current stack as it will increase the maintenance effort much more. Instead we should start learning with the approach and already start decoupling the system components from the monitoring chart, so applying the feature for the internal components only.

I created a PR which is doing exactly that. It introduced a new scrape job not based on serviceMonitors which discovers service endpoints by annotation. The job discovers only the system namespaces. Then I replaced the servicemonitors by annotations. See https://github.com/kyma-project/kyma/pull/16247

The only downside I found is that it is not possible to keep the container and endpoint label on the metric, this is not relevant for the existing metrics.

skhalash:
[...]

PR #16247, opened 2022-12-02
Replace ServiceMonitors in kyma-system by annotation based scraping

a-thaler:
**Description**
Requires release notes as migration script is needed!

The main motivation is to decouple the components from the monitoring component. It also fits the future strategy of not scraping kyma components via ServiceMonitors anymore.

- removed `servicemonitors` from all components except `monitoring`
- removed `prometheusrule` -> no Prometheus Operator CRDs are used anymore by kyma components (except monitoring)
- annotated all kyma services which were scraped by a `servicemonitor`
- removed annotations on pod level when needed. As the reconciler will not remove them on upgrade, I disabled the scrape annotation first
- defined a scrape job hardcoded in Prometheus config to discover the service endpoints in `kyma-system` namespace by annotation
- I compared the targets before and after. All are included, additionally, there is a target for the `ory-postgresql`
- fixed `Eventing / NATS Delivery` dashboard, that are using many-to-one vector matching (group_left). Enabling scraping via annotations duplicates the affected timeseries until the service monitors are cleaned up. It means that for some period of time, the abovementioned dashboard remains broken. To fix it, a `max` function is used to duplicate the timeseries.

**Related issue(s)**

https://github.com/kyma-project/kyma/issues/11300


kwiatekus (dismissed):

cnvergence (approved):

skhalash (approved):

docs/migration-guide.md
```diff
@@ -0,0 +1,5 @@
+---
+title: Migration Guide
+---
+
+As preparation for the upcoming modularization and having a reduced set of dependencies on a module, Kyma switched to a annotation based scraping of metrics for system components. With that the ServiceMonitors of the components needs to get cleaned up. When you upgrade from Kyma 2.9 to 2.10, either run the script [cleanup.sh](./assets/cleanup.sh) or run the commands from the script manually.
```
grego952, line 5:
```suggestion
In preparation for the upcoming modularization and having a reduced set of dependencies on a module, Kyma switched to an annotation-based scraping of metrics for system components. With that, the ServiceMonitors of the components need to get cleaned up. When you upgrade from Kyma 2.9 to 2.10, either run the script [cleanup.sh](./assets/cleanup.sh) or run the commands from the script manually.
```

grego952 (approved):

kwiatekus (dismissed):

grego952 (dismissed):

skhalash (dismissed):

kwiatekus (approved):

triffer (dismissed):

skhalash:
Failing eventing upgrade tests uncover the following problem: after a cluster upgrade, the service monitors remain untouched (the reconciler cannot delete resources). That's why certain targets become duplicated, which breaks some PromQL queries. That's the reason why the tests fail. But it also breaks some Grafana dashboards (e.g. `Published Event Types` with a `Subscription and Published Event Types without Subscription` panels in the `Kyma / Eventing / NATS Delivery dashboard`).  Cleaning up the service monitors fixes the problem. However, it will not fix the data in TSDB, so the dashboards remain broken if the selected time window overlaps with the pre-upgrade time.

skhalash:
Will split the PR into isolated small PRs per component

Issue #16208, opened 2022-11-29
Implement orphaned image cleanup for internal docker registry


kwiatekus:
**Description**

Implement a periodically running cleanup job that identifies orphaned function images and deletes them.

 - the clean up job should be part of internal docker registry manifests - deployed/configurable only in case internal docker registry is enabled
 - It should be possible to disable/enable cleanup job (via values.yaml / via Serverless CRD)

**Reason**
The current implementation for the Internal Registry deployed with Kyma Functions doesn't include any garbage collection logic. In dynamic environments with several Functions with multiple versions, if the Internal Registry is used with a cluster PVC storage backend, the volume disk space can fill up very quickly, resulting in build failures for new Function versions.

**Attachments**
[proposal](https://github.com/kyma-project/kyma/blob/main/components/function-controller/design/internal_registry_garbage_collection.md#internal-registry-garbage-collection)

kwiatekus:
IMO currennt state of the garbage collector at main branch removes too much. 

I changed the config of the garbage-collect container and the schedule of the cron job to run every 5 minutes.

I saw in the logs  of the garbage-collect container that some of the blobs got removed after a function was updated.

However this function cannot be replicated anymore.. as kubernetes cannot pull the image


![Image](https://user-images.githubusercontent.com/6783567/209810158-203a5305-f89e-40f3-b75d-aa03db71a0c8.png)

Moreover, I see that the gc cronjob schedulej jobs and they fail in such scenario as they cannot remove something ('unknown manifest'). The cron job tries again and again until the job reaches the max back off retries.

```
{"level":"info","ts":1672229447.1506612,"logger":"internal registry gc","msg":"reading configuration"}
I1228 12:10:48.202191 1 request.go:665] Waited for 1.02762516s due to client-side throttling, not priority and fairness, request: GET:https://api.kk-gc.wookiee.internal.canary.k8s.ondemand.com:443/apis/autoscaling/v2?timeout=32s
{"level":"info","ts":1672229448.4251747,"logger":"internal registry gc","msg":"removing function images"}
{"level":"info","ts":1672229449.720219,"logger":"internal registry gc","msg":"removing cached layers"}
{"level":"error","ts":1672229450.241133,"logger":"internal registry gc","msg":"while getting registry cached layers","error":"manifest unknown: manifest unknown","stacktrace":"runtime.main\n\t/usr/local/go/src/runtime/proc.go:250"} 
```

kwiatekus:
[...]

PR #16262, opened 2022-12-06
Implement Serverless internal registry image garbage collector

moelsayed:
**Description**

- Implement the function-registry-gc tool
- Add a cronjob manifest and rbac configuration needed to run the tool.
- Add sidecar container to the registry deployment to run the registry `garbage-collect` command

**Related issue(s)**

[16208](https://github.com/kyma-project/kyma/issues/16208)


components/function-controller/internal/registry/registry_test.go
```diff
@@ -0,0 +1,79 @@
...
+	tests := []struct {
+		name    string
+		opts    *RegistryClientOptions
+		want    RegistryClient
+		wantErr bool
+	}{
+		{
+			name: "valid options",
+			opts: &RegistryClientOptions{
+				Username: "username",
```
anoipm, line 22: You very often use different values for different tests (e.g. "test1", "test2"). In this case you always use "username" and "password". Hypothetically the function being tested might always return "username" and "password" and tests will be green :)

components/function-controller/internal/registry/repository.go
```diff
@@ -0,0 +1,79 @@
...
+	GetImageTag(tag string) (*Tag, error)
+	DeleteImageTag(tagDigest digest.Digest) error
+}
+
+type repositoryClient struct {
+	ctx context.Context
+
+	namedImage reference.Named
+
+	tagSservice     distribution.TagService
```
anoipm, line 28: Is the double "S" (tag**Ss**ervice) in this name intentional?

components/function-controller/cmd/registry-gc/main.go
```diff
@@ -0,0 +1,85 @@
...
+	ctrl "sigs.k8s.io/controller-runtime"
+	ctrlzap "sigs.k8s.io/controller-runtime/pkg/log/zap"
+)
+
+type config struct {
+	Namespace               string `envconfig:"default=kyma-system"`
+	RegistryConfigSecreName string `envconfig:"default=serverless-registry-config-default"`
+}
+
+func main() {
```
anoipm, line 20: IMO this function is hard to read because is too long. It has at least two functionalities: collecting image names from deployments, removing images with known name and unknown tag. What do you think about extracting these parts to separate functions?

components/function-controller/cmd/registry-gc/main.go
```diff
@@ -0,0 +1,85 @@
...
+		if err != nil {
+			mainLog.Error(err, "while creating repository client")
+			os.Exit(1)
+		}
+		registryTags, err := repoCli.ListTags()
+		if err != nil {
+			mainLog.Error(err, "while getting image tags")
+			os.Exit(1)
+		}
+		for _, tagStr := range registryTags {
```
anoipm, line 68: `for` inside `for` (with `if` inside) - it looks like good place to extract to separated function.

components/function-controller/cmd/registry-gc/main.go
```diff
@@ -0,0 +1,85 @@
...
+	mainLog := ctrlzap.New().WithName("internal registry gc")
+	mainLog.Info("reading configuration")
+	cfg := &config{}
+	if err := envconfig.Init(cfg); err != nil {
+		mainLog.Error(err, "while reading env variables")
+		os.Exit(1)
+
+	}
+	restConfig := ctrl.GetConfigOrDie()
+	registryConfig := registry.ReadRegistryConfigSecretOrDie(restConfig, cfg.RegistryConfigSecreName, cfg.Namespace)
```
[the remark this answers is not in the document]
moelsayed: Yes, just grouping all the config together, and dying as quickly as possible if something goes wrong, as it doesn't make sense to do anything if we don't have these configurations.

components/function-controller/internal/registry/repository.go
moelsayed: typo :)

components/function-controller/cmd/registry-gc/main.go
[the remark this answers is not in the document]
anoipm: ok

resources/serverless/charts/docker-registry/templates/job.yaml
```diff
@@ -0,0 +1,37 @@
...
+    spec:
+      template:
+        metadata:
+          annotations:
+            sidecar.istio.io/inject: "false"
+        spec:
+          serviceAccountName: {{ template "docker-registry.name" . }}
+          restartPolicy: Never
+          containers:
+          - name: function-image-gc
```
dbadura, line 18: I have problem with this name. It doesn't clearly says what this job do. 
I though about name like `function-image-untagger`.

resources/serverless/charts/docker-registry/templates/job.yaml
```diff
@@ -0,0 +1,37 @@
+{{- if eq .Values.storage "filesystem" }}
+apiVersion: batch/v1
+kind: CronJob
+metadata:
+  name: {{ template "docker-registry.name" . }}-image-gc
+spec:
+  schedule: "42 * * * *"
```
dbadura, line 7: Why `42`?

resources/serverless/charts/docker-registry/values.yaml
```diff
@@ -56,9 +56,14 @@ persistence:
   accessMode: 'ReadWriteOnce'
   enabled: true
   size: 20Gi
+  deleteEnabled: true
```
dbadura, line 59: Why do we use two variables to control the deletion?
Can we enable `garbageCollection` and disable `deleteEnabled`?
I think that we should use `garbageCollection` only to control this behavior.

dbadura (changes requested):

resources/serverless/charts/docker-registry/templates/deployment.yaml
```diff
@@ -207,6 +206,33 @@ spec:
 {{- with .Values.extraVolumeMounts }}
             {{- toYaml . | nindent 12 }}
 {{- end }}
+
+{{- if ( and ( .Values.garbageCollection.enabled )  ( eq .Values.storage "filesystem" ) )}}
+        - name: garbage-collect
```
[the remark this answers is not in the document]
moelsayed: I did consider it. But it seemed more complicated to implement. I prefer the current loosely-coupled implementation instead. This way I don't have to worry about where to run the job, scheduling and capacity issues..etc.

However,  we are planing to add a controller to this tool to watch for deleted functions and remove their images. That would be a continuously running binary that make sense to add as a container in the same pod as the registry.

resources/serverless/charts/docker-registry/values.yaml
moelsayed: Fixed. Thank you for catching this. `deleteEnabled` was referenced in the deployment manifest, so I just added it and forgot to rename it later.

components/function-controller/deploy/registry-gc/Dockerfile
```diff
@@ -0,0 +1,33 @@
...
+
+WORKDIR ${BASE_APP_DIR}
+
+#
+# copy files allowed in .dockerignore
+#
+COPY . ${BASE_APP_DIR}/
+
+RUN go build -ldflags "-s -w" -a -o registry-gc cmd/registry-gc/main.go \
+    && mkdir /app \
```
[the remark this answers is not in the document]
moelsayed: Technically, that's true. However, you would need to reference the full path of the file, or redefine the `${BASE_APP_DIR}` env in the second container block as it's not accessible from the build container.

I think it's easier to read this way.

resources/serverless/charts/docker-registry/templates/job.yaml
moelsayed: It's the answer to life, the universe and everything!

Also, I tend to cron things away from common time-slices to avoid load spikes.

components/function-controller/deploy/registry-gc/Dockerfile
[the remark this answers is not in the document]
dbadura: Would use of `ARG` help with double `ENV`?

resources/serverless/charts/docker-registry/templates/job.yaml
dbadura: Could you add comment above the `schedule`?
It's clever solution.

resources/serverless/charts/docker-registry/templates/job.yaml
```diff
@@ -0,0 +1,37 @@
+{{- if eq .Values.storage "filesystem" }}
```
dbadura, line 1: Can we name this file `cronjob`?

components/function-controller/deploy/registry-gc/Dockerfile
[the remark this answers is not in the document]
moelsayed: Unfortunately, no. And using it would be abusive TBH :smile:. I will revert to the original configuration.

anoipm (approved):

Issue #16200, opened 2022-11-28
Deploy and update simple fluentbit daemonset by telemetry-operator


a-thaler:
**Description**
As part of https://github.com/kyma-project/kyma/issues/15955 the telemetry-operator should deploy and update a fluentbit installation if a LogPipeline is present.

**Criteria**
- Have a feature flag to enable the new feature in the operator and disable the helm sub-chart deployment
- If there is no LogPipeline defined, then there is no fluenbit instance running (only initial, deletion will be implemented later)
- If there is a LogPipeline, a fluenbit instance gets created and configured with all resources as deployed before via the helm chart
- Ignore the ServiceMonitor

skhalash:
The feature is hidden behind a feature flag and is not enabled by default. It should be enabled when daemonset deletion is implemented. For now, I added a job that tests the operator with the. feature flag enabled https://status.build.kyma-project.io/?job=kyma-integration-k3d-telemetry-alpha.

PR #16263, opened 2022-12-06
[Telemetry Operator] - Operator managed Fluent Bit

lindnerby:
**Description**

- Make parts of the Fluent Bit charts switchable by global value `operatorManagedFluentBit`
- If enabled Telemetry Operator itself will apply Fluent Bit resources of DaemonSet, ConfigMaps and Service 
- LogPipeline reconciliation is managing the existence of these resources 

**Related issue(s)**
* https://github.com/kyma-project/kyma/issues/16200


resources/telemetry/charts/fluent-bit/templates/daemonset.yaml
```diff
@@ -1,3 +1,5 @@
+{{- if not .Values.global.operatorManagedFluentBit }}
+{{- if .Values.networkPolicy.enabled }}
```
skhalash, line 2: I don't think it should depend on the `networkPolicy.enabled`

lindnerby: Don't know how this sneaked in 😆 Removed

components/telemetry-operator/controller/logpipeline/controller.go
```diff
@@ -107,6 +109,16 @@ func (r *Reconciler) doReconcile(ctx context.Context, pipeline *telemetryv1alpha
 		return err
 	}
 
+	if r.config.ManageFluentBit && pipeline.DeletionTimestamp.IsZero() {
+		config := resourceConfig{
+			BaseName:  fmt.Sprintf("telemetry-%s", r.config.DaemonSet.Name),
```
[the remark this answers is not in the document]
lindnerby: Or do we just take one of the `fluent-bit-ds-name` or `fluent-bit-cm-name` since they have the same default values. Feels weird to introduce a third name arg that has the same default

chrkl (approved):

skhalash (approved):

skhalash (approved):

Issue #15685, opened 2022-09-30
Implement the protocol settings flow with v1alpha2 CRD for EventMesh backend


mfaizanse:
In v1aplha1 subscription CRD, we have a `spec.ProtocolSettings` field which we can use to override the protocol setting for event mesh. The provided `spec.ProtocolSettings` is then read in `GetInternalView4Ev2()` ([here](https://github.com/kyma-project/kyma/blob/efd0d663d2b8a3e761b760fcf173e9fef73c7a4b/components/eventing-controller/pkg/backend/utils/utils.go#L67)) method while creating the corresponding subjects on event mesh server.

Now, the new v1alpha2 subscription CRD has a new config section (i.e. `spec.config`) which will be providing the optional protocol settings for EventMesh backend as `key[interface]`. We need to read these protocol settings from `spec.config` and convert it to protocol settings struct and use it while creating subjects on eventMesh server. 

Currently, the backend for eventMesh is implemented but the implementation of using this custom protocol settings is missing in [ConvertKymaSubToEventMeshSub()](https://github.com/kyma-project/kyma/blob/efd0d663d2b8a3e761b760fcf173e9fef73c7a4b/components/eventing-controller/pkg/backend/utils/eventmesh_utils.go#L126), and its using the default values.

**Tasks:**
- [x] Implement the new flow for protocol settings for EventMesh backend.
    - [x] Update the [`ConvertKymaSubToEventMeshSub`](https://github.com/kyma-project/kyma/blob/eb254926dfb8802d2f7b769925f6eb2e0baf1095/components/eventing-controller/pkg/backend/utils/eventmesh_utils.go#L113) to use protocol settings from new CRD.
    - [x] If any of the protocol settings field is not provided by user, then use a default value for it.
    - [x] Decide if the structs for protocol settings should reside in `v1alpha2` package [here](https://github.com/kyma-project/kyma/blob/eb254926dfb8802d2f7b769925f6eb2e0baf1095/components/eventing-controller/api/v1alpha2/eventmesh_types.go).
[...]

PR #16264, opened 2022-12-06
Protocol and auth settings for EventMesh in v1alpha2 subscritpion

mfaizanse:
**Description**

- Implemented the flow for overriding protocol and webhook auth settings for EventMesh backend in Subscription v1alpha2 CRD.

**Related issue(s)**

- https://github.com/kyma-project/kyma/issues/15685


components/eventing-controller/pkg/backend/utils/eventmesh_types.go
```diff
@@ -1,4 +1,4 @@
-package v1alpha2
+package utils
```
marcobebway, line 1: The contents of this file is an exact copy of https://github.com/kyma-project/kyma/blob/main/components/eventing-controller/api/v1alpha1/subscription_types.go#L14-L53. Is there a reason for the duplication?

mfaizanse: The structs defined here are related to EventMesh, so I moved them from `api` to `eventMesh_types`. 

Since we also need them in definition of subscription `v1alpha1`, they are also defined there because we cannot import the utils pkg in `v1alpha1` (circular dependency). 

So the idea is to keep the old code in `v1alpha1` and when we remove the code for `v1alpha1` later, the duplication will be removed.

marcobebway: Let's add a todo above the duplicate structs to remove then when possible.

mfaizanse: added

marcobebway: Please add the todos here as well `components/eventing-controller/pkg/backend/utils/eventmesh_types.go`.

components/eventing-controller/pkg/backend/utils/eventmesh_utils.go
```diff
@@ -156,29 +142,80 @@ func ConvertKymaSubToEventMeshSub(subscription *eventingv1alpha2.Subscription, t
...
+			return nil, fmt.Errorf("invalid GrantType: %v, required: %v", grantType,
+				string(types.GrantTypeClientCredentials))
+		}
+		auth.GrantType = types.GrantTypeClientCredentials
+	}
+
+	// check if auth was provided in subscription CR
+	if auth.Type != "" {
+		return auth, nil
+	}
```
marcobebway, line 215: I think it is better to do the check in the beginning of the function instead of the end:

```golang
if auth.Type == "" {
    return defaultWebhookAuth, nil
}
```

components/eventing-controller/pkg/backend/utils/eventmesh_utils.go
```diff
@@ -156,29 +142,80 @@ func ConvertKymaSubToEventMeshSub(subscription *eventingv1alpha2.Subscription, t
...
+		auth.ClientSecret = clientSecret
+	}
+
+	if grantType, ok := subscription.Spec.Config[eventingv1alpha2.WebhookAuthGrantType]; ok {
+		if grantType != string(types.GrantTypeClientCredentials) {
+			return nil, fmt.Errorf("invalid GrantType: %v, required: %v", grantType,
+				string(types.GrantTypeClientCredentials))
+		}
+		auth.GrantType = types.GrantTypeClientCredentials
+	}
```
marcobebway, line 210: Does it make sense to move this check in the beginning of the function so it can fail fast?

components/eventing-controller/pkg/backend/utils/eventmesh_utils_test.go
```diff
@@ -61,55 +62,48 @@ func TestConvertKymaSubToEventMeshSub(t *testing.T) {
...
-	//	gotBEBSubscription, err := ConvertKymaSubToEventMeshSub(
-	//		subscription, eventTypeInfos, apiRule, defaultWebhookAuth,
-	//		defaultProtocolSettings, defaultNamespace, defaultNameMapper,
-	//	)
-	//
-	//	// when
-	//	g.Expect(err).To(BeNil())
-	//	g.Expect(*expectedBEBSubscription).To(Equal(*gotBEBSubscription))
-	//})
+	t.Run("subscription with protocol settings and webhook auth", func(t *testing.T) {
```
marcobebway, line 65: (nit) There are multiple `t.Run` here, does it make sense to use table tests here?

components/eventing-controller/pkg/backend/utils/eventmesh_utils.go
```diff
@@ -112,7 +113,7 @@ func getEventMeshEvents(typeInfos []EventTypeInfo, typeMatching eventingv1alpha2
 
 func ConvertKymaSubToEventMeshSub(subscription *eventingv1alpha2.Subscription, typeInfos []EventTypeInfo,
```
marcobebway, line 114: The `ConvertKymaSubToEventMeshSub` sometimes returns an error, are there unit-tests to cover those scenarios?

components/eventing-controller/pkg/backend/utils/eventmesh_utils.go
mfaizanse: updated

components/eventing-controller/pkg/backend/utils/eventmesh_utils_test.go
mfaizanse: changed it to table tests

components/eventing-controller/pkg/backend/utils/eventmesh_utils.go
marcobebway: As you like, but the current implementation already knows the answer to line 213 `if auth.Type != ""` when the line 189 `auth.Type = types.AuthTypeClientCredentials` is evaluated, then does some checks/computations for the (TokenURL, ClientID and ClientSecret) then if it might throw away those checks/computations if `auth.Type == "" is true`.

mfaizanse: valid point, I updated

marcobebway (approved):

PR #16278, opened 2022-12-07
Change the sequence in the execution of the eventing tests 

VladislavPaskar:
**Description**

This pr changes the sequence in the execution of the eventing tests in order that the check of function reachability is being executed before the in-cluster eventing test. The reason is that the function reachability is the requisite for the in-cluster eventing. The same sequence is also maintained in the eventing fast integration tests, see [here](https://github.com/kyma-project/kyma/blob/main/tests/fast-integration/eventing-test/eventing-test.js#L102-L109).


raypinto (approved):
Nice find!

jeremyharisch (approved):

PR #16295, opened 2022-12-08
Remove simple fn cfg from integration tests

dbadura:
**Description**

- Remove unnecesary config from test code

**Related issue(s)**


tests/function-controller/pkg/function/function.go
```diff
@@ -44,12 +45,19 @@ func NewFunction(name string, c shared.Container) *Function {
 		},
 	}
 
+	//TODO: implement kubectl proxy
```
dbadura, line 48: I will be done in another PR

dbadura: https://github.com/kyma-project/kyma/pull/16296

PR #16298, opened 2022-12-08
improve epp metrics

k15r:
**Description**

- rename epp metrics to follow the pattern: eventting_epp-<backend>-<metric>
- update dashboards to use renamed metrics
- introduce metric for total requests to epp

**Related issue(s)**

### DIFF
--- a/components/event-publisher-proxy/pkg/handler/handler.go
+++ b/components/event-publisher-proxy/pkg/handler/handler.go
@@ -78,6 +78,7 @@ func NewHandler(receiver *receiver.HTTPMessageReceiver, sender sender.GenericSen
 // setupMux configures the request router for all required endpoints.
 func (h *Handler) setupMux() {
 	router := mux.NewRouter()
+	router.Use(h.collector.MetricsMiddleware(h.Sender.URL()))
 	router.HandleFunc(PublishEndpoint, h.maxBytes(h.publishCloudEvents)).Methods(http.MethodPost)
 	router.HandleFunc(LegacyEndpointPattern, h.maxBytes(h.publishLegacyEventsAsCE)).Methods(http.MethodPost)
 	if h.Options.EnableNewCRDVersion {
@@ -208,12 +209,12 @@ func (h *Handler) sendEventAndRecordMetrics(ctx context.Context, event *cev2even
 	result, err := h.Sender.Send(ctx, event)
 	duration := time.Since(start)
 	if err != nil {
-		h.collector.RecordError()
+		h.collector.RecordBackendError()
 		return nil, err
 	}
 	h.collector.RecordEventType(event.Type(), event.Source(), result.HTTPStatus())
-	h.collector.RecordLatency(duration, result.HTTPStatus(), host)
-	h.collector.RecordRequests(result.HTTPStatus(), host)
+	h.collector.RecordBackendLatency(duration, result.HTTPStatus(), host)
+	h.collector.RecordBackendRequests(result.HTTPStatus(), host)
 	return result, nil
 }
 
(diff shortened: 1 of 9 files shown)


components/event-publisher-proxy/pkg/metrics/collector.go
```diff
@@ -2,32 +2,53 @@ package metrics
...
+	BackendLatencyKey = "eventing_epp_backend_duration_milliseconds"
+	// backendLatencyHelp help text for the backendLatencyHelp metric
+	backendLatencyHelp = "The duration of sending events to the messaging server in milliseconds"
+
+	// BackendRequestsKey name of the eventRequests metric
+	BackendRequestsKey = "eventing_epp_backend_requests_total"
+	// backendRequestsHelp help text for event backendRequests metric
+	backendRequestsHelp = "The total number of backend requests"
+
+	// durationKey name of the durationMetric metric
```
friedrichwilken, line 31:
```suggestion
	// durationKey name of the duration metric
```

raypinto:
Not a review but a heads up: We need to update the docs: https://kyma-project.io/docs/kyma/latest/04-operation-guides/operations/evnt-02-eventing-metrics/

components/event-publisher-proxy/pkg/metrics/collector.go
```diff
@@ -123,6 +169,20 @@ func (c *Collector) RecordEventType(eventType, eventSource string, statusCode in
 }
 
 // RecordRequests records an eventRequests metric
-func (c *Collector) RecordRequests(statusCode int, destSvc string) {
-	c.requests.WithLabelValues(fmt.Sprint(statusCode), destSvc).Inc()
+func (c *Collector) RecordBackendRequests(statusCode int, destSvc string) {
+	c.backendRequests.WithLabelValues(fmt.Sprint(statusCode), destSvc).Inc()
+}
+
+func (c *Collector) MetricsMiddleware(destService string) mux.MiddlewareFunc {
```
nachtmaar, line 176: nit: Can you add a comment to the method ? Describe what it does ?

nachtmaar (commented):
LGTM. Only few nits. Found also one text occurrence of the old metric name:

```
pkg/metrics/metricstest/metricstest.go
15:// EnsureMetricErrors ensures metric eventing_epp_errors_total exists.
```

franpog859 (commented):
Hmm, I think something has gone wrong while resolving some conflicts. I see some changes that shouldn't be needed here, like changes in Dockerfile from for example Compass Runtime Agent. Do you really need them? And therefore, do you need my review here, @k15r ?

k15r:
> Hmm, I think something has gone wrong while resolving some conflicts. I see some changes that shouldn't be needed here, like changes in Dockerfile from for example Compass Runtime Agent. Do you really need them? And therefore, do you need my review here, @k15r ?

no clue what happened here... i rebased the pr and now everything looks normal again

nachtmaar (approved):

components/event-publisher-proxy/pkg/handler/handler.go
```diff
@@ -208,12 +209,12 @@ func (h *Handler) sendEventAndRecordMetrics(ctx context.Context, event *cev2even
 	result, err := h.Sender.Send(ctx, event)
 	duration := time.Since(start)
 	if err != nil {
-		h.collector.RecordError()
+		h.collector.RecordBackendError()
```
vvxxvvxx, line 212: Hi Korbinian, can we collect both `RecordBackendLatency` and `RecordBackendRequests` here and assign an http response code (5xx or 4xx) to them? as we also need the latency for the request with the error response code. and for the request number of `RecordBackendRequests`, it should also count the requests number with error response code.

Issue #15837, opened 2022-10-18
Proper base setup for otel-collector


a-thaler:
**Description**
The otel-collector deployment for trace collection requires a solid base setup. Fo that there are some variables to consider:
- For a specific amount of time, the traces data should be able to be buffered when a downtime of the connected backend happens.
- The trace data (spans) have an average size which can vary and may have extreme sizes
- The amount of CPU and memory is limited
- The input rate can vary over time

Assuming a fixed maximum backend downtime, an average data size and maximum resource limits, the goal is to find the maximum input rate an instance can handle. Here, usually you would start scaling out the instances for being able to handle a higher rate. As in the beginning there will be no auto-scaling realized, the limit needs to be safe-guarded by tools like the memoryLimiter, so that the instance is functional at any time.

**Criteria**
- Derive assumptions and by that fix the numbers for these criterias:
  - maximum downtime of the configured backend (check recommended defaults and SLAs of offerings)
  - resource limits (depends a bit on expected load volume, probably something around 1GB-2GB memory)
  - average span size (check average spans of the otel-demo app)
- Out of the assumptions a limit for the input rate is derived by load tests
- The limit gets safe-guarded by the memoryLimiter to refuse input on a higher rate
- Assumptions/Limitations are documented
- All derived settings are incorporated into the sources in a hardcoded way for now
- The instance is never crashing out of resource problems
- A batch processor got tested and decided if it should be part of the initial base setup

**Reasons**

**Attachments**
Some first investigations were done and documented here: https://github.com/kyma-project/community/tree/main/concepts/observability-strategy/configurable-tracing/pocs

hisarbalik:
Documentation PR https://github.tools.sap/huskies/knowledge-hub/pull/39

PR #16304, opened 2022-12-09
Trace collector default resource request values update

hisarbalik:
**Description**

- Default values of tracing collector resource request updated

**Related issue(s)**
https://github.com/kyma-project/kyma/issues/15837


resources/telemetry/values.yaml
```diff
@@ -27,4 +27,3 @@ global:
   log:
     format: "json"
     level: "info"
-  operatorManagedFluentBit: false
```
skhalash, line 30: Maybe let's keep this flag?

a-thaler: Might be useful :P

skhalash (approved):

PR #16343, opened 2022-12-12
fix conversion bug for protocol settings

raypinto:
**Description**

- Initialize ProtocolSettings and WebhookAuth in all cases

**Related issue(s)**


components/eventing-controller/api/v1alpha1/subscription_conversion.go
```diff
@@ -136,56 +136,79 @@ func (src *Subscription) setV2ProtocolFields(dst *v1alpha2.Subscription) {
 		}
 		// webhookAuth fields
 		if src.Spec.ProtocolSettings.WebhookAuth != nil {
-			dst.Spec.Config[v1alpha2.WebhookAuthType] = src.Spec.ProtocolSettings.WebhookAuth.Type
+			if src.Spec.ProtocolSettings.WebhookAuth.Type != "" {
```
mfaizanse, line 139: Any particular reason, why we are only checking for empty string for `Type`? Why not also for other fields e.g. `GrantType`, `ClientID` etc.?

components/eventing-controller/api/v1alpha1/subscription_conversion.go
```diff
@@ -136,56 +136,79 @@ func (src *Subscription) setV2ProtocolFields(dst *v1alpha2.Subscription) {
...
 		src.Spec.Protocol = protocol
 	}
 
 	if currentMode, ok := dst.Spec.Config[v1alpha2.ProtocolSettingsContentMode]; ok {
-		src.Spec.ProtocolSettings = &ProtocolSettings{}
+		src.initializeProtocolSettingsIfNil()
 		src.Spec.ProtocolSettings.ContentMode = &currentMode
 	}
 	if qos, ok := dst.Spec.Config[v1alpha2.ProtocolSettingsQos]; ok {
+		src.initializeProtocolSettingsIfNil()
```
mfaizanse, line 177: (nit) what if we call the `initializeProtocolSettingsIfNil` and `initializeWebhookAuthIfNil` at the start of this method, instead of doing it in every if block?

mfaizanse (approved):

components/eventing-controller/api/v1alpha1/subscription_conversion.go
raypinto: I do not want to initialise it if its not specified. If it was nil I want to retain it as nil

mfaizanse (approved):

PR #16193, opened 2022-11-28
Register template for keda-v0.0.1-alpha

kwiatekus:
**Description**

Changes proposed in this pull request:

- add module template for keda module (v0.0.1-alpha)

How to use it:

[Get latest (unstable) binary for kyma cli](https://github.com/kyma-project/module-manager/issues/157)
```
kyma-unstable alpha deploy

kubectl apply -k modules/alpha                                                                     
moduletemplate.operator.kyma-project.io/moduletemplate-keda created
```

**Related issue(s)**
https://github.com/kyma-project/keda-manager/issues/58
https://github.com/kyma-project/module-manager/issues/157

PR #16347, opened 2022-12-12
Register keda-module 0.0.1 in alpha channel

m00g3n:
**Description**

- add module template for keda module (v0.0.1-alpha)

How to use it:

[Get latest (unstable) binary for kyma cli](https://github.com/kyma-project/module-manager/issues/157)
```
kyma-unstable alpha deploy

kubectl apply -k modules/alpha                                                                     
moduletemplate.operator.kyma-project.io/moduletemplate-keda created
```

**Related issue(s)**
https://github.com/kyma-project/keda-manager/issues/58
https://github.com/kyma-project/module-manager/issues/157

See also:
- https://github.com/kyma-project/kyma/pull/16193

### DIFF
--- a/modules/alpha/kustomization.yaml
+++ b/modules/alpha/kustomization.yaml
@@ -1,3 +1,5 @@
 apiVersion: kustomize.config.k8s.io/v1alpha1
 kind: Component
 
+resources:
+- moduletemplate-keda.yaml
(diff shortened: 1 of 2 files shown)


modules/alpha/kustomization.yaml
```diff
@@ -1,3 +1,5 @@
 apiVersion: kustomize.config.k8s.io/v1alpha1
 kind: Component
 
+resources:
+- moduletemplate-keda.yaml
```
adityabhatia, line 5: not sure if it will have an effect but the target file is called `moduletemplate-keda..yaml` with en extra dot before ext.

m00g3n: will check it out, thx

m00g3n: fixed

PR #16351, opened 2022-12-13
Print error when jetstream dispatch fails

k15r:
**Description**

- Print the full error message in case a message cannot be dispatched to the configured sink

**Related issue(s)**


muralov (approved):
👍🏼

PR #16357, opened 2022-12-14
Disable TLS when non https endpoint used

hisarbalik:
**Description**

- OTLP gRPC exporter use by default TLS enabled and required cert file, this cause TLS hand shake error with Kyma Jaeger collector. TLS will be disabled when used endpoint is not a https endpoint

**Related issue(s)**

### DIFF
--- a/components/telemetry-operator/controller/tracepipeline/resources.go
+++ b/components/telemetry-operator/controller/tracepipeline/resources.go
@@ -109,7 +109,7 @@ func getOutputType(output v1alpha1.TracePipelineOutput) string {
 
 func makeExporterConfig(output v1alpha1.TracePipelineOutput) map[string]any {
 	outputType := getOutputType(output)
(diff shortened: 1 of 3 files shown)


components/telemetry-operator/controller/tracepipeline/resources.go
```diff
@@ -108,6 +109,7 @@ func getOutputType(output v1alpha1.TracePipelineOutput) string {
 
 func makeExporterConfig(output v1alpha1.TracePipelineOutput) map[string]any {
 	outputType := getOutputType(output)
+	isInsecure := len(strings.TrimSpace(output.Otlp.Endpoint.Value)) > 0 && !strings.HasPrefix(output.Otlp.Endpoint.Value, "https://")
```
chrkl, line 112: Let's reverse the condition. TLS should be the default and only if the `http://` prefix is set explicitly, we set the insecure flag. Please also add a unittest.

hisarbalik: sure, but we want to make https explicit, in case of no scheme provided should be treated as http

chrkl (approved):

Issue #16331, opened 2022-12-09
Eventing unit/integration tests are flaky


mfaizanse:
The eventing unit and integration tests are too flaky, especially on local machine.

The most flaky tests are:
- `FAIL: TestJSSubscriptionWithDuplicateFilters (0.13s)`
- `FAIL: TestJSSubscriptionWithMaxInFlightChange (0.13s)`
- `Should recreate BEB Subscription again [It]`
        - `/home/prow/go/src/github.com/kyma-project/kyma/components/eventing-controller/controllers/subscription/beb/reconciler_integration_test.go:1000`

**Prow jobs:**
   - https://status.build.kyma-project.io/view/gs/kyma-prow-logs/pr-logs/pull/kyma-project_kyma/16264/pull-eventing-controller-unit-test/1601230832385658880
   - https://status.build.kyma-project.io/view/gs/kyma-prow-logs/pr-logs/pull/kyma-project_kyma/16308/pull-eventing-controller-unit-test/1601230451882594304

**On local machine:**
1. In eventing dir: `cd components/eventing-controller`
2. Run: `make test-local`

Also, on local machine, its leaving behind dangling `api-server` and `etcd` instances.

mfaizanse:
Logs from local machine on `make test-local`:
```
--- FAIL: TestCreateSubscription (139.88s)
    --- FAIL: TestCreateSubscription/create_and_delete (10.04s)
        reconciler_internal_integration_test.go:289: creating the k8s subscription
        reconciler_internal_integration_test.go:291: testing the k8s subscription
        test_utils.go:98:
            Timed out after 10.000s.
            Failed to match the eventing subscription
            Expected
                <[]v1alpha1.Condition | len:0, cap:0>: nil
            to contain element matching
                <*gstruct.FieldsMatcher | 0xc0009c1560>: {
                    Fields: {
                        "Status": <*matchers.EqualMatcher | 0xc000f167b0>{
                            Expected: <v1.ConditionStatus>"True",
                        },
                        "Type": <*matchers.EqualMatcher | 0xc000f16760>{
                            Expected: <v1alpha1.ConditionType>"Subscription active",
                        },
                        "Reason": <*matchers.EqualMatcher | 0xc000f16780>{
                            Expected: <v1alpha1.ConditionReason>"NATS Subscription active",
                        },
                        "Message": <*matchers.EqualMatcher | 0xc000f16790>{Expected: <string>""},
                    },
                    IgnoreExtras: true,
[...]

PR #16358, opened 2022-12-14
Made JetStream tests use unique stream names

mfaizanse:
**Description**

Problem is that even if you run multiple NATS test servers on different ports, the underlying data conflicts and hence the test fails. Also, using unique `storeDir` for stream did not help. So, to keep the test environment isolated, we are now using unique stream names and subject prefixes in each test. (Because subjects cannot overlap with an existing stream, we have to use unique subject prefixes as well [`nats: subjects overlap with an existing stream`]).

- The `JetStreamSubjectPrefix` was a const before and could not be defined variably in the tests, so first this PR changes the `JetStreamSubjectPrefix` to NATS config variable.
- Second, this PR changes the JetStream tests to use unique stream name and subject prefixes.

**Related issue(s)**

- https://github.com/kyma-project/kyma/issues/16331

### DIFF
--- a/components/eventing-controller/pkg/backend/jetstreamv2/test_helpers.go
+++ b/components/eventing-controller/pkg/backend/jetstreamv2/test_helpers.go
@@ -24,9 +24,10 @@ import (
 )
 
 const (
-	DefaultStreamName    = "kyma"
-	DefaultMaxReconnects = 10
-	DefaultMaxInFlights  = 10
+	DefaultStreamName             = "kyma"
+	DefaultJetStreamSubjectPrefix = "kyma"
+	DefaultMaxReconnects          = 10
+	DefaultMaxInFlights           = 10
 )
 
 // TestEnvironment provides mocked resources for tests.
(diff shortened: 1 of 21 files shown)


components/eventing-controller/pkg/backend/jetstreamv2/test_helpers.go
```diff
@@ -24,9 +24,10 @@ import (
 )
 
 const (
-	DefaultStreamName    = "kyma"
-	DefaultMaxReconnects = 10
-	DefaultMaxInFlights  = 10
+	DefaultStreamName             = "kyma"
+	DefaultJetStreamSubjectPrefix = "kyma"
```
raypinto, line 28: (nit) this will not clash with any other test right?

raypinto (commented):
LGTM

components/eventing-controller/pkg/backend/jetstreamv2/test_helpers.go
mfaizanse: nope, this is just a default prefix for stream name, so in the tests I am using like:
```
fmt.sprintf("%s%d", DefaultStreamName, natsPort)
```

raypinto (approved):

Issue #16299, opened 2022-12-08
Activate configurable tracing feature


a-thaler:
**Description**
After [configurable tracing](https://github.com/kyma-project/kyma/issues/11231) is implemented as MVP, it is time to release it. For that it needs to be activated by default, still having the option to disable it via feature flag for a rollout. Furthermore, the old behaviour with the tracing module should kept stable to have enough time to deprecate the module and adjust testing and so on.

**Criterias**
- CRD gets deployed always
- feature flag is activated by default (not removed yet to allow a phased rollout)
- tracing module activates tracing in istio - in evaluation profile it configures a 100% sampling rate
- tracing module deploys a default TracePipeline

chrkl:
PR #16372 activates configurable tracing in the telemetry-operator. The tracing component activates Istio tracing by creating a Telemetry CR and activates trace shipping to Jaeger with a TracePipeline.

Issue #11638, opened 2021-07-06
Enable w3c-tracecontext support in istio


a-thaler:
**Description**
As part of https://github.com/kyma-project/kyma/issues/11231 the otel-collector will allow support for opencensus and with that will enable us to switch the tracer in istio to be based on opencensus, supporting [w3c-tracecontext](https://www.w3.org/TR/trace-context/) and current B3. As the w3c finally gets adopted widely and is the only vendor-neutral and open protocol in that area, Kyma should switch to that protocol by default.

Goal: Change the default istio configuration to use the w3c-tracecontext protocol as default.

Criteria:
- Requests via the istio ingressgateway will be enriched with proper w3c-tracecontext headers if not existent
- Envoys are pushing trace data to the central collector/backend
- Tracing is disabled by default and can be activated by the user
- Sampling rate can be configured by the user and is using the istio default by default

**Reasons**

**Attachments**

chrkl:
PR #16372 introduce an OpenCensus extension provider for Istio that configures the W3C trace context. Istio tracing has to be activated using a Telemetry resource.

PR #16372, opened 2022-12-15
Activate configurable tracing

chrkl:
**Description**

Enable configurable tracing and integrate with Istio and Jaeger

- Enable TracePipeline controller in telemetry-operator
- Create Istio extension provider for pushing traces to otel-collector
- Switch to W3C trace context
- Change tracing component to integrate with otel-collector

**Related issue(s)**

https://github.com/kyma-project/kyma/issues/11638
https://github.com/kyma-project/kyma/issues/16299


resources/istio/values.yaml
```diff
@@ -41,15 +41,19 @@ meshConfig:
...
-    tracing:
-      sampling: 1
-      zipkin:
-        address: "zipkin.kyma-system:9411"
   enablePrometheusMerge: false
   enableTracing: "{{ .Values.global.tracing.enabled }}"
   extensionProviders:
+    - name: kyma-traces
+      opencensus:
+        service: "telemetry-trace-collector-internal.kyma-system.svc.cluster.local"
```
rakesh-garimella, line 53: should we pass this also from the values yaml ? I am afraid if we update the service name then we might forget to change it here

a-thaler: and if it is in the values.yaml of the istio component you will not forget about it?

chrkl: This is values.yaml ;-)

rakesh-garimella (approved):

resources/tracing/values.yaml
```diff
@@ -235,4 +235,8 @@ authorizationPolicy:
 #      - cluster.local/ns/kiali/sa/kiali-server-service-account
   monitoring:
      extraPrincipals:
-#      - cluster.local/ns/kiali/sa/kiali-server-service-account
\ No newline at end of file
+#      - cluster.local/ns/kiali/sa/kiali-server-service-account
+
+istio:
+  samplingRate: 100.0
```
a-thaler, line 241: I would turn it around to have the same behaviour as before. The istio chart used 0.1 by default and only changed it for evaluation to 100.0

chrkl: Done

rakesh-garimella (approved):

dbadura (approved):

triffer (approved):

resources/serverless/values.yaml
```diff
@@ -217,7 +217,7 @@ containers:
       functionJaegerServiceEndpoint:
         value: "http://tracing-jaeger-collector.kyma-system.svc.cluster.local:14268/api/traces"
```
kwiatekus, line 218: @chrkl would this legacy endpoint continue to work ?

chrkl: For now it still works, but would not recommend to use it since the otel collector adds some workload related attributes to the spans (pod name, namespace, etc.). On long term, Jaeger will be removed from Kyma and be replaced by a tutorial.

Is it still in use? Our otel collector does not provide a Jaeger endpoint that can be used as a replacement.

dbadura: If it's going to be deleted, we need to reload all functions @kwiatekus

kwiatekus: For all new/Updated functions the new collector endpoint will be  used. However there might be still functions that were injected with the old value of the url and will continue to use it even after kyma upgrade.. unless we enforce reloading of those envs somehow

a-thaler: We will not remove jaeger now, so the old endpoint will continue to work for a while. For an SKR scenario it is a new feature to have tracing at all. I think it is fine to not do any additional actions around that.

PR #16397, opened 2022-12-20
Implement update test

VOID404:
**Description**

- Add a test case that updates an app in compass, then verifies if it gets updated correctly on cluster

**Related issue(s)**

#15037

### DIFF
--- a/tests/components/application-connector/test/compass-runtime-agent/synchronisation_test.go
+++ b/tests/components/application-connector/test/compass-runtime-agent/synchronisation_test.go
@@ -45,8 +45,7 @@ func (cs *CompassRuntimeAgentSuite) TestApplication() {
 	})
 
 	cs.Run("Update app", func() {
-		err = cs.updateAndWaitForCompare(applicationInterface, synchronizedCompassAppName, applicationID, updatedAppName)
-		cs.Require().NoError(err)
+		_ = cs.updateAndWaitForCompare(applicationInterface, synchronizedCompassAppName, applicationID, updatedAppName)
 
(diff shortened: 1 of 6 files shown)


franpog859 (commented):
The image is not updated. Therefore, I believe this test is not running on the pipeline yet. Let's change it and check it out! Also, please, update the PR description so we know what actually is done here in regards to the epic acceptance criteria

tests/components/application-connector/test/compass-runtime-agent/synchronisation_test.go
```diff
@@ -49,6 +58,32 @@ func (cs *CompassRuntimeAgentSuite) TestApplication() {
...
+		return err
+	}
+
+	verify := func() bool {
+		app, err := appReader.Get(context.Background(), compassAppName, v1.GetOptions{})
+		if err != nil {
+			t.Logf("Couldn't get updated: %v", err)
+		}
+
+		return err == nil && app.Spec.Description == "Updated"
```
franpog859, line 76: I'd use some constant value like `UPDATED_APP_DESCRIPTION` or something like this just to show explicitly that it's a hardcoded test value. What do you think?

franpog859: Do we really need to check the description here though? Isn't it a job for the App Comparator and the `Compare` method executed later?

VOID404: We do have to somehow wait for sync. App Comparator fails the tests if the app doesn't match so it cannot be used for waiting.

franpog859 (approved):

PR #16406, opened 2022-12-20
enable startup healthcheck for nats in production profile

raypinto:
**Description**

Startup probe checks that the JS server is enabled, is current with the meta leader, and that all streams and consumers assigned to this JS server are current. It must be enabled when nats clustering is enabled.

[_Source_](https://github.com/nats-io/k8s/blob/977343b835fcca19d1e2d583ba6778d7ed59aad1/helm/charts/nats/templates/statefulset.yaml#L480)

- Enabled startup probe for while using production profile

**Related issue(s)**

### DIFF
--- a/resources/eventing/profile-production.yaml
+++ b/resources/eventing/profile-production.yaml
@@ -48,3 +48,6 @@ nats:
     logging:
       debug: false
       trace: false
+    healthcheck:
+      startup:
+        enabled: true


raypinto:
Closing this PR until [this issue](https://github.com/nats-io/k8s/issues/622) is addressed.

Issue #15836, opened 2022-10-18
Singularity by simple status for TracePipelines


a-thaler:
**Description**
In the beginning a TracePipeline resource should exist only once as this will cover the most common use case already. Support for multiple resources should be added at a later time.
As a validation webhook cannot fully guarantee such singularity for a resource type, a safe and acceptable solution is to introduce a status for a pipeline and bring it into an error status if there is already a pipeline defined.

**Criterias**
- Have a status subresource with the typical fields for conditions
- Have a Running and Error condition
- The error condition is used when there is already an existing pipeline in Running state
- Otherwise it is Running

**Reasons**

**Attachments**

chrkl:
The concept to enfore singularity has been documented here: https://github.com/kyma-project/community/pull/748. Keeping this issue open to track the implementation status.

skhalash:
The singularity concept has been implemented in https://github.com/kyma-project/kyma/pull/16413

PR #16413, opened 2022-12-21
Ensure single trace pipeline

skhalash:
**Description**

- Implement the lock mechanism that ensures that only one trace pipeline can be active (described in https://github.com/kyma-project/community/pull/748)
- Move the reason strings from the API pkg to the respective controllers, similar to what Kubernetes does

**Related issue(s)**
https://github.com/kyma-project/kyma/issues/15836


components/telemetry-operator/controller/tracepipeline/controller.go
```diff
@@ -169,3 +177,37 @@ func (r *Reconciler) doReconcile(ctx context.Context, pipeline *telemetryv1alpha
...
+	if err := r.Get(ctx, lockName, &lock); err != nil {
+		if apierrors.IsNotFound(err) {
+			return r.createLock(ctx, lockName, pipeline)
+		}
+		return fmt.Errorf("failed to get lock: %v", err)
+	}
+
+	for _, ref := range lock.GetOwnerReferences() {
+		if ref.Name == pipeline.Name && ref.UID == pipeline.UID {
+			// I am the owner
```
lindnerby, line 193: Can be removed

lindnerby (changes requested):

components/telemetry-operator/controller/tracepipeline/controller.go
skhalash: True

lindnerby (approved):

lindnerby (approved):

Issue #16186, opened 2022-11-25
Central Application Connectivity Validator returns code 500 after application is removed


koala7659:
**Description**

During investigation of [incident](https://github.tools.sap/kyma/backlog/issues/3238) we discovered that Central Application Connectivity Validator returns code 500 (Internal server error) in a situation when processing event from a previously connected system which been removed later from Formation or subaccount. As a result application CR is removed from Kyma cluster and event is rejected. The behaviour is correct but returned HTTP code suggests that the service itself has a failure.

**Logs:**
`
{"level":"ERROR","timestamp":"2022-11-24T10:15:02Z","caller":"httptools/response.go:16","message":"Error: while getting application ClientIds: application with name mp-ct075nck5-franketec1-s1-740797407 is not found in the cache. Please retry","traceid":"unknown","spanid":"unknown","context":{"handler":"validation_proxy_handler","applicationName":"mp-ct075nck5-franketec1-s1-740797407"}}
`

**Expected result**

1.  Central Application Connectivity Validator returns 40x code when previously connected application is no longer available on the cluster.
Suggested response codes might be:
[401 Unauthorized](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/401)
[403 Forbidden](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/403)
[404 Not Found](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/404)

2. Consider to lower the level of logged information to Warning.

franpog859:
It's done in this PR - https://github.com/kyma-project/kyma/pull/16414

PR #16414, opened 2022-12-21
Change status code for not found in Validators cache

franpog859:
**Description**

- Change status code for not found in Validators cache
- I didn't change the log severity as I think it's actually an error from the architecture point of view. Using non-existing Applications is not expected

**Related issue(s)**

See the issue - https://github.com/kyma-project/kyma/issues/16186


mvshao (approved):
/lgtm

Issue #16346, opened 2022-12-12
NATS goes OOM after dying due to full storage


k15r:
Scenario

- Stream usage close to limits (900MB on 1Gb storage)
- NATS dies for no obvious reason

Expected behaviour
- NATS can be restarted and load all stream data from the storage

Actual:
 NATS goes OOM while loading the stream

marcobebway:
I could not reproduce the OOM issue with NATS 2.9.9.

However, the NATS cluster breaks when the stream size reaches the configured max-bytes value of 900Mi, and restarting the NATS cluster pods does help recovering from this.

Testing a lower max-bytes value of 700Mi seems to be more stable and although the NATS cluster could still break under high load when the stream size reaches the 700Mi, but restarting the NATS pods was sufficient to recover the NATS cluster.

Here is a [PR](https://github.com/kyma-project/kyma/pull/16423) to change the max-bytes value to be 70% of the stream storage.

PR #16423, opened 2022-12-21
Change JetStream maxBytes to be 70% of the total file storage

marcobebway:
**Description**

As part of this [ticket](https://github.com/kyma-project/kyma/issues/16346), we discovered that the max-bytes value of 900Mi of the stream size 1G is not suitable and causes the NATS cluster to break beyond repair. The root cause of this is not yet decided, but we suspect that NATS needs more storage buffer to store extra stream meta-data. This PR changes the max-bytes value to be 70% of the stream size to give more room for NATS meta storage.

**Related issue(s)**

- https://github.com/kyma-project/kyma/issues/16346

### DIFF
--- a/resources/eventing/profile-evaluation.yaml
+++ b/resources/eventing/profile-evaluation.yaml
@@ -1,7 +1,7 @@
 global:
   jetstream:
     storage: file
-    maxBytes: 900Mi
+    maxBytes: 700Mi
     fileStorage:
       size: 1Gi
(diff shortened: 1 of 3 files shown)


raypinto (dismissed):
LGTM

friedrichwilken (approved):
/lgtm

mfaizanse (approved):

PR #16442, opened 2022-12-23
Mitigate OAuth2Client reapply

barchw:
**Description**

- Adds metadata: null  and finilizers fields to chart, which clean up caused  errors in hydra-maester

### DIFF
--- a/resources/eventing/charts/controller/templates/oauth2-client.yaml
+++ b/resources/eventing/charts/controller/templates/oauth2-client.yaml
@@ -4,8 +4,6 @@ kind: OAuth2Client
 metadata:
   name: {{ include "controller.fullname" . }}
   labels: {{- include "controller.labels" . | nindent 4 }}
-  finalizers:
-    - finalizer.ory.hydra.sh
 spec:
   grantTypes:
     - client_credentials
--- a/resources/eventing/charts/controller/templates/oauth2-client.yaml
+++ b/resources/eventing/charts/controller/templates/oauth2-client.yaml
@@ -4,9 +4,12 @@ kind: OAuth2Client
 metadata:
   name: {{ include "controller.fullname" . }}
   labels: {{- include "controller.labels" . | nindent 4 }}
+  finalizers:
+    - finalizer.ory.hydra.sh
 spec:
   grantTypes:
     - client_credentials
   hydraAdmin: {}
+  metadata: null
   scope: read write beb uaa.resource
   secretName: {{ include "controller.fullname" . }}{{ .Values.bebSecret.nameSuffix }}


resources/eventing/charts/controller/templates/oauth2-client.yaml
```diff
@@ -4,9 +4,12 @@ kind: OAuth2Client
 metadata:
   name: {{ include "controller.fullname" . }}
   labels: {{- include "controller.labels" . | nindent 4 }}
+  finalizers:
```
jakobmoellersap, line 7: I suggest to never ever add finalisers coming from a controller since it will block deletion of a resource if the controller is gone. Rather the edited resource should be patched via strategic merge, ignoring the finalisers. I suggest to review the logic that forces this field to be present

barchw: Would it be possible then to configure reconciler not to remove the `finalizers` if they were applied by hydra-maester controller? For the sake of fixing the error that would also be OK

k15r: 😕 reconciler removes finalizers? In what scenario?

jakobmoellersap: TBH also surprising for me. Finalisers outside of the reconciler should be ignored imo. Also setting nil metadata would result in them being removed on every reconcile. @tobiscr are you aware of the behaviour here? I almost cannot imagine that this never was an issue before...

tobiscr: The finalisers are ignored by the old reconciler until we are deleting the cluster. During the deletion, one of the final steps is the cleanup of finalisers to make sure every resources gets dropped (just to make sure that there are no "dead bodies" left).

tobiscr: If you have to deal with finalisers during runtime, you can either handle such cases in a custom reconciler (but this will update finalisers only during the regular reconciliation runs) or you handle it over an operator during runtime.

barchw: Removed the finalizers as those will be handled by https://github.com/kyma-incubator/reconciler/pull/1274

resources/eventing/charts/controller/templates/oauth2-client.yaml
```diff
@@ -8,5 +8,6 @@ spec:
   grantTypes:
     - client_credentials
   hydraAdmin: {}
+  metadata: null
```
k15r, line 11: Is this change really necessary?

k15r: well then ill approve this noop change :D

k15r (approved):

PR #16474, opened 2022-12-30
Add NATS ConnectionBuilder

nachtmaar:
**Description**

- introduces a `ConnectionBuilder` to **ease** the **creation** of a **NATS connection** and to **allow better testing** (via the new interface)
- The ConnectionBuilder is used in this upcoming PR https://github.com/kyma-project/kyma/pull/15720:

**Code**
Screenshot 2022-12-30 at 15 57 51
Screenshot 2022-12-30 at 15 58 43
**Tests**
It allows to stub/mock the connection and test the case that the connection creation/establishment fails:
Screenshot 2022-12-30 at 16 00 04

**Related issue(s)**

See also: https://github.com/kyma-project/kyma/pull/15720


components/eventing-controller/pkg/backend/jetstreamv2/connection.go
```diff
@@ -0,0 +1,45 @@
...
+	conn, err := nats.Connect(config.URL, jsOptions...)
+	if err != nil || !conn.IsConnected() {
+		return nil, pkgerrors.MakeError(ErrConnect, err)
+	}
+
+	return conn, nil
+}
+
+// ConnectionInterface is a contract for a NATS connection object.
+type ConnectionInterface interface {
```
friedrichwilken, line 40: I am thinking about the name. I know there is this `-er` convention for Interfaces, but I am also not always super convinced about that. 

I like the approach in [this](https://medium.com/@dotronglong/interface-naming-convention-in-golang-f53d9f471593) short article; especially what is written at point # 3. Let me know what you think.

components/eventing-controller/pkg/backend/jetstreamv2/connection.go
```diff
@@ -0,0 +1,45 @@
...
```
friedrichwilken, line 40: Okay, one last try to find better names:
Name the `ConnectionInterface` -> `Connector`
and the `Build` function `Connect`.
Let me know what you think. Other than that, this looks good to me.

nachtmaar: If I change the Build() method to Connect, then the `Builder` interface needs to be renamed to `Connector` :-D

The `ConnectionInterface` is simply an interface for the `nats.Conn` (nats sdk does not offer an interface)

nachtmaar: `jetstreamv2.Connection`

friedrichwilken (approved):
/lgtm

components/eventing-controller/pkg/backend/jetstreamv2/connection.go
friedrichwilken: Never mind, maybe I am making it too complicated.

Issue #15863, opened 2022-10-20
Separate unit, integration and component test in EC


friedrichwilken:
- [ ] separate unit, integration and component tests where needed into different `_test.go` files
- [ ] add corresponding [build tags](https://mickey.dev/posts/go-build-tags-testing/) to all the test files
- [ ] create make targets to run only unit, integration or component tests

k15r:
individual test can already be executed using go test / goland. no need to invest so much time to introduce this feature

PR #16476, opened 2023-01-02
Separate between unit and integration tests for jetstreamv2 package

nachtmaar:
**Description**

- The changes in this PR allow to run the unit and integration tests independent of each other.
- This makes sense if one wants to see the coverage for unit or integrations tests only. Or if the integration tests are too slow.
- They can still be run all together 😃

```
# Run unit tests:
go test ./... --tags=unit -v
# Run integration tests:
go test ./... --tags=integration -v
```
**Related issue(s)**

See also https://github.com/kyma-project/kyma/issues/15863


friedrichwilken (approved):
/lgtm

PR #16510, opened 2023-01-09
improve application name parsing from request path

k15r:
**Description**

- instead of parsing the entire path into a slice and then returning the first element, we just cleanup the string and cut off the first element.

**Related issue(s)**


friedrichwilken (approved):
/lgtm

PR #16517, opened 2023-01-09
Fix otel collector restart

chrkl:
**Description**

- Fix setting deployment annotation

**Related issue(s)**


skhalash (commented):
Would it be possible to add a unit test to cover this scenario?

chrkl:
> Would it be possible to add a unit test to cover this scenario?

Absolutely, planning to do that.

skhalash (approved):

Issue #16528, opened 2023-01-10
Legacy Events endpoint should support non-clean application names


marcobebway:
**Description**

While testing 2.10.0-rc1 we discovered that the legacy events endpoint is broken and does not accept events from applications with non-clean name.

**Expected result**

The publish requests should succeed.

**Actual result**

The publish requests fail.

**Steps to reproduce**

- Provision a Kyma 2.10.0-rc1.
- Publish a legacy event to no-app endpoint `http://eventing-publisher-proxy.kyma-system/no-app/v1/events`.
- The publish request fails with error `{"status":500,"message":"application name should be cleaned from non-alphanumeric characters"}`.

PR #16535, opened 2023-01-11
Release 2.10

muralov:
**Description**
Fix legacy event publish not clean app name issue. 

- Don't validate app name while extracting request data, as it is checked during CE conversion

**Related issue(s)**
#16528

### DIFF
--- a/components/event-publisher-proxy/pkg/handler/handler.go
+++ b/components/event-publisher-proxy/pkg/handler/handler.go
@@ -128,7 +128,7 @@ func (h *Handler) maxBytes(f http.HandlerFunc) http.HandlerFunc {
 	}
 }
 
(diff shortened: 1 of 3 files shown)


resources/eventing/values.yaml
```diff
@@ -9,7 +9,7 @@ global:
       pullPolicy: "IfNotPresent"
     publisher_proxy:
       name: event-publisher-proxy
-      version: v20221230-5e8bc6ca
+      version: v20230111-18887898
```
mfaizanse, line 12: lets keep it same as `v20221230-5e8bc6ca`, and once this PR is merged then bump the image. How did you get a version before PR is merged?

muralov: good point! It is the build of main, which shouldn't be released.

mfaizanse (approved):

PR #16547, opened 2023-01-11
Add PeerAuthentication to scrape Fluent Bit via plain http

skhalash:
**Description**

- The peer authentication will allow scraping via HTTP. The metrics do not contain sensitive data, so it is acceptable from the security point of view
- Modify the service monitor to scrape via HTTP

**Related issue(s)**


a-thaler:
That resource needs to be managed by code as well, we want to switch soon

chrkl (approved):

PR #16553, opened 2023-01-12
change epp linter config

k15r:
**Description**

- add new linter config
- fix go cognit lint issue
- fix godot lint issues
- fix lint issues lll (where it made sense)
- fix gomnd issues
- fix variable naming
- run tests in parallel
- ignore shadowing `err`
- remove namedreturns
- fix more lint issues

**Related issue(s)**


components/event-publisher-proxy/pkg/handler/handler_test.go
```diff
@@ -418,15 +425,20 @@ func TestHandler_publishLegacyEventsAsCE(t *testing.T) {
...
+					NewApplicationListerOrDie(context.Background(), "testapp")),
+				collector:        metrics.NewCollector(latency),
+				eventTypeCleaner: eventtypetest.CleanerStub{},
 			},
 			args: args{
 				request: legacytest.ValidLegacyRequestOrDie(t, "v1", "testapp", "object.created"),
 			},
 			wantStatus: 200,
 			wantOk:     true,
+
```
VladislavPaskar, line 440: with so many nolint directives, you could just apply the `nolint:lll` for the whole file in the first line of this file. That will save some line of code. The drawback of this of course, is that it will be applied for all the code including the function signatures.

components/event-publisher-proxy/pkg/handler/handler_v1alpha2_test.go
```diff
@@ -74,6 +76,7 @@ func TestHandler_publishCloudEventsV1Alpha2(t *testing.T) {
 				request: CreateValidStructuredRequest(t),
 			},
 			wantStatus: 204,
+			//nolint:lll // output directly from prometheus
```
VladislavPaskar, line 79: the same here

components/event-publisher-proxy/pkg/legacy/legacy.go
```diff
@@ -240,18 +239,18 @@ func (t *Transformer) convertPublishRequestToCloudEvent(appName string, publishR
 }
 
 // combineEventNameSegments returns an eventName with exactly two segments separated by "." if the given event-type
-// has more than two segments separated by "." (e.g. "Account.Order.Created" becomes "AccountOrder.Created")
+// has more than two segments separated by "." (e.g. "Account.Order.Created" becomes "AccountOrder.Created").
 func combineEventNameSegments(eventName string) string {
 	parts := strings.Split(eventName, ".")
-	if len(parts) > 2 {
+	if len(parts) > 1 {
```
VladislavPaskar, line 245: what is the reason of changing this ? the comment says also `as more than two segments`

components/event-publisher-proxy/.golangci.yaml
```diff
@@ -0,0 +1,281 @@
+# This code is licensed under the terms of the MIT license.
```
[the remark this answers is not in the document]
k15r: it is the same. the licenses are compatible.

components/event-publisher-proxy/pkg/handler/handler_test.go
k15r: sounds like a valid solution 👍

components/event-publisher-proxy/pkg/legacy/legacy.go
k15r: i fixed the comment. 
`> 2` is marked by the linter as a `magic number`
the code works exactly the same with `> 1` so that change make sense anyway

mfaizanse (approved):

Issue #16435, opened 2022-12-22
Runtime debuggability of telemetry-operator


a-thaler:
**Description**
Have a way to change operator settings (mainly related to reconcilation) at runtime without getting reconciled like for disabling reconcilation or changing log levels.
For that, we need a dynamic configuration option with some new features for improving operations.

**Criterias**
- At runtime I can pass a config file with settings
- I can add/remove settings at any time without getting the operator restarted
- I can pause reconcilation for Trace/LogPipeline
  - With that I can manipulate the otel-collector log level directly
    ```
    service:
      telemetry:
        logs:
          level: info #info | debug
    ```
  - With that I can manipulate the verbosity of the logging exporter
    ```
    exporters:  
      logging:
         verbosity: basic # basic | normal | detailed
    ```

a-thaler:
There is still an open PR: https://github.com/kyma-project/kyma/pull/16616

Also please outline what to in order to enable pausing, what is the new API?

rakesh-garimella:
Debugging capabilities have been implemented by having a `pause` and changing the `loglevel` on fly feature. These debuggability features have been documented in the Readme.
Additionally logging exporter has also been enabled.

PR #16567, opened 2023-01-16
Add debug ability to the telemetry operator

rakesh-garimella:
**Description**

- The telemetry operator would look for override configmap at the start of reconciliation and would:
    * pause if the the pause flag is set in either logging or tracing config.
    * Change log level of controller runtime on fly.
- To keep code simple a lazy implementation has been implemented, where one needs to trigger reconciliation after adding the configmap to the kyma-system namespace
- Fast integration tests are implemented [here](https://github.com/kyma-project/kyma/pull/16616)

**Related issue(s)**

https://github.com/kyma-project/kyma/issues/16435


components/telemetry-operator/controller/logpipeline/controller.go
```diff
@@ -54,6 +57,11 @@ type DaemonSetAnnotator interface {
 	SetAnnotation(ctx context.Context, name types.NamespacedName, key, value string) error
 }
 
+type GlobalConfigHandler interface {
```
chrkl, line 60: Let's move this interface to the overrides package (same for tracepipeline/controller.go)

components/telemetry-operator/internal/kubernetes/configmap.go
```diff
@@ -0,0 +1,33 @@
...
+	logf "sigs.k8s.io/controller-runtime/pkg/log"
+)
+
+type ConfigmapProber struct {
+	client.Client
+}
+
+const overrideConfigFileName = "override-config"
+
+func (cmp *ConfigmapProber) ReadConfigMapIfPresent(ctx context.Context, name types.NamespacedName) (string, error) {
```
chrkl, line 19: The behaviour of returning an empty string if the CM does not exist might be not expected for that method name. Maybe let's name it `ReadConfigMapOrEmpty`

components/telemetry-operator/internal/overrides/overrides.go
```diff
@@ -0,0 +1,81 @@
...
+	if len(config) == 0 {
+		return overrideConfig, nil
+	}
+
+	err = yaml.Unmarshal([]byte(config), &overrideConfig)
+	if err != nil {
+		return overrideConfig, err
+	}
+
+	log.V(1).Info(fmt.Sprintf("Override Config is: %+v", overrideConfig))
```
chrkl, line 70: Follow our [conventions](https://github.com/kyma-project/community/blob/main/concepts/observability-consistent-logging/improvement-of-log-messages-usability.md#log-structure) for log messages. Maybe better have a message like `Using override config ...`

components/telemetry-operator/internal/configureLogger/configurelogger.go
```diff
@@ -0,0 +1,32 @@
+package configurelogger
```
chrkl, line 1: I miss the `Good package names are short and clear.` aspect from [this](https://go.dev/blog/package-names) article. How about moving this to the `logger` package?

components/telemetry-operator/controller/logparser/controller.go
```diff
@@ -46,22 +49,29 @@ type DaemonSetAnnotator interface {
 	SetAnnotation(ctx context.Context, name types.NamespacedName, key, value string) error
 }
 
+type GlobalConfigHandler interface {
```
chrkl, line 52: Can be removed here, right?

components/telemetry-operator/internal/configureLogger/configurelogger.go
rakesh-garimella: fixed

chrkl (approved):

Issue #16053, opened 2022-11-09
Odata apis are STILL URL encoded by Kyma causing the API call to fail


ArthurPerry:
**Description**
Related to #5164

**Description**

In that case, the OData object had a simple string as a key.  In this case I am requesting an OData object from Commerce which has a composite key: 
`Online|apparelProductCatalog|1667409165852|29533|reviewer1@hybris.com`.  

The composite key elements are delimited by `|` which should be escaped, but the Central API Gateway also encodes the parenthesis `(` and `)` which causes the Commerce OData API to be unable to parse the query.

Version
Kubernetes: v1.23.12
Kyma: 2.7.3

API Server Address
https://api.c-81b58c5.kyma.shoot.live.k8s-hana.ondemand.com

**Expected result**

Here is the Commerce OData URL which succeeds:
https://backoffice.crz8nbwca-internalc6-d30-public.model-t.cc.commerce.ondemand.com/odata2webservices/CustomerProductReview/CustomerReviews('Online%7CapparelProductCatalog%7C1667409165852%7C29533%7Creviewer1%2540hybris.com')

**Actual result**

`"Modified request url : 'https://backoffice.crz8nbwca-internalc6-d30-public.model-t.cc.commerce.ondemand.com/odata2webservices/CustomerProductReview/CustomerReviews%28%27Online%7CapparelProductCatalog%7C1667409165852%7C29533%7Creviewer1@hybris.com%27%29', schema : 'https', path : '/odata2webservices/CustomerProductReview/CustomerReviews('Online|apparelProductCatalog|1667409165852|29533|reviewer1@hybris.com')'\""`

**Steps to reproduce**

Any call to the Central API Gateway for this API: i.e.

http://central-application-gateway.kyma-system.svc.cluster.local:8082/mp-demo-ccv2-d30-tst-eu20/customerproductreview-integration-api/customerproductreview-integration-api/CustomerReviews('Online|apparelProductCatalog|1667409165852|29533|reviewer1@hybris.com')

**Troubleshooting**

I have tried to encode the URL path and object key different ways. but I cannot influence how the Central Application Gateway handles the encoding.

VOID404:
#14783 got closed with #16577. The feature is present on `main`. You can now disable encoding with `encodeUrl` in Application CR.

ArthurPerry:
Is this fixed in Kyma 2.10.1?  I still get the issue.  Do I need to change something with my Application CR?

VOID404:
[...]

Issue #14783, opened 2022-07-11
Add a possibility to disable URL decoding in Application Gateway


janmedrek:
**Description**

There should be a way to enable/disable URL decoding in the Application Gateway. Right now - the URL is always decoded so there is no possibility to pass on encoded path segments (i.e. `curl http://central-application-gateway.kyma-system:8080/application/service/path/segment/abc%2Fsome%2Fvalue%2F`)

**Reasons**

Sometimes the user would like to have the undecoded part sent in the proxied request via Application Gateway.

**Acceptance**
- [ ] There is a switch added in Application CR that enables/disables URL decoding in Application Gateway for the whole Application
- [ ] By default, decoding is enabled (behavior used so far)
- [ ] Application Gateway passes the request without decoding the whole URL when decoding is disabled

janmedrek:
I have used the https://webhook.site/ to investigate the issue, the single-encoded path segments are decoded but there is no decoding for the double-encoded segments (i.e. `/%252Fabc%252Fdef`).

My guess would be that somehow URL parsing in Go automatically decodes the request's path: https://github.com/kyma-project/kyma/blob/9ec2d8f628c9f7c141c89fe276525c7772da4fcd/components/central-application-gateway/internal/proxy/reverseproxy.go#L25

VOID404:
Solved by #16577

PR #16577, opened 2023-01-17
Parametrise gateway URL encoding

VOID404:
**Description**

This PR adds and option to honour the original URL encoding

- Add config option to `AppCR`
- Use URL for URLs, instead of `string`
- Add relevant unit tests

**Related issue(s)**

#14783 #16053


components/central-application-gateway/internal/proxy/proxy.go
```diff
@@ -50,14 +49,20 @@ func (p *proxy) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	r.URL.Path = path
+	// r.URL.Path = path
+	// r.URL.RawPath = path
```
franpog859, line 53: I believe these comments could be removed

components/central-application-gateway/internal/proxy/factory_test.go
```diff
@@ -39,21 +39,24 @@ func TestProxyFactory(t *testing.T) {
...
 
-	createMockServiceDeffService := func(apiIdentifier model.APIIdentifier, targetURL string, credentials *authorization.Credentials) metadatamocks.ServiceDefinitionService {
-		serviceDefServiceMock := metadatamocks.ServiceDefinitionService{}
-		serviceDefServiceMock.On("GetAPIByServiceName", apiIdentifier.Application, apiIdentifier.Service).Return(&metadatamodel.API{
-			TargetUrl:   targetURL,
-			Credentials: credentials,
-		}, nil).Once()
-
-		return serviceDefServiceMock
+	createMockServiceDeffService := func(encodeURL bool) func(apiIdentifier model.APIIdentifier, targetURL string, credentials *authorization.Credentials) metadatamocks.ServiceDefinitionService {
```
franpog859, line 49: What an abomination :'D Is there a more readable way to do it? I mean passing `createMockServiceDeffService(true)` as a function could be misleading at a first glance. And there is no real indication of what this `true` is unless you read this

components/central-application-gateway/internal/proxy/proxy.go
```diff
@@ -77,10 +82,10 @@ func (p *proxy) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	cacheEntry.Proxy.ServeHTTP(w, newRequest)
 }
 
-func (p *proxy) extractPath(u *url.URL) (model.APIIdentifier, string, *url.URL, apperrors.AppError) {
+func (p *proxy) extractPath(u *url.URL) (model.APIIdentifier, *url.URL, *url.URL, apperrors.AppError) {
```
franpog859, line 85: These return types are getting quite hard to guess what is what ;-; It's not a part of this task so I don't think you should fix it here, just noticed stuff :d

components/central-application-gateway/internal/proxy/factory.go
```diff
@@ -41,7 +42,10 @@ func New(
 		gwURL := *u
 		gwURL.Path = "/" + strings.Join(split[:2], "/")
 
-		return apiIdentifier, targetAPIPath, &gwURL, nil
+		// TODO: make sure
+		targetURL, _ := url.Parse(targetAPIPath) // it shouldn't be possible for this to fail?
```
franpog859, line 46: Is it done?

installation/resources/crds/application-connector/applications.applicationconnector.crd.yaml
```diff
@@ -43,6 +43,9 @@ spec:
                   type: boolean
                 skipVerify:
                   type: boolean
+                encodeUrl:
+                  type: boolean
+                  default: true
```
franpog859, line 48: Did you test the behavior when you apply a CR without this field? And some upgrade scenarios where there are already CRs present in the cluster and you upgrade the CRD?

franpog859 (commented):
Do you plan to add any integration tests to this PR?

Also it'd be great to update the official documentation and describe this new field and it's meaning

components/central-application-gateway/internal/proxy/reverseproxy.go
```diff
@@ -39,8 +40,9 @@ func newProxy(targetURL string, requestParameters *authorization.RequestParamete
 		req.Host = target.Host
 
 		combinedPath := joinPaths(target.Path, req.URL.Path)
-		req.URL.RawPath = combinedPath
+		combinedPathE := joinPaths(target.Path, req.URL.RawPath)
```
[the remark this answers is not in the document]
VOID404: Variable name verbosity should be proportional to scope of use, so I didn't want to name it `combinedPathEscaped` considering it's only used in the next line, but I see your point. I'll change it to the long version, but I am open to suggestions.

components/central-application-gateway/internal/proxy/factory_test.go
VOID404: image
In an IDE with typehints it seems quite readable, but I can change it

components/central-application-gateway/internal/proxy/factory.go
VOID404: I'd much rather handle the potential error, but I am not sure how it'd happen...
```go
if err != nil {
    return model.APIIdentifier{}, nil, nil, apperrors.WrongInput("Path has to be a valid path")
}
```

components/central-application-gateway/internal/proxy/factory_test.go
franpog859: Probably yes, but in an IDE without it, it's quite mysterious

components/central-application-gateway/internal/proxy/reverseproxy.go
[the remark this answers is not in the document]
franpog859: True. My point was `combinedPathE` says nothing about why it's there :D By just looking at the way of creating it's logical to name it `combinedRawPath` as it would be consistent with the `42` line. I assume that it's more about the escaping rather than the rawness :D Therefore, `combinedPathEscaped` sounds fine

installation/resources/crds/application-connector/applications.applicationconnector.crd.yaml
VOID404: I did try upgrading from `main` to my PR, with `Application` present and it does get updated with default value for this field (although after some research I don't know if this is the cleanest way to go about it? Should I bump the version?).

franpog859 (approved):
LGTM

grego952 (approved):

PR #16592, opened 2023-01-17
Refactor creating otel collector configuration

chrkl:
**Description**

Refactor the code that generates the otel collector configuration in the telemetry-operator to improve test coverage.

- Decouple the creation of the collector configuration from Kubernetes objects
- Introduce types with yaml marshalling for collector configuration
- Add missing tests for collector processor configuration

**Related issue(s)**

### DIFF
--- a/components/telemetry-operator/controller/tracepipeline/config_builder.go
+++ b/components/telemetry-operator/controller/tracepipeline/config_builder.go
@@ -30,9 +30,14 @@ type OTLPExporterConfig struct {
 	RetryOnFailure RetryOnFailureConfig `yaml:"retry_on_failure,omitempty"`
 }
 
+type LoggingExporterConfig struct {
+	Verbosity string `yaml:"verbosity"`
+}
+
 type ExporterConfig struct {
-	OTLP     OTLPExporterConfig `yaml:"otlp,omitempty"`
-	OTLPHTTP OTLPExporterConfig `yaml:"otlphttp,omitempty"`
+	OTLP     OTLPExporterConfig    `yaml:"otlp,omitempty"`
+	OTLPHTTP OTLPExporterConfig    `yaml:"otlphttp,omitempty"`
+	Logging  LoggingExporterConfig `yaml:"logging,omitempty"`
 }
 
 type ReceiverConfig struct {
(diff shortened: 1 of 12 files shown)


components/telemetry-operator/controller/tracepipeline/config_builder.go
```diff
@@ -0,0 +1,223 @@
...
+	"fmt"
+
+	"github.com/kyma-project/kyma/components/telemetry-operator/apis/telemetry/v1alpha1"
+)
+
+type TlsConfig struct {
+	Insecure bool `yaml:"insecure"`
+}
+
+type OtlpExporterConfig struct {
```
skhalash, line 13: As discussed, let's use all caps for acronyms: `OTLP` and `TLS`

components/telemetry-operator/controller/tracepipeline/config_builder.go
```diff
@@ -0,0 +1,223 @@
...
+	Exporters  ExporterConfig    `yaml:"exporters"`
+	Processors ProcessorsConfig  `yaml:"processors"`
+	Extensions ExtensionsConfig  `yaml:"extensions"`
+	Service    OtlpServiceConfig `yaml:"service"`
+}
+
+func makeReceiverConfig() ReceiverConfig {
+	return ReceiverConfig{
+		OpenCensus: map[string]any{},
+		Otlp: map[string]any{
```
skhalash, line 69: Would it make sense to have a type, for instance, for receivers, exporters, etc? Is there any benefit in having it untyped (as a map[string]any)?

skhalash (changes requested):

components/telemetry-operator/controller/tracepipeline/config_builder.go
skhalash: Ok, let's leave it like this for now.

components/telemetry-operator/controller/tracepipeline/config_builder.go
chrkl: Done

components/telemetry-operator/controller/tracepipeline/config_builder.go
chrkl: Introduced types, except for the expected empty maps.

skhalash (approved):

Issue #16533, opened 2023-01-11
Detecting TLS settings from OTLP endpoint does not work when using Secret reference


chrkl:
**Description**

PR #16357 added a detection for automatically disabling TLS based on the URL scheme of  the OTLP endpoint in a TracePipeline. This detection is only effective when the TracePipeline contains the endpoint value, but not when using a secret reference.

**Expected result**

Using an non TLS connection should work when a TracePipeline consumes a http endpoint from a Secret.

**Actual result**

The scheme is ignored and TLS is used by default.

**Steps to reproduce**

Create Secret with in-cluster Jaeger endpoint:
```
apiVersion: v1
data:
  endpoint: aHR0cDovL3RyYWNpbmctamFlZ2VyLWNvbGxlY3Rvci5reW1hLXN5c3RlbS5zdmMuY2x1c3Rlci5sb2NhbDo0MzE3
kind: Secret
metadata:
  name: jaeger-endpoint
  namespace: default
```

Create TracePipeline for in-cluster Jaeger:

```
apiVersion: telemetry.kyma-project.io/v1alpha1
kind: TracePipeline
metadata:
  name: jaeger
spec:
  output:
    otlp:
      endpoint:
        valueFrom:
          secretKeyRef:
            key: endpoint
            name: jaeger-endpoint
            namespace: default
```

Exporter settings in `telemetry-trace-collector` ConfigMap does not have insecure flag set:
```
    exporters:
        otlp:
            endpoint: ${OTLP_ENDPOINT}
            headers: {}
            retry_on_failure:
                enabled: true
                initial_interval: 5s
                max_elapsed_time: 300s
                max_interval: 30s
            sending_queue:
                enabled: true
                queue_size: 512
            tls:
                insecure: false
```

Otel collector cannot ingest traces to Jaeger:
```
2023-01-11T09:12:03.496Z	warn	zapgrpc/zapgrpc.go:191	[core] [Channel #1 SubChannel #2] grpc: addrConn.createTransport failed to connect to {
  "Addr": "tracing-jaeger-collector.kyma-system.svc.cluster.local:4317",
  "ServerName": "tracing-jaeger-collector.kyma-system.svc.cluster.local:4317",
  "Attributes": null,
  "BalancerAttributes": null,
  "Type": 0,
  "Metadata": null
}. Err: connection error: desc = "transport: Error while dialing dial tcp 100.109.114.12:4317: connect: connection refused"	{"grpc_log": true}
```

**Troubleshooting**

PR #16598, opened 2023-01-18
Fix wrong TLS setting for endpoints defined in a secret

hisarbalik:
**Description**

- TLS insecure should be enabled when endpoint coming from a secret and has a scheme http

**Related issue(s)**
https://github.com/kyma-project/kyma/issues/16533


skhalash (approved):

skhalash (changes requested):
What if the endpoint is hard-coded and not coming from a secret? I think the old implementation has to be combined with a new implementation:
1. If the endpoint is a secret ref, check the secret value
2. Otherwise check the raw endpoint value
Would be also nice to warp into a utility function and cover it with unit tests :)

hisarbalik:
> What if the endpoint is hard-coded and not coming from a secret? I think the old implementation has to be combined with a new implementation:
> 
> 1. If the endpoint is a secret ref, check the secret value
> 2. Otherwise check the raw endpoint value
>    Would be also nice to warp into a utility function and cover it with unit tests :)

endpoint can be either raw value or value from secret, this check happen in the function `fetchSecretData`, test of insecure flag already covered by existing test

skhalash (approved):

PR #16628, opened 2023-01-20
Move protocolsettings checks to the validating webhook

raypinto:
**Description**

- Move QoS, webhook Auth type and grant type validations to the validating webhook 

**Related issue(s)**


components/eventing-controller/pkg/ems/api/events/types/grant_type.go
```diff
@@ -3,3 +3,11 @@ package types
...
 
 const GrantTypeClientCredentials GrantType = "client_credentials"
+
+func IsInvalidGrantType(value string) bool {
+	return value != string(GrantTypeClientCredentials)
+}
+
+func GetGrantType(_ string) GrantType {
+	return GrantTypeClientCredentials
+}
```
marcobebway, line 13: Why do we need this function? We can use the already exported const `GrantTypeClientCredentials`.

components/eventing-controller/pkg/ems/api/events/types/auth_type.go
```diff
@@ -5,3 +5,11 @@ type AuthType string
...
 	AuthTypeClientCredentials AuthType = "oauth2"
 )
+
+func IsInvalidAuthType(value string) bool {
+	return value != string(AuthTypeClientCredentials)
+}
+
+func GetAuthType(_ string) AuthType {
+	return AuthTypeClientCredentials
+}
```
marcobebway, line 15: Why do we need this function? We can use the already exported const `AuthTypeClientCredentials`.

components/eventing-controller/api/v1alpha2/subscription_webhook.go
```diff
@@ -155,6 +170,13 @@ func (s *Subscription) validateSubscriptionSink() *field.Error {
 	return nil
 }
 
+func (s *Subscription) ifKeyExistsInConfig(key string) bool {
+	if _, ok := s.Spec.Config[key]; ok {
+		return true
+	}
+	return false
```
marcobebway, line 177: No need for the if-statement, the following code is enough:

```golang
func (s *Subscription) ifKeyExistsInConfig(key string) bool {
    _, ok := s.Spec.Config[key]
    return ok
}
```

components/eventing-controller/api/v1alpha2/subscription_webhook.go
```diff
@@ -115,11 +117,24 @@ func (s *Subscription) validateSubscriptionTypes() *field.Error {
...
+	if s.ifKeyExistsInConfig(WebhookAuthType) && types.IsInvalidAuthType(s.Spec.Config[WebhookAuthType]) {
+		allErrs = append(allErrs, MakeInvalidFieldError(ConfigPath, s.Name, InvalidAuthTypeErrDetail))
+	}
+	if s.ifKeyExistsInConfig(WebhookAuthGrantType) && types.IsInvalidGrantType(s.Spec.Config[WebhookAuthGrantType]) {
+		allErrs = append(allErrs, MakeInvalidFieldError(ConfigPath, s.Name, InvalidGrantTypeErrDetail))
+	}
+	if len(allErrs) == 0 {
+		return nil
+	}
+	return allErrs
```
marcobebway, line 137: I think we can remove the if-statement so just always return `return allErrs`.

components/eventing-controller/pkg/ems/api/events/types/grant_type.go
raypinto: The idea was to keep it extensible for future. If there are more values tomorrow, we pass the string value as a parameter and then return the `GrantType` value

components/eventing-controller/pkg/ems/api/events/types/auth_type.go
raypinto: The idea was to keep it extensible for future. If there are more values tomorrow, we pass the string value as a parameter and then return the `AuthType` value

components/eventing-controller/api/v1alpha2/subscription_webhook.go
raypinto: that's right! will change

components/eventing-controller/api/v1alpha2/subscription_webhook.go
raypinto: I think that's right! An empty slice would be nil (need to double check it)

raypinto: done

components/eventing-controller/api/v1alpha2/subscription_webhook.go
raypinto: done

marcobebway (approved):

marcobebway (approved):

Issue #16514, opened 2023-01-09
[Telemetry Traces] Filter system traces by default


a-thaler:
**Description**
Having istio trace reporting enabled, by default the kyma-system components are reporting spans repeatedly. That spans are not relevant for any business use case and also not for kyma operations.
For now, let's filter them out hardcoded till there is a better understanding how the API desin will look like.

**Criterias**
- Filter out traces for fluent-bit -> loki communication
- Filter out traces for prometheus and vm_promscrape scraping (reported by the destination potentially)
- Filter out traces targetting the telemetry-otel service (maybe test once more if the span reporting uses a dedicated traceID independent from the initial e2e request chain, if so we can be sure to filter it out)
- document filtering in limitation section
- assure that additional resource consumption is acceptable

**Implementation Idea**
Use the filterprocessor with OTTL. Try to be as specific as possible to not have false positives, include namespace info were possible.
```yaml
    filter:
      traces:
        span:
          - '(attributes["http.method"] == "GET") and (attributes["component"] == "proxy") and (attributes["OperationName"] == "Ingress") and (IsMatch(attributes["http.url"], ".+/metrics") == true)'
          - '(attributes["http.method"] == "GET") and (attributes["component"] == "proxy") and (attributes["OperationName"] == "Ingress") and (IsMatch(attributes["http.url"], ".+/healthz(/.*)?") == true)'
          - '(attributes["http.method"] == "GET") and (attributes["component"] == "proxy") and (attributes["OperationName"] == "Ingress") and (attributes["user_agent"] == "vm_promscrape")'
          - '(attributes["http.method"] == "POST") and (attributes["component"] == "proxy") and (attributes["OperationName"] == "Egress") and (IsMatch(attributes["http.url"], "http(s)?:\\/\\/telemetry-trace-collector\\.kyma-system(\\..*)?:(4318|4317).*") == true)'
          - '(attributes["http.method"] == "POST") and (attributes["component"] == "proxy") and (attributes["OperationName"] == "Ingress") and (resource.attributes["service.name"] == "loki.kyma-system")'
          - '(attributes["http.method"] == "POST") and (attributes["component"] == "proxy") and (attributes["OperationName"] == "Egress") and (resource.attributes["service.name"] == "telemetry-fluent-bit.kyma-system")'
 ```

PR #16634, opened 2023-01-20
Tracing filter processor integration test

hisarbalik:
**Description**

- Integration test for filter processor

**Related issue(s)**
https://github.com/kyma-project/kyma/issues/16514


tests/fast-integration/telemetry-test/suite.js
```diff
@@ -409,6 +411,35 @@ describe('Telemetry Operator', function() {
           await k8sDelete(secondPipeline);
         });
       });
+
+
+      context('Filter Processor', function() {
+        const testApp = loadTestData('tracepipeline-test-app.yaml');
```
chrkl, line 417: We should try to reduce the dependencies to other Kyma components. Let's make it a regular deployment and not a function. Also the chain of different services that are called seems to be not really necessary for this test (just more things that can break and cause flakyness). Maybe just deploy a simple webserver (nginx image or similar) and expose it as a VirtualService.

tests/fast-integration/tracing/client.js
```diff
@@ -1,9 +1,31 @@
...
 const {proxyGrafanaDatasource} = require('../monitoring/client');
 
+const axios = require('axios');
+const https = require('https');
+const httpsAgent = new https.Agent({
+  rejectUnauthorized: false, // curl -k
+});
+axios.defaults.httpsAgent = httpsAgent;
+
+async function getTestAppUrl() {
```
chrkl, line 16: Let's move everything that is related to calling the application to the telemetry-test and only have the helpers to query Jaeger in the tracing module.

pbochynski (dismissed):

tests/fast-integration/telemetry-test/helpers.js
```diff
@@ -109,3 +110,26 @@ function waitForPodWithLabel(
   );
 }
 
+function waitForDeploymentWithLabel(
```
rakesh-garimella, line 113: may be we can move this to here: https://github.com/kyma-project/kyma/blob/main/tests/fast-integration/utils/index.js it can be useful for other tests also.

tests/fast-integration/telemetry-test/suite.js
```diff
@@ -458,6 +481,55 @@ describe('Telemetry Operator', function() {
...
+
+          assert.isFalse(services.data.includes('grafana.kyma-system'), 'spans are present for grafana');
+          assert.isFalse(services.data.includes('jaeger.kyma-system'), 'spans are present for jaeger');
+          assert.isFalse(services.data.includes('telemetry-fluent-bit.kyma-system'),
+              'spans are present for fluent-bit');
+          assert.isFalse(services.data.includes('loki.kyma-system'), 'spans are present for loki');
+        });
+
+        it(`Should delete test setup`, async function() {
+          await k8sApply(testAppIstioCleanup);
```
rakesh-garimella, line 530: can we reuse the variable:` testAppIstioPatch` and update the value. To keep the number of files in testdata under control.

tests/fast-integration/telemetry-test/helpers.js
hisarbalik: moved to the utils

tests/fast-integration/telemetry-test/suite.js
hisarbalik: istio cleanup resource removed

rakesh-garimella (approved):

PR #16642, opened 2023-01-23
Improve webhook logs

pPrecel:
**Description**

- Add most basic logs for all serverless webhooks to easily track eventual request lifecycle

### DIFF
--- a/components/function-controller/internal/webhook/conversion_webhook.go
+++ b/components/function-controller/internal/webhook/conversion_webhook.go
@@ -47,17 +47,21 @@ func NewConvertingWebhook(client ctrlclient.Client, scheme *runtime.Scheme, log 
 }
 
 func (w *ConvertingWebhook) ServeHTTP(resp http.ResponseWriter, req *http.Request) {
+	log := w.log
 	convertReview := &apix.ConversionReview{}
 	err := json.NewDecoder(req.Body).Decode(convertReview)
 	if err != nil {
-		w.log.Error(err, "failed to read conversion request")
+		log.Error(err, "failed to read conversion request")
 		resp.WriteHeader(http.StatusBadRequest)
 		return
 	}
 
+	log = w.log.With("uid", convertReview.Request.UID)
+	log.Debug("strting conversion")
+
 	conversionResponse, err := w.handleConvertRequest(convertReview.Request)
 	if err != nil {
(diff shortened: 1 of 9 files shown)


components/function-controller/internal/webhook/defaulting_webhook.go
```diff
@@ -21,23 +22,34 @@ type DefaultingWebHook struct {
...
 		configAlphaV1: configV1Alpha1,
 		configAlphaV2: configV1Alpha2,
 		client:        client,
+		log:           log,
 	}
 }
 
 func (w *DefaultingWebHook) Handle(_ context.Context, req admission.Request) admission.Response {
+	log := w.log.With("name", req.Name, "namespace", req.Namespace, "kind", req.Kind.Kind)
+	log.Debug("strting conversion")
```
dbadura, line 39: starting

components/function-controller/internal/webhook/conversion_webhook.go
```diff
@@ -47,17 +47,21 @@ func NewConvertingWebhook(client ctrlclient.Client, scheme *runtime.Scheme, log
...
 	convertReview := &apix.ConversionReview{}
 	err := json.NewDecoder(req.Body).Decode(convertReview)
 	if err != nil {
-		w.log.Error(err, "failed to read conversion request")
+		log.Error(err, "failed to read conversion request")
 		resp.WriteHeader(http.StatusBadRequest)
 		return
 	}
 
+	log = w.log.With("uid", convertReview.Request.UID)
```
dbadura, line 59: Do we need to know the uid of request?
Can we get the name of converted resource?

pPrecel: I set up the logger with UID because this is the only info from the request struct that informs us from which request this log is (to help with cases when we would have a few requests at the same time). What do you think?

components/function-controller/internal/webhook/conversion_webhook.go
```diff
@@ -47,17 +47,21 @@ func NewConvertingWebhook(client ctrlclient.Client, scheme *runtime.Scheme, log
...
+	log.Debug("strting conversion")
```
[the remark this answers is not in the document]
pPrecel: What would be the pros of this change?

dbadura (approved):

PR #16659, opened 2023-01-24
Submit Keda Module version 0.0.1-6cd5086 to alpha channel

kwiatekus:
**Description**

- promote latest version of keda module to alpha channel to include latest fixes


kwiatekus:
> Unfortunately I have to request a change here. While testing I have noticed that you use default kustomisations which causes the manager to be named `operator-controller-manager` while the actual keda deployment is named `keda-manager`.
> 
> I think we should change: `operator-controller-manager` => `keda-manager` `keda-manager` => `keda`
We have a separate issue for that. I'd like to ask you to accept it for now
https://github.com/kyma-project/keda-manager/issues/7

jakobmoellersap (approved):

Issue #16570, opened 2023-01-16
[Telemetry Operator] Finalize fluent-bit managed by operator


a-thaler:
**Description**
As part of https://github.com/kyma-project/kyma/issues/15955 the kyma fluent-bit instances are manageable by code already but the feature is not activated yet and will be still realized via a helm chart. Finalize the efforts by testing the migration fully and fixing leftovers.

**Criterias:**
- All resources of the old fluent-bit sub-chart are either managed by code (new telemetry-operator functionality), are moved to the umbrella telemetry chart (and will be managed by old logic) or are deleted (automated or manual migration script?)
    - Potential resources of interest are: Grafana Dashboards, ServiceMonitors, PeerAuthentication
    - Agreement for ServiceAccount is to also manage it by the operator with least privileges (remember to limit the namespace). It includes the ServiceAccount, a Role and a RoleBinding
- Feature flag got removed
- temporary integration test got merged into regular test suite
- Old fluent-bit sub-chart does not exist anymore
- Migration/Upgrade from helm chart approach to managed by code approach is tested
- Try to verify that automated upgrade tests will not fail
- add missing test cases
  - no logpipeline reveals in no fluentbit 
  - deletion of logpipelines will remove fluentbit when reaching an amount pipelines of 0

**Reasons**

**Attachments**

rakesh-garimella:
Following tasks have been performed:

- Enable managed fluent-bit and removed the feature flag
- Deleted fluent bit charts
- Moved servicemonitor, peerauthentication and dashboards in flunet-bit folder
- Changed the logic to have fluent bit daemonset created only after the checksum is calculated.
- Fixed deletion logic
- Restricted the rights of telemetry-operator

The automated tests will be updated via a separated PR

rakesh-garimella:
PR for automated tests have been added here: https://github.com/kyma-project/kyma/pull/16726/files

a-thaler:
production profile is ignored, will re-open till it is respected

a-thaler:
The containerPort 2021 is exposed by both containers, fluentbit and directory-size-exporter. WHich one will be used? I'm very sure that fluentbit should expose only the 2020 port and that this must be a mistake, please verify that

a-thaler:
The generated luascript is broken.
The line `local mapped_key = string.gsub(key, "[%/%.]", "_")` contains escaping characters which breaks fluentbit startup `local mapped_key = string.gsub(key, \"[%/%.]\", \"_\")`

[...]

PR #16668, opened 2023-01-26
Enable managed fluent-bit

rakesh-garimella:
**Description**

- Enable managed fluent-bit and removed the feature flag
- Deleted fluent bit charts
- Moved servicemonitor, peerauthentication and dashboards in `managed-flunet-bit` folder
- Changed the logic to have fluent bit daemonset created only after the checksum is calculated.
- Changed the secret roles to be restrict controller to read secrets from all namespaces and write to only `kyma-system`
- Delete the finalizer after the fluent-bit is reconciled as it may lead to race conditions sometimes where api server might remove the CR before we validate if fluent-bit daemon set should be deleted. And cause failure of deletion of fluent-bit resources when last pipeline is deleted.
- Restricted the rights of telemetry-operator
- Delete flies and sections configmaps as well

**Related issue(s)**

[16570](https://github.com/kyma-project/kyma/issues/16570)


components/telemetry-operator/controller/logpipeline/controller_test.go
```diff
@@ -305,7 +252,7 @@ var _ = Describe("LogPipeline controller", func() {
 
 			Expect(k8sClient.Delete(ctx, logPipeline)).Should(Succeed())
 
-			// Fluent Bit daemon set should rollout-restarted (generation changes from 1 to 2)
+			// Fluent Bit daemon set should rollout-restarted (generation is 1)
```
chrkl, line 255: This comment seem to be wrong now. We validate that a DaemonSet has been created, right?

components/telemetry-operator/internal/kubernetes/daemonset.go
```diff
@@ -19,6 +20,11 @@ func (dsp *DaemonSetProber) IsReady(ctx context.Context, name types.NamespacedNa
 
 	var ds appsv1.DaemonSet
 	if err := dsp.Get(ctx, name, &ds); err != nil {
+		if apierrors.IsNotFound(err) {
+			// The status change of loki happens before the creation of daemonset
```
chrkl, line 24: This is not loki specific, but would happen for any LogPipeline, right?

components/telemetry-operator/internal/resources/logpipeline/resources_test.go
```diff
@@ -52,3 +53,34 @@ func TestMakeLuaConfigMap(t *testing.T) {
...
+func TestMakeServiceAccount(t *testing.T) {
+	name := types.NamespacedName{Name: "telemetry-fluent-bit", Namespace: "kyma-system"}
+	svcAcc := MakeServiceAccount(name)
+
+	require.NotNil(t, svcAcc)
+	require.Equal(t, svcAcc.Name, name.Name)
+	require.Equal(t, svcAcc.Namespace, name.Namespace)
+}
+
+func TestMakeClusterRole(t *testing.T) {
```
chrkl, line 66: Let's validate the rules of the ClusterRole as well.

components/telemetry-operator/internal/kubernetes/utils.go
```diff
@@ -141,6 +193,36 @@ func DeleteFluentBit(ctx context.Context, c client.Client, name types.Namespaced
...
+
+	var clusterRole v1.ClusterRole
+	err = c.Get(ctx, name, &clusterRole)
+	if err == nil {
+		if err = c.Delete(ctx, &clusterRole); err != nil && !errors.IsNotFound(err) {
+			return err
+		}
+	} else if !errors.IsNotFound(err) {
+		return err
+	}
```
skhalash, line 224: This function gets too long and repetitive. Can we maybe somehow extract the deletion and reuse it?

components/telemetry-operator/internal/kubernetes/utils_test.go
```diff
@@ -162,3 +164,61 @@ func TestMergeChecksumAnnotations(t *testing.T) {
...
+	mockClient.On("Get", mock.Anything, mock.Anything, mock.Anything).Return(notFoundErr)
+	mockClient.On("Create", mock.Anything, mock.Anything).Return(nil)
+
+	resServiceAcc := corev1.ServiceAccount{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "foo",
+			Namespace: "foo-namespace",
+		},
+	}
+	err := CreateOrUpdateServiceAccount(context.Background(), mockClient, &resServiceAcc)
```
skhalash, line 221: Just wondering - what are we testing here?

components/telemetry-operator/internal/resources/logpipeline/resources.go
```diff
@@ -12,7 +13,46 @@ import (
...
+	resClusterRoleBinding := v1.ClusterRoleBinding{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      name.Name,
+			Namespace: name.Namespace,
+		},
+		Subjects: []v1.Subject{{Name: "telemetry-fluent-bit", Namespace: "kyma-system", Kind: "ServiceAccount"}},
+		RoleRef: v1.RoleRef{
+			APIGroup: "rbac.authorization.k8s.io",
+			Kind:     "ClusterRole",
+			Name:     "telemetry-fluent-bit",
```
skhalash, line 38: We should use `name.Name` instead

resources/telemetry/charts/operator/values.yaml
```diff
@@ -47,6 +47,25 @@ readinessProbe:
   initialDelaySeconds: 5
   periodSeconds: 10
 
+managedfluentbit:
```
skhalash, line 50: Why do we need a `managed` prefix? Isn't it implicitly clear? (I also means the subchart name)

skhalash (changes requested):

components/telemetry-operator/controller/logpipeline/controller_test.go
rakesh-garimella: In the code it validating the generation but yeah gen 1 means that it has been created and not updated

components/telemetry-operator/internal/kubernetes/utils_test.go
rakesh-garimella: Yeah true everything is mocked will remove this

components/telemetry-operator/internal/resources/logpipeline/resources.go
rakesh-garimella: done

components/telemetry-operator/internal/kubernetes/utils_test.go
chrkl: Switching to the fake client from controller runtime might make this test a bit more meaningful.

skhalash: No-no, it might make sense to test that either `Create` or `Update` is called on the mock.

skhalash: @chrkl Fully agree

chrkl (approved):

skhalash (approved):

Issue #15618, opened 2022-09-23
Implement new publishing logic in event-publisher-proxy (EPP)


raypinto:
Tasks:

- [ ] ~~Implement new version for publishing (v2) using cloud events (CE)~~
- [x] Support the older version of the CRD as well for CE and update the logic for the new version
- [x] Update the logic for the existing Legacy endpoint, without changing the endpoint or the response.

Note: EPP would still support the application lister.

raypinto:
Changes expected:

- [convertPublishRequestToCloudEvent](https://github.com/kyma-project/kyma/blob/a7bb098cee1df9ab59a90eef6f324269ada0f244/components/event-publisher-proxy/pkg/legacy/legacy.go#L159)
- [add new publish endpoint for v1alpha2](https://github.com/kyma-project/kyma/blob/a7bb098cee1df9ab59a90eef6f324269ada0f244/components/event-publisher-proxy/pkg/handler/handler.go#L80)
- [modify subscribed endpoint](https://github.com/kyma-project/kyma/blob/a7bb098cee1df9ab59a90eef6f324269ada0f244/components/event-publisher-proxy/pkg/handler/handler.go#L82)

raypinto:
> Note: Here `prefix` stands for `sap.kyma.custom`

| Scenario      | Subscription |  Backend |   Publish |
| ----------- | ----------- | ----------- | ----------- |
| JS v1alpha1      | source: ""  <br/> et: “prefix.noapp.order.created.v1”      | kyma.prefix.noapp.order.created.v1 | source: does not matter <br/>  et: “prefix.noapp.order.created.v1” |
| JS v1alpha2 - standard     | source: "test"  <br/> et: “order.created.v1”      | kyma.test.order.created.v1 | source: "test" <br/>  et: “order.created.v1” |
| JS v1alpha2 - exact <br/> (Should not be removed as soon as we deprecate v1alpha1)     | source: does not matter  <br/> et: “order.created.v1”      | kyma.order.created.v1 | source: does not matter <br/>  et: “order.created.v1” |
| JS upgrade     | source: "" <br/> et: “prefix.noapp.order.created.v1”      | kyma.prefix.noapp.order.created.v1 | source: does not matter <br/>  et: “prefix.noapp.order.created.v1” |
| | | |
[...]

PR #16671, opened 2023-01-26
validate source for valid URI

raypinto:
**Description**

- In the EPP validate the cloud event after building it
- In the EC, validate that the source provided is accepted by the CE Validator

**Related issue(s)**


mfaizanse:
image bumps?

components/event-publisher-proxy/pkg/cloudevents/builder/generic_test.go
```diff
@@ -76,6 +84,13 @@ func Test_Build(t *testing.T) {
 			givenApplicationName: "appName1",
 			wantError:            true,
 		},
+		{
+			name:                 "should return error if empty source",
+			givenSource:          "",
+			givenType:            "order.created.v1",
+			givenApplicationName: "appName1",
+			wantError:            true,
+		},
```
mfaizanse, line 93: Is this only for v1alpha2 CRD? Also what if in case of `exact`? As in some cases source does not matter. ([reference](https://github.com/kyma-project/kyma/issues/15618#issuecomment-1310364282))

components/eventing-controller/api/v1alpha2/errors.go
```diff
@@ -18,6 +18,7 @@ var (
 	NSPath     = field.NewPath("metadata").Child("namespace")
 
 	EmptyErrDetail          = "must not be empty"
+	InvalidURIErrDetail     = "must be a valid URI reference"
```
mfaizanse, line 21: Are we asking user to provide URI reference? We are mostly expecting just a string as source? Maybe it can cause confusion for the user?

components/event-publisher-proxy/pkg/cloudevents/builder/generic_test.go
raypinto: it is only for v1alpha2 jetstream. There is no `exact` in EPP right?

components/eventing-controller/api/v1alpha2/errors.go
raypinto: a URI reference is also a string with certain limitations. The source should be a valid URI reference. That's how the error will look like.

components/eventing-controller/utils/utils.go
```diff
@@ -95,3 +100,10 @@ func GetSinkData(sink string) (string, []string, error) {
 	subDomains := strings.Split(trimmedHost, ".")
 	return trimmedHost, subDomains, nil
 }
+
+func GetCloudEvent() event.Event {
+	newEvent := cloudevents.NewEvent()
+	newEvent.SetType(GetRandString(randStringlength))
+	newEvent.SetID(GetRandString(randStringlength))
+	return newEvent
+}
```
mfaizanse, line 109: (nit) Its setting a random type in CE, maybe mention it in name e.g. `GetCloudEventWithRandType()`?

components/event-publisher-proxy/pkg/cloudevents/builder/generic_test.go
raypinto: As discussed, the `exact` flow in EPP does not go through this function.

components/eventing-controller/api/v1alpha2/errors.go
raypinto: Agreed to string: `must be a valid source as per RFC 3986`

components/eventing-controller/utils/utils.go
raypinto: Agreed to provide actual type from the subscription instead of a random one

mfaizanse (approved):

Issue #15919, opened 2022-10-27
Delete PSP from Eventing components for k8s 1.25


mfaizanse:
**Description**

Since PSP will be removed with k8s 1.25, remove it from following kyma components:
- Eventing controller
- Event-publisher-proxy
- NATS
- KMC

More info: https://github.com/kyma-project/kyma/issues/15897

marcobebway:
The following security context config is applied for all Eventing Pods/Containers:

```yaml
  podSecurityContext:
    fsGroup: 10001
    runAsUser: 10001
    runAsGroup: 10001
    runAsNonRoot: true
    seccompProfile:
      type: RuntimeDefault
    sysctls:
      - name: kernel.shm_rmid_forced
        value: ''
      - name: net.ipv4.ip_local_port_range
        value: ''
      - name: net.ipv4.ip_unprivileged_port_start
        value: ''
      - name: net.ipv4.tcp_syncookies
        value: ''
      - name: net.ipv4.ping_group_range
        value: ''
```

```yaml
  containerSecurityContext:
    privileged: false
    allowPrivilegeEscalation: false
    runAsNonRoot: true
    capabilities:
      drop:
        - ALL
      add:
        - NET_BIND_SERVICE
    seLinuxOptions:
      type: container_t
    procMount: default
```

marcobebway:
The following security context config is applied for all KMC Pods/Containers:

```yaml
podSecurityContext:
  fsGroup: 10001
  runAsUser: 10001
  runAsGroup: 10001
  runAsNonRoot: true
  seccompProfile:
    type: RuntimeDefault
```

```yaml
containerSecurityContext:
  privileged: false
  allowPrivilegeEscalation: false
  runAsNonRoot: true
  capabilities:
    drop:
      - ALL
    add:
      - NET_BIND_SERVICE
  seLinuxOptions:
    type: container_t
  procMount: default
```

mfaizanse:
- Disabled SecurityContext for NATS in this PR: https://github.com/kyma-project/kyma/pull/16728
- Because the k3d upgrade job was failing.
    - https://status.build.kyma-project.io/?job=kyma-upgrade-k3d-kyma2-to-main
    - https://status.build.kyma-project.io/view/gs/kyma-prow-logs/logs/kyma-upgrade-k3d-kyma2-to-main/1621438209999245312

- Error logs from NATS:

```
## kubectl logs -n kyma-system eventing-nats-0 -c nats
[...]

PR #16681, opened 2023-01-27
Configure PSA for Eventing components

marcobebway:
**Description**

Configure PSA for the following Eventing components:

- NATS.
- Eventing controller.
- Eventing publisher.

**Related issue(s)**

- https://github.com/kyma-project/kyma/issues/15919

### DIFF
--- a/resources/eventing/values.yaml
+++ b/resources/eventing/values.yaml
@@ -5,7 +5,7 @@ global:
   images:
     eventing_controller:
       name: eventing-controller
-      version: PR-16646
+      version: PR-16681
       pullPolicy: "IfNotPresent"
     publisher_proxy:
       name: event-publisher-proxy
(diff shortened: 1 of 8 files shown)


resources/eventing/values.yaml
```diff
@@ -45,9 +45,37 @@ global:
...
+      - name: net.ipv4.ip_local_port_range
+        value: ''
+      - name: net.ipv4.ip_unprivileged_port_start
+        value: ''
+      - name: net.ipv4.tcp_syncookies
+        value: ''
+      - name: net.ipv4.ping_group_range
+        value: ''
+
+  containerSecurityContext:
```
mfaizanse, line 67: Is NATS already running as non-root user? would now it would work with all NATS commands? storage etc.

marcobebway: Yes, it is verified offline.

mfaizanse (approved):

Issue #16664, opened 2023-01-25
Published Event Types Summary widget in NATS Delivery grafana dashboard is not working as expected for Subscription v1alpha2


mfaizanse:
![image](https://user-images.githubusercontent.com/9897945/214608570-c14af101-7344-498b-a1a2-bb67019993f2.png)

The Published Event Types Summary widget is not updating the event types for v1alpha2 subscription as `Published Event Types with a Subscription` even if they are part of a subscription.

PR #16720, opened 2023-02-02
Fix Eventing Metrics for new CRD

muralov:
**Description**
Use original event type without event type prefix for EPP published event total metric

**Related issue(s)**
#16664

### DIFF
--- a/components/event-publisher-proxy/pkg/handler/handler.go
+++ b/components/event-publisher-proxy/pkg/handler/handler.go
@@ -3,6 +3,7 @@ package handler
 import (
 	"context"
 	"errors"
+	"fmt"
 	"net/http"
 	"strings"
 	"time"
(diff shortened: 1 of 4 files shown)


components/event-publisher-proxy/pkg/handler/handler.go
```diff
@@ -317,7 +318,12 @@ func (h *Handler) sendEventAndRecordMetrics(ctx context.Context, event *cev2even
 		h.collector.RecordBackendError()
 		return nil, err
 	}
-	h.collector.RecordEventType(event.Type(), event.Source(), result.HTTPStatus())
+	originalEventType, ok := event.Extensions()[builder.OriginalTypeHeaderName].(string)
+	if !ok {
+		h.namedLogger().With().Errorw("failed to convert event original event type extension value to string", builder.OriginalTypeHeaderName, event.Extensions()[builder.OriginalTypeHeaderName])
+		return nil, fmt.Errorf("failed to convert event extension %s value %d to string", builder.OriginalTypeHeaderName, event.Extensions()[builder.OriginalTypeHeaderName])
```
mfaizanse, line 324: 1. please also check if its not empty value.
2. Instead of returning error, why dont we log a warning and use the prefixed type like:
```
if !ok && originalEventType == "" {
    h.namedLogger().With().Warningw("failed to convert event original event type extension value to string", builder.OriginalTypeHeaderName, event.Extensions()[builder.OriginalTypeHeaderName])

    originalEventType = event.Type()
}
```

mfaizanse: and also please add unit tests.

muralov: done

mfaizanse (approved):

Issue #16847, opened 2023-02-20
Update HPA configuration in Istio charts


cnvergence:
**Description**
The autoscaling/v2beta1 API version of HorizontalPodAutoscaler is no longer served as of v1.25 Kubernetes.
We will need to align our Helm chart configuration and migrate it to autoscaling/v2.
Example config change:
```bash
      hpaSpec:
        maxReplicas: 10
        minReplicas: 3
        metrics:
        - resource:
            name: cpu
            target:
              type: Utilization
              averageUtilization: 80
          type: Resource
        - resource:
            name: memory
            target:
              type: Utilization
              averageUtilization: 80
          type: Resource
```
autoscaling/v2 version is available since K8s 1.23

**Attachments**

- https://kubernetes.io/docs/reference/using-api/deprecation-guide/#horizontalpodautoscaler-v125

PR #16848, opened 2023-02-20
Update Istio HPA to autoscaling/v2

cnvergence:
- update hpa to autoscaling/v2
- remove hpa config from cni, because it does not exist

**Related issue(s)**

See also:
- https://github.com/kyma-project/kyma/issues/16847

### DIFF
--- a/resources/istio/values.yaml
+++ b/resources/istio/values.yaml
@@ -1,7 +1,6 @@
 ---
 helmValues:
   global:
-    autoscalingv2API: true
     priorityClassName: "kyma-system-priority"
     imagePullPolicy: IfNotPresent
     proxy:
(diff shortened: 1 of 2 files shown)


cnvergence:
Hold for now

resources/istio/values.yaml
```diff
@@ -1,6 +1,7 @@
 ---
 helmValues:
   global:
+    autoscalingv2API: true
```
triffer, line 4: We shouldn't add this as this is marked for internal usage only ([ref](https://github.com/istio/istio/pull/37872#discussion_r847543725)). This field is set depending on the configuration in the IOP.

cnvergence: Good point, confirmed it is working

triffer (approved):

PR #16856, opened 2023-02-21
Replace usernames with a team name in CODEOWNERS

kasiakepka:
**Description**

- Replace usernames with a team name
**Related issue(s)**


a-thaler:
I recommend to use group name similar to the domain/are the people are responsible for

kasiakepka:
Thanks Andreas, we will consider that but I am afraid that we have to wait for Magda (till Monday) before we make final decision.

pbochynski (approved):

PR #16890, opened 2023-02-23
Fix nodejs16 returned value formatting regression

### LINKED ISSUE #16860: Serverless Functions: Node16 runtime with server.js framework breaks return values handling/formatting
Related to https://github.com/kyma-project/kyma/issues/16673

**Description**

With nodejs base serverless the return values formatting is not repespected when the return value is passed to the framework

1. returnJSON.stringify(jsonData, 0, 2) will appear exactly as returnJSON.stringify(jsonData) - no formatting
2. returnJSON.stringify(jsonData) and return jsonData yields the same result
3. return xmlData will yield unpredictable result if the xml is quite voluminous or otherwise the result in formatted in a strange way

**Expected result**

The only workaround is to revert to nodejs14 with the kubeless.js framework

**Actual result**

Both sever.js and  kubeless.js  frameworks should handle the return values the same way.

**Steps to reproduce**
use some jsonData for testing:

use some xmlData for testing:

**Troubleshooting**

kwiatekus:
**Related issue(s)**
Fixes #16860 
Includes https://github.com/kyma-project/kyma/pull/16870/files

### DIFF
--- a/resources/serverless/values.yaml
+++ b/resources/serverless/values.yaml
@@ -92,7 +92,7 @@ global:
       version: "v20230202-93374cde"
     function_runtime_nodejs16:
       name: "function-runtime-nodejs16"
-      version: "v20230202-93374cde"
+      version: "v20230223-4c2a2fb9"
     function_runtime_python39:
       name: "function-runtime-python39"
       version: "v20230202-93374cde"


m00g3n (approved):
LGTM

Issue #16981, opened 2023-03-03
regression in serverless after 2.11 upgrade. telemetry error


remi-sap:
**Description**

Hello, shortly after 2.11 upgrade, newly built python serverless function report plenty of errors such as:

<pre>HTTPConnectionPool(host='telemetry-otlp-traces.kyma-system.svc.cluster.local', port=4318): 
Max retries exceeded with url: /v1/traces 
(Caused by NewConnectionError('<urllib3.connection.HTTPConnection object at 0x7ff663568e20>: 
Failed to establish a new connection: [Errno -2] Name does not resolve'))
</pre>

We can't exploit the function traces as they are completely polluted by those messages.
I don't see a service called **telemetry-otlp-traces** in the kyma-system namespace. How can I disable telemetry or add the service to fix those errors?

**Steps to reproduce**
 Build a new serverless python function and call it.

a-thaler:
Hi @remi-sap,
thanks for the report, we will look into it and see how to solve the problem and workaround it

a-thaler:
Hi @remi-sap, we can reproduce the problem and investigating on this. As a workaround you could define a [TracePipeline](https://kyma-project.io/docs/kyma/latest/01-overview/main-areas/telemetry/telemetry-03-traces/#step-1-create-a-trace-pipeline-with-an-output) which will dynamicaaly provision an otel-collector and serve the missing endpoint. The TracePipeline could point to a dumy endpoint even causing the otel-collector to drop the data at some point

kwiatekus:
This was happening because in case any spans (manually created or resulting from auto-instrumentation) were pushed by the exporter to a non-existing collector endpoint.

This was mitigated by removing default auto-instrumentation for outgoing requests.
Now, if users are interested in observing traces (and have trace pipeline configured) they will need to also enable the instrumentation for requests in the function code, i.e:

```py
import requests

from opentelemetry.instrumentation.requests import RequestsInstrumentor

def main(event, context):
    
    RequestsInstrumentor().instrument()

    print("doing some work...")
    r = requests.get('https://swapi.dev/api/people/2')
    return r.json()
```

A change in documentation will follow : https://github.com/kyma-project/kyma/issues/17021

remi-sap:
What's the easiest way to get observability working in the current version 2.11.2? 
The doc for version 2.11 mentions the service telemetry-otlp-traces which doesn't exist, either the doc is wrong or I should somehow activate the service? Is observability broken until version 2.12?

[...]

PR #16990, opened 2023-03-03
Exclude default `request` auto-instrumentation in python functions

kwiatekus:
**Description**

- Do not auto-instrument `request` library by default as it causes error logs in case trace pipeline is missing 
- Use latest opentelemetry SDK version for python
- configure sampling strategy

**Related issue(s)**
https://github.com/kyma-project/kyma/issues/16981

### DIFF
--- a/components/function-runtimes/python39/kubeless/tracing.py
+++ b/components/function-runtimes/python39/kubeless/tracing.py
@@ -1,57 +1,45 @@
-import logging
 from contextlib import contextmanager
 from typing import Iterator
 
-import requests
 from opentelemetry import trace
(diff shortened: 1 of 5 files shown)


components/function-runtimes/python39/kubeless/tracing.py
```diff
@@ -74,4 +60,4 @@ def set_req_context(req) -> Iterator[trace.Span]:
     try:
         yield span
     finally:
-        context_api.detach(token)
+        context_api.detach(token)
```
pPrecel, line 63: 🥺 
```suggestion
        context_api.detach(token)
        
```

pPrecel (approved):

PR #17009, opened 2023-03-07
Add target to run tests as binary

dbadura:
**Description**

- run tests as binary

**Related issue(s)**
https://github.com/kyma-project/serverless-manager/issues/44


tests/function-controller/run-local.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+kubectl proxy &
+printenv  | grep DAMIAN
```
anoipm, line 3: `grep DAMIAN` ???

dbadura: Fixed :)

anoipm (approved):

Issue #13697, opened 2022-03-23
Use Exponential Message Delivery Backoff in JetStream dispatcher


mfaizanse:
**Pre-requisites**
- JetStream supports [Exponential Message Delivery Backoff](https://github.com/nats-io/nats-server/issues/2042)
- Exponential backoff or `msg.NakWithDelay(time.Second * 10)` feature is released in https://github.com/nats-io/nats.go

**Description**

Currently, the JetStream dispatcher will not NAK the msg if it fails to deliver it to the sink, so that it timeouts after AckWait period and then redeliver. It will keep on trying for some time with a constant backoff. 

Once Exponential Message Delivery Backoff is supported by JetStream, we should use the exponential backoff in our JetStream dispatcher instead of constant backoff.

- [ ] configurable maximum redelivery
- [ ]  configurable delivery timeout

**Attachments**

Related issue: https://github.com/kyma-project/kyma/issues/13625

VladislavPaskar:
seems that the backoff functionality was already introduced in the v2.7.1: https://github.com/nats-io/nats-server/releases/tag/v2.7.1,

see the comment in the [original NATS issue](https://github.com/nats-io/nats-server/issues/2042)

grischperl:
Now that JetStream supports both exponential `backoff` and `NakWithDelay`, it was decided to use the latter as the EC will respond with a Nak. The msg is put back in the queue and waits previously configured time (30sec) until the msg is send again (max. 100 retries).

Additionally, in case `NakWithDelay` fails and no response is sent, `AckWait` is configured to redeliver every 30sec.
#17102

PR #17102, opened 2023-03-16
NakWithDelay Message Delivery in JetStream

grischperl:
**Description**
Instead of using an exponential `Backoff` where the EC does not respond, this PR changes the currently used `AckWait` (EC also does not respond) to a `NakWithDelay`. 

This improves the memory management compared to `AckWait`. `NakWithDelay` results in a consistent Stream Data Size at about 100 KiB (when sending 2000 events). `AckWait` often spikes up to 300-350 KiB.

- EC responds with a Nak
- msg is put back in the queue and waits previously configured time (30sec) until the msg is send again (max. 100 retries)

**Related issue(s)**

#13697


grischperl:
**Stream Data Size with AckWait:**
_1 Consumer:_
    1 Event: 476 B
    10 Events: 4.65 KiB
    20 Events: 9.30 KiB
    100 Events: 46.5 KiB
    200 Events: 93.0 KiB
    
    after 10min:
    Pending msg: 170
    Pending re-delivery msg: 30
    
**Stream Data Size with NakWithDelay:**
_1 Consumer:_
    1 Event: 476 B
    10 Events: 4.65
    20 Events: 9.30 KiB
    100 Events: 46.5 KiB
    200 Events: 93.0 KiB
    
    after 10min:
    Pending msg: 190
    Pending re-delivery msg: 10

grischperl:
Loadtester: 1500 events 

- target eps = ack
- undelivered = 0
- nack = 0
- after loadtest-stop msg are successfully redelivered

components/eventing-controller/pkg/backend/nats/jetstream/jetstream.go
```diff
@@ -651,7 +651,6 @@ func (js *JetStream) getDefaultSubscriptionOptions(consumer SubscriptionSubjectI
 		toJetStreamConsumerDeliverPolicyOptOrDefault(js.Config.JSConsumerDeliverPolicy),
 		nats.MaxAckPending(subConfig.MaxInFlightMessages),
 		nats.MaxDeliver(jsConsumerMaxRedeliver),
-		nats.AckWait(jsConsumerAcKWait),
```
muralov, line 654: what do you think should we not keep this in case the call `msg.NakWithDelay(jsConsumerNakDelay)` fails?
if the nack function call fails for any reason, it will not be redelivered anymore.

components/eventing-controller/pkg/backend/jetstreamv2/utils.go
```diff
@@ -47,7 +47,6 @@ func (js *JetStream) getDefaultSubscriptionOptions(consumer SubscriptionSubjectI
 		toJetStreamConsumerDeliverPolicyOptOrDefault(js.Config.JSConsumerDeliverPolicy),
 		nats.MaxAckPending(maxInFlightMessages),
 		nats.MaxDeliver(jsConsumerMaxRedeliver),
-		nats.AckWait(jsConsumerAcKWait),
```
muralov, line 50: Should we not keep `ackWait` in case the call `msg.NakWithDelay(jsConsumerNakDelay)` fails?
if the nack function call fails for any reason, it will not be redelivered anymore.

grischperl: done

muralov (approved):

PR #17190, opened 2023-03-28
add info on Kyma import certs

### LINKED ISSUE #16871: Kyma API Rule exposed cannot be opened in browser due to invalid certificate presented.
**Description**

Exposing API through the gateway and trying to access it through browser seems that it does not work because of maybe invalid certificate.

**Expected result**

Expose service through API Rules and being able to access it through browser even if the https certificate is a self signed one

**Actual result**

Endpoint cannot be accessed within browser because it is blocked.
See: 
image
image

**Steps to reproduce**

1. Install kyma locally.
2. Deploy function or microservice.
3. Create API Rule. Something like this:
```yaml
apiVersion: gateway.kyma-project.io/v1alpha1
kind: APIRule
metadata:
  labels:
    app: commerce-mock
  name: commerce-mock
spec:
  gateway: kyma-gateway.kyma-system.svc.cluster.local
  service:
    name: commerce-mock
    port: 10000
    host: commerce
  rules:
    - path: /.*
      methods: ["GET","POST","PUT","DELETE","HEAD"]
      accessStrategies:
        - handler: allow
 ```

**Troubleshooting**

I did not followed any steps to solve this issue, but I guess there might be some documentation about how to create your self signed certificates locally that might work.

grego952:
**Description**

- Add a Note in Deploy and expose a Function about accessing API through the browser.

**Related issue(s)**
Resolves #16871

### DIFF
--- a/docs/02-get-started/02-deploy-expose-function.md
+++ b/docs/02-get-started/02-deploy-expose-function.md
@@ -174,3 +174,6 @@ The operation was successful if the call returns `Hello Serverless`.
 
   </details>
 </div>
+
+> **NOTE:** Local installation provides the self-signed certificates out of the box, but if you want to access your API through your browser, you must add them to your local trust store. 
+To do this, call the `kyma import certs` command with proper permissions. For more information, read [Kyma import certs](https://github.com/kyma-project/cli/blob/main/docs/gen-docs/kyma_import_certs.md). 


grego952:
@a-thaler, Can you look at this note about Kyma import certs?
Thanks!

a-thaler (approved):

a-thaler:
Sounds good, approved

nataliasitko (approved):

PR #17218, opened 2023-03-31
Api-gateway 1.5.0

barchw:
**Description**

- Update api-gateway to release 1.5.0
- Add rbac for Pods

**API Gateway 1.5.0 changelog**
- https://github.com/kyma-project/api-gateway/releases/tag/1.5.0


triffer (approved):

videlov:
@barchw should we also change version in here: `resources/api-gateway/Chart.yaml`

triffer (approved):

resources/eventing/values.yaml
```diff
@@ -12,7 +12,7 @@ global:
       version: PR-17124
     certHandler:
       name: api-gateway-webhook-certificates
-      version: "1.4.1"
+      version: "1.5.0"
```
triffer, line 15: We should not bump image in eventing, but rather inform them, that there is a new version available.

triffer (changes requested):

resources/eventing/values.yaml
barchw:
```suggestion
      version: "1.4.0"
```

triffer (approved):

PR #17253, opened 2023-04-05
Use nodeport as pull addr for internal registry

dbadura:
**Description**

- use nodeport to pull images from internal registry

**Related issue(s)**


pPrecel (commented):
LGTM

resources/serverless/values.yaml
```diff
@@ -64,7 +64,8 @@ global:
     app.kubernetes.io/instance: "{{ .Release.Name }}"
     app.kubernetes.io/version: "{{ .Values.global.images.function_controller.version }}"
     helm.sh/chart: '{{ include "chart" . }}'
-  dockerServicePort: 5000
```
kwiatekus, line 67: why changing the name of the port config ?

resources/serverless/templates/virtualservice.yaml
```diff
@@ -1,3 +1,4 @@
+{{/*TODO remove it*/}}
```
kwiatekus, line 1: why not to remove it now in this PR ?
Having this makes istio still required

dbadura: I have another PR with istio deletion: https://github.com/kyma-project/kyma/pull/17257/files

resources/serverless/values.yaml
dbadura: Because it's confusing. It's not docker port, but it's registry port.

Cortey (approved):

PR #17290, opened 2023-04-12
Execa Logs Streaming

ralikio:
**Description**

- Improved logging of `execa` - before the change logs were printed only when a command finished. Because of that in the case of timeouts the logs were not visible.
- Prompt confirmation - because of introduce confirmation prompt in kcp, the tests require sending `Y` to invoked subprocesses.
- Switch params for the above.

**Related issue(s)**


tests/fast-integration/skr-aws-upgrade-integration/index.js
```diff
@@ -44,6 +44,11 @@ describe('SKR-Upgrade-test', function() {
     return options;
   };
 
+  it(`Perform kcp login`, async function() {
+    console.log('Performing kcp login');
+    await kcp.login();
```
PK85, line 49: how it will help? kcp.login() is called in the getOrProvisionSKR function

ralikio: You are right.

ralikio:
/test 
pre-main-kyma-gardener-gcp-eventing-upgrade

Issue #17103, opened 2023-03-16
Subscription webhook is failing if a subscription has different prefix


mfaizanse:
**Description**

The Subscription webhook fails to convert a subscription if it has different prefix from the one defined in `eventing` configMap.

The effect of this is that the `kubectl get subscriptions -A` is not working:
```
-> kubectl get subscriptions -A
Error from server: conversion webhook for eventing.kyma-project.io/v1alpha1, Kind=Subscription failed: prefix not found`
```

The `kubectl get subscriptions.v1alpha1.eventing.kyma-project.io -A` is working.

**Steps to reproduce**

1. Provision a cluster with v1alpha1 (kyma v2.11.2) and default prefix `sap.kyma.custom`.
2. Create the following subscription:
```
apiVersion: eventing.kyma-project.io/v1alpha1
kind: Subscription
metadata:
  name: event-xxx-test
  namespace: test
spec:
  filter:
    filters:
    - eventSource:
        property: source
        type: exact
        value: /xxx/xxx/xxx
      eventType:
        property: type
        type: exact
        value: sap.a4.abc.xxx.v1.test.Changed.v1
  sink: http://event-consumer.dev.svc.cluster.local
```
3. Upgrade the cluster to kyma v2.12.0.
4. Run `kubectl get subscriptions -A`.

marcobebway:
Closing this bug because it does not affect existing workload clusters.

marcobebway:
It is intentional to fail converting misconfigured subscriptions.

PR #17291, opened 2023-04-12
Fix the subscription migration from v1alpha1 to v1alpha2 if the prefix is not supported

marcobebway:
**Description**

Fix the subscription migration from v1alpha1 to v1alpha2 if the prefix is not supported.

**Related issue(s)**

- https://github.com/kyma-project/kyma/issues/17103


marcobebway:
Not needed for now.

Issue #17194, opened 2023-03-29
Serverless Controller should rebuild functions which use outdated base image


kwiatekus:
**Description**

Serverless Controller should automatically rebuild functions which use outdated base image. Currently when a new function runtime  image version is rolled out, functions do not pick them up automatically.

**Expected result**

When an image version for function runtime changes (i.e during kyma version upgrades), functions using this runtime should be scheduled for reconciliation and rebuilt (based on the new runtime base image version).

**Actual result**

When an image version for function runtime changes (i.e during kyma version upgrades), functions using this runtime are not automatically rebuilt and continue using previous runtime image version.

PR #17307, opened 2023-04-13
Rebuild the function when the base image is changed

MichalKalke:
**Description**
Serverless Controller should automatically rebuild functions which use outdated base image. Currently when a new function runtime image version is rolled out, functions do not pick them up automatically.

- Rebuild function when the `base_image` in the configmap is changed for given runtime
- Test if the function is reconciled when the `base_image` is changed 

**Related issue(s)**
https://github.com/kyma-project/kyma/issues/17194


components/function-controller/internal/controllers/serverless/function_reconcile_test.go
```diff
@@ -1337,6 +1337,39 @@ func TestFunctionReconciler_Reconcile(t *testing.T) {
...
+		inFunction := newFixFunctionWithCustomImage(testNamespace, "custom-runtime-fn-image", "", 1, 2)
+		g.Expect(resourceClient.Create(context.TODO(), inFunction)).To(gomega.Succeed())
+		defer deleteFunction(g, resourceClient, inFunction)
+
+		request := ctrl.Request{NamespacedName: types.NamespacedName{Namespace: inFunction.GetNamespace(), Name: inFunction.GetName()}}
+
+		//WHEN
+		t.Log("should detect runtimeImage change")
+
+		function := &serverlessv1alpha2.Function{}
```
dbadura, line 1352: Is it needed?

components/function-controller/internal/controllers/serverless/function_reconcile_test.go
```diff
@@ -1337,6 +1337,39 @@ func TestFunctionReconciler_Reconcile(t *testing.T) {
...
+		g.Expect(resourceClient.Get(context.TODO(), request.NamespacedName, function)).To(gomega.Succeed())
+
+		result, err := reconciler.Reconcile(ctx, request)
+		g.Expect(err).To(gomega.BeNil())
+		g.Expect(result.Requeue).To(gomega.BeFalse())
+		g.Expect(result.RequeueAfter).To(gomega.Equal(time.Second * 1))
+
```
dbadura, line 1359: I would check if `runtime` is set correctly to be sure that it will be changed.

components/function-controller/internal/controllers/serverless/function_reconcile_test.go
```diff
@@ -1337,6 +1337,39 @@ func TestFunctionReconciler_Reconcile(t *testing.T) {
...
+		configMap := changeDockerfileForRuntime(rtm)
+
+		g.Expect((resourceClient.Update(ctx, configMap))).To(gomega.Succeed())
```
dbadura, line 1362:
```suggestion
		g.Expect(resourceClient.Update(ctx, configMap)).To(gomega.Succeed())
```

components/function-controller/internal/controllers/serverless/function_reconcile_test.go
MichalKalke: It is not, the log was deleted.

components/function-controller/internal/controllers/serverless/function_reconcile_test.go
MichalKalke: Thanks for that, here it comes.

dbadura (approved):

PR #17352, opened 2023-04-19
switch prometheus pushgateway to upstream docker image

a-thaler:
**Description**
The prometheus-pushgateway is provided by a sub-chart of the monitoring component. The gateway was never enabled by default and not really maintained actively, especially the image is not maintained by the kyma team actively. That should be made transparent by not serving it anymore by the kyma specific registry.

- switch the image used by the disabled pushgateway chart to use the upstream image directly
- ...
- ...

**Related issue(s)**


a-thaler:
Will keep it as is and better remove the feature as soon as the deprecation period is over

Issue #16232, opened 2022-12-01
User should be able to manage function runtime pod's labels and annotations


kwiatekus:
**Description**

As a function developer I want to define metadata (labels and annotations) for function runtime pod.

**AC**

 - [ ] User should be able to manage function runtime pod's labels and annotations from kyma CLI
 - [x] User should be able to manage function runtime pod's labels and annotations from kyma UI
 - [x] Validation webhook shouldn't allow to add labels from serverless.kyma-project.io group

**Reason**

Labels are used to target workloads in kubernetes. User should have the possibility to add custom labels to function pod to look it up, display logs or target it for any other reason.
Various k8s tools are using annotations to control their features against the workload (i.e istio, fluentbit). 
( For Example: Custom fluent bit log parser is enabled for a given pod by an annotation: `fluentbit.io/parser: my-regex-parser`). User should be in control of that.

**Attachments**
See the agreed [proposal](https://github.com/kyma-project/kyma/blob/main/components/function-controller/design/extend_function_spec_proposal.md#labels-and-annotations) for the CRD design:

```yaml
apiVersion: serverless.kyma-project.io/v1alpha2
kind: Function
metadata:
  name: my-function
  namespace: default
  labels:
    app.kubernetes.io/name: my-function # <-- this label will be inherited by deployment, job, HPA and config map
spec: 
  ...
  labels:
    app: my-app # <-- this label will be used in Deployment's PodTemplate and will be applied on the function's runtime pod 
  annotations:
    fluentbit.io/parser: my-regex-parser # <-- those annotations will be used in Deployment's PodTemplate and will be applied on the function's runtime pod
    istio-injection: enabled

```

**Backward compatibility considerations**

Currently the `function.serverless.kyma-project.io/v1alpha2` offers:

```yaml
spec:
  template:
    labels: # <--read by controller to use for function's pod labels
    annotations # <-- not used
```

The desired changes should include:
```yaml
spec:
  template: # <--deprecated
    labels: # <-- still read by controller / deprecated in favor of spec.labels
    annotations # <-- still not used

  labels: # read by controller and merged with the deprecated ones
  annotations: # read by controller
```

jalibu:
Please also add the metadata to the deployments (not only to the pods)

kwiatekus:
The missing CLI PR will be delivered with 2.15

PR #17357, opened 2023-04-19
add runtime labels and annotations of serverless functions to busola …

anoipm:
**Description**

- add runtime labels and annotations of serverless functions to busola extension

**Related issue(s)**
See also #16232

### DIFF
--- a/resources/serverless/templates/busola-serverless-extension.yaml
+++ b/resources/serverless/templates/busola-serverless-extension.yaml
@@ -380,12 +380,14 @@ data:
       children:
         - path: labels
           widget: KeyValuePair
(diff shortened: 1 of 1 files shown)


kwiatekus (changes requested):
The warning buble should not be shown by default.
Validation webhook should reject such labels anyways.. 
Screenshot 2023-04-21 at 15 23 51

kwiatekus (changes requested):
Can you also add the labels and annotations to the details view? so that users can see those in the read-only mode  (not only when editing )

kwiatekus (approved):

Issue #16584, opened 2023-01-17
Remove tracing component


a-thaler:
**Description**
After jaeger got deprecated and three releases are over it gets time to remove it.

**Criterias:**
- kyma repo cleanup
  - delete resources folder
  - remove usage of global.tracing.enabled setting
  - remove component from all component lists
- 3party repo cleanup
  - remove folder/module in third-party-image repo
  - remove prow jobs
- Surroundings
  - remove/adjust docu
  - Prepare release notes and migration script
  - Remove entry in SKR feature matrix
  - keep the tutorial and links to the tutorial
  - verify tutorial once more
- Tests
  - replace jaeger deployment in integration tests
  - remove manual test cases
  - verify periodic jobs for healthyness

**Reasons**

**Attachments**

https://github.com/kyma-project/kyma/issues/15852

rakesh-garimella:
Removing from components.yaml in control-plane: https://github.com/kyma-project/control-plane/pull/2708

rakesh-garimella:
```
kyma repo cleanup
[x]  delete resources folder
[x]  remove usage of global.tracing.enabled setting
[x]  remove component from all component lists
3party repo cleanup
  [x] remove folder/module in third-party-image repo
  [x] remove prow jobs
Surroundings
  [] remove/adjust docu (PR in review)
  [x] Prepare release notes and migration script
  [x] Remove entry in SKR feature matrix
  [x] keep the tutorial and links to the tutorial
  [x] verify tutorial once more
Tests
[x] replace jaeger deployment in integration tests
[x] remove manual test cases
[] verify periodic jobs for healthyness (in Progress)
```

PR #17389, opened 2023-04-26
Cleanup script for jaeger removal

hisarbalik:
**Description**

- Shell script to cleanup remaining resources after Jaeger tracing removal

**Related issue(s)**
https://github.com/kyma-project/kyma/issues/16584

### DIFF
--- a/docs/assets/2.13.-2.14-cleanup-jaeger.sh
+++ b/docs/assets/2.13.-2.14-cleanup-jaeger.sh
@@ -1,0 +1,24 @@
+#!/usr/bin/env bash
+
+kubectl delete -n kyma-system tracepipelines.telemetry.kyma-project.io jaeger
+kubectl delete -n kyma-system telemetries.telemetry.istio.io kyma-traces
+kubectl delete -n kyma-system jaegers.jaegertracing.io tracing-jaeger
+kubectl delete -n kyma-system serviceaccounts tracing-auth-proxy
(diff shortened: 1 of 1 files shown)


NHingerl (commented):
as a TW, I cannot judge whether this script does what it should. However, I can approve if you need to merge.

NHingerl (approved):

docs/assets/2.13.-2.14-cleanup-jaeger.sh
```diff
@@ -0,0 +1,24 @@
+#!/usr/bin/env bash
+
+kubectl delete -n kyma-system tracepipelines.telemetry.kyma-project.io jaeger
+kubectl delete -n kyma-system telemetries.telemetry.istio.io kyma-traces
```
a-thaler, line 4: Why no review happened here from a developer?
The resource is not placed in the kyma-system namespace and the command will fail, see https://github.com/kyma-project/kyma/blob/2.13.0/resources/tracing/templates/kyma-additions/istio-telemetry.yaml#L6

Why cluster scoped resources are using a namespace in he script?
Also the migration guide related to the script is missing, see for example https://kyma-project.io/docs/kyma/2.12/migration-guide-2.11-2.12

NHingerl: I, too, was wondering why the only mandatory review was by TW. 
Is the code owners assignment wrong?

PR #17419, opened 2023-04-27
Change eu.gcr.io/ to europe-docker.pkg.dev/ for ory and istio images

videlov:
**Description**

- Change eu.gcr.io/ to europe-docker.pkg.dev/ for ory and istio images

**Related issue(s)**

https://github.tools.sap/kyma/test-infra/issues/140


resources/istio/values.yaml
```diff
@@ -171,7 +171,7 @@ components:
 
 global:
   containerRegistry:
-    path: eu.gcr.io/kyma-project
+    path: europe-docker.pkg.dev/kyma-project/prod
```
triffer, line 174: This is overwritten in the images, but why do we add `prod` here? If we remove the `containerRegistryPath` it would fail, since the directory already has `prod`defined, right? 🤔

videlov: `/prod` is a mistake in here, we actually do not use this property `containerRegistry` at all in istio helm templates. let see if it fails without it.

triffer (approved):

Ressetkk (approved):

a-thaler (approved):

nataliasitko (approved):

PR #17461, opened 2023-05-09
fix description for busola extension

mrCherry97:
**Description**

- fix description, because it is rendering with additional `"}}`

**Related issue(s)**


kwiatekus:
We cant merge it I'm affraid. Those `{{ }}` characters are interpreted by helm when installing serverless module

```
    - lastTransitionTime: "2023-05-09T11:41:23Z"
      message: 'could not render manifest from chart: while templating chart: parse
        error at (serverless/templates/busola-serverless-extension.yaml:398): unexpected
        "[" in command'
```

Issue #17426, opened 2023-04-28
Automate CRD documentation


VOID404:
Description:
---
- Add descriptions to fields in CRDs
- Modify CRD's README
- Add generation to Makefile

Also described in detail [here](https://github.com/kyma-project/kyma/issues/17347#issuecomment-1514494302).

Resources:
---
- [ ] [applications.applicationconnector.kyma-project.io](https://github.com/kyma-project/kyma/blob/fe23dd930eeb7e45df010cfd5419002023712acf/installation/resources/crds/application-connector/applications.applicationconnector.crd.yaml) 
  #17432
- [x] [compassconnections.compass.kyma-project.io](https://github.com/kyma-project/kyma/blob/fe23dd930eeb7e45df010cfd5419002023712acf/installation/resources/crds/compass-runtime-agent/compass-connection.crd.yaml) 
  #17428

Related Issues:
---
- #17347
- #17428
- #17432

Issue #17347, opened 2023-04-19
Automate CRD documentation


mmitoraj:
**Description**

Kyma documentation in the **Technical Reference** section focuses on [Kyma Custom Resources](https://kyma-project.io/docs/kyma/latest/05-technical-reference/00-custom-resources/). The documents include tables listing all the possible parameters of a given resource together with their descriptions. As the parameters and CRDs change, the documentation may easily become out-of-date and difficult to maintain. Huskies came up with the idea to auto-generate the content of the tables. For example, see the auto-generated table for the [Telemetry - LogPipeline CR parameters](https://kyma-project.io/docs/kyma/latest/05-technical-reference/00-custom-resources/telemetry-01-logpipeline/#custom-resource-parameters) CR. The solution could be improved, adjusted, and implemented for all CRs descriptions. 

Steps:

1. [x] Ask @a-thaler to explain the mechanism so that other teams could confirm if the solution fits their use cases and requirements
2. [x] Discuss with @kyma-project/technical-writers if the solution meets Kyma documentation requirements
3. [x] Contact POs to consult if auto-generating the tables' content with their teams' CRDs description is possible
4. [x] Improve and adjust the mechanism if needed
5. [ ] Automate the docu creation

**Area**

Area | Custom resource
-- | --
Application Connectivity | Application,  CompassConnection
API Gateway | APIRule
Eventing | Subscription, EventingBackend
Istio | Istio
Serverless | Function, GitRepository
Telemetry | LogPipeline, LogParser, TracePipeline

**Reasons**

Up-to-date documentation.

**Assignees**

@mmitoraj 
@a-thaler 

**Related issues**
#17421 #17380

a-thaler:
- every domain need to provide good and complete descriptions on the CRD attributes
- prepare your MD files by removing the relevant tables and replacing them with the placeholders as described [here](https://github.com/kyma-project/kyma/tree/main/hack/table-gen)
- add new mappings for your CRDs to that [Makefile](https://github.com/kyma-project/kyma/blob/main/hack/table-gen/Makefile), defining to map which CRD to which file
- when providing a new CRD version to the kyma folder, execute first the "make generate" command
- a prow job will remind you if you forgot the previous step as it will execute the same make target and check the git status afterwards

mmitoraj:
Dear All,

To drive the topic forward, we would need your and your teams' engagement at that point. I created a table listing Kyma CRDs. **Please:**
[...]

PR #17476, opened 2023-05-11
Update the CRD descriptions of the CompassConnection

mmitoraj:
**Description**

- Unified the CompassConnection CRD descriptions. Fixed punctuation.

**Related issue(s)**
#17426 #17347

### DIFF
--- a/docs/05-technical-reference/00-custom-resources/ra-01-compassconnection.md
+++ b/docs/05-technical-reference/00-custom-resources/ra-01-compassconnection.md
@@ -60,31 +60,31 @@ This table lists all the possible parameters of the CompassConnection custom res
 | Parameter | Type | Description |
 | ---- | ----------- | ---- |
 | **managementInfo** (required) | object |  |
-| **managementInfo.&#x200b;connectorUrl** (required) | string | Connector URL used for maintaining secure connection. |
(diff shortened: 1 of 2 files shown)


VOID404 (approved):
It'd be nice if `Specifies the date` and `Specifies when` were normalized. 
I don't think dots are an improvement. There are no dots in the example from [Kubernetes docs](https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#additional-printer-columns), but it's fine, I'll approve with dots.

Other than that, looks good 👍 
/hold

mmitoraj:
@VOID404 I wouldn't change all the sentences to either `Specifies when...` or 'Specifies the date of...`. Sometimes it depends on the context. When there is a verb later in the sentence it is better to say `Specifies **when** the certificate was acquired` instead of the longer `Specifies **the date when** the certificate was acquired`. However, if we can use a noun later in the sentence, I'd say `Specifies **the date of** the last certificate renewal.` And in this case, there is also no sense in adding passive voice to the sentence and saying `Specifies when the certificate was renewed for the last time.` I'd leave the descriptions as they are now. I've also added a link to the specification file itself in the paragraph with the table. Thank you :)

grego952 (approved):

PR #17479, opened 2023-05-11
add checking changes of runtime annotations

anoipm:
**Description**

- add checking changes of runtime annotations in serverless function

**Related issue(s)**
see also #16232

### DIFF
--- a/components/function-controller/internal/controllers/serverless/deployment_test.go
+++ b/components/function-controller/internal/controllers/serverless/deployment_test.go
@@ -205,7 +205,7 @@ func TestFunctionReconciler_equalDeployments(t *testing.T) {
 			want: false,
 		},
 		{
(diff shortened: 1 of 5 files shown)


components/function-controller/internal/controllers/serverless/deployment_test.go
```diff
@@ -199,7 +205,7 @@ func TestFunctionReconciler_equalDeployments(t *testing.T) {
 			want: false,
 		},
 		{
-			name: "different pod annotations - spec.template.metadata.annotations is ignored so that kubectl rollout restart works properly",
+			name: "different pod annotations", //TODO: remove this? - spec.template.metadata.annotations is ignored so that kubectl rollout restart works properly",
```
dbadura, line 208: Remove it as `spec.template.metadata.annotation` is not ignored

dbadura (commented):
Looks good, just remove one test

dbadura (approved):

PR #17563, opened 2023-05-24
Increase memory limit for telemetry operator

rakesh-garimella:
**Description**

- Telemetry operator is watching many resources in the cluster like secrets, configmaps etc. It can happen that when in a cluster there are many of these resources it would cause the cache to be big. Current limit 256 Mi would not be enough in such cases so increase it to 384Mi

- Change is telemetry manager repo [here](https://github.com/kyma-project/telemetry-manager/pull/197)
**Related issue(s)**

### DIFF
--- a/resources/telemetry/charts/operator/values.yaml
+++ b/resources/telemetry/charts/operator/values.yaml
@@ -69,7 +69,7 @@ fluentbit:
 resources:
   limits:
     cpu: 100m
-    memory: 256Mi
+    memory: 384Mi
   requests:
     cpu: 5m
     memory: 20Mi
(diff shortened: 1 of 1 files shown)


resources/telemetry/charts/operator/values.yaml
```diff
@@ -69,7 +69,7 @@ fluentbit:
 resources:
   limits:
     cpu: 100m
-    memory: 256Mi
+    memory: 384Mi
   requests:
     cpu: 5m
     memory: 20Mi
```
chrkl, line 75: Let's increase the requests as well. It seems unrealistic that the operator runs with only 20Mi.

chrkl (approved):

PR #17600, opened 2023-06-01
Fast integration BTP Manager Secret tests

szwedm:
**Description**

add tests for BTP Manager Secret.

**Related issue(s)**

### DIFF
--- a/tests/fast-integration/skr-test/btp-manager-secret/index.js
+++ b/tests/fast-integration/skr-test/btp-manager-secret/index.js
@@ -1,14 +1,32 @@
 const os = require('os');
 const {expect} = require('chai');
-
 const {
     initializeK8sClient,
+    getSecret,
 } = require('../../utils');
+
+const secretName = 'sap-btp-manager';
+const ns = 'kyma-system';
+
+let expectedSecret, modifiedSecret;
 
 function btpManagerSecretTest() {
     describe('BTP Manager Secret Test', function () {
         before('REMOVE LATER: Initialize kubeconfig', async function(){
             await initializeK8sClient({kubeconfigPath: `${os.homedir()}/k3d_kubeconfig`});
+        });
+        // Check if BTP Manager Secret with BTP Operator credentials is created properly
+        it('should check if Secret exists and contains the expected data keys', async function() {
+            console.log(`Checking the data keys of the "sap-btp-manager" Secret`);
+            expectedSecret = await getSecret(secretName, ns);
+            expect(expectedSecret).to.not.be.empty;
+            expect(expectedSecret.metadata.labels['app.kubernetes.io/managed-by']).to.equal('kcp-kyma-environment-broker');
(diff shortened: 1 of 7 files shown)


tests/fast-integration/skr-test/btp-manager-secret/index.js
```diff
@@ -0,0 +1,115 @@
...
+      console.log(`Changing data in the "sap-btp-manager" Secret`);
+      prepareSecretForApply(modifiedSecret);
+      changeSecretData(modifiedSecret);
+      console.log(`Applying edited Secret`);
+      await k8sApply([modifiedSecret], ns);
+      console.log(`Waiting ${reconciliationTimeout} ms until edited Secret is created`);
+      await waitForSecret(secretName, ns, reconciliationTimeout);
+      let actualSecret = await getSecret(secretName, ns);
+      console.log(`Waiting for the reconciliation for ${reconciliationTimeout} ms`);
+      await waitForK8sObject(
```
jaroslaw-pieszka, line 57: Why don't we extract this part as a separate function, like we do with `waitForSecret`. 
Then we would have the same level of abstraction (more or less) in this assertion.

tests/fast-integration/skr-test/btp-manager-secret/index.js
```diff
@@ -0,0 +1,115 @@
...
+            return (
+              watchObj.object.metadata.name.includes(secretName) &&
+                        watchObj.object.metadata.resourceVersion !== actualSecret.metadata.resourceVersion
+            );
+          },
+          reconciliationTimeout,
+          `Waiting for ${secretName} Secret reconciliation timeout (${reconciliationTimeout} ms)`,
+      );
+      console.log(`Secret has been reconciled. Checking Secret's data`);
+      actualSecret = await getSecret(secretName, ns);
```
jaroslaw-pieszka, line 70: This sequence of four lines is duplicated at least twice. So it could be extracted.
As a matter of fact, if we join the first two assertions we could use extracted function for the third time.

tests/fast-integration/skr-test/btp-manager-secret/index.js
szwedm: I used a closure instead of a defined function to avoid defining what we can treat as a Secret reconciliation. In this case, we know that KEB is the only entity to modify the Secret, but resourceVersion can change due to user actions like patch or update and it won't be a reconciliation. If I would define a func like `waitForSecretReconcile` other developers could assume that the func always waits for a reconciliation when it doesn't.

tests/fast-integration/skr-test/btp-manager-secret/index.js
szwedm: I agree, we can wrap this sequence in a separate func to assert a Secret

jaroslaw-pieszka (approved):

Issue #17168, opened 2023-03-24
Eventing pods request much more CPU and memory than they need


pbochynski:
**Description**

The default installation of eventing requests a lot of resources:
- publisher proxy: 410m CPU and 448Mi memory x 2 pods
- controller: 410m CPU and 704Mi memory x 1 pod
- nats: 400m CPU and 576Mi memory x 3 pods

What gives together: **2,43 CPU and 3,33 GB memory** 
The actual usage of eventing pods in the idle cluster is: **0.015 CPU and 0.35 GB** 

Could you adjust requests settings to allow better cluster resources utilization, please?

**Acceptance**
- [x] combine settings used in MPS-Config and kyma/production profile and make them as low as possible
- [x] validate using loadtester that with minimal setting eventing still comes up and is able to run a low workload (10eps)

pbochynski:
Still valid.

k15r:
production profile states:
```
controller:
  resources:
    limits:
      cpu: 1000m
      memory: 512Mi
    requests:
      cpu: 10m
      memory: 256Mi
  publisherProxy:
    resources:
      limits:
        cpu: 500m
        memory: 512Mi
      requests:
        cpu: 10m
        memory: 256Mi

nats:
    resources:
      limits:
        cpu: 500m
        memory: 1Gi
      requests:
        cpu: 10m
        memory: 512Mi
    logging:
      debug: false
      trace: false
```

mfaizanse:
@k15r There is a [PR](https://github.com/kyma-project/kyma/pull/17651/files), which is even increasing the request resources for eventing.

marcobebway:
On SKRs, we reduced the Eventing **requested** resources as follows:

EC(1 replica):
- CPU: **40m**
- MEM: **64Mi**

EPP(sum of 2 replicas):
- CPU: **80m** (2 * 40m)
- MEM: **128Mi** (2 * 64Mi)

NATS(sum of 3 replicas):
- CPU: **120m** (3 * 40m)
- MEM: **192Mi** (3 * 64Mi)

Total:
- CPU: **240m** (6 * 40m)
- MEM: **384Mi** (6 * 64Mi)

marcobebway:
OSS Production:

https://github.com/kyma-project/kyma/blob/35a75e206717e6f0733c34e7b061c4a86860973e/resources/eventing/profile-production.yaml#L8-L51

OSS Evaluation

[...]

PR #17651, opened 2023-06-12
runtime overrides-2

iamps5:
**Description**

With the Kyma modularization the overrides to Kyma modules on SKRs supplied by management-plane-config shall be kept absolutely minimum. So, We are moving all the latest overrides for **eventing, ory & serverless** here

**Related issue(s)**


iamps5:
@ebensom can you please take a look?

resources/ory/profile-evaluation.yaml
```diff
@@ -3,26 +3,34 @@ global:
   ory:
     hydra:
       persistence:
-        enabled: false
+        enabled: true
```
ebensom, line 6: I'm not sure what to do with these ory changes. These are related to the fact that the azure_lite runtime plan uses the evaluation profile but it needs persisted ory-hydra unlike trial/free runtimes. 

@kyma-project/goat 
Should these changes be made here in the evaluation profile, or should we just keep them as runtime overrides for azure_lite until ory is fully deprecated and removed from runtimes? AFAIK ory is not getting modularized as being a deprecated component, right?

Also, imo changing the persistence is not allowed for existing ory deployments as it would cause statefuset to be changed in immutable fields. So this changeset on ory/profile-evaluation.yaml might not be even valid.

resources/ory/profile-production.yaml
```diff
@@ -30,10 +30,11 @@ hydra:
   deployment:
     resources:
       requests:
-        cpu: "250m"
-        memory: "256Mi"
+        cpu: 250m
```
ebensom, line 33: CPU limit is not set on purpose, we should remove our artificial CPU limit setting of 10 cores in runtime-overrides. Feel free to revert this change.

resources/ory/profile-production.yaml
```diff
@@ -44,10 +45,11 @@ oathkeeper:
   deployment:
     resources:
       limits:
-        memory: "512Mi"
+        cpu: 10m
```
ebensom, line 48: CPU limit is not set on purpose, we should remove our artificial CPU limit setting of 10 cores in runtime-overrides. Feel free to revert this change.

resources/serverless/values.yaml
```diff
@@ -176,8 +176,8 @@ containers:
   manager:
     resources:
       limits:
-        cpu: 500m
-        memory: 160Mi
+        cpu: 600m
```
ebensom, line 179: I would even accommodate bigger CPU spikes for serverless-ctrl-mngr

```suggestion
        cpu: 1500m
```

resources/ory/profile-evaluation.yaml
triffer: @ebensom 
Yes, Hydra is deprecated and will be removed in the future. We agree that it would be better to just keep the overrides for azure_lite in place.

resources/eventing/profile-production.yaml
```diff
@@ -13,19 +13,20 @@ controller:
...
       memory: 256Mi
   publisherProxy:
+    replicas: 2
     resources:
       limits:
         cpu: 500m
         memory: 512Mi
       requests:
-        cpu: 10m
-        memory: 256Mi
```
friedrichwilken, line 28: the request should not be higher than the limits. why are the cpu numbers so high?

friedrichwilken (approved):

resources/eventing/profile-production.yaml
```diff
@@ -13,19 +13,20 @@ controller:
     maxMessages: -1
   resources:
     limits:
-      cpu: 1000m
-      memory: 512Mi
+      cpu: 2000m
```
mfaizanse, line 16: Are you trying to move these [values](https://github.tools.sap/kyma/management-plane-config/blob/main/resources/runtimes/2.15/base/eventing.yaml)?

resources/ory/profile-evaluation.yaml
iamps5: @triffer okay. I've disabled ory hydra persistence

resources/eventing/profile-production.yaml
iamps5: My bad, I supposed to raise the limit & instead I increased request. I have corrected. Now cpu request is moved backed to 50m & limit is 1000m, we are keeping it high to allow bigger bursts.

resources/eventing/profile-production.yaml
iamps5: @mfaizanse Yes, but we want to increase those values also. That's why here cpu limit is bit more than the override in management plane

friedrichwilken (dismissed):

resources/ory/profile-evaluation.yaml
```diff
@@ -5,24 +5,32 @@ global:
       persistence:
         enabled: false
         postgresql:
-          enabled: false
+          enabled: true
```
triffer, line 8: I think we should also keep false since this is also an azure-lite override.
```suggestion
          enabled: false
```

resources/ory/profile-evaluation.yaml
```diff
@@ -5,24 +5,32 @@ global:
...
 oathkeeper:
   replicaCount: 1
   deployment:
     resources:
       requests:
-        cpu: "10m"
+        cpu: 10m
     autoscaling:
       enabled: false
+postgresql:
```
triffer, line 29: The postgresql config is also part of the ory azure-lite override and can be removed.

resources/ory/profile-evaluation.yaml
```diff
@@ -5,24 +5,32 @@ global:
...
         postgresql:
-          enabled: false
+          enabled: true
         gcloud:
           enabled: false
 hydra:
   replicaCount: 1
   hydra:
-    autoMigrate: false
+    autoMigrate: true
```
triffer, line 14: This is also part of the ory azure-lite override.
```suggestion
    autoMigrate: false
```

iamps5:
@triffer Updated the recommended changes, please review.

triffer (dismissed):

friedrichwilken (approved):

kwiatekus (dismissed):

friedrichwilken (dismissed):

marcobebway:
Please don't change the Eventing overrides. We already reduced the Eventing requested resources as part of this [ticket](https://github.com/kyma-project/kyma/issues/17168)

marcobebway (changes requested):
Please don't change the Eventing overrides. We already reduced the Eventing requested resources as part of this [ticket](https://github.com/kyma-project/kyma/issues/17168)

PR #17652, opened 2023-06-12
Update production profile for monitoring component

iamps5:
**Description**

Updating the production profile for the monitoring component with the latest settings used on the SAP based managed offering

**Related issue(s)**


iamps5:
@ebensom can you please approve this?

resources/monitoring/profile-evaluation.yaml
```diff
@@ -3,13 +3,61 @@ prometheus:
...
+        memory: 4Gi
       requests:
         cpu: 75m
         memory: 1000Mi
     retention: 6h
     retentionSize: 1GB
 
+prometheusOperator:
+  namespaces:
+    releaseNamespace: "true"
```
ebensom, line 15: I think this is specific to runtimes (SKRs) only where we don't want other namespaces to be watched by prometheus-operator.

resources/monitoring/profile-evaluation.yaml
```diff
@@ -3,13 +3,61 @@ prometheus:
...
+      protocol: TCP
+      targetPort: 15020
+  sidecar:
+    dashboards:
+      searchNamespace: "kyma-system"
+    datasources:
+      searchNamespace: "kyma-system"
+  kyma:
+    authProxy:
+      configDocsLink: "https://help.sap.com/docs/BTP/65de2977205c403bbc107264b8eccf4b/3e4299cfd0884c428e6b4774225638e8.html"
```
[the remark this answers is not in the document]
a-thaler: Yes, the overrides are specific for SKRs and need to stay. The component will not be migrated to a module but instead will be deleted at some point

resources/monitoring/profile-evaluation.yaml
```diff
@@ -3,13 +3,61 @@ prometheus:
...
+      requests:
+        cpu: 200m
+        memory: 768Mi
+      limits:
+        cpu: 1000m
+        memory: 8Gi
+
+defaultRules:
+  disabled:
+    PrometheusNotConnectedToAlertmanagers: "true"
```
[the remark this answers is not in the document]
iamps5: @ebensom then should I add `alertmanager: { enable: "false"}`, or remove the above rule

resources/monitoring/profile-evaluation.yaml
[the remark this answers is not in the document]
iamps5: okay, I'll remove all these overrides

a-thaler (changes requested):
That are mainly incompatible changes (like desabling the alertmanager by default), so the behaviour of the component will change fundamentally. If we think it is time to do such changes even before the removal of the component, we will do that in a controlled way. Please do not just apply changes to the open source project out of nothing.
As I said in a different comment, we do not plan to transform the component into a module.

iamps5:
> That are mainly incompatible changes (like desabling the alertmanager by default), so the behaviour of the component will change fundamentally. If we think it is time to do such changes even before the removal of the component, we will do that in a controlled way. Please do not just apply changes to the open source project out of nothing. As I said in a different comment, we do not plan to transform the component into a module.

Okay, I'll remove the `deafultRules` section

iamps5:
@a-thaler can you please review this now?

resources/monitoring/profile-production.yaml
```diff
@@ -35,4 +51,30 @@ kube-state-metrics:
...
+      protocol: TCP
+      targetPort: 15020
+  sidecar:
+    dashboards:
+      searchNamespace: "kyma-system"
+    datasources:
+      searchNamespace: "kyma-system"
+  kyma:
+    authProxy:
+      configDocsLink: "https://help.sap.com/docs/BTP/65de2977205c403bbc107264b8eccf4b/3e4299cfd0884c428e6b4774225638e8.html"
```
a-thaler, line 80: That whole block should not get in as it is specific to the SAP setup and will break compatibility from OS perspective

resources/monitoring/profile-production.yaml
```diff
@@ -17,13 +17,29 @@ prometheus:
               storage: 20Gi
 
 prometheusOperator:
+  namespaces:
+    releaseNamespace: "true"
```
a-thaler, line 21: That block should not get in as it is specific to the SAP setup and will break compatibility from OS perspective

resources/monitoring/profile-production.yaml
```diff
@@ -17,13 +17,29 @@ prometheus:
...
+      requests:
+        cpu: 200m
+        memory: 768Mi
+      limits:
+        cpu: 1000m
+        memory: 8Gi
+
+defaultRules:
+  disabled:
+    PrometheusNotConnectedToAlertmanagers: "true"
```
a-thaler, line 41: That whole block should not get in as it is specific to the SAP setup and will break compatibility from OS perspective

iamps5:
@a-thaler My bad yesterday I made changes only in evaluation. Now, I have added recommend changes in production profile also. Please review

resources/logging/profile-evaluation.yaml
```diff
@@ -1,8 +1,8 @@
 loki:
   resources:
     limits:
-      cpu: 100m
```
a-thaler, line 4: The logging component got already removed from main branch

resources/monitoring/profile-evaluation.yaml
```diff
@@ -3,13 +3,19 @@ prometheus:
...
         cpu: 600m
-        memory: 2000Mi
+        memory: 4Gi
       requests:
         cpu: 75m
         memory: 1000Mi
     retention: 6h
     retentionSize: 1GB
 
-grafana:
```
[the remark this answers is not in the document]
a-thaler: I agree, why we should remove it?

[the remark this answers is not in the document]
iamps5: Added back

resources/logging/profile-evaluation.yaml
iamps5: removed

a-thaler (approved):

PR #17690, opened 2023-06-20
Security patch Istio to 1.17.3

barchw:
**Description**

- Upgrade Istio version to 1.17.3


triffer (commented):
I think we should also update the istio version in the dashbaords`resources/istio-resources/files/dashboards`. 
I think for the future it would be better to just have major and minor version in the dasboard description.

triffer (approved):

PR #17725, opened 2023-06-22
Fix protecode issue for golang-runtime

muralov:
**Description**
bump image


marcobebway:
Please update the images of the controller and publisher.

marcobebway (changes requested):
Please update the images of the controller and publisher.

marcobebway (dismissed):

PR #17764, opened 2023-06-29
Remove PR images from Application Connector resources

mvshao:
**Description**
Remove PR images from Application Connector resources


koala7659 (approved):
LGTM

Issue #17749, opened 2023-06-27
Fix Eventing controller logic to detect secret changes


marcobebway:
**Description**

As per this [logic](https://github.com/kyma-project/kyma/blob/a6d77e18431c42fa1ecb397256b3fabd3f896828/components/eventing-controller/controllers/backend/reconciler.go#L328-L330), the Eventing controller reports that the secret is changed during the controller startup which is not always true. As a side effect of that, the Eventing controller invokes the active backend to [stop](https://github.com/kyma-project/kyma/blob/a6d77e18431c42fa1ecb397256b3fabd3f896828/components/eventing-controller/controllers/backend/reconciler.go#L336) and do cleanup which is not desired.

**Steps to reproduce**

1. Switch the Eventing backend to EventMesh.
2. Create a Kyma subscription.
3. Check the creation time of the EvenMesh webhook.
4. Restart the Eventing controller.
5. Check the creation time of the EvenMesh webhook (notice it is changed).

**Acceptance criteria**

- [x] The bug is fixed.

mfaizanse:
**Before:**
![image](https://github.com/kyma-project/kyma/assets/9897945/a3f0f616-fae9-4027-9ac3-9b8a6c23cdb8)

PR #17781, opened 2023-06-30
Fixed EventMesh subscription deletion when OAuth not initialized

mfaizanse:
**Description**

- Fixed EventMesh subscription deletion when OAuth not initialized

**Related issue(s)**

- https://github.com/kyma-project/kyma/issues/17749

### DIFF
--- a/components/eventing-controller/controllers/backend/reconciler.go
+++ b/components/eventing-controller/controllers/backend/reconciler.go
@@ -324,7 +324,7 @@ func (r *Reconciler) syncOauth2ClientIDAndSecret(ctx context.Context, backendSta
 	}
 	oauth2CredentialsNotFound := k8serrors.IsNotFound(err)
 	oauth2CredentialsChanged := false
-	if err == nil {
+	if err == nil && r.isOauth2CredentialsInitialized() {
 		oauth2CredentialsChanged = !bytes.Equal(r.credentials.clientID, credentials.clientID) ||
 			!bytes.Equal(r.credentials.clientSecret, credentials.clientSecret) ||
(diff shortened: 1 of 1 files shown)


components/eventing-controller/controllers/backend/reconciler.go
```diff
@@ -1000,6 +1000,10 @@ func (r *Reconciler) getMutatingAndValidatingWebHookConfig(ctx context.Context)
 	return &mutatingWH, &validatingWH, nil
 }
 
+func (r *Reconciler) isOauth2CredentialsInitialized() bool {
+	return len(r.credentials.clientID) > 0 && len(r.credentials.clientSecret) > 0
```
marcobebway, line 1004: Does it make sense to also check the tokenURL and certsURL?

marcobebway: Also please write a unit test for this function.

mfaizanse: Not sure atm, that `certsURL` would be non-empty when oauth2 feature flag is disabled. Thats why I only checked the fields which are used in both cases (i.e. oauth2 feature flag enabled/disabled)

mfaizanse: I will add unit test in the next PR.

marcobebway (approved):

components/eventing-controller/controllers/backend/reconciler.go
marcobebway: Please also check the tokenURL and certsURL (if possible) in the followup PR.

PR #17802, opened 2023-07-06
record backend-errors with latency in epp

k15r:
**Description**

- record also metrics for failed calls to the backend

**Related issue(s)**


muralov (commented):
I see in all cases code="500" `eventing_epp_backend_requests_total{code="500",destination_service=""}`, but wantStatus is different in some test case. 
Is this correct?
Is the case where backend http status code is none 500tested?

k15r:
> I see in all cases code="500" `eventing_epp_backend_requests_total{code="500",destination_service=""}`, but wantStatus is different in some test cases. Is this correct? Is the case where backend http status code is none 500 tested?

this pr is just about the `eventing_epp_backend_request_total` to actually track errors. Presenting failed calls here with 500 is already an improvement. The HTTP Code used here is the one returned from the backend. In case of NATS we did some special handling of the error earlier to provide a meaningful return code to the caller. This will be synchronized later

muralov (approved):

PR #17812, opened 2023-07-07
Bump kiwigrid k8s-sidecar to 1.24.6 version

a-thaler:
**Description**
We did not update the sidecar version for a long time as we observed increased resource consumption.
I tried it once more and using the additional flag IGNORE_ALREADY_PROCESSED the consumption is very similar as before. The big advantage is that logs are now in JSON and the info logs are reduced to a minimum. As the resources are not getting processed constantly anymore but only on changes, the amount of logs is reduced heavily.

- update to latest version 1.24.6
- add new flag which works from k8s 1.19 on, to process only changed resources

**Related issue(s)**

https://github.com/kyma-project/third-party-images/pull/364


skhalash:
Let's convert it to a regular PR?

a-thaler:
I updated to the final image, ready for review

skhalash (approved):

skhalash (approved):

PR #17822, opened 2023-07-13
simplify sender interface

k15r:
**Description**

previously the sender had to return
- a PublishResult
- an Error

The problem here was that PublishResult was treated as an error
indicator for the eventMesh Backend, and the error was for other senders
such as the on Nats

The new interface just expects an Error to be returned in case of an
error. Any successful call will result in a 204 response with an empty
message.

This means now the sender has to map its errors or error responses onto
an error object, and any other function in the chain can just treat all
errors equally.

**Related issue(s)**

### DIFF
--- a/components/event-publisher-proxy/pkg/handler/handler.go
+++ b/components/event-publisher-proxy/pkg/handler/handler.go
@@ -315,8 +315,6 @@ func (h *Handler) sendEventAndRecordMetrics(ctx context.Context, event *cev2even
 			code = pubErr.Code()
 		}
 		h.collector.RecordBackendLatency(duration, code, host)
-		h.collector.RecordBackendRequests(code, host)
-		h.collector.RecordBackendError()
 		return err
 	}
 	originalEventType := event.Type()
@@ -334,7 +332,6 @@ func (h *Handler) sendEventAndRecordMetrics(ctx context.Context, event *cev2even
 	}
 	h.collector.RecordEventType(originalEventType, event.Source(), http.StatusNoContent)
 	h.collector.RecordBackendLatency(duration, http.StatusNoContent, host)
-	h.collector.RecordBackendRequests(http.StatusNoContent, host)
 	return nil
 }
 
--- a/components/event-publisher-proxy/pkg/handler/handler.go
+++ b/components/event-publisher-proxy/pkg/handler/handler.go
@@ -67,26 +67,29 @@ type Handler struct {
 	OldEventTypePrefix string
 }
 
-// NewHandler returns a new HTTP Handler instance.
-func NewHandler(receiver *receiver.HTTPMessageReceiver, sender sender.GenericSender, healthChecker health.Checker,
+// New returns a new HTTP Handler instance.
+func New(receiver *receiver.HTTPMessageReceiver, sender sender.GenericSender, healthChecker health.Checker,
 	requestTimeout time.Duration, legacyTransformer legacy.RequestToCETransformer, opts *options.Options,
 	subscribedProcessor *subscribed.Processor, logger *logger.Logger, collector metrics.PublishingMetricsCollector,
 	eventTypeCleaner eventtype.Cleaner, ceBuilder builder.CloudEventBuilder, oldEventTypePrefix string,
 	activeBackend env.ActiveBackend) *Handler {
 	return &Handler{
+		Name:                "",
 		Receiver:            receiver,
 		Sender:              sender,
 		HealthChecker:       healthChecker,
-		RequestTimeout:      requestTimeout,
+		Defaulter:           nil,
 		LegacyTransformer:   legacyTransformer,
+		RequestTimeout:      requestTimeout,
 		SubscribedProcessor: subscribedProcessor,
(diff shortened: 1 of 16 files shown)


mmitoraj (approved):

components/event-publisher-proxy/pkg/sender/eventmesh/eventmesh.go
```diff
@@ -36,47 +37,43 @@ func (s *Sender) URL() string {
...
+		return e
+	}
+	if resp.StatusCode >= http.StatusOK && resp.StatusCode < http.StatusMultipleChoices {
+		return nil
 	}
 	body, err := io.ReadAll(resp.Body)
 	defer func() { _ = resp.Body.Close() }()
 	if err != nil {
-		return nil, err
+		return common.ErrInternalBackendError
```
mfaizanse, line 74: should we log the `err` object here? or wrap it with the returned error?

components/event-publisher-proxy/pkg/sender/jetstream/jetstream.go
```diff
@@ -58,48 +60,60 @@ func NewSender(ctx context.Context, connection *nats.Conn, envCfg *env.NATSConfi
...
+func (s *Sender) Send(_ context.Context, event *event.Event) sender.PublishError {
 	if s.ConnectionStatus() != nats.CONNECTED {
-		return nil, ErrNotConnected
+		return ErrNotConnected
 	}
 
 	jsCtx, jsError := s.connection.JetStream()
 	if jsError != nil {
-		return nil, jsError
+		return common.ErrClientNoConnection
```
mfaizanse, line 77: should we log the jsError object here? or wrap it with the returned error?

components/event-publisher-proxy/pkg/handler/handler.go
```diff
@@ -331,10 +330,9 @@ func (h *Handler) sendEventAndRecordMetrics(ctx context.Context, event *cev2even
 			originalEventType = event.Type()
 		}
 	}
-	h.collector.RecordEventType(originalEventType, event.Source(), result.HTTPStatus())
-	h.collector.RecordBackendLatency(duration, result.HTTPStatus(), host)
-	h.collector.RecordBackendRequests(result.HTTPStatus(), host)
```
mfaizanse, line 336: Dont we need this metric `h.collector.RecordBackendRequests(result.HTTPStatus(), host)`?

like `h.collector.RecordBackendRequests(http.StatusNoContent, host)`?

components/event-publisher-proxy/pkg/metrics/collector.go
```diff
@@ -13,21 +13,11 @@ import (
...
 	// BackendLatencyKey name of the backendLatencyHelp metric.
 	BackendLatencyKey = "eventing_epp_backend_duration_milliseconds"
 	// backendLatencyHelp help text for the backendLatencyHelp metric.
 	backendLatencyHelp = "The duration of sending events to the messaging server in milliseconds"
 
-	// BackendRequestsKey name of the eventRequests metric.
-	BackendRequestsKey = "eventing_epp_backend_requests_total"
-	// backendRequestsHelp help text for event backendRequests metric.
-	backendRequestsHelp = "The total number of backend requests"
-
```
mfaizanse, line 30: We dont need these metrics? or do you think the info is already covered by the latency metric?

components/event-publisher-proxy/pkg/sender/eventmesh/eventmesh.go
k15r: yeah, logging is probably a good idea

components/event-publisher-proxy/pkg/sender/jetstream/jetstream.go
k15r: yeah, logging is probably a good idea

components/event-publisher-proxy/pkg/metrics/collector.go
k15r: the latency metric is a histogram.. it already provides you a count 
so figuring out the failed ones is just about filtering by the statusCode

components/event-publisher-proxy/pkg/handler/handler.go
k15r: BackendRequests is the total count of requests done to the backend -> 
eventing_epp_backend_duration_milliseconds_count

as the latency metric has destination_service and code as labels we can filter by those and get:
- all successful requests (filter by 2XX)
- all failed requests (filter by non2XX)
- all requests (dont filter)

components/event-publisher-proxy/pkg/sender/jetstream/jetstream.go
mfaizanse: Do we log here now?

k15r: yes :)

NHingerl (approved):

mfaizanse (approved):

PR #17852, opened 2023-07-18
Adapt EC for eventing manager

muralov:
**Description**
Adapt Eventing Controller to be able to be used by Eventing Manager

- Make set EPP owner EC setting with flag so that we can set Eventing CR as owner
- Add some methods used by EM to update NATS and publisher config

**Related issue(s)**
[issue](https://github.com/kyma-project/eventing-manager/issues/16)

### DIFF
--- a/components/eventing-controller/controllers/backend/reconciler.go
+++ b/components/eventing-controller/controllers/backend/reconciler.go
@@ -119,6 +119,14 @@ func NewReconciler(ctx context.Context, natsSubMgr subscriptionmanager.Manager, 
 	}
 }
 
+func (r *Reconciler) SetNatsConfig(natsConfig env.NATSConfig) {
+	r.natsConfig = natsConfig
+}
+
+func (r *Reconciler) SetBackendConfig(backendCfg env.BackendConfig) {
+	r.cfg = backendCfg
+}
+
 // +kubebuilder:rbac:groups="",resources=secrets,verbs=get;list;watch;update;patch;create;delete
 // +kubebuilder:rbac:groups=apps,resources=deployments,verbs=get;list;watch;update;patch;create;delete
 // +kubebuilder:rbac:groups=eventing.kyma-project.io,resources=eventingbackends,verbs=get;list;watch;create;update;patch;delete
@@ -621,6 +629,13 @@ func getSecretStringData(clientID, clientSecret, tokenEndpoint, grantType, publi
 }
 
 func (r *Reconciler) CreateOrUpdatePublisherProxy(ctx context.Context, backend eventingv1alpha1.BackendType) (*appsv1.Deployment, error) {
+	return r.CreateOrUpdatePublisherProxyDeployment(ctx, backend, true)
+}
+
+func (r *Reconciler) CreateOrUpdatePublisherProxyDeployment(
+	ctx context.Context,
+	backend eventingv1alpha1.BackendType,
+	withOwnerReference bool) (*appsv1.Deployment, error) {
 	var desiredPublisher *appsv1.Deployment
 	// set backend type here so that the function can be used in eventing-manager
 	r.backendType = backend
@@ -634,8 +649,10 @@ func (r *Reconciler) CreateOrUpdatePublisherProxy(ctx context.Context, backend e
 		return nil, fmt.Errorf("unknown EventingBackend type %q", backend)
 	}
 
-	if err := r.setAsOwnerReference(ctx, desiredPublisher); err != nil {
-		return nil, errors.Wrapf(err, "set owner reference for Event Publisher failed")
+	if withOwnerReference {
+		if err := r.setAsOwnerReference(ctx, desiredPublisher); err != nil {
+			return nil, errors.Wrapf(err, "set owner reference for Event Publisher failed")
+		}
 	}
 
 	currentPublisher, err := r.getEPPDeployment(ctx)
--- a/components/eventing-controller/controllers/backend/reconciler.go
+++ b/components/eventing-controller/controllers/backend/reconciler.go
@@ -103,8 +103,15 @@ type Reconciler struct {
 	credentials oauth2Credentials
 }
 
-func NewReconciler(ctx context.Context, natsSubMgr subscriptionmanager.Manager, natsConfig env.NATSConfig,
-	envCfg env.Config, backendCfg env.BackendConfig, bebSubMgr subscriptionmanager.Manager, client client.Client, logger *logger.Logger,
+func NewReconciler(
+	ctx context.Context,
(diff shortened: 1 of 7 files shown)


components/eventing-controller/pkg/deployment/publisher_deployment.go
```diff
@@ -43,7 +43,7 @@ const (
 	PublisherSecretEMSURLKey        = "ems-publish-url"
 	PublisherSecretBEBNamespaceKey  = "beb-namespace"
 
-	configMapName               = "eventing"
+	configMapName               = "eventing-configmap"
```
muralov, line 46: I really had to do this change because I was not able to create a configmap with `eventing`

components/eventing-controller/pkg/env/nats_config.go
```diff
@@ -9,7 +9,7 @@ import (
 // NATSConfig represents the environment config for the Eventing Controller with Nats.
 type NATSConfig struct {
 	// Following details are for eventing-controller to communicate to Nats
-	URL           string `envconfig:"NATS_URL" required:"true"`
```
muralov, line 12: I had to make this change to to make it usable for Eventing Manager as it builds the nats url on the fly.

components/eventing-controller/controllers/backend/reconciler.go
```diff
@@ -116,10 +122,18 @@ func NewReconciler(ctx context.Context, natsSubMgr subscriptionmanager.Manager,
 		Client:     client,
 		logger:     logger,
 		record:     recorder,
-		cfg:        cfg,
+		cfg:        backendCfg,
 	}
 }
 
+func (r *Reconciler) SetNatsConfig(natsConfig env.NATSConfig) {
```
friedrichwilken, line 129: Why is this public? Where is this used?

components/eventing-controller/controllers/backend/reconciler.go
```diff
@@ -103,10 +103,16 @@ type Reconciler struct {
...
-	envCfg env.Config, bebSubMgr subscriptionmanager.Manager, client client.Client, logger *logger.Logger,
+func NewReconciler(
+	ctx context.Context,
+	natsSubMgr subscriptionmanager.Manager,
+	natsConfig env.NATSConfig,
+	envCfg env.Config,
+	backendCfg env.BackendConfig,
+	bebSubMgr subscriptionmanager.Manager,
+	client client.Client,
+	logger *logger.Logger,
```
friedrichwilken, line 114: nice!

components/eventing-controller/controllers/backend/reconciler.go
```diff
@@ -622,7 +636,16 @@ func getSecretStringData(clientID, clientSecret, tokenEndpoint, grantType, publi
 }
 
 func (r *Reconciler) CreateOrUpdatePublisherProxy(ctx context.Context, backend eventingv1alpha1.BackendType) (*appsv1.Deployment, error) {
+	return r.CreateOrUpdatePublisherProxyDeployment(ctx, backend, true)
+}
+
+func (r *Reconciler) CreateOrUpdatePublisherProxyDeployment(
```
friedrichwilken, line 642: can we make `CreateOrUpdatePublisherProxyDeployment` private?

components/eventing-controller/controllers/backend/reconciler.go
muralov: It is used in Eventing Auth Manager. Eventing Manager updates NATS config to apply the desired NATS configuration from Eventing CR.

components/eventing-controller/controllers/backend/reconciler.go
muralov: this is a new function used by Eventing Manager, and therefore it must be public. The reason I created a new function is that the old one `CreateOrUpdatePublisherProxy` sets EC deployment as owner, but we want Eventing CR as owner in case of Eventing Manager.

components/eventing-controller/controllers/backend/reconciler.go
friedrichwilken: Thank you!

components/eventing-controller/controllers/backend/reconciler.go
friedrichwilken: Thank you!

components/eventing-controller/controllers/backend/reconciler.go
```diff
@@ -633,8 +656,10 @@ func (r *Reconciler) CreateOrUpdatePublisherProxy(ctx context.Context, backend e
 		return nil, fmt.Errorf("unknown EventingBackend type %q", backend)
 	}
 
-	if err := r.setAsOwnerReference(ctx, desiredPublisher); err != nil {
-		return nil, errors.Wrapf(err, "set owner reference for Event Publisher failed")
+	if withOwnerReference {
```
friedrichwilken, line 659:
```suggestion
	if hasOwnerReference {
```
see above

components/eventing-controller/controllers/backend/reconciler.go
```diff
@@ -622,7 +636,16 @@ func getSecretStringData(clientID, clientSecret, tokenEndpoint, grantType, publi
 }
 
 func (r *Reconciler) CreateOrUpdatePublisherProxy(ctx context.Context, backend eventingv1alpha1.BackendType) (*appsv1.Deployment, error) {
+	return r.CreateOrUpdatePublisherProxyDeployment(ctx, backend, true)
+}
+
+func (r *Reconciler) CreateOrUpdatePublisherProxyDeployment(
+	ctx context.Context,
+	backend eventingv1alpha1.BackendType,
+	withOwnerReference bool) (*appsv1.Deployment, error) {
```
friedrichwilken, line 645:
```suggestion
	withOwnerReference bool) (*appsv1.Deployment, error) {
```
let's give this a more 'booly' name; we use the 'with' prefix usually for options like [here](https://github.com/kyma-project/nats-manager/blob/9241d3b926bd16da195988aeeeac54ec9d8e23d5/testutils/options.go#L53)

friedrichwilken (changes requested):

components/eventing-controller/controllers/backend/reconciler.go
muralov: the flag is to say whether set owner reference or not. `hasOwnerReference` would mean whether it has owner reference. I can rename it to `setOwnerReference` if you this to the existing one?

friedrichwilken: that sounds like a satisfactory solution.

friedrichwilken (approved):

PR #17861, opened 2023-07-19
Release 2.16 app con val performance bump

Disper:
**Description**

- bump that contain https://github.com/kyma-project/kyma/pull/17842 
- ...
- ...

/area application-connector
/kind bug


Disper:
attached one too many commit

PR #17911, opened 2023-07-31
Update dev tutorial with kyma default domain

ECasio:
**Description**

- As developer who has used the default kyma domain it would be more helpful to already provide it in the Developers Tutorial so that you simply copy and paste it


strekm (changes requested):
hello @ECasio our tutorials are written in most generic way to support all domains such as local domain, gardener domain or user provided domain. We want to support all cases in our tutorials not only local scenario.

strekm:
> hi @strekm okay I agree but would it not make sense to provide a small hint there at least as a "info" or "note" something like "in case you followed the tutorial step by step and you have not adjusted the cluster domain, the default domain is "local.kyma.dev"" ? I could adjust my PR and add this there something similar like this note here: image

i'm sorry but i can't agree with that. default domain is not "local.kyma.dev". domain depends on user setting and even for local environment it can be set differently. i don't want to introduce too much noise into this already not trivial topic. in addition this note would need to be repeated for every tutorial.

PR #18114, opened 2023-09-05
Remove profiles from webhook

MichalKalke:
**Description**

Cleanup [presets](https://github.com/kyma-project/kyma/blob/4785615d7d9c07c5cffd94754dcbe2d33491d915/resources/serverless/charts/webhook/values.yaml#L100) for function pod replication.

- Delete `presets` from `webhook_config.yaml`
- Delete `ReplicasPreset`
- Delete `defaultScaling` function
- Add label to skip tests

**Related issue(s)**
https://github.com/kyma-project/serverless-manager/issues/252


components/function-controller/pkg/apis/serverless/v1alpha2/function_defaults_test.go
```diff
@@ -390,10 +393,6 @@ func TestSetDefaults(t *testing.T) {
 func fixDefaultingConfig() *DefaultingConfig {
 	return &DefaultingConfig{
```
dbadura, line 394: You can write `t.Skip()` instead of build flags

components/function-controller/hack/webhook_config.yaml
```diff
@@ -8,26 +8,6 @@ reservedEnvs:
 function:
   replicas:
     minValue: "1"
-    defaultPreset: "L"
```
dbadura, line 11: You can remove it from webhook chart. (values.yaml file in serverless/charts/webhook)

dbadura (changes requested):

components/function-controller/hack/webhook_config.yaml
pPrecel: done

dbadura (approved):

PR #18266, opened 2023-10-05
fix docsify-tabs regex

dennis-ge:
**Description**

- The `div tabs` sections are broken in the telemetry module, e.g., https://kyma-project.io/#/telemetry-manager/user/03-traces?id=step-3a-add-authentication-details-from-plain-text or https://kyma-project.io/#/telemetry-manager/user/04-metrics?id=step-2b-add-authentication-details-from-secrets
- The reason is that the regex converting the markdown `div tabs` sections to proper [docsify-tabs](https://github.com/jhildenbiddle/docsify-tabs) sections is failing when the markdown sections are badly formatted. Badly formatted means that a `\n` or tab or whitespace is at the wrong place
- I formatted the telemetry markdown `div tabs` sections in https://github.com/kyma-project/telemetry-manager/pull/439
- This PR improves the regex that converts to markdown `div tabs` sections regex to docsify-tabs sections
- I also had a look at some other tabs sections and verified the new regex does not break them

**Related issue(s)**

### DIFF
--- a/docs/index.html
+++ b/docs/index.html
@@ -43,11 +43,11 @@
           // remove title metadata section
           hook.beforeEach(function (markdown) {
             markdown = markdown.replaceAll(/---.*\ntitle: (.*)\n---.*\n/gm, "# $1");
-            markdown = markdown.replace(/(<div tabs.*>)((.|\n)*?)(<\/div>)/gm,function(match,g1,g2,g3,g4){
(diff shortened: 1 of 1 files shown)


a-thaler (approved):

pbochynski:
@NHingerl @dennis-ge @a-thaler Please consider not using html tags for tabs. I introduced that plugin only to speed up migration from legacy website to docsify. This way we could render new website and old website from the same sources. As the old website is retired now we should use docsify syntax directly in the markdown files. Now we write in the old syntax and use magic regexp to convert it to the docsify syntax. Please change the documentation sources to [the markdown syntax](https://jhildenbiddle.github.io/docsify-tabs/#/)

PR #18281, opened 2023-10-06
Remove the application lister and cleaner from the eventing-controller

marcobebway:
**Description**

Remove the application lister and cleaner from the eventing-controller.

**Related issue(s)**

- https://github.com/kyma-project/eventing-manager/issues/124


components/eventing-controller/config/crd/kustomization.yaml
```diff
@@ -5,7 +5,6 @@ resources:
 - bases/eventing.kyma-project.io_subscriptions.yaml
 - bases/eventing.kyma-project.io_eventingbackends.yaml
 - external/apirules-gateway-kyma-project-io.yaml
```
mfaizanse, line 7: Should this also be deleted? since you are deleting the `components/eventing-controller/config/crd/external/apirules-gateway-kyma-project-io.yaml`?

marcobebway: Why should we delete the APIRules?

mfaizanse (approved):

PR #18312, opened 2023-10-18
Move application CRD to compass-runtime-agent

koala7659:
It is required as part of preparation to release Application Connector module.


m00g3n (approved):
LGTM

Issue #18497, opened 2023-12-18
Add more linting for the .md files in `docs` folder


mmitoraj:
**Description**

Add more linting for the `.md` files in the `docs` folder. Check: https://github.com/igorshubovych/markdownlint-cli#readme

Example:
https://github.tools.sap/perfx/cloud-logging-service/blob/main/.github/linters/.markdownlint.yml

The solution would require a config per repo.

**AC**
- [x] kyma (@grego952) (https://github.com/kyma-project/kyma/pull/18621)
- [x] community (@grego952) (https://github.com/kyma-project/community/pull/904)
- [x] documentation (@grego952)
- [ ] api-gateway (@nataliasitko)
- [ ] istio (@nataliasitko)
- [ ] btp-manager (@IwonaLanger)
- [ ] kyma-environment-broker  (@IwonaLanger)
- [x] eventing-manager (@grego952) (https://github.com/kyma-project/eventing-manager/pull/541)
- [x] nats-manager (@grego952) (https://github.com/kyma-project/nats-manager/pull/354)
- [x] telemetry-manager (@grego952) (https://github.com/kyma-project/telemetry-manager/pull/916)
- [x] application-connector-manager (@grego952) (https://github.com/kyma-project/application-connector-manager/pull/209)
- [x] keda-manager (@grego952) (https://github.com/kyma-project/keda-manager/pull/376)
- [x] serverless (@grego952) (https://github.com/kyma-project/serverless/pull/834)
- [ ] test-infra (@IwonaLanger)
- [x] lifecycle-manager (@mmitoraj)
- [x] Write instructions on how to use markdownlint-cli

**Assignees**

@kyma-project/technical-writers

**Attachments**

IwonaLanger:
The Gopher team is working on a GitHub action that will run the job on their repositories.

PR #18621, opened 2024-03-18
Apply markdowlint rules in the kyma repository

grego952:
**Description**

- Apply markdowlint rules in the kyma repository

**Related issue(s)**
See also https://github.com/kyma-project/kyma/issues/18497


mmitoraj (approved):

a-thaler:
Could you maybe add the new config files to the codeowner file, so that you are a defined owner of that

grego952:
@a-thaler 
Thanks for pointing this out.
I've added TWs as codeowners and removed the milv.config.yaml file, as it was deleted some time ago already.

heading_capitalization.js
```diff
@@ -0,0 +1,28 @@
+module.exports = [{
```
a-thaler, line 1: From where the file gets consumed? Is it picked up by naming convention? Could we maybe prefix it with something like ".markdown*" as well?
Could you at least put a comment in the file for what it is used

a-thaler (approved):

mmitoraj (approved):

